PR #34, opened 2015-02-12
Alpha fixes

suhr:
Subj.


badboy:
Do not merge this. I intentionally did not submit a PR as I rely on my personal forks of rust-sha1 and rust-url.
This PR now relies on the upstream repos, but we best want to rely on the released crates (otherwise redis-rs can't be released as a crate)

suhr:
> we best want to rely on the released crates

Fixed.

badboy:
Better, though unsure if he might want to retain the git retain dependencies as a comment.

PR #35, opened 2015-02-15
Update to compile with current Rust

panicbit:

PR #36, opened 2015-02-16
Update for rustc-2015-02-14

hugoduncan:
This removes all warnings except `or_else` deprecations using
rustc-2015-02-14.

It is based on https://github.com/mitsuhiko/redis-rs/pull/35


panicbit:
I'll close #35 then

Issue #40, opened 2015-04-03
Does not compile with rustc beta


vasily-kirichenko:
```
C:\Users\kot\.cargo\registry\src\github.com-1ecc6299db9ec823\redis-0.2.0\src\macros.rs:11:20: 11:55
error: unresolved name `std::error::FromError::from_error`
C:\Users\kot\.cargo\registry\src\github.com-1ecc6299db9ec823\redis-0.2.0\src\macros.rs:11         re
turn Err(::std::error::FromError::from_error($expr));

         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I think it's important to make it compilable with beta compiler because it'll soon become release 1.0.

PR #41, opened 2015-04-03
Replace FromError by convert::From

badboy:
This fixes the error mentioned in #40, but it brings up the next error:

```
src/types.rs:50:21: 50:28 error: the trait `core::clone::Clone` is not implemented for the type `std::io::error::Error` [E0277]
src/types.rs:50     InternalIoError(IoError),
```

`Clone`, `Eq` and `PartialEq` where removed from `io::Error`, see https://github.com/rust-lang/rust/pull/23919

Not sure how to go from here.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -8,7 +8,7 @@ macro_rules! ensure {
 
 macro_rules! fail {
     ($expr:expr) => (
-        return Err(::std::error::FromError::from_error($expr));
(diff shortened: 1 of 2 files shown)


daboross:
@badboy That error means that Clone is no longer implemented for io::Error. Removing the `#[derive(Clone)]` and `#[derive(Copy)]` from ErrorKind and RedisError should fix it.

PR #52, opened 2015-07-06
Pass "--yes" to rustup.sh for non-interactive Travis CI build

56quarters:
Looks like rustup.sh became interactive by default with https://github.com/rust-lang/rustup/commit/5e5bf08d5c7523ca5645aa61a797a9d37932d5d3 . This change passes the "--yes" option to make sure it doesn't prompt when running from Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ before_install:
   - sudo apt-get update
 
 install:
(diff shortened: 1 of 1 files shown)


56quarters:
Saw that this duplicates #49

PR #78, opened 2016-01-13
after running cargo rustfmt on the source tree

harrydevnull:
source tree formatted using cargo rustfmt


badboy:
I'm not a big fan of some of these changes.

PR #79, opened 2016-01-23
Shutdown connection upon protocol error

arthurprs:
Shutdown connection upon protocol error

Solves #77

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,7 +1,7 @@
 #[cfg(feature="unix_socket")]
 use std::path::PathBuf;
 use std::io::{Read, BufReader, Write};
-use std::net::TcpStream;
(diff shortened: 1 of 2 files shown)


badboy:
LGTM

badboy:
The test is not optimal though. We will always have an `IoError` on the second call, simply because at that point the server is already gone.

arthurprs:
Oh, I see. I'll get to it later today.

arthurprs:
I updated the commit. I reused the library parser to make the test correct.

badboy:
Cool! Don't worrry about the test failure, I'll fix that one and then merge it.

badboy:
Merged as bba92d93ef832a15a631a35b07b487a2f6818017

PR #88, opened 2016-04-26
Update to url 1.0.0

badboy:
rust-url v1.0.0 had some major breaking changes.
The biggest one was the change from struct members to getter methods
and the fact that the scheme mapper was removed.
With it, the `InvalidScheme` error was removed.

redis-rs used this to fail with schemes other than `redis` and `unix`.

I now changed the return type to a RedisResult, so we can handle the
error ourselves. It is only called at one position anyway.

`parse_redis_url` _is_ part of the publicly exposed API though, so
changing the return type is a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,6 @@ readme = "README.md"
 
 [dependencies]
 sha1 = "0.1.1"
-url = "0.5.4"
+url = "1.0.0"
 rustc-serialize = "0.3.16"
 unix_socket = { version ="0.5.0", optional = true }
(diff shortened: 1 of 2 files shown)


dogabudak:
We are using redis-rs plugin for our production systems. Its crucial for us. Can you merge it to master ASAP?

badboy:
It will be merged ASAP. ;)

ksaritek:
Hi, we already have 10 projects depends on that project at prod. Is there a time for that :) is there any help needed that we can do?

PR #98, opened 2016-07-24
split encode_pipeline in two functions to remove boolean flag for ato…

MarkusJais:
currently reading the src code. I like it. This is just a small suggestion to remove a boolean flag and make atomic pipelines more explicit.


badboy:
Very late thanks, but this is a small internal function, so I don't bother with changing it now.

PR #102, opened 2016-08-09
Fix compilation when unix socket features are disabled

reset:
I am pretty certain this was just a find and replace error, thank you for the amazing library!


badboy:
It was indeed. Thanks!

PR #106, opened 2016-09-29
Fix FromRedisValue for bool

pzread:
The `ToRedisArgs` will convert bool value to the string "true" or "false".
But the `FromRedisValue` for bool type won't convert them back to bool value.

This patch makes `FromRedisValue` convert `Value::Data` of bool to the corresponding bool value.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -742,6 +742,13 @@ impl FromRedisValue for bool {
                 }
             }
             Value::Okay => Ok(true),
+            Value::Data(ref bytes) => {
+                match try!(from_utf8(bytes)).parse::<bool>() {
+                    Ok(rv) => Ok(rv),
+                    Err(_) => invalid_type_error!(v,
+                        "Could not convert from string.")
+                }
+            },
             _ => invalid_type_error!(v,
                 "Response type not bool compatible."),
         }
--- a/tests/test_basic.rs
+++ b/tests/test_basic.rs
@@ -294,6 +294,20 @@ fn test_json() {
 }
 
 #[test]
+fn test_bool() {
+    let ctx = TestContext::new();
+    let con = ctx.connection();
+
+    redis::cmd("SET").arg("true").arg(true).execute(&con);
+    redis::cmd("SET").arg("false").arg(false).execute(&con);
+
+    let true_val : bool = redis::cmd("GET").arg("true").query(&con).unwrap();
+    let false_val : bool = redis::cmd("GET").arg("false").query(&con).unwrap();
+    assert_eq!(true_val, true);
+    assert_eq!(false_val, false);
+}
+
+#[test]
 fn test_scanning() {
     let ctx = TestContext::new();
     let con = ctx.connection();


mitsuhiko:
I wonder why we convert to `true`/`false` in `ToRedisArgs`. That seems wrong.

pzread:
For now, ToRedisArgs just use `to_string` to convert bool, which will output `true` / `false` by rust implementation.

Maybe converting bool to `0` / `1` is a better idea ?

przygienda:
generally, we have a problem with Vec<u8>, #116 has the fix on 

5f20465a915561861fe681be24782f284d9e31e0

pzread:
I reviewed the commit
2e3c25f8703c769f89262758d1104b3316d35b82
I prefer to representing the `bool` by using `"0"/ "1"` instead of `"true" / "false"`
So I think it's better than this one.

Geobert:
From Rust's doc:
> The bool represents a value, which could only be either true or false. If you cast a bool into an integer, true will be 1 and false will be 0.

So if this idea is going on, 0/1 seems better to me as well :)

jwilm:
One consideration with such a change is that switching serialization from "true"/"false" to "1"/"0" could silently break dependent projects.

badboy:
> One consideration with such a change is that switching serialization from "true"/"false" to "1"/"0" could silently break dependent projects.

Absolutely, that's why I'm considering _not_ encoding bools at all and let the user handle that (that means: removing the implementation of `ToRedisArgs for bool`)

mitsuhiko:
Are there any redis commands where the bool even appears? I feel like deprecating and removing it would be sensible.

Marwes:
> Are there any redis commands where the bool even appears? I feel like deprecating and removing it would be sensible.

Arguably in the return value of https://redis.io/commands/setnx#return-value (etc), as `1` and `0`.

mitsuhiko:
@Marwes i would not remove the bool support for `FromRedisValue` just `ToRedisArgs `.

jwilm:
> Are there any redis commands where the bool even appears? I feel like deprecating and removing it would be sensible.

I get the impression this is more about ser/de for custom types rather than Redis commands. For instance, can define structs with bools and ser/de with `hmset` and `hgetall`.

In OneSignal/serde-redis, [this](https://github.com/OneSignal/serde-redis/pull/13) was the approach we took for deserialization. Didn't see a harm in generous interpretation of bools for decoding.

Terkwood:
Aside from the design decision that's been discussed here, and in #324, this particular pull request doesn't seem valid for a nitpicky reason:  the forked repo tied to this PR [has been deleted](https://github.com/pzread/redis-rs).

unkrepo

Without wanting to discourage efforts here, this PR is further complicated by the fact that there's a merge conflict present.

Because of these nitpicks, I would suggest closing this particular PR.  Depending on the decision made in #324, we can still refer to the closed PR and its code changes if needed, for inspiration.

pzread:
I haven't tracked on this issue for a long time. Closed the PR and waiting for the design decision.

Terkwood:
> @Marwes i would not remove the bool support for `FromRedisValue` just `ToRedisArgs `.

FWIW, I vote for this option.  We could potentially follow-on with support for `ToRedisArgs` as an opt-in feature.

We might want to consider releasing the library as v1.0.0 since this would be a breaking change.  I have a feeling we're already running in some production systems, whether we like it or not :wink:

PR #114, opened 2017-01-26
Fixed setbit command. True/False caused server to error

hanscj1:
Also bumped the crate version to 0.8.1.


badboy:
`value as u8` should work as well, not sure which is more idiomatic.

Also please don't bump the version number, we will take care of that when actually cutting a release.

badboy:
Added tests in #117 and removed version change.

PR #116, opened 2017-01-31
implemented BTreeMap/Set conversions ...

przygienda:
added the lacking BTreeMap/Set conversions ... 

I don't know where to plug in the to_redis_args based on the HashSet/Map implementation. Somehow BTreeMap has more traits and the implementation doesn't catch ...


badboy:
Please don't commit your IDE's project files.
Please add tests.
Test failure unrelated, I restarted it.

przygienda:
OK ... 

Can you tell me 

a) WHERE the implementation for Hash for the to_redis is? I can't find it since I need to include one that catches BTreeMap 

b) which test do you want the stuff to go into ?

badboy:
I'm not entirely sure what you are asking. Do you mean [the already existing implementations for `HashMap` and `HashSet`](https://github.com/mitsuhiko/redis-rs/blob/b2ab3686b4b5f898720cb39e0285d4416c42649f/src/types.rs#L775-L806)?

There are [basic tests for both implementations](https://github.com/mitsuhiko/redis-rs/blob/b2ab3686b4b5f898720cb39e0285d4416c42649f/tests/test_basic.rs#L221-L254). You would need two similar test functions for your new implementations.

przygienda:
OK, extended significantly, now the whole thing deals with maps as well gracefully generating them from BTreeSet and so on for HMGET and similar things. Tests added. IDE removed. Pls review ...

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
badboy: This belongs into your global gitignore, not in a project-specific one.

src/types.rs
```diff
@@ -558,6 +570,58 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
     }
 }
 
+impl<T: ToRedisArgs + Hash + Eq> ToRedisArgs for HashSet<T> {
+    fn to_redis_args(&self) -> Vec<Vec<u8>> {
+        let mut rv : Vec<&T> = Vec::with_capacity(self.len());
+
+        rv.extend(self.iter());
+        ToRedisArgs::make_arg_vec_ref(rv.as_slice())
```
badboy: This looks like we are allocating twice. Once for the Vector in here, once for the Vector in `make_arg_vec_ref`, which seems kinda wasteful.

src/types.rs
```diff
@@ -558,6 +570,58 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+}
+
+/// this flattens BTreeMap into something that goes well with HMSET
+impl<T: ToRedisArgs + Hash + Eq + Ord,
+     V: ToRedisArgs> ToRedisArgs for BTreeMap<T,V> {
+    fn to_redis_args(&self) -> Vec<Vec<u8>> {
+        let mut rv = Vec::with_capacity(self.len());
+
+        rv.extend(self.keys()
+            .zip(self.values())
```
badboy: Overly complicated to zip keys and values. That's what the default `into_iter` already does. Use that.

src/types.rs
```diff
@@ -558,6 +570,58 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+            .into_iter()
+            .flat_map(|e| {
```
badboy: Use value destructuring here: `|(key,value)|`. Makes the code below far easier to read.

src/types.rs
```diff
@@ -558,6 +570,58 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+                // otherwise things like HMSET will simply NOT work
+                assert!(e.0.is_single_arg() &&
+                        e.1.is_single_arg());
+
+                vec![e.0.to_redis_args(),
+                               e.1.to_redis_args()].into_iter() } )
+            .collect::<Vec<_>>());
```
badboy: Avoid allocation. `extend` already expects an iterator.

src/types.rs
```diff
@@ -558,6 +570,58 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+        ToRedisArgs::make_arg_vec(rv.as_slice())
```
badboy: `rv.as_slice()` is just `&rv[..]`. Here simply `&rv` also works and is more idiomatic (IMO)

src/types.rs
```diff
@@ -558,6 +570,58 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+    fn is_single_arg(&self) -> bool {
+        self.len() <= 1
+    }
+}
+
+/// this flattens BTreeMap into something that goes well with HMSET
+impl<T: ToRedisArgs + Hash + Eq + Ord,
+     V: ToRedisArgs> ToRedisArgs for BTreeMap<T,V> {
+    fn to_redis_args(&self) -> Vec<Vec<u8>> {
+        let mut rv = Vec::with_capacity(self.len());
```
badboy: We actually need `self.len()*2` (key & value)

src/types.rs
badboy: `make_arg_vec_ref` was only added for this new use case and in both instances we double-allocate. We can simply inline this functionality for now

src/types.rs
```diff
@@ -437,6 +438,17 @@ pub trait ToRedisArgs: Sized {
         rv
     }
 
+    /// This only exists internally as a workaround for the lack of
+    /// specialization.
+    #[doc(hidden)]
+    fn make_arg_vec_ref(items: &[&Self]) -> Vec<Vec<u8>> {
+        let mut rv = vec![];
+        for item in items.iter() {
+            rv.extend(item.to_redis_args().into_iter());
```
badboy: No need for an explicit `.into_iter()` here.

przygienda:
thanks. will attend to it. Fighting some other issues right now with the package on top ... Once that settles will come back here ...

src/types.rs
przygienda: done

src/types.rs
przygienda: done, just cruft that accumulated while I was experimenting

src/types.rs
przygienda: done

src/types.rs
przygienda: done, cool stuff, didn't know ...

src/types.rs
przygienda: done

src/types.rs
```diff
@@ -558,6 +570,57 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+/// this flattens BTreeMap into something that goes well with HMSET
+impl<T: ToRedisArgs + Hash + Eq + Ord,
+     V: ToRedisArgs> ToRedisArgs for BTreeMap<T,V> {
+    fn to_redis_args(&self) -> Vec<Vec<u8>> {
+        let mut rv = Vec::with_capacity(self.len()*2);
+
+        rv.extend(self
+            .into_iter()
+            .flat_map(|e| {
+                let (key,value) = e;
```
badboy: Why not `|(key,value)|`?

.gitignore
badboy: This still needs to be removed.

przygienda:
normalized bool representation. The test should actually write _all_ types into Redis and read back since right now they assume somehow the right "value type" shows up ...

badboy:
Uhm, @przygienda, the last commit looks like it is related to #106 so it should be kept separate from the *Map implementations in this PR.

przygienda:
sigh, I have no time to keep track of multiple branches so it's here. I have 3-4 different projects I'm fixing stuff to get it to work ... So please let that one just go in because as master is now it's buggy.

badboy:
If you'd take a look at that other issue you'd see that using the strings `true` or `false` are probably the wrong way to go.
Thus I can't let this sit here as well. However, I can rip out that commit manually before merging.

przygienda:
what's the "other issue".  I commit "0" and "1" as [u8;1] which is what you had as "status". redis on read does return Value::Data(..) which is what I check while I left your checks in place ...

badboy:
#106 

Oh sorry. At first glance I somehow overlooked the `"0"` and `"1"` and just assumed it was true/false. In that case it might be more likely to be correct. I'll still pull this apart from the initial issue

przygienda:
cool, cherry pick or whatever you like ... It's just one diff ... thanks much ...

src/types.rs
przygienda: done

przygienda:
@badboy: ok, beta fails due to unrelated stuff ... otherwise e'thing reworked. Added more bug-fixes while working with it ... Let me know what you need more ..

badboy:
I'll hope to do a final review this week and then merge it.

badboy:
You didn't address the double-allocation at all. Could you take another look?

.gitignore
przygienda: done

przygienda:
so what's still outstanding ? Took care of pretty much everything requested ...

badboy:
My last final review. I got a bit swamped with other work the last two weeks. I'll take care of this tomorrow.

przygienda:
Have it working now under pretty heavy load in my app, 10K+ resulting key writes and the only gotcha was the empty sets/maps that Redis does not support. I put a comment in ... running it over redis 3.2.6 up to 3.2.8 ... Thanks for the outstanding package ...

badboy:
I reordered & rebased your commits. Merging in #119

PR #127, opened 2017-06-12
Don't crash if the rustc version can't be determined

Eijebong:
This was causing troubles with `cargo clippy` as they seem to replace
the RUSTC env variable (see https://github.com/Manishearth/rust-clippy/issues/1517)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,27 +1,41 @@
 use std::env;
 use std::ffi::OsString;
 use std::process::Command;
+use std::error::Error;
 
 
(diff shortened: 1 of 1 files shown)


Eijebong:
Ping ? :)

Eijebong:
Re-ping, it's been a while now

badboy:
Thanks, will take a look next week

Eijebong:
Thank you ! :)

badboy:
I have to say thank you.

PR #129, opened 2017-07-08
Nom parser

gsquire:
This is an *extremely* rough start to using nom. I am sharing it to get thoughts on the layout as well as how to debug the unit tests. Perhaps I have written the parsers in a completely wrong way.

If this seems like it will not fit into this library, then feel free to close it. Thanks for looking.

Related to #81

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1,11 +1,13 @@
-use std::io::{Read, BufReader};
+use std::io::Read;
+use std::str::{self, FromStr};
 
 use types::{RedisResult, Value, ErrorKind, make_extension_error};
 
+use nom::{IResult, line_ending};
 
 /// The internal redis response parser.
 pub struct Parser<T> {
-    reader: T,
+    reader: T
 }
 
 /// The parser can be used to parse redis responses into values.  Generally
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -27,164 +29,149 @@ impl<'a, T: Read> Parser<T> {
...
     pub fn parse_value(&mut self) -> RedisResult<Value> {
-        let b = try!(self.read_byte());
-        match b as char {
-            '+' => self.parse_status(),
-            ':' => self.parse_int(),
-            '$' => self.parse_data(),
-            '*' => self.parse_bulk(),
-            '-' => self.parse_error(),
-            _ => fail!((ErrorKind::ResponseError, "Invalid response when parsing value")),
+        let mut buf = [0; 4];
```
gsquire: I need to figure out the best way to read from the socket. This is just for testing locally now.

gsquire:
@badboy Sorry to ping you directly, but I am curious if you find this PR valuable? If so I will keep iterating on it.

gsquire:
In any case I will keep working on it. I wanted to see if it would be a hard no or not :)

No rush on the feedback by the way! Whenever is convenient.

jwilm:
@gsquire I am personally curious about performance difference between the current parser and your nom based one. Any chance you can provide some numbers?

gsquire:
@jwilm I don't have any numbers I am confident in now and am more focused on correctness. If it proves to be the same or worse then I will abandon this.

gsquire:
I have to say...I am puzzled why reading from the socket causes the test to hang. The next step to try is a tcpdump of the traffic.

src/parser.rs
```diff
@@ -27,164 +29,149 @@ impl<'a, T: Read> Parser<T> {
...
-        let b = try!(self.read_byte());
-        match b as char {
-            '+' => self.parse_status(),
-            ':' => self.parse_int(),
-            '$' => self.parse_data(),
-            '*' => self.parse_bulk(),
-            '-' => self.parse_error(),
-            _ => fail!((ErrorKind::ResponseError, "Invalid response when parsing value")),
+        let mut buf = vec![];
+        match self.buf_reader.read_to_end(&mut buf) {
```
badboy: reading a stream to end will wait for an EOF, meaning for the connection to close.
That's definitely not what you want here.
You would want to read as much as possible, then try to parse it and if necessary read more of it.

src/parser.rs
```diff
@@ -27,164 +29,149 @@ impl<'a, T: Read> Parser<T> {
...
-            "NOSCRIPT" => ErrorKind::NoScriptError,
-            code => {
-                fail!(make_extension_error(code, pieces.next()));
-            }
-        };
-        match pieces.next() {
-            Some(detail) => fail!((kind, desc, detail.to_string())),
-            None => fail!((kind, desc)),
+fn parse_error(line: &str) -> RedisResult<Value> {
+    let desc = "An error was signaled by the server";
```
badboy: _signalled_

badboy: Hm, okay, both spellings are ok. Don't mind then.

badboy:
It's definitely off to a good start.

> @jwilm I don't have any numbers I am confident in now and am more focused on correctness. If it proves to be the same or worse then I will abandon this.

The current parser is pretty simple and not optimized at all. I'm sure there is some low-hanging fruit to speed it up.
However, we currently don't have any good benchmark for the parser. The 3 bundled benchmarks are are a starting point, but don't exercise most parts of the parser.

src/parser.rs
gsquire: I am working on this now, and am trying to figure out the most efficient way to do so.

gsquire:
I need to read up on nom and why I am getting `Alt` errors when feeding input. The unit tests I had written and tested separately all seemed happy before migrating it to this crate. I will keep digging into it.

gsquire:
Another status update...

It's just hanging on the transaction tests now. I know the tests are failing but I will attend to that once I am done figuring out why it's not reading all that it should.

gsquire:
@badboy I haven't had much time to look at this as I just moved into a new place. I still plan on getting this working and will return to it in due time. If you have any free time, you can edit my PR too. :)

gsquire:
The transaction tests are failing on this line: https://github.com/mitsuhiko/redis-rs/pull/129/files#diff-2c09afcdc3c420ab0678ba9b5e83959cR35

The read is blocking for some reason and I cannot figure out why. After this I will see what the parser errors are.

gsquire:
I have squashed and rebased my changes.

For now I moved back to the old way of reading byte by byte since the buffered reader would just block for transactions. I would love to hear ideas as to why this could have been or what could be done to fix it.

Let me know your thoughts and thanks for your patience with this PR.

PR #130, opened 2017-07-11
Support for Geospatial Commands

ayosec:
Add support for geospatial commands, available since Redis 3.2.0.

## Added commands

* [`GEOADD`](https://redis.io/commands/geoadd)
* [`GEODIST`](https://redis.io/commands/geodist)
* [`GEOHASH`](https://redis.io/commands/geohash)
* [`GEOPOS`](https://redis.io/commands/geopos)
* [`GEORADIUS`](https://redis.io/commands/georadius)
* [`GEORADIUSBYMEMBER`](https://redis.io/commands/georadiusbymember)

## Changes in Travis

The Redis installed by default in Travis is 3.0, so these commands are not available.

I tried to use `ppa:chris-lea/redis-server`, but it still [installs Redis 3.0](https://travis-ci.org/ayosec/redis-rs/jobs/251570164).

To use Redis 3.2.9, it is [downloaded and built from sources](https://github.com/ayosec/redis-rs/blob/21968af4b9b73f11f73d96599701b0f4655e6fbd/.travis.yml#L8-L12). The build time varies between 20 and 50 seconds.

## `common` module for tests

In be773d728475d6ecc6b9f98dcf60a03b2374a077, I moved some code from `tests/test_basic.rs` to a common module, so it can be used in a new `tests/test_geospatial.rs` test. It does not change anything, but add some noise in the diff.

## Example

Full example is available in [`examples/geospatial.rs`](https://github.com/ayosec/redis-rs/blob/e99439fe76c77a377f9b9cf99138f4fd44396714/examples/geospatial.rs).

```rust
// Add some members to the geospatial index.

let added: isize = con.geo_add(
    "gis",
    &[
        (geo::Coord::lon_lat("13.361389", "38.115556"), "Palermo"),
        (geo::Coord::lon_lat("15.087269", "37.502669"), "Catania"),
        (geo::Coord::lon_lat("13.5833332", "37.316667"), "Agrigento"),
    ],
)?;

println!("[geo_add] Added {} members.", added);

// Get the position of one of them.

let position: Vec<geo::Coord<f64>> = con.geo_pos("gis", "Palermo")?;

println!("[geo_pos] Position for Palermo: {:?}", position);

// Search members near (13.5, 37.75)

let options = geo::RadiusOptions::default()
    .order(geo::RadiusOrder::Asc)
    .with_dist()
    .limit(2);

let items: Vec<geo::RadiusSearchResult> =
    con.geo_radius("gis", 13.5, 37.75, 150.0, geo::Unit::Kilometers, options)?;

for item in items {
    println!(
        "[geo_radius] {}, dist = {} Km",
        item.name,
        item.dist.unwrap_or(f64::NAN)
    );
}
```


badboy:
Wow, thanks for the effort. That's quite some code and I'd like to get this in, but I currently don't have all the capacity to review this, so it might take some time.

I rekicked the failed test. It was an unrelated problem it seems.

categulario:
It seems like the names of the new functions are not consistent with the existent ones, for example:

* `geo_add` instead of `geoadd`
* `geo_add` is adding multiple items. Maybe `geoadd_multiple` and `geoadd`?

just to keep things regular

grippy:
@badboy There was a lot of work that looks like it went into this PR and almost 20 months later it's still sitting here. At least 75% of this PR is related to tests, examples, and docs. So the actual code review is pretty small -- everything should be fixable in followup PR's. 

Obviously, this PR has a few merge conflicts now, but what's the hold up on solving those and merging this? I understand it takes time to review and at this point @ayosec might've moved onto other things by now. But sans feedback it's hard for anyone else to pick this up and drive it to completion. 

Anyway, I'm mostly asking because I'm evaluating Redis Streams and noticed there's an issue for adding this set of commands to this lib. Just want to make sure if I start working on this issue that the PR won't sit idle once it's ready.

badboy:
I do appreciate the effort, yet I can't make any promises on any timeline.

I work on redis-rs in my free time, I am not a heavy Redis user anymore and quite honestly have never once used the geo functionality. That's why reviewing this was always rather low on my priority list.

It doesn't matter if 75% of the PR is not the actual code, I will still end up reviewing it (after all, if it lands, people will expect me to maintain it anyway). Especially given the fact that the tests are the way I can verify the code does what it is supposed to.
By now this PR unfortunately bit-rotted and is not mergeable in its current state, even if reviewed. I don't have the time to rebase it & fix against the current code.

I can't make any promises for your work either for similar reasons. If you see the need for that I don't want to discourage you.
Small PRs are always much more welcome than a huge PR that includes everything of course.

grippy:
@badboy Thanks for the reply. I completely understand not wanting to maintain something you don't use anymore -- especially if you don't have the experience with the new commands. Have you considered adding more maintainers or handing this project off to someone else?

There's at least a dozen new commands for Streams (with varying complexity). I don't know how beneficial it would be to add them piecemeal over a series of PR's. However, if it makes reviewing them easier, I'm willing to give it a shot for one command.

badboy:
There's actually more people with commit rights on this repository and I have some docs in the pipeline to explain better how to actually do releases and stuff, so that anyone could easily do that.

Re streams: discussing is better placed in the relevant issue, but IMO it should be possible to implement streams as its own crate on top of this, this way you wouldn't even be blocked on me or other maintainers.

ayosec:
I updated the patch with the current master.

Tests in AppVeyor failed because the Redis server is too old, and doesn't support geospatial commands.

badboy:
Thanks for the updating.

> Tests in AppVeyor failed because the Redis server is too old, and doesn't support geospatial commands.

That means we need to somehow disable the tests on AppVeyor somehow (back then for redis-rb we did version detection for the tests, hmmm...)

Marwes:
I would suggest adding a `test-geospatial` default feature and then disable default features on appveyor and explicitly enable every feature but `test-geospatial`. That way the problem gets limited to appveyor only, without needing to do any version detection.

examples/geospatial.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::process::exit;
+
+fn redis_url() -> String {
+    match env::var("REDIS_URL") {
+        Ok(url) => url,
+        Err(..) => "redis://127.0.0.1/".to_string(),
+    }
+}
+
+fn run() -> RedisResult<()> {
```
badboy: When we add a feature to conditionnaly disable the geo functionality, we also need to stub this out:

```rust
#[cfg(feature = "geo")]
fn run() -> RedisResult<()> {
 ...
}
#[cfg(not(feature = "geo"))]
fn run() -> RedisResult<()> { Ok(()) }
```

src/geo.rs
```diff
@@ -0,0 +1,354 @@
...
+/// * To keep the raw value from Redis, use `String`.
+#[derive(Debug, PartialEq)]
+pub struct Coord<T> {
+    pub longitude: T,
+    pub latitude: T,
+}
+
+impl<T> Coord<T> {
+    /// Create a new Coord with the (longitude, latitude)
+    pub fn lon_lat(longitude: T, latitude: T) -> Coord<T> {
```
badboy: Hm, it's unfortunate that there's no way to restrict it to f64 or f32 without us including another crate or having another trait.

badboy: However, we can ... maybe make the docs stronger implying this is useless with any other `T` and then we can always go the trait way sometime later.

src/geo.rs
```diff
@@ -0,0 +1,354 @@
...
+
+impl<T: FromRedisValue> FromRedisValue for Coord<T> {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let values: Vec<T> = FromRedisValue::from_redis_value(v)?;
+        if values.len() != 2 {
+            invalid_type_error!(v, "Expect a pair of numbers");
+        }
+        let mut values = values.into_iter();
+        Ok(Coord {
+            longitude: values.next().unwrap(),
```
badboy: Unwrap here could panic. We can return an error instead.

src/geo.rs
```diff
@@ -0,0 +1,354 @@
...
+        self
+    }
+
+    /// Sort the returned items
+    pub fn order(mut self, o: RadiusOrder) -> Self {
+        self.order = o;
+        self
+    }
+
+    /// Store the results in a sorted set at key, instead of returning them.
```
badboy:
```suggestion
    /// Store the results in a sorted set at `key`, instead of returning them.
```

src/geo.rs
```diff
@@ -0,0 +1,354 @@
...
+
+    /// Return the distance of the returned items from the specified center.
+    /// The distance is returned in the same unit as the unit specified as the
+    /// radius argument of the command.
+    pub fn with_dist(mut self) -> Self {
+        self.with_dist = true;
+        self
+    }
+
+    /// Return the longitude,latitude coordinates of the matching items.
```
badboy:
```suggestion
    /// Return the `longitude, latitude` coordinates of the matching items.
```

src/geo.rs
```diff
@@ -0,0 +1,354 @@
...
+
+    /// Store the results in a sorted set at key, instead of returning them.
+    ///
+    /// This feature can't be used with any `with_*` method.
+    pub fn store<K: ToRedisArgs>(mut self, key: K) -> Self {
+        self.store = Some(ToRedisArgs::to_redis_args(&key));
+        self
+    }
+
+    /// Similar to `store`, but includes the distance of every member from the
```
badboy: Can we repeat what this function does?

```
Store the results in a sorted key with the distance from the center as its score.
```

This makes the docs independent of those of `store`

src/geo.rs
```diff
@@ -0,0 +1,354 @@
...
+
+    /// Sort the returned items
+    pub fn order(mut self, o: RadiusOrder) -> Self {
+        self.order = o;
+        self
+    }
+
+    /// Store the results in a sorted set at key, instead of returning them.
+    ///
+    /// This feature can't be used with any `with_*` method.
```
badboy: _Maybe_ we can later add some way to verify that at runtime (or statically? but that requires a bit more code, not too much a fan of)

src/geo.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    #[test]
+    fn test_radius_options() {
+        // Without options, should not generate any argument
+        let empty = RadiusOptions::default();
+        assert_eq!(ToRedisArgs::to_redis_args(&empty).len(), 0);
+
+        // Some combinations with WITH* options
+        let opts = || RadiusOptions::default();
```
badboy:
```suggestion
        let opts = RadiusOptions::default;
```

src/commands.rs
```diff
@@ -752,6 +752,195 @@ implement_commands! {
...
+        cmd("GEORADIUS")
+            .arg(key)
+            .arg(longitude)
+            .arg(latitude)
+            .arg(radius)
+            .arg(unit)
+            .arg(options)
+    }
+
+    /// This command is exactly like [`geo_radius`](#method.geo_radius) with the sole
```
badboy: As in another place let's explain this command on its own.

src/commands.rs
```diff
@@ -752,6 +752,195 @@ implement_commands! {
...
+    /// difference that instead of taking, as the center of the area to query, a
```
badboy: This whole "instead of" makes it very hard to understand IMO

examples/geospatial.rs
badboy: Scrap that. We only need to disable the test that is run, not any actual implementation code.

src/commands.rs
ayosec: It is a copy of the [original description](https://redis.io/commands/georadiusbymember).

Do you prefer the [description in Lettuce](https://github.com/lettuce-io/lettuce-core/blob/c98108dd7fbe9bee8702e468efff4b085a976374/src/main/java/io/lettuce/core/api/reactive/RedisGeoReactiveCommands.java#L100-L101)?

>  Retrieve members selected by distance with the center of `member`. The member itself is always contained in the results.

badboy: Sounds better, yes

Cargo.toml
```diff
@@ -23,12 +23,18 @@ tokio-io = "0.1"
 tokio-codec = "0.1"
 tokio-sync = "0.1"
 
+[features]
+default = [ "geospatial" ]
```
badboy: Ah hm, turns out we don't necessarily need to disable the code, but only the test.
What do you think?

src/geo.rs
ayosec: The [documentation of `Coord`](https://github.com/mitsuhiko/redis-rs/blob/3d2b2e053e894bd84689ef5cc26572e14b61c1c6/src/geo.rs#L49-L50) mentions it.

Using strings is also useful if you want to keep the raw value.

Cargo.toml
ayosec: I saw the suggestion in https://github.com/mitsuhiko/redis-rs/pull/130#issuecomment-478644177, but I think that it can be confusing for users of the crate. In other crates that I have seen, *features* are something intended for users of the library, and not an internal detail of the implementation (like configuring the CI).

A `geospatial` feature is easier to recognise by users, and it is still possible to control the CI.

src/geo.rs
badboy: 👍

Cargo.toml
badboy: Makes sense.
I'm fine with that, as long as the feature is enabled by default.

ayosec:
I think that all suggestions are applied. Is there anything else to do?

badboy:
> I think that all suggestions are applied. Is there anything else to do?

A final round of review. Doing that now!

.travis.yml
```diff
@@ -13,11 +13,6 @@ matrix:
     - rust: nightly
       env: RUST_VERSION=nightly
 
-addons:
-  apt:
-    packages:
-      - redis-server
```
badboy: Oh we don't need this anymore at all? Nice!

Cargo.toml
```diff
@@ -23,12 +23,18 @@ tokio-io = "0.1"
...
+
+geospatial = []
```
badboy: We should document this feature, we can do that in a follow-up

badboy:
Merged and [added to the changelog](https://github.com/mitsuhiko/redis-rs/commit/1a6bf195ed21b84354b9237f720ab11967cb7ff9)

PR #132, opened 2017-08-29
Fix unix socket path parsing

### LINKED ISSUE #113: unix socket path parsing is broken
`Url::to_file_path()` is used here http://mitsuhiko.github.io/redis-rs/src/redis/src/connection.rs.html#125. 
This is incorrect because `to_file_path` only works for `file` scheme urls - see https://docs.rs/url/1.3.0/url/struct.Url.html#method.to_file_path.

Instead, `Url::path()` should be used.

jwilm:
`to_file_path()` only work works for file:// URLs.

resolves #113

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -123,9 +123,7 @@ fn url_to_tcp_connection_info(url: url::Url) -> RedisResult<ConnectionInfo> {
 #[cfg(any(feature="with-unix-sockets", feature="with-system-unix-sockets"))]
 fn url_to_unix_connection_info(url: url::Url) -> RedisResult<ConnectionInfo> {
     Ok(ConnectionInfo {
-        addr: Box::new(ConnectionAddr::Unix(unwrap_or!(url.to_file_path().ok(),
-                                                       fail!((ErrorKind::InvalidClientConfig,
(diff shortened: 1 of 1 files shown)


minus7:
The documentation shows that you can pass it URLs in the format `redis+unix:///[:<passwd>@]<path>[?db=<db>]` which does not quite work. It seems to be `redis+unix://[:<passwd>@]<path>[?db=<db>]` where path must be absolute/starting with a slash, or `redis+unix:<path>[?db=<db>]` where path may be relative as well (but password doesn't work). The correct format should probably be `redis+unix://[:<passwd>@]/<path>[?db=<db>]` (double slash to mark the start; optionally password; a slash for separating host (always empty) and path; path, starting with an extra slash if absolute), since that allows using a password on sockets with relative paths too. The `url` crate does not support that though. It'd be nice if you could adjust [the documentation](https://github.com/mitsuhiko/redis-rs/blob/01d13e818ac46496791e6315cd03904513d0828a/src/lib.rs#L80-L85) accordingly though.

mark-i-m:
I wasn't able to get either format working, and I ended up using `unix:/tmp/redis.sock` instead.

PR #133, opened 2017-10-04
Make FromRedisValue impl for HashMap/HashSet generic over hash builder

sd2k:
This allows use of (for example) the FnvHashMap/FnvHashSet provided by [fnv](https://crates.io/crates/fnv) when getting maps/sets back from Redis.

Let me know if you'd like me to squash!


badboy:
Oof, that was laying around for long. Sorry about that, merged as https://github.com/mitsuhiko/redis-rs/commit/13932a2d248e6bec8043a5b84797e0bb084c1aad

PR #136, opened 2017-10-24
Implement flexible PubSub API

### LINKED ISSUE #135: Scoped PubSub API
Currently, pubsub support is only available only by creating a client and then asking for a `PubSub` API. This neatly avoids the issue that only certain commands are available once calling `subscribe` or `psubscribe`. However, it's very unfriendly to connection poolers such as r2d2.

I'd like to propose adding a `PubSubCommands` trait which will make pub/sub usable from a borrowed connection.

```rust
/// Allows pubsub callbacks to stop receiving messages.
///
/// Arbitraty data may be returned from `Break`.
enum ControlFlow<U> {
    Continue,
    Break(U),
}

/// The PubSub trait allows subscribing to one or more items
/// and receiving a callback whenever a message arrives.
///
/// The trait is generic so that it may be implemented for multiple
/// types such as `&str` and `&[u8]`. The `self` argument is a mutable
/// reference to prevent the implementing type from being used
/// while subscribed to messages.
///
/// Each method handles subscribing to the list of keys, waiting for
/// messages, and unsubscribing from the same list of keys once
/// a ControlFlow::Break is encountered.
///
/// Once (p)subscribe returns Ok(U), the connection is again safe to use
/// for calling other methods.
trait PubSub<T> {
    fn subscribe<F, U>(&mut self, _: &[T], _: F) -> Result<U>
        where F: FnMut(Msg) -> ControlFlow<U>;
        
    fn psubscribe<F, U>(&mut self, _: &[T], _: F) -> Result<U>
        where F: FnMut(Msg) -> ControlFlow<U>;
}
```

Another design I'm considering is to have an API which borrows a `&mut Connection(Like)` and returns a new type which only has pubsub commands. Again, this supports connection pooling but would be a lot more manual than the above proposition. It is, however, much closer to the current implementation.

For example,

```rust
impl Connection {
    /// Get a PubSub API for this connection
    ///
    /// Lifetimes are included to clearly show ownership
    fn as_pubsub<'a>(&'a mut self) -> PubSubRef<'a>;
}
```

`PubSubRef` would then basically be today's `PubSub` type but using a borrowed connection. Again, using a mutable borrow prevents the connection from having disallowed commands called on it while a subscription is active.

Both of these approaches can coexist! The scoped `subscribe` and `psubscribe` can be provided for convenience while also exposing the ownership based approach. In fact, the scoped convenience methods would probably be implemented using a `PubSubRef`.

I wanted to get some feedback about this before going too far down an implementation. Now that I've gone through the design, this actually doesn't seem like *too* much work, so I may just put something together and see how it feels.

jwilm:
Originally, pubsub support as only available only by creating a client
and then asking for a pubsub API. The original `PubSub` type owned a
connection. This neatly avoided the issue that only certain commands are
available once calling subscribe or psubscribe; however, it's very
unfriendly to connection poolers such as r2d2

This commit makes it so that a borrowed connection may be used for
PubSub. Available commands are limited while in pubsub mode by borring
the connection mutably into a type that only exposes the allowed
commands. This API may be accessed by calling `as_pubsub` on a
`Connection`. To go back to being able to use *any* command, this
wrapper type may be dropped, and the connection should become usable
again for other purposes.

Additionally, convenience methods are added to `Connection`: `subscribe`
and `psubscribe`. They accept a `ToRedisArgs` and a closure which will
be called every time a message is received. The subscription status can
be managed via the return value of the closure which is either
`ControlFlow::Continue` or `ControlFlow::Break`.

This is a breaking change; the original PubSub type was removed in favor
of this more flexible API and to avoid code bloat by having multiple
implementations or additional abstractions to allow an implementation to
be shared.

Resolves #135.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -2,7 +2,7 @@
 #![cfg_attr(rustfmt, rustfmt_skip)]
 use types::{FromRedisValue, ToRedisArgs, RedisResult, NumericBehavior};
 use client::Client;
-use connection::{Connection, ConnectionLike};
+use connection::{Connection, ConnectionLike, Msg};
 use cmd::{cmd, Cmd, Pipeline, Iter};
 
 
@@ -734,8 +734,99 @@ implement_commands! {
     }
 }
 
+/// Allows pubsub callbacks to stop receiving messages.
+///
+/// Arbitraty data may be returned from `Break`.
+pub enum ControlFlow<U> {
+    Continue,
+    Break(U),
+}
+
+/// The PubSub trait allows subscribing to one or more items
+/// and receiving a callback whenever a message arrives.
+///
+/// Each method handles subscribing to the list of keys, waiting for
+/// messages, and unsubscribing from the same list of keys once
+/// a ControlFlow::Break is encountered.
+///
+/// Once (p)subscribe returns Ok(U), the connection is again safe to use
+/// for calling other methods.
+///
+/// # Examples
+///
+/// ```rust,no_run
+/// # fn do_something() -> redis::RedisResult<()> {
+/// use redis::{PubSubCommands, ControlFlow};
+/// let client = redis::Client::open("redis://127.0.0.1/")?;
+/// let mut con = client.get_connection()?;
+/// let mut count = 0;
+/// con.subscribe(&["foo"], |msg| {
+///     // do something with message
+///     assert_eq!(msg.get_channel(), Ok(String::from("foo")));
+///
+///     // incerement messages seen counter
+///     count += 1;
+///     match count {
+///         // stop after receiving 10 messages
+///         10 => ControlFlow::Break(()),
+///         _ => ControlFlow::Continue,
+///     }
+/// });
+/// # Ok(()) }
+/// ```
+//
+// In the future, it would be nice to implement Try such that `?` will work
+// within the closure.
+pub trait PubSubCommands {
+    /// Subscribe to a list of channels using SUBSCRIBE and run the provided
+    /// closure for each message received.
+    fn subscribe<'a, C, F, U>(&mut self, _: C, _: F) -> RedisResult<U>
+        where F: FnMut(Msg) -> ControlFlow<U>,
+              C: ToRedisArgs;
+
+    /// Subscribe to a list of channels using PSUBSCRIBE and run the provided
+    /// closure for each message received.
+    fn psubscribe<'a, P, F, U>(&mut self, _: P, _: F) -> RedisResult<U>
+        where F: FnMut(Msg) -> ControlFlow<U>,
+              P: ToRedisArgs;
+}
+
 impl Commands for Connection {}
 impl Commands for Client {}
+
+impl PubSubCommands for Connection {
+    fn subscribe<'a, C, F, U>(&mut self, channels: C, mut func: F) -> RedisResult<U>
+        where F: FnMut(Msg) -> ControlFlow<U>,
+              C: ToRedisArgs
+    {
+        let mut pubsub = self.as_pubsub();
+        pubsub.subscribe(channels)?;
+
+        loop {
+            let msg = pubsub.get_message()?;
+            match func(msg) {
+                ControlFlow::Continue => continue,
+                ControlFlow::Break(value) => return Ok(value),
+            }
+        }
+    }
+
+    fn psubscribe<'a, P, F, U>(&mut self, patterns: P, mut func: F) -> RedisResult<U>
+        where F: FnMut(Msg) -> ControlFlow<U>,
+              P: ToRedisArgs
+    {
+        let mut pubsub = self.as_pubsub();
+        pubsub.psubscribe(patterns)?;
+
+        loop {
+            let msg = pubsub.get_message()?;
+            match func(msg) {
+                ControlFlow::Continue => continue,
+                ControlFlow::Break(value) => return Ok(value),
+            }
+        }
+    }
+}
 
 impl PipelineCommands for Pipeline {
     fn perform(&mut self, cmd: &Cmd) -> &mut Pipeline {
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -777,14 +777,14 @@ pub enum ControlFlow<U> {
 //
 // In the future, it would be nice to implement Try such that `?` will work
 // within the closure.
-pub trait PubSubCommands {
+pub trait PubSubCommands: Sized {
     /// Subscribe to a list of channels using SUBSCRIBE and run the provided
     /// closure for each message received.
     ///
     /// For every `Msg` passed to the provided closure, either
     /// `ControlFlow::Break` or `ControlFlow::Continue` must be returned. This
     /// method will not return until `ControlFlow::Break` is observed.
-    fn subscribe<'a, C, F, U>(&mut self, _: C, _: F) -> RedisResult<U>
+    fn subscribe<'a, C, F, U>(self, _: C, _: F) -> RedisResult<(U, Self)>
         where F: FnMut(Msg) -> ControlFlow<U>,
               C: ToRedisArgs;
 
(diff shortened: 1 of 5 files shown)


jwilm:
Test failures were "Connection reset by peer" errors. Maybe just needs to be rerun?

dbettin:
Is this close to being merged? If not, anything specifically blocking it?

src/commands.rs
```diff
@@ -734,9 +734,100 @@ implement_commands! {
...
+/// });
+/// # Ok(()) }
+/// ```
+//
+// In the future, it would be nice to implement Try such that `?` will work
+// within the closure.
+pub trait PubSubCommands {
+    /// Subscribe to a list of channels using SUBSCRIBE and run the provided
+    /// closure for each message received.
+    fn subscribe<'a, C, F, U>(&mut self, _: C, _: F) -> RedisResult<U>
```
badboy: It might be good to explicitly state here that this is blocking until `Break` is returned. What do you think?

src/commands.rs
```diff
@@ -734,9 +734,100 @@ implement_commands! {
...
+        where F: FnMut(Msg) -> ControlFlow<U>,
+              C: ToRedisArgs;
+
+    /// Subscribe to a list of channels using PSUBSCRIBE and run the provided
+    /// closure for each message received.
```
badboy: As above.

src/connection.rs
```diff
@@ -512,6 +492,13 @@ impl PubSub {
     }
 }
 
+impl<'a> Drop for PubSub<'a> {
+    fn drop(&mut self) {
+        let _ = cmd("UNSUBSCRIBE").query::<()>(self.con);
+        let _ = cmd("PUNSUBSCRIBE").query::<()>(self.con);
```
badboy: Unsubscribing will trigger a message per channel that's unsubscribed.
Will this cause problems for the underlying connection if it is later reused?
Having a test for this would be great.

src/connection.rs
```diff
@@ -427,44 +423,28 @@ impl ConnectionLike for Connection {
...
-        chan
-    }
-
+impl<'a> PubSub<'a> {
     /// Subscribes to a new channel.
     pub fn subscribe<T: ToRedisArgs>(&mut self, channel: T) -> RedisResult<()> {
-        let chan = self.get_channel(&channel);
-        let _: () = try!(cmd("SUBSCRIBE").arg(&*chan).query(&self.con));
-        self.channels.insert(chan);
+        let _: () = try!(cmd("SUBSCRIBE").arg(channel).query(self.con));
```
badboy: While we're at it, can we switch to the postfix question mark operator here (and below)?

src/commands.rs
jwilm: Good idea, I'll add it.

src/connection.rs
jwilm: Yep, happy to do so. I believe I went with `try!` to be consistent with the rest of the code base.

src/connection.rs
jwilm: It shouldn't because `query` will already expect to receive a result from the server and parse it. I agree that we should have a test for this including that the connection is no longer subscribed to anything.

jwilm: As it turns out, a message per channel actually means a distinct message per channel at a protocol level. This will take a bit of work to support. Nice catch!

jwilm:
@badboy I've addressed the issues you pointed to in your review. Please see the description on the final commit for details.

One point worth reiterating is that the updated PubSub API now consumes the connection, and it's only returned upon a successful call to `PubSub::into_inner`. There are two drawbacks to this:

1. Connections from a connection pooler won't be able to utilize the PubSub API
2. A failed call to `PubSub::into_inner` will cause the connection to be dropped.

On the flip side, this allows us to guarantee that the connection will be in a usable state when going in and out of PubSub state.

jwilm:
Reflecting a bit on the updated implementation, the ownership requirements imposed are too extreme, and they are contrary to the goals set out at the start of this PR. However, I have another idea about how to implement this.

Instead of having PubSub take ownership of connection, we can return to taking a mutable reference. The connection would itself gain a flag that indicates whether it is in the pubsub state. When entering this state, the flag would be raised. On an unsuccessful exit from this state, the flag would remain. Subsequent commands could check this flag and attempt to exit the pubsub state, or else return an appropriate error.

This solution seems to be the best of both worlds -- good ergonomics, connection pooling support, and pubsub constraints can be enforced while making it clear that a connection is not usable by returning an appropriate error on subsequent usage if exiting the pubsub state fails.

What do you think?

badboy:
@jwilm I haven't yet fully grasped your new proposal, but wanted to throw in a comment first:
Is there a problem to use your first implementation and consume all unsubscribe-messages on drop? (I really liked your first implementation and think it's quite ergonomic)

jwilm:
@badboy the initial implementation *could* leave the connection in an unusable state if it encounters an error while running the unsubscribe process. I made the changes proposed in my previous comment. It has the same API as the original proposal, but it ensures that we don't accidentally miss an error in the Drop impl and leave the connection in an unknown state.

badboy:
I will look into this soon.

jwilm:
No rush, take your time.

src/connection.rs
```diff
@@ -427,44 +499,32 @@ impl ConnectionLike for Connection {
...
 /// ```
-impl PubSub {
-    fn get_channel<T: ToRedisArgs>(&mut self, channel: &T) -> Vec<u8> {
-        let mut chan = vec![];
-        for item in channel.to_redis_args().iter() {
-            chan.extend(item.iter().cloned());
-        }
-        chan
+impl<'a> PubSub<'a> {
+    pub fn new(con: &'a mut Connection) -> Self {
```
badboy: Do we need to expose this constructor? I'd favor to only have `as_pubsub` on the connection.

src/connection.rs
```diff
@@ -375,10 +371,82 @@ impl Connection {
...
+        // cancelled *and* all unsubscribe messages were received.
+        Ok(())
+    }
 }
 
 impl ConnectionLike for Connection {
     fn req_packed_command(&self, cmd: &[u8]) -> RedisResult<Value> {
+        if self.pubsub.get() {
+            self.exit_pubsub()?;
+        }
```
badboy: This now forces an additional check for each command sent, right? I guess the performance implications are still negligible?

src/connection.rs
jwilm: Nope, this was a mistake. Nice catch.

src/commands.rs
```diff
@@ -734,9 +734,108 @@ implement_commands! {
...
+///     match count {
+///         // stop after receiving 10 messages
+///         10 => ControlFlow::Break(()),
+///         _ => ControlFlow::Continue,
+///     }
+/// });
+/// # Ok(()) }
+/// ```
+//
+// In the future, it would be nice to implement Try such that `?` will work
```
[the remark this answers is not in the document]
jwilm: I did not, but from what I understand about the Try trait, it should be possible.

src/connection.rs
jwilm: It is an additional check, but I'm guessing the branch predictor will behave favorably here. In our benchmarking of this library at OneSignal, the parser ends up being a much bigger bottleneck than anything else (we have at least 1 PR forthcoming in that regard).

jwilm:
Feedback addressed; can you take another look?

jwilm:
Note that I replied to one or two of the comments you left which have now been folded.

src/connection.rs
badboy: Looking forward to that PR.

jwilm: There's [a PR](https://github.com/mitsuhiko/redis-rs/pull/151) up for it as of a minute ago.

jwilm:
@badboy is there anything blocking this PR now?

badboy:
Just me hitting the big green button. Will do so today.

PR #146, opened 2018-04-10
Update dependencies and bump version

Eijebong:


Cargo.toml
```diff
@@ -22,13 +22,13 @@ with-unix-sockets = ["unix_socket"]
 with-system-unix-sockets = []
 
 [dependencies]
-sha1 = "0.2.0"
+sha1 = "0.6.0"
```
jwilm: If these updates don't require code changes, it would be good to accept a range of versions rather than potentially forcing dependent crates update rand/sha1.

Eijebong: Since those aren't public dependencies, I don't see the point of having a range there.

jwilm: In this case it could prevent multiple instances of the libs from being compiled which takes time and also bloats binaries. It's not a big deal, but it's certainly nice to have.

Eijebong: Woups, this fell out my radar, sorry for the delay

PR #148, opened 2018-04-25
Remove the need for RefCell by requiring `&mut ConnectionLike`

Marwes:
There isn't anything in redis-rs that requires the `RefCell` so users can always replicate the earlier behavior by themselves wrapping the connection in a `RefCell`.

This might also protect against programmer errors when using `iter` to query as one should not issue another `query` while still using the iterator.

BREAKING CHANGE

Changes the signature of several functions that took a `&ConnectionLike` to take `&mut ConnectionLike`

### DIFF
--- a/benches/bench_basic.rs
+++ b/benches/bench_basic.rs
@@ -22,13 +22,13 @@ fn get_client() -> redis::Client {
 
 fn bench_simple_getsetdel(b: &mut Bencher) {
     let client = get_client();
-    let con = client.get_connection().unwrap();
+    let mut con = client.get_connection().unwrap();
 
     b.iter(|| {
         let key = "test_key";
-        redis::cmd("SET").arg(key).arg(42).execute(&con);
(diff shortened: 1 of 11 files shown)


badboy:
I have a rebased version of this [in the `mut_connection`](https://github.com/mitsuhiko/redis-rs/commit/53297239010c99c8ae34f0fb3b7692996a96f827) branch and I'm trying it out on various reverse dependencies to see its impact.

Marwes:
@badboy I rebased this using your `mut_connection` branch and fixed some additional compile errors in case you want to move forward with this.

badboy:
I'll spend some time on this tomorrow and see if we go ahead with it.

badboy:
Rebased and merged as 531415e9f17b618abfcf96133a0fbd5f21c8705a.
Changelog: 03a9611a3feaebe31fc7e2522cdca5ec87dba1e9

PR #147, opened 2018-04-24
chore: Cache cargo on travis

Marwes:
Should speedup the travis builds a bit.

PR #149, opened 2018-04-27
fix: Use write_all to guarantee that all bytes gets written 

Marwes:
Based on #147 to avoid the test flakiness.

Closes #147


badboy:
Let's see how github deals with all these duplicated commits.

PR #151, opened 2018-05-15
Increase performance

coder543:
These are some small changes that should improve performance, more details are provided in the commit messages.


jwilm:
@badboy we were hoping to see this pulled into the mainline -- why close?

PR #153, opened 2018-05-30
Fix handling of passwords with special characters

CoRfr:
The url module automatically `percent_encode` the password, so we need to `percent_decode` before using it or character such as `=` are left as `%3D`.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -115,7 +115,15 @@ fn url_to_tcp_connection_info(url: url::Url) -> RedisResult<ConnectionInfo> {
                            fail!((ErrorKind::InvalidClientConfig, "Invalid database number")))
             }
         },
-        passwd: url.password().and_then(|pw| Some(pw.to_string())),
+        passwd: {
+            match url.password() {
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -115,7 +115,11 @@ fn url_to_tcp_connection_info(url: url::Url) -> RedisResult<ConnectionInfo> {
                            fail!((ErrorKind::InvalidClientConfig, "Invalid database number")))
             }
         },
-        passwd: url.password().and_then(|pw| Some(pw.to_string())),
+        passwd: {
+            let passwd_encoded = url.password().unwrap();
+            let passwd = url::percent_encoding::percent_decode(passwd_encoded.as_bytes()).decode_utf8().unwrap();
+            Some(passwd.to_string())
+        }
```
CoRfr: This is the first time that I wrote in rust.
Although this code seems to work (I'm using it through sccache), I have basically no idea what I'm doing, so any feedback is very much welcome :)

badboy:
Closed by #163

PR #165, opened 2018-09-07
perf: Optimize the command encoding

Marwes:
Not as important as `IO` latency but can still be a decent throughput improvement.

(Lots of variance in the benchmarks but there is still some clear improvements)
```
encode/pipeline         time:   [727.07 us 731.89 us 737.21 us]
                        change: [-59.635% -58.716% -57.668%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
encode/pipeline_nested  time:   [246.61 us 249.05 us 251.81 us]
                        change: [-71.597% -70.950% -70.333%] (p = 0.00 < 0.05)
                        Performance has improved.
encode/integer          time:   [693.79 us 698.95 us 703.98 us]
                        change: [-63.633% -63.125% -62.651%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
encode/small            time:   [1.1458 us 1.1580 us 1.1701 us]
                        change: [-40.826% -39.741% -38.664%] (p = 0.00 < 0.05)
                        Performance has improved.
```

I could probably refactor to remove these breaking changes but for the moment...

BREAKING CHANGE

`ToRedisArgs` has been changed to take take an instance of `RedisWrite`  instead of `Vec<Vec<u8>>`. Use the `write_arg` method instead of `Vec::push`.

`impl Iterator` is used which requires 1.26

`iter` now takes `self` by value instead of cloning `self` inside the method.

### DIFF
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -1,21 +1,23 @@
 use connection::ConnectionLike;
 use types::{
-    from_redis_value, ErrorKind, FromRedisValue, RedisFuture, RedisResult, ToRedisArgs, Value,
+    from_redis_value, ErrorKind, FromRedisValue, RedisFuture, RedisResult, RedisWrite, ToRedisArgs,
+    Value,
 };
 
 use futures::Future;
 
 #[derive(Clone)]
-enum Arg<'a> {
-    Simple(Vec<u8>),
+enum Arg<D> {
+    Simple(D),
     Cursor,
-    Borrowed(&'a [u8]),
 }
 
 /// Represents redis commands.
 #[derive(Clone)]
 pub struct Cmd {
-    args: Vec<Arg<'static>>,
+    data: Vec<u8>,
+    // Arg::Simple contains the offset that marks the end of the argument
+    args: Vec<Arg<usize>>,
     cursor: Option<u64>,
     is_ignored: bool,
 }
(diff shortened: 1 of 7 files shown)


Marwes:
cc #81

Marwes:
If #167 is merged a breaking release is needed anyways so perhaps it is fine to use `impl Iterator` then as well.

Marwes:
ping @badboy

badboy:
I won't get to review it this week

Marwes:
I rebased and squashed this since there was some files accidentally included and to fix the conflicts.

badboy:
> I rebased and squashed this since there was some files accidentally included and to fix the conflicts.

I was just about to ask :D

badboy:
Test failures due to formatting (because I merged #167). I'm fine with merging as is and fixing afterwards.

src/cmd.rs
```diff
@@ -214,7 +263,8 @@ impl Cmd {
     /// # let client = redis::Client::open("redis://127.0.0.1/").unwrap();
     /// # let con = client.get_connection().unwrap();
     /// let mut cmd = redis::cmd("SSCAN");
-    /// let mut iter : redis::Iter<isize> = cmd.arg("my_set").cursor_arg(0).iter(&con).unwrap();
+    /// let mut iter : redis::Iter<isize> =
+    ///     cmd.arg("my_set").cursor_arg(0).clone().iter(&con).unwrap();
```
badboy: Hm, is this clone here now required?
That _could_ break current users.

Marwes: `iter` used to `clone` the command inside the method call before so I tweaked it to be more honest by making it take `self` by value instead. It is a bit more verbose but it is the only real way to remove that unnecessary overhead.

Marwes: (That is, users could write the command construction to avoid the clone by moving the command instead of cloning, but that is a bit more verbose)

badboy: Hm, this only breaks when using `.iter()` then? Small break, but nothing too serious.

I'd still cut a release now and move that into the release afterwards and include other breaking changes there.

Marwes: There are other breaking changes in this PR as well though, so even if I revert the need for `clone` it would still need a breaking semver bump.

badboy: And I think I'm fine with that for the next release. We can accumulate a few of the breaking changes and provide proper docs on what changed and how to use the new way.

badboy:
Ah damn, of course this has merge conflicts now and thus will need a rebase (and resolving those conflicts).

But otherwise I would be fine with merging it now.

Marwes:
Rebased!

PR #167, opened 2018-09-12
perf(async): Use impl Future instead of boxing where possible

Marwes:
BREAKING CHANGE

The crate now requires rust 1.26


Marwes:
ping @badboy

badboy:
I won't get to review it this week

PR #176, opened 2018-12-30
Allow a more efficient pattern of Pipeline memory allocation

tsnoam:
This change allows a performance sensitive tight loop to reuse a
Pipeline object instead of allocating a new one each time.

 - `with_capacity` allows to preallocate the internal `commands` vector
   to the specified capacity.
 - `Pipeline.query_clear()` clears the internal `commands` vector from
   values after the query (but keeps the allocated capacity).
 - The unitests show the expected behaviour of `query` vs `query_clear`.

### DIFF
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -502,7 +502,7 @@ impl Pipeline {
         )
     }
 
-    /// Clear the internal data strcture of `Pipeline`, allowing it to be reused.
+    /// Clear the internal data structure of `Pipeline`, allowing it to be reused.
     #[inline]
     pub fn clear(&mut self) {
         self.commands.clear();
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -355,8 +355,14 @@ impl Pipeline {
     /// Creates an empty pipeline.  For consistency with the `cmd`
     /// api a `pipe` function is provided as alias.
     pub fn new() -> Pipeline {
+        Self::with_capacity(0)
+    }
+
+    /// Creates an empty pipeline with pre-allocated capacity.  For consistency with the `cmd`
+    /// api a `pipe` function is provided as alias.
+    pub fn with_capacity(capacity: usize) -> Pipeline {
         Pipeline {
-            commands: vec![],
+            commands: Vec::with_capacity(capacity),
             transaction_mode: false,
         }
     }
@@ -491,6 +497,15 @@ impl Pipeline {
                 try!(self.execute_pipelined(con))
             }),
         )
+    }
+
+    /// Like ``query()`` but clears the pooled commands from memory.  This allows reusal of a Pipeline
+    /// object with minimal re-allocation of memory.
+    #[inline]
+    pub fn query_clear<T: FromRedisValue>(&mut self, con: &ConnectionLike) -> RedisResult<T> {
+        let res = self.query(con);
+        self.commands.clear();
+        res
     }
 
     fn execute_pipelined_async<C>(self, con: C) -> RedisFuture<(C, Value)>
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -358,8 +358,7 @@ impl Pipeline {
         Self::with_capacity(0)
     }
 
-    /// Creates an empty pipeline with pre-allocated capacity.  For consistency with the `cmd`
-    /// api a `pipe` function is provided as alias.
+    /// Creates an empty pipeline with pre-allocated capacity.
     pub fn with_capacity(capacity: usize) -> Pipeline {
         Pipeline {
             commands: Vec::with_capacity(capacity),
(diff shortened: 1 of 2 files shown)


badboy:
IMO the fact that we don't clear the internal command vector is a bug and we should just do that for query. That removes the need for any `query_clear` function.

badboy:
That will require `&mut self` on `query` and `execute` on `Pipeline`, but I think it's worthwhile (but should also be documented clearly)

Marwes:
Having `query` clear the pipeline seems like really confusing behaviour to me. Just having a separate `clear` method on the pipeline seems more straightforward.

badboy:
@Marwes Why is that confusing behavior? What use-case do you think of to re-use the filled pipeline?

Marwes:
> Why is that confusing behavior? 

Query then does more than just querying which makes it confusing if I don't care about reuse.

If I am interested in clearing the pipeline for later reuse I am not going to expect that this is done automatically either so I will need to look up how to do that. A separate `clear` method would be more discoverable in that case.


> What use-case do you think of to re-use the filled pipeline?


A pipeline might be pre-created and used multiple for a commonly used query (incrementing/querying a global counter). Even disregarding that, changing the API to take `&mut` would prevent usage when only a `&` is available (which does not necessarily mean that the pipeline is used more than once, it might just be a part of a another object that is shared).

tsnoam:
@badboy 
I'd be happy to change my PR according to your decision. So what would it be?

badboy:
sorry for the silence for so long, this got a bit buried.

I am fine with not changing `query`, but introducing a `clear` method instead. In addition documenting that `query` doesn't clear would be needed.

tsnoam:
@badboy 
I've changed the code as requested.

src/cmd.rs
```diff
@@ -355,8 +355,14 @@ impl Pipeline {
     /// Creates an empty pipeline.  For consistency with the `cmd`
     /// api a `pipe` function is provided as alias.
     pub fn new() -> Pipeline {
+        Self::with_capacity(0)
+    }
+
+    /// Creates an empty pipeline with pre-allocated capacity.  For consistency with the `cmd`
+    /// api a `pipe` function is provided as alias.
```
badboy: The second part here is not true. The `pipe` function is only equivalent to `new()` above.

src/cmd.rs
```diff
@@ -480,6 +486,9 @@ impl Pipeline {
     ///     .cmd("GET").arg("key_1")
     ///     .cmd("GET").arg("key_2").query(&con).unwrap();
     /// ```
+    ///
+    /// NOTE: In order to clear (and reuse) the `Pipeline` object after calling `query()`, one must
```
badboy: I think the `reuse` here is slightly misleading.
Without clear the pipeline can be reused with the existing commands in it, `clear` gives you a clean pipeline without losing the allocation.

Maybe something like:

> NOTE: In order to clear and reuse an empty `Pipeline` object ... 

What do you think?

src/cmd.rs
```diff
@@ -493,6 +502,12 @@ impl Pipeline {
         )
     }
 
+    /// Clear the internal data strcture of `Pipeline`, allowing it to be reused.
```
badboy: Same here as before: make it more clear that it's empty afterwards.

src/cmd.rs
```diff
@@ -554,6 +569,9 @@ impl Pipeline {
     /// # let con = client.get_connection().unwrap();
     /// let _ : () = redis::pipe().cmd("PING").query(&con).unwrap();
     /// ```
+    ///
+    /// NOTE: In order to clear (and reuse) the `Pipeline` object after calling `execute()`, one
+    ///       must explicitly call `clear()`.
```
badboy: As above.

src/cmd.rs
tsnoam: @badboy 
Would you be kind enough to suggest a phrasing for this?

src/cmd.rs
tsnoam: @badboy 
as above... :)

src/cmd.rs
tsnoam: oh... i see you did. didn't scroll up enough...

src/cmd.rs
tsnoam: oopss. my bad. will fix.

src/cmd.rs
tsnoam: What about:

> NOTE: A `Pipeline` object may be reused after `query()` with all the commands as were inserted to them. In order to clear a `Pipeline` object with minimal memory released/allocated, one should call the `clear()`method before inserting new commands.

src/cmd.rs
tsnoam: What about:

> Clear a `Pipeline` object internal data structure. This allows reusing a `Pipeline` object as a clear object but with a minimal amount of memory released/allocated.

src/cmd.rs
tsnoam: Probably "function" is better than "method".

tsnoam:
@badboy 
Please see my comments on the review.

badboy:
I don't see any changes since my last review comments.

tsnoam:
@badboy
I replied on your comments (started a discussion), please take a look there.

tsnoam:
@badboy 
I've pushed a commit with the documentation changes as I suggested them on the comments.
I hope you'll find it appropriate.

badboy:
👍

tsnoam:
@badboy 
Not to be blunt or anything, but when would it be possible to push a new release (with this merged PR) to crates.io?

badboy:
I plan to have one more round of going through the open PRs and then cut a release, so probably tonight or tomorrow.

PR #181, opened 2019-02-05
style: Run cargo fmt

Marwes:
The formatting should be (more or less) stable now so it would be nice
to have the project updated to it.

### DIFF
--- a/tests/test_types.rs
+++ b/tests/test_types.rs
@@ -27,7 +27,8 @@ fn test_info_dict() {
 
     let d: InfoDict = FromRedisValue::from_redis_value(&Value::Status(
         "# this is a comment\nkey1:foo\nkey2:42\n".into(),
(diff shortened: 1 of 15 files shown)


badboy:
I'm definitely a fan of doing that (and then also enforce it via CI), but I'd like to hold off of this until #176 is merged to avoid churn.
Unfortunately I'm a bit busy this week running an event, but I'll try to give that one a final review to get it in.

tests/test_types.rs
```diff
@@ -157,7 +162,8 @@ fn test_types_to_redis_args() {
             .cloned()
             .collect::<HashSet<_>>()
             .to_redis_args()
-            .len() > 0
+            .len()
+            > 0
```
badboy: Yikes, but oh well.

PR #182, opened 2019-02-05
refactor(async): Simplify the implicit pipelining

Marwes:
Modeling it as a `Sink` instead lets us use the `forward` combinator
which removes a big chunk of complicated code.

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -315,6 +315,8 @@ where
         loop {
             let item = match self.sink_stream.poll() {
                 Ok(Async::Ready(Some(item))) => Ok(item),
+                // The redis response stream is not going to produce any more items so we `Err`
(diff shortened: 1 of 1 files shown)


src/async.rs
```diff
@@ -356,7 +315,7 @@ where
         loop {
             let item = match self.sink_stream.poll() {
                 Ok(Async::Ready(Some(item))) => Ok(item),
-                Ok(Async::Ready(None)) => return Ok(Async::Ready(())),
+                Ok(Async::Ready(None)) => return Err(()),
```
badboy: I'm still not really familiar with futures.
Why is `Ready(None)` turned into an error here?

Marwes: Yeah, I should have explained that. Added a comment.

badboy: 👍

PR #186, opened 2019-02-17
Remove try

badboy:


badboy:
I messed up the travis config previously. Fixed now, once it's green I'll merge

PR #189, opened 2019-03-01
refactor: Rename the async module to aio

Marwes:
`async` is a reserved keyword in rust 2018 so this avoids the need to
write `r#async` in it.

Suggested by @jwilm in https://github.com/mitsuhiko/redis-rs/issues/180

Closes https://github.com/mitsuhiko/redis-rs/issues/180

BREAKING CHANGE

Rename uses of the `async` module to `aio`


badboy:
I wonder if we can re-export under the old name so it's ... less breaking?
Or is that just unnecessary duplication and given that this is all experimental anyway, it's fine with breaking it now?

Marwes:
There are some other breaking changes in the queue such as https://github.com/mitsuhiko/redis-rs/pull/165 so I think it may be fair to bundle this in with that?

badboy:
Bundle it is!

badboy:
Merged as 0913426ee380110b0fc93868c50140d98ac46054 (after a rebase)

PR #193, opened 2019-03-09
Expose ValueFuture struct as public API

frol:
Currently, `parse_async` returns `ValueFuture`, which is private, so this makes `parse_async` completely useless as far as I can see. Thus, either `ValueFuture` should be public or `parse_async` should be private, shouldn't they?

I use the parser to play around with Redis protocol, so it is quite useful to have `ValueFuture`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,7 +389,7 @@ pub use connection::{
     parse_redis_url, transaction, Connection, ConnectionAddr, ConnectionInfo, ConnectionLike,
     IntoConnectionInfo, Msg, PubSub,
 };
-pub use parser::{parse_async, parse_redis_value, Parser};
+pub use parser::{parse_async, parse_redis_value, Parser, ValueFuture};
 pub use script::{Script, ScriptInvocation};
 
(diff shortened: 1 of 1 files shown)


Marwes:
You can still use use the function though, as long as you don't try to name the returned future (similarily to `impl Future`). I would consider `ValueFuture` an implementation detail that could be replaced with `impl Future` if necessary so exposing may not be a good idea.

frol:
I have already found a standalone RESP (REdis Serialization Protocol) parser [redis-protocol.rs](https://github.com/aembke/redis-protocol.rs/), so I don't have a need to use `redis-rs` just to parse RESP.

> could be replaced with impl Future if necessary so exposing may not be a good idea.

Indeed! Yet, it is not immediately clear from the documentation what can be the use of `parse_async` since you don't see that `ValueFuture::poll` returns `(impl BufReader, Value)`

Marwes:
That's  a good point, made a PR to change it to `impl Future` https://github.com/mitsuhiko/redis-rs/pull/196

PR #194, opened 2019-03-11
perf(async): Only allocate one oneshot per request

Marwes:
Changes `SharedConnection` to collect the values in the worker task
before sending sending them back instead of sending each response value
in it's own `oneshot`.

This should generally improve performance for pipelined queries since
both allocations and synchronization is reduced. The only downside is
that there is no way to read responses as each one comes in but this
isn't used today.

(I wouldn't put to much stock in the exact numbers as they have pretty high variance, however this PR should always do less work than before).
```
query/simple_getsetdel_async
                        time:   [190.66 us 196.74 us 203.39 us]a
                        change: [-10.222% -6.5229% -2.8090%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

query_pipeline/shared_async_implicit_pipeline
                        time:   [4.7329 ms 4.7641 ms 4.8001 ms]
                        thrpt:  [208.33 Kelem/s 209.90 Kelem/s 211.29 Kelem/s]
                 change:
                        time:   [-24.133% -23.438% -22.678%] (p = 0.00 < 0.05)
                        thrpt:  [+29.329% +30.613% +31.809%]
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
query_pipeline/shared_async_long_pipeline
                        time:   [2.5665 ms 2.5807 ms 2.5998 ms]
                        thrpt:  [384.64 Kelem/s 387.49 Kelem/s 389.63 Kelem/s]
                 change:
                        time:   [-23.663% -23.095% -22.537%] (p = 0.00 < 0.05)
                        thrpt:  [+29.093% +30.030% +30.999%]
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  3 (3.00%) low mild
  4 (4.00%) high mild
  6 (6.00%) high severe
```

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -13,7 +13,7 @@ use tokio_io::{self, AsyncWrite};
 use tokio_tcp::TcpStream;
 
 use futures::future::Either;
-use futures::{future, stream, Async, AsyncSink, Future, Poll, Sink, StartSend, Stream};
+use futures::{future, Async, AsyncSink, Future, Poll, Sink, StartSend, Stream};
 use tokio_sync::{mpsc, oneshot};
 
 use cmd::cmd;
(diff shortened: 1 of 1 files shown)


badboy:
What's the thrpt line in the benchmark output?
(needs a rebase now I guess)

badboy:
> (I wouldn't put to much stock in these)

To much stock into what?

Marwes:
> What's the thrpt line in the benchmark output?

It is for `Throughput` (requests/second).

> To much stock into what?

Reworded it as `I wouldn't put to much stock in the exact numbers as they have pretty high variance, however this PR should always do less work than before`

badboy:
👍

PR #196, opened 2019-03-11
doc: Document what `parse_async` returns with `impl Future`

Marwes:
`ValueFuture` is a hidden type so it isn't possible to see what it
returns. We can instead use `impl Future` to document that while keeping
it hidden.

cc #193


badboy:
👍

PR #197, opened 2019-03-11
perf: Remove redundant BufReader in parse_redis_value

Marwes:


badboy:
That ... kinda makes sense.

PR #198, opened 2019-03-13
 perf(parser): Use more performant operations for line parsing

Marwes:
`take_until_bytes` uses `memchr` to find the next end of line + some
minor changes to use the fact that we know that `line` contains `\r\n`
at the end.

```
decode                  time:   [2.1868 us 2.2129 us 2.2494 us]
                        change: [-30.297% -29.538% -28.706%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ sha1 = ">= 0.2, < 0.7"
 url = "1.2"
 rustc-serialize = { version = "0.3.16", optional = true }
 unix_socket = { version = "0.5.0", optional = true }
-combine = "3.8.0"
+combine = "3.8.1"
 bytes = "0.4"
 futures = "0.1"
 tokio-executor = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ sha1 = ">= 0.2, < 0.7"
 url = "1.2"
 rustc-serialize = { version = "0.3.16", optional = true }
 unix_socket = { version = "0.5.0", optional = true }
-combine = "3.2.0"
+combine = "3.8.0"
 bytes = "0.4"
 futures = "0.1"
 tokio-executor = "0.1"
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -9,13 +9,13 @@ use tokio_io::codec::{Decoder, Encoder};
 use tokio_io::AsyncRead;
 
 use combine;
-use combine::byte::{byte, crlf, newline};
+use combine::byte::{byte, crlf, take_until_bytes};
 use combine::combinator::{any_send_partial_state, AnySendPartialState};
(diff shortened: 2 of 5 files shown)


badboy:
First off: really appreciate all this perf work here! Awesome!

This is causing issues at the moment though:

```
thread 'main' panicked at 'Could not read enough bytes', tests/parser.rs:102:9
thread 'main' panicked at '[quickcheck] TEST FAILED (runtime error). Arguments: (ArbitraryValue(bulk(string-data('""'))), PartialWithErrors { items: [Limited(7)], _marker: PhantomData })
Error: "Could not read enough bytes"', /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-0.6.2/src/tester.rs:179:28
```

Marwes:
Thought I didn't need to add a partial parse test `take_until_bytes` specifically since I just generalized an existing parser to cope with it. Seems that one lacked a test or two though. Good thing it was caught here at least!

Cargo.toml
```diff
@@ -26,7 +26,7 @@ sha1 = ">= 0.2, < 0.7"
 url = "1.2"
 rustc-serialize = { version = "0.3.16", optional = true }
 unix_socket = { version = "0.5.0", optional = true }
-combine = "3.8.0"
+combine = "3.8.1"
```
badboy: Wait, what? *checks combine commits*

src/parser.rs
```diff
@@ -56,13 +56,11 @@ where
 parser! {
     type PartialState = AnySendPartialState;
     fn value['a, I]()(I) -> RedisResult<Value>
-        where [I: RangeStream<Item = u8, Range = &'a [u8]> ]
+        where [I: FullRangeStream<Item = u8, Range = &'a [u8]> ]
     {
-        let end_of_line: fn () -> _ = || crlf().or(newline());
-        let line = || recognize(take_until_range(&b"\r\n"[..]).with(end_of_line()))
+        let line = || recognize(take_until_bytes(&b"\r\n"[..]).with(take(2).map(|_| ())))
```
badboy: Just for my understanding (haven't used combine myself at all yet): The `\r\n` will be included in the final line and thus we can safely assume `line.len() below is at least 2, so the `-2` is safe?

Marwes: Exactly, `recognize` returns the entire range (`&[u8]`) covered by what its argument has parsed. Since the argument parses everything up until `\r\n` and then takes two more elements (which is the actual `\r\n` bytes) it contains at least two bytes.

badboy: Ah, that explains the "with" then. All good to go!

PR #199, opened 2019-03-20
Add support for connection with SENTINEL

gferon:
This is an attempt to introduce support for connecting to a cluster of redis instances configured in high-availability mode.

I understand this kinda breaks the current paradigm that the `Client` creation does not perform any operations/connections, so feel free to point me to a different solution.

(I tested this with the [redis-ha Kubernetes helm chart](https://github.com/helm/charts/tree/master/stable/redis-ha) that I'm currently using with my app)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -41,8 +41,8 @@ impl Client {
     /// return a client pointing to the current master.
     /// This opens a short-lived connection to the sentinal server but does not open
     /// yet a connection to the master.
-    pub fn with_sentinel<T: IntoConnectionInfo>(
-        cluster_name: &str,
+    pub fn open_with_sentinel<T: IntoConnectionInfo>(
+        master_group_name: &str,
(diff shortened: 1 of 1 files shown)


badboy:
This looks like it can easily live outside of the core client, e.g. in your own code or a separate crate for now?

gferon:
Actually sounds like a good idea, I'll keep it like this for now, thanks!

alecks:
@gferon Did you ever get round to publishing a crate for this? I've been on a bit of a search for Sentinel support and this is the only attempt I've found. 😄

gferon:
No, I didn't. I do have this code lying around in a project though, but it's not really smart in any ways (it will not renew the connection when it fails, etc.). I'll extract it into a repository and maybe we can discuss there how to make it useful to you / better and then eventually publish it as a crate. How does that sound?

PR #202, opened 2019-03-29
add GETRANGE to Commands trait

christian-blades-cb:
Adds [GETRANGE](https://redis.io/commands/getrange) and SUBSTR (for redis versions <= 2.0)

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -166,6 +166,18 @@ implement_commands! {
         cmd("GETSET").arg(key).arg(value)
     }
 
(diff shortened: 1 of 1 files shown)


badboy:
Let's ... not support Redis <2.0

christian-blades-cb:
@badboy Sounds good to me. Is there a redis version target for redis-rs that I should put add to the documentation? I could do that while I'm in there.

badboy:
Apparently AppVeyor comes with Redis 3.0, so technically that's what we can guarantee to support.

PR #203, opened 2019-04-08
Expose RedisWrite

grippy:
To implement `ToRedisArgs` outside this crate,  RedisWrite needs to be publicly exposed.


badboy:
That ... was probably an oversight. Thanks for fixing!

PR #206, opened 2019-04-20
feat: Script.invoke_async

emschwartz:
resolves https://github.com/mitsuhiko/redis-rs/issues/175
also changes Script and ScriptInvocation to use an Arc so that the value lasts for the execution of the future

### DIFF
--- a/src/script.rs
+++ b/src/script.rs
@@ -1,11 +1,29 @@
 use sha1::Sha1;
 
+use aio::ConnectionLike as AsyncConnectionLike;
 use cmd::cmd;
 use connection::ConnectionLike;
-use types::{ErrorKind, FromRedisValue, RedisResult, ToRedisArgs};
+use futures::{
+    future::{err, Either},
+    Future,
+};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+use types::{ErrorKind, FromRedisValue, RedisFuture, RedisResult, ToRedisArgs};
 
 /// Represents a lua script.
+#[derive(Clone)]
 pub struct Script {
+    inner: Arc<ScriptInner>,
+    // This stores whether the script has already been loaded into Redis
+    // by this instance. It is possible that the script could have been loaded
+    // by another instance before even if this value is false.
+    already_loaded: Arc<AtomicBool>,
+}
+
+struct ScriptInner {
     code: String,
     hash: String,
 }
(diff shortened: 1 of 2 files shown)


emschwartz:
@mitsuhiko @badboy @ayosec any feedback on this? If it looks okay, would you be open to merging it? Thanks!

badboy:
I'm currently busy and will only get to it in a week or 2.

src/script.rs
```diff
@@ -146,4 +176,61 @@ impl<'a> ScriptInvocation<'a> {
...
+    where
+        T: FromRedisValue + Send + 'static,
+        C: AsyncConnectionLike + Send + Clone + 'static,
+    {
+        let code = self.script.inner.code.to_string();
+
+        // Check if the script has already been loaded into Redis by this instance
+        // Note that storing this boolean makes it so that we do not need to clone
+        // the ScriptInvocation in the normal case when we know that the script has
+        // already been loaded.
```
Marwes: Rather than using an atomic to avoid the need clone we could create a custom future for this (as long as we move `self` in this method). Something like this

```rust
let future = Either::Left(EVAL_FUTURE);
let mut did_load = false;
poll_fn(move || loop {
    match &mut future {
        Either::Left(f) => match f.poll() {
            Err(err) => {
                if did_load {
                    return Err(err);
                }
                if error.kind() == NoScriptError {
                    future = Either::Right(LOAD_FUTURE);
                    did_load = true; // Only retry the eval once after loading
                }
            }
        },
        Either::Right(f) => {
           try_ready!(f.poll());
           future = Either::Left(EVAL_FUTURE);
        }
    }
})
```

While it has a different sort of complexity I think I prefer that since it limits it to only the async version.

emschwartz: Good point! I implemented it as a custom future. Happy to use `poll_fn` instead if you prefer

emschwartz:
@Marwes Do you have permission to merge this into master? Is there anyone else that needs to take a look at it before it's merged in? Thanks!

Marwes:
@emschwartz I can merge it, but I can't do a release so I figure I may as well wait for @badboy to approve. If you need to use it now you would need to point it a git dependency anyway so.

badboy:
If @Marwes took a look at the code and approved we can merge it!

We should get a release out the door soon. My next week clears up a bit so I hope to push it out by then.

badboy:
Oh btw, feel free to send a PR [adding it to the changelog](https://github.com/mitsuhiko/redis-rs/blob/master/CHANGELOG.md). Otherwise I'm happy to fold that in before I cut a release.

PR #211, opened 2019-05-20
Update .gitignore to remove redundant target path

czarneckid:
Just getting started with Rust and was looking at some popular libraries and noticed this redundant path.


Marwes:
Thanks!

PR #217, opened 2019-07-01
fix(async): Don't block the executor from shutting down

Marwes:
The task that drives `SharedConnection` is supposed to shutdown
automatically once the last reference to it is dropped. It did not do
that due to the `poll_read` call in `close` which registered another
`read` on the socket. Since this socket only produces data in response
to any requests done to redis and the last `SharedConnection` reference
were dropped there will only be more data to read while there are
requests in flight.

This ensures that we only drive the socket until those requests are done
whereafter we shut it down.

Tests did not catch this because they only waited for the main task to
finish which were fixed with `block_on_all`.

### DIFF
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -412,10 +412,14 @@ where
     }
 
     fn close(&mut self) -> Poll<(), Self::SinkError> {
-        try_ready!(self.sink_stream.close().map_err(|err| {
+        // No new requests will come in after the first call to `close` but we need to complete any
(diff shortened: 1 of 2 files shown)


badboy:
Still need to review, but this seems important enough to include in the next release (I did now realize I published a beta a while ago, but never the full release).

do you agree?

Marwes:
Yep, definintely. It is a non breaking change so no issues there.

PR #223, opened 2019-07-20
Fix "Trait object without an explicit dyn is deprecated" warnings

Bunogi:
A recent-ish version of Rust nightly deprecated bare trait objects. This PR adds `dyn` to all trait objects in order to get rid of the warning.


badboy:
This bumps our minimum required Rust version to 1.27.
I'm fine with that. Would you be so kind changing the note in the README in this PR?

Bunogi:
@badboy No problem at all, done.

PR #225, opened 2019-07-23
Fix SINTERSTORE wrapper name

akx:
The wrapper for `SINTERSTORE` was incorrectly `sdinterstore`, probably a copy-paste mishap since the previous ones in the series are `sdiff...`


Marwes:
Thanks!

PR #229, opened 2019-08-15
Allow running SharedConnection with any other runtime

hadashiA:
Currently, SharedConnection only works on tokio 0.1 runtime.
Because it depends on `tokio_executor::spawn` (with tokio 0.1.x).
( https://github.com/mitsuhiko/redis-rs/blob/master/src/aio.rs#L437

I want to use SharedConnection with `async/await` in tokio 0.3.

In this PR, for that purpose, make it can be specified Executor from outside without depending on a specific Executor.
( This way is also adopted for [hyper](https://github.com/hyperium/hyper/blob/c71abe5c208de4f1a488cad5f3c68509e7a89266/src/server/conn.rs#L313) , [rusoto](https://github.com/rusoto/rusoto/pull/1472), and more.

For example, 0.3 Executor will work with 0.1 adapter as below:

```rust
let client = redis::Client::open("redis://127.0.0.1:6379").unwrap();
let executor = DefaultExecutor::current(); // This is tokio 0.3 Executor
let executor = Executor03As01::new(DefaultExecutor::current()); // This is tokio 0.1 compatible  Executor
let conn = client.get_shared_async_connection_with_executor(executor).compat().await?; // Works with await 
```

Compatibility layers is:

```rust
use std::sync::Mutex;
use futures01::future::{Executor as Executor01, ExecuteError as ExecuteError01};
use futures01::Future as Future01;
use futures::compat::Future01CompatExt;
use futures::FutureExt;
use tokio_executor::Executor;

pub struct Executor03As01<Ex> {
    inner: Mutex<Ex>,
}

impl<Ex> Executor03As01<Ex> {
    pub fn new(inner: Ex) -> Self {
        Executor03As01 { inner: Mutex::new(inner) }
    }
}

impl<Ex, Fut> Executor01<Fut> for Executor03As01<Ex>
    where
        Ex: Executor,
        Fut: Future01<Item = (), Error = ()> + Send + 'static,
{
    fn execute(&self, future: Fut) -> Result<(), ExecuteError01<Fut>> {
        let mut future = future.compat().map(|_| ());
        let mut pin = Box::pin(future);

        let mut guard = self.inner.lock().unwrap();
        (&mut guard).spawn(pin).expect("unable to spawn future from Compat executor");
        Ok(())
    }
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,16 +18,17 @@ url = "1.2"
 combine = "3.8.1"
 bytes = "0.4"
 futures = "0.1"
-tokio-executor = "0.1"
+tokio-executor = { version = "0.1", optional = true }
 tokio-tcp = "0.1"
 tokio-io = "0.1"
 tokio-codec = "0.1"
 tokio-sync = "0.1"
 
 [features]
-default = [ "geospatial" ]
+default = [ "executor", "geospatial" ]
 
 geospatial = []
+executor = ["tokio-executor"]
 
 [target.'cfg(unix)'.dependencies]
 tokio-uds = { version = "0.2" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ tokio-sync = "0.1"
 default = [ "executor", "geospatial" ]
 
 geospatial = []
-executor = ["tokio-executor"]
+executor = [ "tokio-executor" ]
 
 [target.'cfg(unix)'.dependencies]
 tokio-uds = { version = "0.2" }
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,4 +20,4 @@ install:
 build: false
 
 test_script:
-  - cargo test --verbose --no-default-features %cargoflags%
+  - cargo test --verbose --no-default-features --features executor %cargoflags%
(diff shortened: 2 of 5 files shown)


hadashiA:
Oh, that makes sense. I agree.
I changed tokio-executor to optional.

Marwes:
Hmm, we also depend on tokio-tcp/tokio-uds 0.1. Isn't that also an issue for you? The connection should really be parameterized by `R: AsyncRead + AsyncWrite`, though those would have versioning issues as well atm.

hadashiA:
> Hmm, we also depend on tokio-tcp/tokio-uds 0.1. Isn't that also an issue for you?

No. In my project, the problem now is that the redis-rs require different runtime. 

For me, mixing tokio 0.1 and tokio 0.3 is not a problem.

Because futures 0.1 and futures 0.3 can be used together.
[Compatibility Layer | Futures-rs](https://rust-lang-nursery.github.io/futures-rs/blog/2019/04/18/compatibility-layer.html)

Therefore, even if redis-rs depends on tokio-tcp / uds 0.1.x, it can be run on the tokio 0.3 runtime.

The only incompatibility is the runtime that is running in the application.
I think that tokio-tcp support is enough even after waiting to become stable. 🤔

PR #231, opened 2019-08-20
feat: Update to std::future::Future

### LINKED ISSUE #184: Futures 0.3 support
I couldn't find an existing tracking issue so opening this to ask about in-progress or planned support for futures 0.3/stdlib futures.

Marwes:
Can't be merged yet as it depends on async/await being stabilized. Also has one test disabled as https://github.com/facebookincubator/rust-partial-io does not work with tokio-io 0.2.

- [ ] Update partial-io and fix test
- [ ] Don't depend on the whole of tokio
- [ ] Wait for async/await

Closes #184

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,21 +18,15 @@ sha1 = ">= 0.2, < 0.7"
 url = "1.2"
 combine = "3.8.1"
 bytes = "0.4"
-futures = "0.1"
-tokio-executor = { version = "0.1", optional = true }
-tokio-tcp = "0.1"
-tokio-io = "0.1"
-tokio-codec = "0.1"
-tokio-sync = "0.1"
+futures = { version = "0.3.0-alpha.18", package = "futures-preview" }
+pin-project = "0.3"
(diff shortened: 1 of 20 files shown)


badboy:
Hey, thanks for all that work. Indeed, we need to wait for the stabilization, but happy to start reviewing this (and #232).

Do you think we could extract the 2018 edition fixes and land them up front? That would make it easier to also merge other things if they come in already.

magnet:
Hi, std futures are already stabilized, and async/await stabilization is coming in a month. Several crates in the async/tokio ecosystem have published alpha versions on async/await (e.g hyper, mysql_async, etc) so that we can test them. Would you consider doing the same?

Marwes:
I think we should jump directly to https://github.com/mitsuhiko/redis-rs/pull/232 which requires async await as that improves the API in several ways. I don\t feel like it is ready just yet though as it has a disabled test and some awkward warts which aren't to difficult to fix when I get the time for it (give it a week or two).

PR #232, opened 2019-08-21
Async-ify the API

Marwes:
With async/await we can take references in the async api and avoid clones and
moves in many cases.

Based on #231


Marwes:
Outcome of https://github.com/rust-lang/rust/issues/63778 might affect what kind of API we want for the query builders as they are quite inconvenient atm

Marwes:
I changed the async `ConnectionLike` trait to take `&Cmd/&Pipeline` instead of the encoded vector. Should make it easier to implement the sharding calculation on the redis-cluster side. Might be a good change for the sync interface as well.

https://github.com/mitsuhiko/redis-rs/pull/232/files/d20c2b23af8211adf38cc4e044659b5cc0968e49..b7561257a23d575803bbece33c00cb9453a11cf4

djc:
Are there any plans around merging/releasing an alpha with this?

Marwes:
This is probably as good as I can make this PR for now so 👍 on releasing an alpha with this. Could use a review however.

benches/bench_basic.rs
```diff
@@ -2,18 +2,16 @@
 extern crate criterion;
 use redis;
 
-use futures;
-
 #[path = "../tests/support/mod.rs"]
 mod support;
 
-use futures::{future, stream, Future, Stream};
+use futures::{prelude::*, stream};
```
djc: I generally prefer to avoid `*` imports.

benches/bench_basic.rs
```diff
@@ -35,21 +33,23 @@ fn bench_simple_getsetdel_async(b: &mut Bencher) {
...
+                let () = redis::cmd("SET")
+                    .arg(key)
+                    .arg(42)
+                    .query_async(&mut con)
+                    .await?;
+                let _: isize = redis::cmd("GET").arg(key).query_async(&mut con).await?;
+                let () = redis::cmd("DEL").arg(key).query_async(&mut con).await?;
+                Ok(())
+            })
+            .map_err(|err: RedisError| err)
```
djc: I'm not sure what this `map_err()` is for?

benches/bench_basic.rs
```diff
@@ -119,35 +119,29 @@ fn bench_long_pipeline(b: &mut Bencher) {
...
 
     b.iter(|| {
-        con = runtime
-            .block_on(future::lazy(|| {
-                pipe.clone()
-                    .query_async(con.take().expect("Connection"))
-                    .map(|(con, ())| Some(con))
-            }))
+        runtime
+            .block_on(async { pipe.query_async(&mut con).map_ok(|()| ()).await })
```
djc: What is this `map_ok()` doing? It reads like a no-op.

src/aio.rs
```diff
@@ -1,222 +1,205 @@
 //! Adds experimental async IO support to redis.
 use std::collections::VecDeque;
-use std::fmt::Arguments;
-use std::io::{self, BufReader, Read, Write};
+use std::io::{self};
```
djc: Just simplify this to `std::io`?

src/aio.rs
```diff
@@ -1,222 +1,205 @@
...
-use futures::{future, try_ready, Async, AsyncSink, Future, Poll, Sink, StartSend, Stream};
+use tokio_codec::Decoder;
+use tokio_io::{AsyncBufRead, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader, BufWriter};
+use tokio_net::tcp::TcpStream;
 use tokio_sync::{mpsc, oneshot};
 
-use crate::cmd::cmd;
-use crate::types::{ErrorKind, RedisError, RedisFuture, Value};
+use futures::{
+    prelude::*,
```
djc: (See above.)

src/aio.rs
```diff
@@ -1,222 +1,205 @@
...
-        })
+/// Represents a stateful redis TCP connection.
+pub struct Connection {
+    con: ActualConnection,
+    db: i64,
+}
+
+impl ActualConnection {
+    pub async fn read_response(&mut self) -> RedisResult<Value> {
+        let value = crate::parser::parse_redis_value_async(self).await?;
```
djc: Maybe just elide the `value` temporary here?

src/aio.rs
```diff
@@ -1,222 +1,205 @@
...
-                        }),
-                )
-            } else {
-                Either::B(future::ok(rv))
-            }
-        })
-    }))
+    if connection_info.db != 0 {
+        let mut cmd = cmd("SELECT");
+        cmd.arg(connection_info.db);
```
djc: Looks like you can chain the `arg()` call on `cmd()` directly?

src/aio.rs
```diff
@@ -226,51 +209,38 @@ pub trait ConnectionLike: Sized {
...
+            self.con.flush().await?;
+
+            for _ in 0..offset {
+                self.con.read_response().await?;
+            }
+
+            let mut rv = Vec::with_capacity(count);
+            for _ in 0..count {
+                let item = self.con.read_response().await?;
+                rv.push(item);
```
djc: I would probably inline the `item` temporary here.

src/aio.rs
```diff
@@ -403,180 +394,168 @@ where
...
 
     // `None` means that the stream was out of items causing that poll loop to shut down.
-    fn send(
-        &self,
-        item: T::SinkItem,
-    ) -> impl Future<Item = T::Item, Error = Option<T::Error>> + Send {
+    async fn send(&mut self, item: SinkItem) -> Result<I, Option<E>> {
         self.send_recv_multiple(item, 1)
-            .map(|mut item| item.pop().unwrap())
+            .map_ok(|mut item| item.pop().unwrap())
```
djc: (Not directly related to your PR, but anyway:) is this `unwrap()` safe? Maybe add some comment explaining why this is okay?

src/aio.rs
```diff
@@ -403,180 +394,168 @@ where
...
             .map_err(|_| None)
             .and_then(|_| {
-                receiver.then(|result| match result {
-                    Ok(result) => result.map_err(Some),
-                    Err(_) => {
-                        // The `sender` was dropped which likely means that the stream part
-                        // failed for one reason or another
-                        Err(None)
-                    }
+                receiver.then(|result| {
```
djc: Instead of `then(|result| { future::ready(...) })`, I think you can simplify to just `.map()`?

src/client.rs
```diff
@@ -1,7 +1,7 @@
-use futures::{future::Executor, Future};
+use futures::{prelude::*, task};
```
djc: (As above.)

src/client.rs
```diff
@@ -45,34 +45,47 @@ impl Client {
...
+                future: future::FutureObj<'static, ()>,
+            ) -> Result<(), task::SpawnError> {
+                use tokio_executor::Executor;
+                tokio_executor::DefaultExecutor::current()
+                    .spawn(future.boxed())
+                    .map_err(|_| task::SpawnError::shutdown())
+            }
+        }
+
+        self.get_async_connection().and_then(move |con| {
```
djc: As above, I think `and_then(move |con| { future::ready(...) })` can be simplified with async/await.

src/client.rs
```diff
@@ -45,34 +45,47 @@ impl Client {
...
-    pub fn get_shared_async_connection(
-        &self,
-    ) -> impl Future<Item = crate::aio::SharedConnection, Error = RedisError> {
-        let executor = tokio_executor::DefaultExecutor::current();
-        self.get_async_connection()
-            .and_then(move |con| crate::aio::SharedConnection::new(con, executor))
+    pub fn get_shared_async_connection<'a>(
+        &'a self,
+    ) -> impl Future<Output = RedisResult<crate::aio::SharedConnection>> + 'a {
+        struct TokioExecutor;
```
djc: Personally I find it pretty unclear what this is trying to achieve. Maybe add some comments?

src/client.rs
```diff
@@ -45,34 +45,47 @@ impl Client {
...
         executor: E,
-    ) -> impl Future<Item = crate::aio::SharedConnection, Error = RedisError>
+    ) -> impl Future<Output = RedisResult<crate::aio::SharedConnection>> + 'a
     where
-        E: Executor<Box<dyn Future<Item = (), Error = ()> + Send>>,
+        E: task::Spawn + 'a,
     {
         self.get_async_connection()
-            .and_then(move |con| crate::aio::SharedConnection::new(con, executor))
+            .and_then(move |con| future::ready(crate::aio::SharedConnection::new(con, executor)))
```
djc: (As above.)

src/cmd.rs
```diff
@@ -1,14 +1,20 @@
...
 use crate::types::{
-    from_redis_value, ErrorKind, FromRedisValue, RedisFuture, RedisResult, RedisWrite, ToRedisArgs,
-    Value,
+    from_redis_value, ErrorKind, FromRedisValue, RedisResult, RedisWrite, ToRedisArgs, Value,
 };
 
-use futures::Future;
+use tokio_io::{AsyncWrite, AsyncWriteExt};
+
+use futures::prelude::*;
```
djc: (As above.)

src/parser.rs
```diff
@@ -181,107 +183,132 @@ impl Decoder for ValueCodec {
     }
 }
 
-pub struct ValueFuture<R> {
-    reader: Option<R>,
-    state: AnySendPartialState,
-    // Intermediate storage for data we know that we need to parse a value but we haven't been able
-    // to parse completely yet
-    remaining: Vec<u8>,
+// https://github.com/tokio-rs/tokio/pull/1687
```
djc: I suspect this is not updated with the latest changes from that PR?

src/types.rs
```diff
@@ -8,7 +8,7 @@ use std::hash::{BuildHasher, Hash};
 use std::io;
 use std::str::{from_utf8, Utf8Error};
 
-use futures::Future;
+use futures::prelude::*;
```
djc: (As above.)

src/parser.rs
Marwes: No, the PR now contains breaking changes so I can't use the new code from there. It would need to wait for a new alpha release of tokio as well.

The unsafe is unfortunate but at least it is just temporary until tokio is updated.

benches/bench_basic.rs
Marwes: Just a remnant from futures-0.1 I tend to shut of my brain while refactoring :)

benches/bench_basic.rs
Marwes: I agree, but I think it is fine for prelude imports. That is what they are for after all (to my knowledge).

Marwes: I find that it is really easy when modifying a module with explicit `StreamExt,Future, etc` imports that you need to add or remove some which just ends up as noise.

djc: I disagree, but it's not up to me anyway. :)

benches/bench_basic.rs
Marwes: It lets rustc know which error type to use (due to `From` use in `?`).

djc:
FWIW, I've just finished my work on [updating the bb8 connection pool to async/await](https://github.com/khuey/bb8/pull/37), which includes support for this branch in the bb8-redis crate.

dbrgn:
I'm currently trying out your branch and will post feedback here as I gain more experience with the library.

The first thing that I noticed: The difference between `redis::aio::Connection` and `redis::aio::SharedConnection` is not really clear, some more explanatory docs (either in the main module, in the `get_async_connection` / `get_shared_async_connection_with_executor` methods or in the connection structs) would be great! Especially some guidance on when to use which connection might be useful.

(If I understand it correctly, with `Connection` you might have multiple connections in a single executor while `SharedConnection` is something similar to a thread-local, but in the context of the executor. Is that correct? If yes, what does it mean in the case of a thread pool executor, will there be a single connection for all threads, or will there be a connection per thread?)

Marwes:
Thanks for the feedback @dbrgn . I renamed it to multiplexed (like https://stackexchange.github.io/StackExchange.Redis/PipelinesMultiplexers.html) instead of shared which should explain it better, and added some more documentation.

dbrgn:
Thanks @Marwes!

src/client.rs
```diff
@@ -72,19 +77,24 @@ impl Client {
...
     }
 
-    /// Returns a async shared connection with a specific executor.
-    pub async fn get_shared_async_connection_with_executor<E>(
+    /// Returns an async multiplexed connection from the client.
+    ///
+    /// A multiplexed connection can be cloned, allowing requests to be be sent concurrently
+    /// on the same underlying connection (tcp/unix socket).
+    ///
+    /// Requires an executor to spawn a task that drives the underlying connection.
```
dbrgn: I wonder if a note should be added that `get_multiplexed_async_connection` is also available, but only if the `executor` feature is enabled. Might be a bit hard to discover otherwise.

Marwes: Changed it around a bit to be `tokio-executor` instead and mentioned the feature.

Marwes:
> However, in this case since the redis commands are strings we cannot prevent the user from calling those commands, right?

We could prevent those and return `Err`, but I don't think it is worth it. It could still be useful, and the only thing that happens for the other requests is that they will be delayed or not return, but that could happen regardless (network problems etc) so everything still works.

dbrgn:
Yep, since the API is rather lower-level we can probably ignore this. However, documenting the limitation in the `MultiplexedConnection` would probably be a good idea!

elpiel:
Hey all! Rust 1.39 is out and `async/await` is in stable! What is the status of the PR?

Marwes:
Waiting for an alpha of tokio which uses futures-0.3 seems prudent (https://github.com/tokio-rs/tokio/issues/1708) otherwise this could be released as an alpha whenever.

appveyor.yml
```diff
@@ -4,7 +4,7 @@ environment:
   REDISRS_SERVER_TYPE: tcp
   RUST_BACKTRACE: 1
   matrix:
-    - channel: stable
+    - channel: stablej
```
elpiel: Typo and that's why the build fails

masonforest:
Thanks for your work here @Marwes ! :D

I'm getting the following error when creating a new project and adding your branch of redis as a dependency:

    redis_test HEAD % cargo run
      Compiling redis v0.13.1-alpha.0 (https://github.com/Marwes/redis-rs?branch=async_await_api#8d8295ed)
    error[E0599]: no method named `map` found for type `tokio_sync::mpsc::bounded::Receiver<_>` in the current scope
      --> /Users/masonf/.cargo/git/checkouts/redis-rs-85639ac6b2e4d531/8d8295e/src/aio.rs:442:14
        |
    442 |             .map(Ok)
        |              ^^^ method not found in `tokio_sync::mpsc::bounded::Receiver<_>`
        |
        = note: the method `map` exists but the following trait bounds were not satisfied:
                `&mut tokio_sync::mpsc::bounded::Receiver<_> : futures_util::future::FutureExt`
                `&mut tokio_sync::mpsc::bounded::Receiver<_> : futures_util::stream::StreamExt`
                `&mut tokio_sync::mpsc::bounded::Receiver<_> : std::iter::Iterator`
                `&tokio_sync::mpsc::bounded::Receiver<_> : futures_util::future::FutureExt`
                `&tokio_sync::mpsc::bounded::Receiver<_> : futures_util::stream::StreamExt`
                `tokio_sync::mpsc::bounded::Receiver<_> : futures_util::future::FutureExt`
                `tokio_sync::mpsc::bounded::Receiver<_> : futures_util::stream::StreamExt`

Adding `tokio = "0.2.0-alpha.6"` as a dependency fixes the error.

Let me know if there's any other information I can provide. Thanks again!

Marwes:
@masonforest  Huh, looks like https://github.com/rust-lang/cargo/issues/5730 . Wasn't aware of that issue

Marwes:
Tokio 0.2 is released now. Updated this to support it so we could release a new verson (minor/breaking) now. (I can merge this if everything is ok but I can't make a release to crates.io).

uvd:
merge plan?

bikeshedder:
I would like this to be merged, too.

For the time being I had to create wrappers for `Cmd` and `Pipeline` in [`deadpool-redis`](https://crates.io/crates/deadpool-redis) in order to provide an `async/.await` interface. Once this is merged and a new version of the `redis` crate released I can remove my own wrappers which work around some of the limitations of the consuming function `query_async`.

elpiel:
Yes, please! I also want this to move forwards and be released. Even if it's in `alpha` release.

Marwes:
Ping @mitsuhiko @badboy . Any chance of a release with this? I have been blocked on other dependencies requiring tokio 0.1 so far but that increasingly looks to be solved. I'd prefer not needing a got dependency or forking this crate

badboy:
I plan to be back working on this stuff over the holidays ... but until then I won't get time before the weekend to take a look.

src/parser.rs
```diff
@@ -181,107 +186,132 @@ impl Decoder for ValueCodec {
...
+async fn fill_buf<R>(reader: &mut R) -> io::Result<&[u8]>
```
zyctree: it can simplify like this, the `transmute` is only to Extend the life time
```rust
async fn fill_buf<R>(reader: &mut R) -> io::Result<&[u8]>
where
    R: AsyncBufRead + Unpin,
{
    future::poll_fn(|cx| unsafe { std::mem::transmute(Pin::new(&mut *reader).poll_fill_buf(cx)) })
        .await
}
```

src/commands.rs
```diff
@@ -121,6 +123,52 @@ macro_rules! implement_commands {
             }
         }
 
+
+        /// Implements common redis commands for connection like objects.  This
```
badboy: Should this documentation mention that it's async?

src/parser.rs
```diff
@@ -181,107 +186,132 @@ impl Decoder for ValueCodec {
...
+where
+    R: AsyncBufRead + Unpin,
+{
+    let mut reader = Some(reader);
+    future::poll_fn(move |cx| match reader.take() {
+        Some(r) => match Pin::new(&mut *r).poll_fill_buf(cx) {
+            // SAFETY We either drop `self.reader` and return a slice with the lifetime of the
+            // reader or we return Pending/Err (neither which contains `'a`).
+            // In either case `poll_fill_buf` can not be called while it's contents are exposed
```
badboy:
```suggestion
            // In either case `poll_fill_buf` can not be called while its contents are exposed
```

dbrgn:
Awesome, thanks to everyone involved for landing this feature! :tada:

PR #233, opened 2019-08-22
Fix Script::invoke_async return type error

emschwartz:
Previously, invoking a script with a complex return type would cause the following error: `Response was of incompatible type: "Not a bulk response" (response was string data('"4b98bef92b171357ddc437b395c7c1a5145ca2bd"'))`

This was because the Future returned when loading the script into the DB returns the hash of the script, and thus the return type of String would not match the intended return type.

This commit adds an enum to account for the different Future return types.

Apologies for not catching this originally!

### DIFF
--- a/src/script.rs
+++ b/src/script.rs
@@ -3,7 +3,7 @@ use sha1::Sha1;
 use aio::SharedConnection;
 use cmd::{cmd, Cmd};
 use connection::ConnectionLike;
-use futures::{Async, Future, Poll};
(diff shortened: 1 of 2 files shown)


Marwes:
Thanks!

Marwes:
The one breaking change merged since 0.11 is a minimum version bump of rustc 1.26 to 1.27. I will let @badboy decide (I can't publish anyway)

PR #234, opened 2019-08-23
Update to url v2

jbg:


Marwes:
Breaking change since `Url` is exposed publicly. Otherwise LGTM

PR #243, opened 2019-10-14
Add get_connection_with_timeout to Client

omidplus:
Add get_connection_with_timeout to Client which acts like get_connection but gets a timeout for connecting

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,5 +1,5 @@
 use std::io::{BufRead, BufReader, Write};
-use std::net::{self, TcpStream};
+use std::net::{self, TcpStream, ToSocketAddrs};
 use std::path::PathBuf;
 use std::str::from_utf8;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


yaa110:
@badboy would you mind reviewing this pull request?

omidplus:
Any comment on this pull request?
Will this be ever reviewed?

src/connection.rs
```diff
@@ -205,11 +205,40 @@ pub struct Msg {
...
         Ok(match *addr {
             ConnectionAddr::Tcp(ref host, ref port) => {
                 let host: &str = &*host;
-                let tcp = TcpStream::connect((host, *port))?;
+                let tcp = match timeout {
+                    None => TcpStream::connect((host, *port))?,
+                    Some(timeout) => {
+                        let mut tcp = None;
+                        let mut last_error = None;
+                        for addr in ((host, *port)).to_socket_addrs()? {
```
badboy: What's the reason to resolve and iterate over the addresses here?

omidplus: This is because host can resolve to multiple IP addresses.
The same process is implicitly done in `TcpStream::connect` function, but `TcpStream::connect_timeout` takes only a single `SocketAddr` as described in [TcpStream::connect_timeout](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.connect_timeout).

badboy: That makes sense, thanks!

PR #255, opened 2019-11-05
Do not require transaction to use FromRedisValue

nate-onesignal:
`redis::transaction` takes a closure that returns a value, and it returns that value itself if the value is Option::Some. There is no use of FromRedisValue in `transaction`, so it's not clear why this trait bound is required. It seems perfectly valid to return something that will never be serialized/deserialized with redis, such as a struct containing metrics about the statements run in the transaction.


Marwes:
Thanks!

PR #261, opened 2020-01-04
Add async/await example.

mattcollier:
If this is accurate and of interest, I would like to add an additional function that demonstrates parallelism using `MultiplexedConnection`.

### DIFF
--- a/examples/async-await.rs
+++ b/examples/async-await.rs
@@ -1,0 +1,18 @@
+use redis::AsyncCommands;
+
+#[tokio::main]
+async fn main() -> redis::RedisResult<()> {
+  let client = redis::Client::open("redis://127.0.0.1/").unwrap();
+  let mut con = client.get_async_connection().await?;
+
+  let () = con.set("key1", b"foo").await?;
+
+  let () = redis::cmd("SET").arg(&["key2", "bar"]).query_async(&mut con).await?;
+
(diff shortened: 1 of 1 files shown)


badboy:
We should extract the example from lib.rs instead: https://github.com/mitsuhiko/redis-rs/blob/master/src/lib.rs#L304-L319

It's more concise and avoids the awful unwraps ;)

mattcollier:
@badboy Done. Thanks for pointing out that code.  Provided this makes it in, should the code in that comment be replaced with a reference to the example?

badboy:
> Provided this makes it in, should the code in that comment be replaced with a reference to the example?

No, having it in lib.rs has the nice advantage of getting rendered on the docs page: https://docs.rs/redis/0.13.0/redis/#async

examples/async-await.rs
```diff
@@ -0,0 +1,18 @@
+use redis::AsyncCommands;
+
+#[tokio::main]
+async fn main() -> redis::RedisResult<()> {
+  let client = redis::Client::open("redis://127.0.0.1/").unwrap();
```
badboy: 4-spaces indent (or just run `rustfmt` on that file)

mattcollier: Fixed.

PR #263, opened 2020-01-04
Add async multiplexed example.

mattcollier:
Can the syntax be streamlined or simplified?

### DIFF
--- a/examples/async-multiplexed.rs
+++ b/examples/async-multiplexed.rs
@@ -3,34 +3,32 @@ extern crate redis;
 use futures::{future, prelude::*};
 use redis::{aio::MultiplexedConnection, RedisResult};
 
-fn test_cmd(con: &MultiplexedConnection, i: i32) -> impl Future<Output = RedisResult<()>> + Send {
+async fn test_cmd(con: &MultiplexedConnection, i: i32) -> RedisResult<()> {
     let mut con = con.clone();
-    async move {
-        let key = format!("key{}", i);
-        let key_2 = key.clone();
-        let key2 = format!("key{}_2", i);
-        let key2_2 = key2.clone();
+    let key = format!("key{}", i);
+    let key_2 = key.clone();
+    let key2 = format!("key{}_2", i);
+    let key2_2 = key2.clone();
 
-        let foo_val = format!("foo{}", i);
+    let foo_val = format!("foo{}", i);
 
-        let () = redis::cmd("SET")
-            .arg(&key[..])
-            .arg(foo_val.as_bytes())
-            .query_async(&mut con)
-            .await?;
-        let () = redis::cmd("SET")
-            .arg(&[&key2, "bar"])
-            .query_async(&mut con)
-            .await?;
-        redis::cmd("MGET")
-            .arg(&[&key_2, &key2_2])
-            .query_async(&mut con)
-            .map(|result| {
-                assert_eq!(Ok((foo_val, b"bar".to_vec())), result);
-                Ok(())
-            })
-            .await
-    }
+    let () = redis::cmd("SET")
+        .arg(&key[..])
+        .arg(foo_val.as_bytes())
+        .query_async(&mut con)
+        .await?;
+    let () = redis::cmd("SET")
+        .arg(&[&key2, "bar"])
+        .query_async(&mut con)
+        .await?;
+    redis::cmd("MGET")
+        .arg(&[&key_2, &key2_2])
+        .query_async(&mut con)
+        .map(|result| {
+            assert_eq!(Ok((foo_val, b"bar".to_vec())), result);
+            Ok(())
+        })
+        .await
 }
 
 #[tokio::main]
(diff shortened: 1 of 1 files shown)


examples/async-multiplexed.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    tokio::spawn(driver);
+    let cmds = (0..100).map(move |i| test_cmd(&con, i));
+    future::try_join_all(cmds).await.unwrap();
+}
+
+#[tokio::main]
+async fn main() {
+    println!("Main starting...");
+    tokio::spawn(async move {
```
Marwes: Spawning and immediataly awaiting is the same as just running the future itself so just `run_multiplexed().await;` is enough.

examples/async-multiplexed.rs
```diff
@@ -0,0 +1,55 @@
...
+            .query_async(&mut con)
+            .map(|result| {
+                assert_eq!(Ok((foo_val, b"bar".to_vec())), result);
+                Ok(())
+            })
+            .await
+    }
+}
+
+async fn run_multiplexed() {
```
Marwes: Could just as well move the contents of this method into `main`

examples/async-multiplexed.rs
```diff
@@ -0,0 +1,55 @@
...
+    let client = redis::Client::open("redis://127.0.0.1/").unwrap();
+    let (con, driver) = client.get_multiplexed_async_connection().await.unwrap();
```
Marwes: Since this uses `tokio` explicitly, just calling the `tokio` variant would be better`.

examples/async-multiplexed.rs
```diff
@@ -0,0 +1,55 @@
+extern crate futures;
+extern crate redis;
+extern crate tokio;
+use futures::{future, prelude::*};
+use redis::{aio::MultiplexedConnection, RedisResult};
+
+fn test_cmd(con: &MultiplexedConnection, i: i32) -> impl Future<Output = RedisResult<()>> + Send {
```
Marwes: This could be an `async` function

examples/async-multiplexed.rs
```diff
@@ -0,0 +1,55 @@
...
+            .await
+    }
+}
+
+async fn run_multiplexed() {
+    let client = redis::Client::open("redis://127.0.0.1/").unwrap();
+    let (con, driver) = client.get_multiplexed_async_connection().await.unwrap();
+
+    tokio::spawn(driver);
+    let cmds = (0..100).map(move |i| test_cmd(&con, i));
```
Marwes: Might be simpler, easier to understand to just run two queries concurrently with `try_join`

examples/async-multiplexed.rs
```diff
@@ -0,0 +1,55 @@
+extern crate futures;
+extern crate redis;
+extern crate tokio;
```
Marwes: `extern crate` is redundant.

examples/async-multiplexed.rs
mattcollier: Done.

examples/async-multiplexed.rs
mattcollier: Done.

examples/async-multiplexed.rs
mattcollier: Done.

examples/async-multiplexed.rs
mattcollier: I'll need some guidance on return values etc.  The compiler wins again.

examples/async-multiplexed.rs
mattcollier: Unless you feel strongly about this, I think it's good.

examples/async-multiplexed.rs
mattcollier: Fixed.

examples/async-multiplexed.rs
```diff
@@ -0,0 +1,50 @@
+extern crate futures;
+extern crate redis;
+use futures::{future, prelude::*};
+use redis::{aio::MultiplexedConnection, RedisResult};
+
+fn test_cmd(con: &MultiplexedConnection, i: i32) -> impl Future<Output = RedisResult<()>> + Send {
```
Marwes:
```suggestion
async fn test_cmd(con: &MultiplexedConnection, i: i32) -> RedisResult<()> {
```

The remove the `async move { }` block in the function and it should work

mattcollier: Right, went down this road.  Pushed the change which does not compile with:
```
cannot infer an appropriate lifetime for borrow expression due to conflicting requirements

note: ...so that closure can access `con`
```
Hoping that lifetime syntax does not need to be introduced here.  Is there a simple way out?

badboy:
Fixed and merged in https://github.com/mitsuhiko/redis-rs/commit/c7da0fd64d11175b7515b6eb043ad50ddbc57c63

PR #267, opened 2020-01-15
Fix all clippy lints

jan-auer:
Clippy was run without a profile, which is not checking anything.

I've enabled `clippy::all` now and disabled individual lints where they didn't make sense.


jan-auer:
@mitsuhiko should be ready for review now.

PR #268, opened 2020-01-16
feat: Add ToRedisArgs for &T

untitaker:
This is purely for ergonomics. Having a variable `x: Vec<_>` and passing that variable as `&x` will infer `&Vec<_>` as type, so one currently has to write:

```
let x: Vec<_> =  ...;
client.get(&x[..])
```

With this change one can write:

```
client.get(&x)
```

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -740,12 +740,20 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
     }
 }
 
-impl <T: ToRedisArgs> ToRedisArgs for &T {
+impl<T: ToRedisArgs> ToRedisArgs for &T {
+    fn to_redis_args(&self) -> Vec<Vec<u8>> {
+        ToRedisArgs::to_redis_args(*self)
+    }
+
     fn write_redis_args<W>(&self, out: &mut W)
     where
         W: ?Sized + RedisWrite,
     {
         ToRedisArgs::write_redis_args(*self, out)
+    }
+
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -749,6 +740,19 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+    fn write_redis_args<W>(&self, out: &mut W)
+    where
+        W: ?Sized + RedisWrite,
+    {
+        ToRedisArgs::write_redis_args(*self, out)
+    }
+
+    fn is_single_arg(&self) -> bool {
+        ToRedisArgs::is_single_arg(*self)
+    }
```
Marwes: I believe this should forward all the other methods as well as `&T` and `T` may behave differently otherwise if `T` has specialized one of the inner methods.

untitaker: Thanks for pointing this out. I left the doc(hidden) stuff to defaults

Marwes: I think those should be forwarded as well

Marwes: If possible

untitaker:
I decided to not do this. `Vec<u8>` is currently specialized to produce a bytestring. `Vec<&u8>` would not `impl ToRedisArgs` prior to this PR. With this PR, it receives an implementation that does not do the same thing as `Vec<u8>`. I cannot make such an implementation because I cannot forward `make_arg_vec` for `&u8`. I cannot forward `make_arg_vec` because I cannot construct an `&[u8]` from an `&[&u8]` efficiently.

PR #272, opened 2020-01-27
feat: Let async dependencies like tokio be avoided in sync code

Marwes:
Updates to `combine-4` which pulls the decoding logic into combine itself, allowing redis to only call a pair of macros (for sync and async code respectively). The buffering has been moved into the `Decoder` struct provided by `combine` which avoids some unnecessary copying that the previous implementation did and the `BufReader` were therefore removed.

BREAKING CHANGE

The `Parser` type now only persists the buffer and takes the `Read` instance in `parse_value`

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,5 +1,5 @@
 use std::fmt;
-use std::io::{BufRead, BufReader, Write};
+use std::io::Write;
 use std::net::{self, TcpStream, ToSocketAddrs};
 use std::path::PathBuf;
 use std::str::from_utf8;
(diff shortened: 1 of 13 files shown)


src/connection.rs
```diff
@@ -283,11 +283,7 @@ impl ActualConnection {
     pub fn send_bytes(&mut self, bytes: &[u8]) -> RedisResult<Value> {
         match *self {
             ActualConnection::Tcp(ref mut connection) => {
-                let res = connection
-                    .reader
-                    .get_mut()
-                    .write_all(bytes)
-                    .map_err(RedisError::from);
+                let res = connection.reader.write_all(bytes).map_err(RedisError::from);
```
badboy: Why did we name this `reader` when we're clearly writing to it?
(I know it wasn't part of this PR, I just stumbled upon it here)

badboy: Ah well, it was a `Buf_Reader_` previously. Let's not tack this on this PR though, I'm happy to do the search-and-replace afterwards.

PR #274, opened 2020-01-29
Add an afl fuzz target

dbrgn:
Here's an afl fuzz target for `parse_redis_value`. See the updated README for instructions on how to run this.

I ran this (with commit 9788c51f83a8f77eda3cbd9e267afae25e579ff7) on 24 cores for 8 days and it didn't turn up any crashes or hangs. (To be honest I started the fuzz run last week in a tmux session and then forgot about it :sweat_smile:)


dbrgn:
Is there any other interesting path which accepts arbitrary data that could/should be fuzzed?

badboy:
> Is there any other interesting path which accepts arbitrary data that could/should be fuzzed?

I don't think so. All data passes through that.

PR #276, opened 2020-01-30
fix: Don't hang in the multiplexed connection if the server shuts down unexpectedly

### LINKED ISSUE #275: MultiplexedConnection hangs on connection loss
Code to reproduce:

```rust
use std::env;
use std::process;
use std::time::Duration;

use redis::aio::ConnectionLike;
use tokio::time::interval;

enum Mode {
    Default,
    Multiplexed,
}

async fn run<C: ConnectionLike>(mut con: C) -> redis::RedisResult<()> {
    let mut interval = interval(Duration::from_secs(1));
    loop {
        interval.tick().await;
        println!("PING");
        let result: redis::RedisResult<String> = redis::cmd("PING").query_async(&mut con).await;
        println!("Query result: {:?}", result);
    }
}

#[tokio::main]
    let mode = match env::args().skip(1).next().as_ref().map(String::as_str) {
        Some("default") => {
            println!("Using default connection mode\n");
            Mode::Default
        }
        Some("multiplexed") => {
            println!("Using multiplexed connection mode\n");
            Mode::Multiplexed
        }
        Some(_) | None => {
            println!("Usage: Pass mode as argument (either 'default' or 'multiplexed')");
            process::exit(1); 
        }
    };

    let client = redis::Client::open("redis://127.0.0.1/").unwrap();
    match mode {
        Mode::Default => run(client.get_async_connection().await?).await?,
        Mode::Multiplexed => run(client.get_multiplexed_tokio_connection().await?).await?,
    };
    Ok(())
}
```

With the default connection, when restarting the Redis server, this happens:

```
Using default connection mode

PING
Query result: Ok("PONG")
PING
Query result: Ok("PONG")
PING
Query result: Err(Could not read enough bytes)
PING
Query result: Err(Broken pipe (os error 32))
PING
Query result: Err(Broken pipe (os error 32))
PING
Query result: Err(Broken pipe (os error 32))
^C
```

But with the multiplexed tokio connection, the command hangs indefinitely:

```
Using multiplexed connection mode

PING
Query result: Ok("PONG")
PING
Query result: Ok("PONG")
PING
Query result: Ok("PONG")
PING
^C
```

Of course I could wrap the future in a timeout, but a broken pipe is not the same thing as "server does not respond" and I'd like to know about connection loss.

Refs #218 

CC @Marwes

Marwes:
Regressed in #232 (there is a comment explaining what should be done,
only it wasn't anymore)

Fixes #275

### DIFF
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -308,14 +308,14 @@ where
     }
 
     // Read messages from the stream and send them back to the caller
-    fn poll_read(mut self: Pin<&mut Self>, cx: &mut task::Context) -> Poll<()> {
+    fn poll_read(mut self: Pin<&mut Self>, cx: &mut task::Context) -> Poll<Result<(), ()>> {
         loop {
             let item = match ready!(self.as_mut().project().sink_stream.poll_next(cx)) {
                 Some(Ok(item)) => Ok(item),
                 Some(Err(err)) => Err(err),
                 // The redis response stream is not going to produce any more items so we `Err`
                 // to break out of the `forward` combinator and stop handling requests
-                None => return Poll::Ready(()),
+                None => return Poll::Ready(Err(())),
             };
             self.as_mut().send_result(item);
         }
@@ -411,7 +411,7 @@ where
             .map_err(|err| {
                 self.as_mut().send_result(Err(err));
             }))?;
-        self.poll_read(cx).map(Ok)
+        self.poll_read(cx)
     }
 
     fn poll_close(


Marwes:
Tested manually, without a way to mock the `AsyncRead + AsyncWrite` there isn't a good way to test this.

dbrgn:
Very nice, thanks for the really fast fix :slightly_smiling_face: 

I also tested it locally, the broken pipe error now seems to propagate as it should.

dbrgn:
Nitpick regarding the commit message: EIther "on server shutdown" or "if the server shuts down".

Issue #218, opened 2019-07-08
Connection pooling support for async operation


suraj99p:
I need to write a redis client working in  async mode. We would like to run inside a mpsc receiver and for each request fire the command to redis. We do not want to create  a connection for each client as it will have overhead for each command, we would like to create a pool and keep a persistent TCP connection and multiplex commands on this.

Marwes:
https://docs.rs/redis/0.11.0-beta.1/redis/aio/struct.SharedConnection.html is meant for this use case, it can be cloned and used from everywhere (already uses channels under the hood).

suraj99p:
How do specify pool size or number of connections?

Marwes:
It is just a single connection but that connection does not block further requests while there are are requests in flight. So you can run as many requests as you want on it (you are bound only by IO and single threaded CPU performance on the client and server).

suraj99p:
Ok how do i ensure that connection is still valid and not broken?

Marwes:
Sending https://redis.io/commands/ping would be the simplest way (but be aware that doing any connection check before sending the actual message is no guarantee that the connection won't be broken when the actual message is sent).

suraj99p:
So how do i manage all these scenarios, can you plz guide me how to design this connection manager?

suraj99p:
So cases i can think of are
1) get a new shared connection when previous connection is closed?
2) ping at a given interval, i cant ping before sending each request. That will be a overkill and impact my throughput
3) graceful shutdown and close socket on program close

Any other cases I need to handle apart from these 3?

Marwes:
You can send messages on the connection and when you receive an error which warrants a reconnection, then re-connect and swap the old connection with the new connection in your "Manager". If a request needs to be ran again after an error you make sure to run that on the new connection.

Marwes:
If you have multiple requests in flight when the connection goes down you should also be careful that you only reconnect once.

Marwes:
> ping at a given interval, i cant ping before sending each request. That will be a overkill and impact my throughput

Can be useful if you have periods of downtime and don't want the first request after downtime to (potentially) need to retry if the connection has closed. Otherwise I wouldn't bother with it as any frequently used connection are more likely to report errors on a normal request instead.

suraj99p:
do you plan to implement a connection manager for above use case in near future?

suraj99p:
> 
> 
[...]

Issue #277, opened 2020-01-30
Logging


dbrgn:
Right now redis-cli does not contain any logging, which may be useful for debugging (e.g. logging connection creation, connection loss, errors, etc).

Would it make sense to include the `log` crate? Potentially as an optional dependency.

Terkwood:
`log` and `env_logger` would make sense here. They're widely used, and easy to understand.

I believe that their addition outweighs the long-term cost associated with including them: compile times will increase. (I'm not super confident this is true, as we may already have transitive deps on these libs) 

The programming effort needed to add these is small. Whoever takes this task should be prepared to engage in a bit of healthy discussion about _where_ the log statements should surface.

badboy:
Redis-rs could use `log`. It should depend on `env_logger` (except if you want for examples). The actualy logger is to be determined by the final application, not by a library such as redis.

maxcountryman:
Curious if more thought has been given to this and would be equally (perhaps even more so) interested in tracing as an alternative to logging. (I think this should ideally be implemented via a feature flag.)

dbrgn:
This issue was closed as "not planned". But if I read the current `Cargo.toml` correctly, then `log` is already included as an optional dependency, and this can be considered resolved, right?

nihohit:
Sorry, this was marked by mistake. A limited part of the crate is logged, but not fully. If anyone wants to contribute a more thorough logging, they're welcome.

PR #278, opened 2020-01-30
Connection manager

dbrgn:
Early draft of a ConnectionManager that would address the reconnect behavior discussed in #218.

Would be great to get some feedback on it. @Marwes you've written something like this in the past, right?

To test, run `cargo run --example async-connection-loss --features="tokio-rt-core" reconnect` and then restart and/or stop redis. Output for a 2s server downtime:

```
PING
Query result: Ok("PONG")
PING
Query result: Ok("PONG")
PING
Connection lost, reconnecting
Reconnecting failed: Connection refused (os error 111)
Query result: Err(broken pipe)
PING
Connection lost, reconnecting
Reconnecting failed: Connection refused (os error 111)
Query result: Err(broken pipe)
PING
Connection lost, reconnecting
Query result: Err(broken pipe)
PING
Query result: Ok("PONG")
```

I'll add some comments inline.

Based on #276.

### DIFF
--- a/examples/async-connection-loss.rs
+++ b/examples/async-connection-loss.rs
@@ -4,7 +4,7 @@
 //! - Async multiplexed connection
 //! - Async connection manager
 //!
-//! It will then send a PING every second and print the result.
+//! It will then send a PING every 100 ms and print the result.
 
 use std::env;
 use std::process;
@@ -20,12 +20,12 @@ enum Mode {
 }
 
 async fn run<C: ConnectionLike>(mut con: C) -> redis::RedisResult<()> {
-    let mut interval = interval(Duration::from_secs(1));
+    let mut interval = interval(Duration::from_millis(100));
     loop {
         interval.tick().await;
-        println!("PING");
+        println!("> PING");
         let result: redis::RedisResult<String> = redis::cmd("PING").query_async(&mut con).await;
-        println!("Query result: {:?}", result);
+        println!("< {:?}", result);
     }
 }
 
--- a/examples/async-connection-loss.rs
+++ b/examples/async-connection-loss.rs
@@ -31,7 +31,7 @@ async fn run<C: ConnectionLike>(mut con: C) -> redis::RedisResult<()> {
 
 #[tokio::main]
 async fn main() -> redis::RedisResult<()> {
-    let mode = match env::args().skip(1).next().as_ref().map(String::as_str) {
+    let mode = match env::args().nth(1).as_ref().map(String::as_str) {
         Some("default") => {
             println!("Using default connection mode\n");
             Mode::Default
--- a/examples/async-connection-loss.rs
+++ b/examples/async-connection-loss.rs
@@ -1,0 +1,60 @@
+//! This example will connect to Redis in one of three modes:
+//!
+//! - Regular async connection
+//! - Async multiplexed connection
+//! - Async connection manager
+//!
+//! It will then send a PING every second and print the result.
+
+use std::env;
+use std::process;
+use std::time::Duration;
+
+use redis::aio::ConnectionLike;
+use tokio::time::interval;
+
+enum Mode {
+    Default,
+    Multiplexed,
+    Reconnect,
+}
+
+async fn run<C: ConnectionLike>(mut con: C) -> redis::RedisResult<()> {
+    let mut interval = interval(Duration::from_secs(1));
+    loop {
+        interval.tick().await;
+        println!("PING");
+        let result: redis::RedisResult<String> = redis::cmd("PING").query_async(&mut con).await;
+        println!("Query result: {:?}", result);
+    }
+}
+
+#[tokio::main]
+async fn main() -> redis::RedisResult<()> {
+    let mode = match env::args().skip(1).next().as_ref().map(String::as_str) {
+        Some("default") => {
+            println!("Using default connection mode\n");
+            Mode::Default
+        }
+        Some("multiplexed") => {
+            println!("Using multiplexed connection mode\n");
+            Mode::Multiplexed
+        }
+        Some("reconnect") => {
+            println!("Using reconnect manager mode\n");
+            Mode::Reconnect
+        }
+        Some(_) | None => {
+            println!("Usage: reconnect-manager (default|multiplexed|reconnect)");
+            process::exit(1);
+        }
+    };
+
+    let client = redis::Client::open("redis://127.0.0.1/").unwrap();
+    match mode {
+        Mode::Default => run(client.get_async_connection().await?).await?,
+        Mode::Multiplexed => run(client.get_multiplexed_tokio_connection().await?).await?,
+        Mode::Reconnect => run(client.get_tokio_connection_manager().await?).await?,
+    };
+    Ok(())
+}
--- a/examples/async-connection-loss.rs
+++ b/examples/async-connection-loss.rs
@@ -25,7 +25,7 @@ async fn run_single<C: ConnectionLike>(mut con: C) -> RedisResult<()> {
     let mut interval = interval(Duration::from_millis(1000));
     loop {
         interval.tick().await;
-        println!("");
+        println!();
         println!("> PING");
         let result: RedisResult<String> = redis::cmd("PING").query_async(&mut con).await;
         println!("< {:?}", result);
@@ -36,7 +36,7 @@ async fn run_multi<C: ConnectionLike + Clone>(mut con: C) -> RedisResult<()> {
     let mut interval = interval(Duration::from_millis(100));
     loop {
         interval.tick().await;
-        println!("");
+        println!();
         println!("> PING");
         println!("> PING");
         println!("> PING");
--- a/examples/async-connection-loss.rs
+++ b/examples/async-connection-loss.rs
@@ -10,7 +10,9 @@ use std::env;
 use std::process;
 use std::time::Duration;
 
+use futures::future;
 use redis::aio::ConnectionLike;
+use redis::RedisResult;
 use tokio::time::interval;
 
 enum Mode {
@@ -19,18 +21,43 @@ enum Mode {
     Reconnect,
 }
 
-async fn run<C: ConnectionLike>(mut con: C) -> redis::RedisResult<()> {
-    let mut interval = interval(Duration::from_millis(100));
+async fn run_single<C: ConnectionLike>(mut con: C) -> RedisResult<()> {
+    let mut interval = interval(Duration::from_millis(1000));
     loop {
         interval.tick().await;
+        println!("");
         println!("> PING");
-        let result: redis::RedisResult<String> = redis::cmd("PING").query_async(&mut con).await;
+        let result: RedisResult<String> = redis::cmd("PING").query_async(&mut con).await;
         println!("< {:?}", result);
     }
 }
 
+async fn run_multi<C: ConnectionLike + Clone>(mut con: C) -> RedisResult<()> {
+    let mut interval = interval(Duration::from_millis(100));
+    loop {
+        interval.tick().await;
+        println!("");
+        println!("> PING");
+        println!("> PING");
+        println!("> PING");
+        let results: (
+            RedisResult<String>,
+            RedisResult<String>,
+            RedisResult<String>,
+        ) = future::join3(
+            redis::cmd("PING").query_async(&mut con.clone()),
+            redis::cmd("PING").query_async(&mut con.clone()),
+            redis::cmd("PING").query_async(&mut con),
+        )
+        .await;
+        println!("< {:?}", results.0);
+        println!("< {:?}", results.1);
+        println!("< {:?}", results.2);
+    }
+}
+
 #[tokio::main]
-async fn main() -> redis::RedisResult<()> {
+async fn main() -> RedisResult<()> {
     let mode = match env::args().nth(1).as_ref().map(String::as_str) {
         Some("default") => {
             println!("Using default connection mode\n");
@@ -52,9 +79,9 @@ async fn main() -> redis::RedisResult<()> {
 
     let client = redis::Client::open("redis://127.0.0.1/").unwrap();
     match mode {
-        Mode::Default => run(client.get_async_connection().await?).await?,
-        Mode::Multiplexed => run(client.get_multiplexed_tokio_connection().await?).await?,
-        Mode::Reconnect => run(client.get_tokio_connection_manager().await?).await?,
+        Mode::Default => run_single(client.get_async_connection().await?).await?,
+        Mode::Multiplexed => run_multi(client.get_multiplexed_tokio_connection().await?).await?,
+        Mode::Reconnect => run_multi(client.get_tokio_connection_manager().await?).await?,
     };
     Ok(())
 }
--- a/examples/async-connection-loss.rs
+++ b/examples/async-connection-loss.rs
@@ -22,7 +22,7 @@ enum Mode {
 }
 
 async fn run_single<C: ConnectionLike>(mut con: C) -> RedisResult<()> {
-    let mut interval = interval(Duration::from_millis(1000));
+    let mut interval = interval(Duration::from_millis(100));
     loop {
         interval.tick().await;
         println!();
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -570,3 +570,91 @@ impl ConnectionLike for MultiplexedConnection {
         self.db
     }
 }
+
+#[cfg(feature = "tokio-rt-core")]
+mod connection_manager {
+    use super::*;
+
+    /// A `ConnectionManager` is a proxy that wraps a multiplexed connection and
+    /// automatically reconnects to the server when necessary.
+    pub struct ConnectionManager {
+        connection_info: ConnectionInfo,
+        connection: MultiplexedConnection,
+    }
+
+    #[cfg(feature = "tokio-rt-core")]
+    impl ConnectionManager {
+        /// Connect to the server and store the connection inside the returned `ConnectionManager`.
+        ///
+        /// TODO: Add `new_tokio`?
+        pub async fn new(connection_info: ConnectionInfo) -> RedisResult<Self> {
+            let (connection, driver) = MultiplexedConnection::new(connect(&connection_info).await?);
+            tokio::spawn(driver);
+            Ok(Self {
+                connection_info,
+                connection,
+            })
+        }
+
+        async fn reconnect(&mut self) -> RedisResult<&mut Self> {
+            let (new_connection, driver) = MultiplexedConnection::new(connect(&self.connection_info).await?);
+            tokio::spawn(driver);
+            let _old_connection = std::mem::replace(&mut self.connection, new_connection);
+            Ok(self)
+        }
+    }
+
+    #[cfg(feature = "tokio-rt-core")]
+    impl ConnectionLike for ConnectionManager {
+        fn req_packed_command<'a>(&'a mut self, cmd: &'a Cmd) -> RedisFuture<'a, Value> {
+            (async move {
+                let result = self.connection.req_packed_command(cmd).await;
+                if let Err(ref e) = result {
+                    if e.is_connection_dropped() {
+                        // TODO: This could lead to multiple reconnection
+                        // attempts being done in parallel... We should
+                        // probably avoid that.
+                        println!("Connection lost, reconnecting");
+                        if let Err(reconnect_failed) = self.reconnect().await {
+                            println!("Reconnecting failed: {}", reconnect_failed);
+                        };
+                    }
+                }
+                result
+            })
+            .boxed()
+        }
+
+        fn req_packed_commands<'a>(
+            &'a mut self,
+            cmd: &'a crate::Pipeline,
+            offset: usize,
+            count: usize,
+        ) -> RedisFuture<'a, Vec<Value>> {
+            (async move {
+                let result = self.connection.req_packed_commands(cmd, offset, count).await;
+                if let Err(ref e) = result {
+                    if e.is_connection_dropped() {
+                        // TODO: This could lead to multiple reconnection
+                        // attempts being done in parallel... We should
+                        // probably avoid that.
+                        println!("Connection lost, reconnecting");
+                        self.reconnect().await;
+                        if let Err(reconnect_failed) = self.reconnect().await {
+                            println!("Reconnecting failed: {}", reconnect_failed);
+                        };
+                    }
+                }
+                result
+            })
+            .boxed()
+        }
+
+        fn get_db(&self) -> i64 {
+            self.connection.db
+        }
+    }
+}
+
+#[cfg(feature = "tokio-rt-core")]
+pub use connection_manager::ConnectionManager;
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -582,14 +582,20 @@ mod connection_manager {
         connection: MultiplexedConnection,
     }
 
-    #[cfg(feature = "tokio-rt-core")]
     impl ConnectionManager {
         /// Connect to the server and store the connection inside the returned `ConnectionManager`.
         ///
-        /// TODO: Add `new_tokio`?
+        /// This requires the `tokio-rt-core` feature as it uses the default tokio executor.
         pub async fn new(connection_info: ConnectionInfo) -> RedisResult<Self> {
-            let (connection, driver) = MultiplexedConnection::new(connect(&connection_info).await?);
+            // Wait for the initial connection to be established
+            let inner_connection = connect(&connection_info).await?;
+
+            // Wrap the connection in a MultiplexedConnection
+            let (connection, driver) = MultiplexedConnection::new(inner_connection);
+
+            // Launch the driver that drives the connection future
             tokio::spawn(driver);
+
             Ok(Self {
                 connection_info,
                 connection,
@@ -599,12 +605,11 @@ mod connection_manager {
         async fn reconnect(&mut self) -> RedisResult<&mut Self> {
             let (new_connection, driver) = MultiplexedConnection::new(connect(&self.connection_info).await?);
             tokio::spawn(driver);
-            let _old_connection = std::mem::replace(&mut self.connection, new_connection);
+            self.connection = new_connection;
             Ok(self)
         }
     }
 
-    #[cfg(feature = "tokio-rt-core")]
     impl ConnectionLike for ConnectionManager {
         fn req_packed_command<'a>(&'a mut self, cmd: &'a Cmd) -> RedisFuture<'a, Value> {
             (async move {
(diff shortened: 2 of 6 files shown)


src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+#[cfg(feature = "tokio-rt-core")]
+mod connection_manager {
+    use super::*;
+
+    /// A `ConnectionManager` is a proxy that wraps a multiplexed connection and
+    /// automatically reconnects to the server when necessary.
+    pub struct ConnectionManager {
+        connection_info: ConnectionInfo,
+        connection: MultiplexedConnection,
+    }
```
dbrgn: The `ConnectionManager` wraps a `MultiplexedConnection`. I still have to add cloning behavior to match the `MultiplexedConnection`, but the idea was that the manager can be passed around to different tasks / connection handlers just like the `MultiplexedConnection` itself.

src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+
+    #[cfg(feature = "tokio-rt-core")]
+    impl ConnectionManager {
+        /// Connect to the server and store the connection inside the returned `ConnectionManager`.
+        ///
+        /// TODO: Add `new_tokio`?
```
dbrgn: I'm not sure if this could easily be implemented non-tokio-specific, the futures from the old and the new connection would need to be proxied somehow. If that's possible, we could split `new` and `new_tokio`.

src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+                connection_info,
+                connection,
+            })
+        }
+
+        async fn reconnect(&mut self) -> RedisResult<&mut Self> {
+            let (new_connection, driver) = MultiplexedConnection::new(connect(&self.connection_info).await?);
+            tokio::spawn(driver);
+            let _old_connection = std::mem::replace(&mut self.connection, new_connection);
+            Ok(self)
```
dbrgn: This seems quite straightforward. The old connection will be dropped, resulting in cleanup. That should work, right?

src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+                let result = self.connection.req_packed_command(cmd).await;
+                if let Err(ref e) = result {
+                    if e.is_connection_dropped() {
+                        // TODO: This could lead to multiple reconnection
+                        // attempts being done in parallel... We should
+                        // probably avoid that.
+                        println!("Connection lost, reconnecting");
+                        if let Err(reconnect_failed) = self.reconnect().await {
+                            println!("Reconnecting failed: {}", reconnect_failed);
+                        };
```
dbrgn: The current logic will trigger a reconnect attempt once after every "connection dropped" error.

It means that right now multiple reconnection attempts could happen in parallel if multiple errors happen simultaneously (which is very likely).

Does someone have an idea on how to solve that cleanly?

src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+        fn req_packed_command<'a>(&'a mut self, cmd: &'a Cmd) -> RedisFuture<'a, Value> {
+            (async move {
+                let result = self.connection.req_packed_command(cmd).await;
+                if let Err(ref e) = result {
+                    if e.is_connection_dropped() {
+                        // TODO: This could lead to multiple reconnection
+                        // attempts being done in parallel... We should
+                        // probably avoid that.
+                        println!("Connection lost, reconnecting");
+                        if let Err(reconnect_failed) = self.reconnect().await {
```
dbrgn: This should probably happen asynchronously, but I had a lifetime issue when I put the `self.reconnect()` call into `tokio::spawn`.

src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+            Ok(Self {
+                connection_info,
+                connection,
+            })
+        }
+
+        async fn reconnect(&mut self) -> RedisResult<&mut Self> {
+            let (new_connection, driver) = MultiplexedConnection::new(connect(&self.connection_info).await?);
+            tokio::spawn(driver);
+            let _old_connection = std::mem::replace(&mut self.connection, new_connection);
```
Marwes: self.connection = new_connection;

src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+        ) -> RedisFuture<'a, Vec<Value>> {
+            (async move {
+                let result = self.connection.req_packed_commands(cmd, offset, count).await;
+                if let Err(ref e) = result {
+                    if e.is_connection_dropped() {
+                        // TODO: This could lead to multiple reconnection
+                        // attempts being done in parallel... We should
+                        // probably avoid that.
+                        println!("Connection lost, reconnecting");
+                        self.reconnect().await;
```
Marwes: This waits for the new connection to successfully reconnect. It should instead start the reconnection (`spawn` it on its own task) and just let this request return the error it encountered. This is where https://docs.rs/futures/0.3.1/futures/future/trait.FutureExt.html#method.shared are useful.  Share the reconnection future, then clone and spawn it once while storing the `Shared` future in the "Manager" `connection_future: Shared<...>`. Then `self.connection_future.await?` can be used to retrieve the new connection by just waiting for it to reconnect.

src/aio.rs
```diff
@@ -577,3 +577,91 @@ impl ConnectionLike for MultiplexedConnection {
...
+                        // TODO: This could lead to multiple reconnection
+                        // attempts being done in parallel... We should
+                        // probably avoid that.
+                        println!("Connection lost, reconnecting");
+                        if let Err(reconnect_failed) = self.reconnect().await {
+                            println!("Reconnecting failed: {}", reconnect_failed);
+                        };
+                    }
+                }
+                result
```
dbrgn: The error will be passed on to the caller unmodified. I think this is the only feasible generic solution, since depending on the application logic, if the connection fails after redis has processed the request, automatic retrying might have unintended consequences.

This way, the application must deal with the error, but reconnecting will happen automatically.

src/aio.rs
Marwes: Store something that uniquely identifies each individual connection and then when calling `reconnect` you check if the connection is the same or has changed. If it has changed then a concurrent request has already started the reconnection and nothing needs to be done.

Could be an integer incremented on every reconnection attempt or you could store the connection behind an `Arc` and compare the pointers.

src/aio.rs
Marwes: You will need to put the contents of `ConnectionManager` behind an `Arc`. As it is you can only issue one request at a time on the `ConnectionManager`.

That will then let you clone `self` when needed to get a 'static value for `tokio::spawn`.

src/aio.rs
dbrgn: Ah, @Marwes mentioned `future::Shared` here: https://docs.rs/futures/0.3.1/futures/future/trait.FutureExt.html#method.shared That could be useful.

src/aio.rs
Marwes: Just need to propagate `driver` out to the caller again for this.

src/aio.rs
dbrgn: D'oh :sweat_smile:

src/aio.rs
dbrgn: But the `driver` is replaced when reconnecting.

Marwes: Aha, that's true. Could perhaps use https://docs.rs/futures/0.3.1/futures/task/trait.Spawn.html but tokio unfortunately does not implement it.

dbrgn:
I have to say I'm a bit stuck here.

In the current version of the code, `ConnectionManager` is not yet cloneable. Once `#[derive(Clone)]` is enabled, it's possible to clone the `ConnectionManager`, which is fine in general because it wraps a `Shared` instance.

However, once a manager instance is cloned, if instance A replaces the shared future with the reconnect future, instance B will still have a reference to the old shared future, right?

That would mean that we would need some kind of wrapper type with interior mutability. For example an Arc:

```rust
#[derive(Clone)]
pub struct ConnectionManager(Arc<ConnectionManagerInner>);

struct ConnectionManagerInner {
    connection_info: ConnectionInfo,
    connection: MultiplexedConnection,
}
```

Now references to the connection need to be updated. For read-only references that's easy:

```rust
fn get_db(&self) -> i64 {
    self.0.connection.db
}
```

However, mutable references now complain that the arc contents cannot be borrowed mutably. For example, the reconnect method now cannot replace the connection:

```rust
error[E0594]: cannot assign to data in an `Arc`
   --> src/aio.rs:620:13
    |
620 |             self.0.connection = new_connection;
    |             ^^^^^^^^^^^^^^^^^ cannot assign
    |
    = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `std::sync::Arc<aio::connection_manager::ConnectionManagerInner>`
```

I could now wrap the entire `ConnectionManagerInner` in something like a `Mutex` or a `RwLock`, but then we lose the advantages that we got from wrapping `MultiplexedConnection`.

Am I on the wrong path? :slightly_smiling_face:

Marwes:
No, you will need some sort of interior mutability. https://docs.rs/arc-swap/0.4.4/arc_swap/index.html works well and you can use `compare_and_swap` to prevent multiple reconnections.

dbrgn:
@Marwes thanks for pointing out `arc_swap`, looks great! However, `req_packed_command` takes `&mut self` as argument. This would mean that I'd either have to lock the `ArcSwap` in write mode for every message dispatched, or to clone the `MultiplexedConnection` for every message dispatched.

ArcSwap about their performance characteristics:

> Write operations are considered expensive. A write operation is more expensive than access to an uncontended Mutex and on some architectures even slower than uncontended RwLock.

Cloning the `MultiplexedConnection` would mean cloning an `mpsc::Sender` as well as the db `i64`. Does that sound reasonable? It would be great if we could opt-in to automatic reconnection without having to pay that cost, but I'm not sure if it's possible. (Caching the cloned connection would mean that the reference could become stale...)

Marwes:
Cloning a `MultiplexedConnection` is pretty cheap, it might be possible to do better though by adding methods which work with `&self` which this could call. But it would need https://github.com/tokio-rs/tokio/pull/2041/files in that case.

Marwes:
Perhaps it could cache the connection, but still do a read on the `arc_swap` to check that the connection isn't stale?

src/aio.rs
dbrgn: Ok, then I guess we have to require Tokio for now.

src/aio.rs
```diff
@@ -577,3 +577,111 @@ impl ConnectionLike for MultiplexedConnection {
...
+            println!("Connect");
+            let (new_connection, driver) = MultiplexedConnection::new(
+                connect(&self.connection_info).await?
+            );
+
+            // Spawn the driver that drives the connection future
+            tokio::spawn(driver);
+
+            // Update the connection in the connection manager
+            self.connection.compare_and_swap(guard, Arc::new(new_connection));
```
dbrgn: @Marwes this is not sufficient to avoid creating multiple connections in parallel, right? The way I understand things, it will only prevent *storing* multiple connections (overwriting the previously established connection). In a multi-threaded scenario, the `.await` call on the `connect()` method may yield control to another task that also launches a connection.

If that is correct, I will add an atomic connection counter to avoid reconnecting multiple times.

I didn't manage to trigger multiple connections when removing the loop delay in the example, but that's probably just because I did not enqueue reconnection tasks from multiple threads at the same time.

dbrgn:
I managed to get things working with `arc_swap`. Very nice library, thanks for the recommendation! Also thanks for the help with the pull request :slightly_smiling_face: 

I'll work on the remaining open issues tomorrow:

- Ensure that multiple connections are not established in parallel
- Reconnect asynchronously (don't `.await` after reconnecting)

Open questions:

- Is triggering a reconnect attempt on every failed command OK, or should there be some kind of throttling? Throttling might result in more failed connection attempts than not doing it, but at the same time the server is not hammered with connection attempts.
- Is the separate `connection-manager` feature good? Should it be named differently?
- This is probably non-trivial to unit-test, right?

src/aio.rs
Marwes: We can't prevent creating the future twice, but it is possible to prevent the reconnection from happening twice. Something like this.

```rust
// We can reuse the guard/connection reference that were used to run the connection
fn reconnect(&self,  current: arc_swap::Guard<_>) {
    let (new_connection, driver) = MultiplexedConnection::new(
        connect(&self.connection_info).await?
    );

    // Update the connection in the connection manager
    let prev = self.connection.compare_and_swap(guard, Arc::new(new_connection));

    if Arc::ptr_eq(&prev, current) {
        tokio::spawn(driver);
        tokio::spawn(new_connection);
    }
}
```

Marwes:
> This is probably non-trivial to unit-test, right?

You could abstract `ConnectionManager<T: ConnectionLike>`, and supply a mocked connection instead of `MultiplexedConnection`

src/aio.rs
dbrgn: Hmm, but `connect(...).await` will already open the TCP socket, right?

Marwes:
> Is triggering a reconnect attempt on every failed command OK, or should there be some kind of throttling? Throttling might result in more failed connection attempts than not doing it, but at the same time the server is not hammered with connection attempts.

A circuitbreaker could be built on top of this but the issue comes with figuring out a good strategy for it or making the strategy pluggable which is complicated in its own right :/

> Is the separate connection-manager feature good? Should it be named differently?

Sounds good to me.

dbrgn:
> A circuitbreaker could be built on top of this but the issue comes with figuring out a good strategy for it or making the strategy pluggable which is complicated in its own right :/

Sounds like something that should ideally be implemented in a separate library, so that other projects could re-use it. Probably out of scope for this PR.

(Btw, I edited my comment and added another open question 11 minutes ago. Not sure if you saw that :slightly_smiling_face:)

Marwes:
> If a reconnection attempt is in progress, should the current connection be considered dead? Or could it be that connection errors are transient? In the former case, commands should be delayed until the new connection is established, while in the latter case all commands should be forwarded to the old connection, with the possibility that they fail as well.

As I see it, as soon as a reconnection attempt is started there is no way to retrieve an old connection since the `ArcSwap` has been updated at that point.

dbrgn:
> As I see it, as soon as a reconnection attempt is started there is no way to retrieve an old connection since the `ArcSwap` has been updated at that point.

Ah, good point. In that case, as soon as a reconnection attempt is in progress, all requests should await the new connection.

src/aio.rs
dbrgn: Also, `new_connection` is a pipeline. Does it need to be spawned? I thought only the driver needs to be spawned.

Marwes:
```
    let new_connection =  async move {
        let (new_connection, driver) = MultiplexedConnection::new(
            connect(&self.connection_info).await?
        );
        tokio::spawn(driver);
        Ok(new_connection)
    }.shared();

    // Update the connection in the connection manager
    let prev = self.connection.compare_and_swap(guard, Arc::new(new_connection));

    if Arc::ptr_eq(&prev, current) {
        // start the connection attempt immediately but do not wait on it
        tokio::spawn(new_connection);
    }
```

dbrgn: Ok, that makes sense. However, now the `ArcSwap` doesn't wrap a `MultiplexedConnection` anymore, but a `dyn Future<MultiplexedConnection>` instead: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6ecee2780e2046452967f70fa97d5f3e

I tried putting a `Shared` in the `ArcSwap` instead. However, with static dispatch this does of course not work due to a type change in the future. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=36e7e78a91db04a403c06ec379c565e6

I tried to use dynamic dispatch instead but couldn't get it to work :confused: I'll try again tomorrow, but if you have any pointer that would be helpful. (These are the situations where I sometimes still bump into the Rust async APIs, even though I've done async stuff for a while now.)

Marwes: You will indeed need to box it and use dynamic dispatch https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=99941c415672f38d43b81c7b7cb67b3b

dbrgn: Great, thanks! Somehow it did not occur to me to put the pinned box *inside* the Shared instance.

I also managed to simplify the code a bit. `boxed()` already returns a pinned box, so the type annotations and `Box::new` calls are not necessary. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b62964c32866c8ab11af0912eff1d4c9

dbrgn: Hmm, it seems that since reconnecting can fail, I'll have to wrap the `RedisError` in an `Arc` since it must be `Clone` to be put inside `Shared`...

dbrgn: @Marwes I hit another blocker: When a reconnection starts, the future is stored. All commands that are issued after that point in time should await the connection future before they can issue the command.

The problem is that `Shared` cannot contain a `Future` that resolves to a `RedisError` because that is not `Clone`. I solved this by using `Result<T, Arc<RedisError>>` instead, which works generally but leads to the following issue: If `req_packed_command` fails, it must return a `RedisError`. From the reconnection we get an `Arc<RedisError>` but cannot clone the error (since it's not `Clone`).

I see three possibilities:

- Put `io::Error` inside `RedisError` into something like an `Arc` so that it can be cloned, or use a cloneable abstraction [like this one (seems unmaintained)](https://crates.io/crates/arc-io-error) instead. However, that would be a breaking change.
- Convert `ErrorRepr::IoError(io::Error)` into `ErrorRepr::WithDescription(ErrorKind::IoError, "Reconnecting failed")` and discard the error details. (They should probably be logged though, see #277.) Main downside of this is that errors returned by initial connection and reconnection are different (since error details are discarded). The user does not see *why* the reconnection failed (except in logs). Note that a failed reconnection could immediately trigger another attempt.
- Change the implementation logic. As long as the connection is not yet established, fail all commands immediately with `ErrorRepr::WithDescription(ErrorKind::IoError, "Connection lost, reconnection in progress") until the connection is re-established.

I tend towards number 2. What do you think?

dbrgn:
TIL that `ArcSwap` does not behave like a pointer and should not be cloned in our case :man_facepalming: :slightly_smiling_face: https://github.com/vorner/arc-swap/pull/24

I guess we should use `connection: Arc<ArcSwap<SharedRedisFuture<MultiplexedConnection>>>` for our connection.

I got a version with async reconnection and non-parallel reconnections to work, will have to clean up the code and push it later.

src/aio.rs
dbrgn: I implemented number 2 for now. Seems to work! We can also pass in a string representation of the I/O error to the cloned error.

Marwes: I think ideally we would return the original `io::Error` to the first one to see the error, while the rest get a generic `Reconnecting failed` one. May need to use a `Mutex` for that though, unsure if there is a better way than that :/

dbrgn:
Ok, I think now I have a version that does what it should! It's now ready to review.

Currently implemented behavior:

- When creating an instance of the `ConnectionManager`, an initial connection will be established and awaited. Connection errors will be returned directly.
- When a command sent to the server fails with an error that represents a "connection dropped" condition, that error will be passed on to the user, but it will trigger a reconnection in the background.
- The reconnect code will atomically swap the current (dead) connection with a future that will eventually resolve to a `MultiplexedConnection` or to a `RedisError`
- All commands that are issued after the reconnect process has been initiated, will have to await the connection future.
- If reconnecting fails, all pending commands will be failed as well. A new reconnection attempt will be triggered if the error is an I/O error.

This can be tested by decreasing the delay in `examples/async-connection-loss.rs` to a very low value like 1ms. (I removed most debug prints from the committed code, but left in a few lines.) Here's a test run (3 futures being resolved in parallel every 1ms):

Initially, the connection still works:

```
> PING
> PING
> PING
< Ok("PONG")
< Ok("PONG")
< Ok("PONG")
```

Then I restarted the Redis server. In the debug output, you can see the pointer to the connection future before and after being swapped.

If the log output contains "Swap happened", then a stale connection was replaced with a connection future. If the log output contains "Swap did not happen", then another task was faster in creating a connection future. In this case, no connection will be established.

```
> PING
> PING
> PING
Connection lost, reconnecting
Prepare reconnect
Swap happened
    Current: 0x558835c06c50
        New: 0x558835c077d0
Connection lost, reconnecting
Prepare reconnect
Swap did not happen
Connection lost, reconnecting
Prepare reconnect
Swap did not happen
< Err(Connection reset by peer (os error 104))
< Err(broken pipe)
< Err(broken pipe)
```

On the next command a few ms later, the server was not yet back up, so the reconnect future fails and is immediately replaced by a new connection future:

```
> PING
> PING
> PING
Reconnect failed, reconnecting
Prepare reconnect
Swap happened
    Current: 0x558835c077d0
        New: 0x558835c01a10
Reconnect failed, reconnecting
Prepare reconnect
Swap did not happen
Reconnect failed, reconnecting
Prepare reconnect
Swap did not happen
< Err(Reconnecting failed: Connection refused (os error 111))
< Err(Reconnecting failed: Connection refused (os error 111))
< Err(Reconnecting failed: Connection refused (os error 111))
```

Sometimes the reconnection future fails so fast that the next (concurrent) task already sees the new, failed future:

```
> PING
> PING
> PING
Reconnect failed, reconnecting
Prepare reconnect
Swap happened
    Current: 0x558835c02ab0
        New: 0x558835c02a10
Reconnect failed, reconnecting
Prepare reconnect
Swap happened
    Current: 0x558835c02a10
        New: 0x558835c07800
Reconnect failed, reconnecting
Prepare reconnect
Swap did not happen
< Err(Reconnecting failed: Connection refused (os error 111))
< Err(Reconnecting failed: Connection refused (os error 111))
< Err(Reconnecting failed: Connection refused (os error 111))
Reconnecting
```
After a few dozen milliseconds, the server is back up and the connection future resolves to a working `MultiplexedConnection`.

```
> PING
> PING
> PING
Reconnect failed, reconnecting
Prepare reconnect
Swap happened
Reconnecting
< Err(Reconnecting failed: Connection refused (os error 111))
< Ok("PONG")
< Ok("PONG")
```

Not yet implemented:

- [Connection caching](https://github.com/mitsuhiko/redis-rs/pull/278#issuecomment-581504576). Is it worth it?
- Throttling of reconnections. Maybe just delaying the reconnection attempt by ~50ms would help?

TODO:

- Remove debug print calls
- Squash commits

Note: The rustfmt failures happen because 1.41 introduced a slight change in formatting rules. Maybe the rustfmt check should be pinned to a specific Rust version. The code can then be reformatted when bumping the used Rust version.

dbrgn:
Rebased against master. I also added a commit that removes the debug prints (should be squashed anyways, but can be temporarily reverted for easier testing).

@Marwes is there anything else for me to do here?

src/types.rs
```diff
@@ -405,6 +405,33 @@ impl RedisError {
...
+    pub(crate) fn clone_mostly(&self, ioerror_description: Option<&'static str>) -> Self {
+        let repr = match self.repr {
+            ErrorRepr::WithDescription(kind, desc) => ErrorRepr::WithDescription(kind, desc),
+            ErrorRepr::WithDescriptionAndDetail(kind, desc, ref detail) => {
+                ErrorRepr::WithDescriptionAndDetail(kind, desc, detail.clone())
+            }
+            ErrorRepr::ExtensionError(ref code, ref detail) => {
+                ErrorRepr::ExtensionError(code.clone(), detail.clone())
+            }
+            ErrorRepr::IoError(ref e) => ErrorRepr::WithDescriptionAndDetail(
```
Marwes: Can't this still be an `io::Error` (preserving the kind would be nice)

https://doc.rust-lang.org/std/io/struct.Error.html#method.new

```rust
io::Error::new(e.kind(), e.to_string())
```

src/types.rs
```diff
@@ -405,6 +405,33 @@ impl RedisError {
...
+    /// Clone the `RedisError`, turning wrapped `io::Error`s into a string
+    /// representation.
+    ///
+    /// Deriving `Clone` is not possible because the wrapped `io::Error` is not
+    /// cloneable.
+    ///
+    /// The `ioerror_description` parameter will be used as the static error
+    /// description in case an `IoError` is found.
+    #[cfg(feature = "connection-manager")] // Used to avoid "unused method" warning
+    pub(crate) fn clone_mostly(&self, ioerror_description: Option<&'static str>) -> Self {
```
Marwes: This is always called with `Some` so the `Option` is redundant.

examples/async-connection-loss.rs
```diff
@@ -0,0 +1,87 @@
...
+    Default,
+    Multiplexed,
+    Reconnect,
+}
+
+async fn run_single<C: ConnectionLike>(mut con: C) -> RedisResult<()> {
+    let mut interval = interval(Duration::from_millis(1000));
+    loop {
+        interval.tick().await;
+        println!("");
```
Marwes: Clippy complains about this

examples/async-connection-loss.rs
```diff
@@ -0,0 +1,87 @@
...
+        let result: RedisResult<String> = redis::cmd("PING").query_async(&mut con).await;
+        println!("< {:?}", result);
+    }
+}
+
+async fn run_multi<C: ConnectionLike + Clone>(mut con: C) -> RedisResult<()> {
+    let mut interval = interval(Duration::from_millis(100));
+    loop {
+        interval.tick().await;
+        println!("");
```
Marwes: Ditto

src/types.rs
dbrgn: You are right, fixed!

src/client.rs
```diff
@@ -89,6 +89,32 @@ impl Client {
...
+    /// The connection manager wraps a
+    /// [`MultiplexedConnection`][multiplexed-connection]. If a command to that
+    /// connection fails with a connection error, then a new connection is
+    /// established in the background and the error is returned to the caller.
+    ///
+    /// This means that on connection loss at least one command will fail, but
+    /// the connection will be re-established automatically if possible. (TODO
+    /// dbrgn: Document what happens if reconnection attempts fail, and what happens
+    /// to commands that are invoked while a reconnection attempt is in
+    /// progress.)
```
dbrgn: Oh, I forgot a TODO in here. Will fix that.

Marwes:
Looks good! Thanks!

dbrgn:
On this occasion, thanks @Marwes for the fantastic mentoring and quick feedback you did here - I probably wouldn't have managed to get to this solution without you :slightly_smiling_face:

Terkwood:
Late to the party, but this looks like a great addition! Thank you!

PR #285, opened 2020-02-16
Follow move

mibes:
On my Redis 5.0.6 cluster the "MOVED" error was not handled as expected. Implementing a similar logic as "ASK" made it work. 

Apologies for the formatting changes. The relevant lines are between 461 and 503.

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -20,7 +20,7 @@
 //!
 //! assert_eq!(rv, "test_data");
 //! ```
-//!features = ["default", "cluster"]
+//!
 //! # Pipelining
 //! ```rust,no_run
 //! use redis::{Commands, pipe};
@@ -459,7 +459,6 @@ impl ClusterConnection {
         let mut excludes = HashSet::new();
         let mut asking = None::<String>;
         let mut moved = None::<String>;
-
         loop {
             // Get target address and response.
             let (addr, rv) = {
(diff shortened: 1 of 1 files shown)


mibes:
Files and commits have been updated as per your feedback.

Thanks!

src/cluster.rs
```diff
@@ -491,10 +495,8 @@ impl ClusterConnection {
                         if kind == ErrorKind::Ask {
                             asking = err.redirect_node().map(|x| x.0.to_string());
                         } else if kind == ErrorKind::Moved {
-                            // Refresh slots and request again.
-                            self.refresh_slots()?;
-                            excludes.clear();
-                            continue;
+                            // follow the Moved hint from Redis
+                            moved = err.redirect_node().map(|x| x.0.to_string());
```
badboy: `redirect_node` does explicitly check for `ErrorKind::Ask` and will return `None` if it isn't.
This usage here is thus essentially a no-op.

mibes: OK. I'll do some more testing and see if I can isolate why this did not work for us. I'm definitely not ruling out that we made an implementation error! 😊

mibes:
Original code should work. Closed the pull request until we can reproduce this consistently
in isolation.

badboy:
> Original code should work. Closed the pull request until we can reproduce this consistently
> in isolation.

👍 please do let me know if you find another bug with this

PR #286, opened 2020-02-16
Add `take` method to `Value`

malte-v:
I know this is really opinionated, but it makes things a lot easier for me (currently writing a RedisGraph client) and probably also for other people working with this on a lower level. It's analogous to `std::option::Option::take()`.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -6,6 +6,7 @@ use std::error;
 use std::fmt;
 use std::hash::{BuildHasher, Hash};
 use std::io;
+use std::mem;
 use std::str::{from_utf8, Utf8Error};
 
 /// Helper enum that is used in some situations to describe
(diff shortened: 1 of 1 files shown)


badboy:
That indeed seems _very_ specific.
You could just do the `mem::replace` in your own code, can't you?

malte-v:
Yes, I could just use `mem::replace(&mut value, Value::Nil)` in my own code. It's just much much less expressive than `value.take()` and would get quite repetitive.

If you don't want to merge this, that's fine. It's just a convenience feature.

badboy:
Yeah, I don't think it is a good idea to expose this as public API here.
It's use seems to infrequent to justify its existence.

(and you can always implement it as an extension trait in your code base).

Thanks for suggesting it though!

PR #287, opened 2020-02-17
Add asynchronous PubSub support

### LINKED ISSUE #183: Plans to add async pubsub?
It seems that pubsub is not implemented in the async version. Is it planned to be and would be it hard to add? I'll try to add it but wanted to know if it's being worked on or if there are specific difficulties to do so?

evdokimovs:
Resolves #183

This PR adds asynchronous support for the `PubSub` mechanism of Redis.

`redis::aio::Connection::into_pubsub` requires `Connection` owning because we don't have ability to asynchronously exit from `PubSub` mode in the `Drop` implementation of `redis::aio::PubSub`. 

As alternative we can add `PubSub::close` method which will asynchronously exit from `PubSub` mode and implement synchronous exit from `PubSub` (if `Connection` is in this mode) on `redis::aio::PubSub` dropping for guarantee that `Connection` will not stuck in `PubSub` mode, but it look weird because implicit blocking in asynchronous code is unexpected behavior and many users may forget to call `PubSub::close` method.

Also, you can look at [`tokio-postgres`] implementation of asynchronous API with [`futures::channel::mpsc`] channels. This implementation will allow to asynchronously exit from `PubSub` mode in `Drop` implementation of `redis::aio::PubSub`. But it requires more global changes than I have proposed.

Currently, while `redis::aio::PubSub::on_message`'s `Stream` is stored somewhere, `redis::aio::PubSub::subscribe`-like function wouldn't be available because they all requires mutable `PubSub`. Theoretically, this can be fixed by splitting `TcpStream` / `UnixStream` into a read part and a write part. 

[`tokio-postgres`]: https://github.com/sfackler/rust-postgres/tree/master/tokio-postgres
[`futures::channel::mpsc`]: https://docs.rs/futures/0.3.4/futures/channel/mpsc/index.html


src/aio.rs
```diff
@@ -86,18 +87,159 @@ impl AsyncRead for ActualConnection {
...
+        // subscriptions. Thus, to accurately drain all unsubscribe messages received from the
+        // server, both commands need to be executed at once.
+        {
+            // Prepare both unsubscribe commands
+            let unsubscribe = cmd("UNSUBSCRIBE").get_packed_command();
+            let punsubscribe = cmd("PUNSUBSCRIBE").get_packed_command();
+
+            // Execute commands
+            self.send_bytes(&unsubscribe).await?;
+            self.send_bytes(&punsubscribe).await?;
```
Marwes: Nit: Should be possible to pack both commands into a pipeline and send that.

evdokimovs:
Done!

Marwes:
Thanks!

tyranron:
@Marwes it would be nice to have this one in a release on crates.io. Is there any blockers for that?

PR #299, opened 2020-03-23
Created a method for consuming the connection and generating a stream…

engylemure:
… based on the on_message method, this is useful in cases where we want to return a stream not bounded to the connection itself and could be used outside of the scope of the lifetime of the connection.


Marwes:
Triggering the missing documentation warning on CI

engylemure:
@Marwes Sorry for delay in solving this.

Issue #241, opened 2019-10-07
TLS support


gferon:
Hi, I'd like to implement TLS support for `redis-rs` and I already have [an initial implementation that works for sync](https://github.com/gferon/redis-rs/tree/redis-tls) that's gated by a feature flag (`--tls`).

I just wanted to open an issue because I have a couple of questions:

* Would this be a welcome addition to `redis-rs`?
* I realized while adding a new variant of ActualConnection that matching on the enum variant everywhere is, to my understanding, very much similar to having dynamic dispatch of a trait implementation. Is there a reason it is done like this, and it shouldn't be done with `Box<dyn>`? I personally find that this makes using the feature flags much easier and this is the path I chose while hacking TLS in.
* How should the URL string look like? I had a quick look around, and there's nothing in the Redis Protocol spec, and it looks like other libraries that support TLS use `redis+tls` - I also thought of using the URI fragments to support insecure mode (no check on domain and certificate chain) which would make an URL look like `redis+tls://127.0.0.1:6380#insecure`.

Any feedback is appreciated.

Marwes:
> I realized while adding a new variant of ActualConnection that matching on the enum variant everywhere is, to my understanding, very much similar to having dynamic dispatch of a trait implementation. Is there a reason it is done like this, and it shouldn't be done with Box<dyn>? I personally find that this makes using the feature flags much easier and this is the path I chose while hacking TLS in.

I looked at parameterize the connection types by any type that implements `Write + Read` and it seems that the only reason that is the way it is, is because connecting to an `url` forces the produced connection to be a known type which forces an enum for each backend. It should be solvable though.

ajmwagar:
Hey everyone, 

Now that `async`/`.await` have standardized, and `redis-rs` has added support, are there any plans or updates for this?

If not, I could look into drafting a PR.

Please let me know.

Best,

Avery

gferon:
Actually sounds like a great side-project, I'll get back at it today! I'm happy to collaborate with you @ajmwagar if you want.

abhirockzz:
Hello! Wanted to check if this will be available in anytime soon? Thanks!

PR #305, opened 2020-03-31
Add TLS support, gated by a feature flag

gferon:
Following up on my earlier attempt in MR #244.

TODO:
- [x] Implement TLS for synchronous client with [native-tls](https://crates.io/crates/native-tls)
- [x] Implement TLS with [tokio-tls](https://crates.io/crates/tokio-tls)
- [x] Implement TLS with [async-native-tls](https://crates.io/crates/async-native-tls)
- [x] Add testing facilities for `redis+tls`

Resolves issue #241

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -156,7 +156,7 @@ fn do_redis_code(url: &str) -> redis::RedisResult<()> {
 fn main() {
     // at this point the errors are fatal, let's just fail hard.
     let url = if env::args().nth(1) == Some("--tls".into()) {
-        "redis+tls://127.0.0.1:6380/#insecure"
+        "rediss://127.0.0.1:6380/#insecure"
     } else {
         "redis://127.0.0.1:6379/"
     };
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -1,6 +1,7 @@
 use redis::{self, transaction, Commands};
 
 use std::collections::HashMap;
+use std::env;
 
 /// This function demonstrates how a return value can be coerced into a
 /// hashmap of tuples.  This is particularly useful for responses like
@@ -134,9 +135,9 @@ fn do_atomic_increment(con: &mut redis::Connection) -> redis::RedisResult<()> {
 }
 
 /// Runs all the examples and propagates errors up.
-fn do_redis_code() -> redis::RedisResult<()> {
+fn do_redis_code(url: &str) -> redis::RedisResult<()> {
     // general connection handling
-    let client = redis::Client::open("redis://127.0.0.1/")?;
+    let client = redis::Client::open(url)?;
     let mut con = client.get_connection()?;
 
     // read some config and print it.
@@ -154,7 +155,12 @@ fn do_redis_code() -> redis::RedisResult<()> {
 
 fn main() {
     // at this point the errors are fatal, let's just fail hard.
-    match do_redis_code() {
+    let url = if env::args().nth(1) == Some("--tls".into()) {
+        "redis+tls://127.0.0.1:6380/#insecure"
+    } else {
+        "redis://127.0.0.1:6379/"
+    };
+    match do_redis_code(url) {
         Err(err) => {
             println!("Could not execute example:");
             println!("  {}: {}", err.category(), err);
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -23,6 +23,13 @@ use tokio::{
 #[cfg(feature = "tokio-comp")]
 use tokio::net::TcpStream as TcpStreamTokio;
 
+#[cfg(feature = "tokio-comp")]
+#[cfg(feature = "tls")]
+use tokio_tls::TlsStream as TlsStreamTokio;
+
+#[cfg(feature = "tls")]
+use native_tls::TlsConnector;
+
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use tokio_util::codec::Decoder;
 
@@ -52,7 +59,14 @@ use crate::aio_async_std;
 pub(crate) trait Connect {
     /// Performs a TCP connection
     async fn connect_tcp(socket_addr: SocketAddr) -> RedisResult<ActualConnection>;
-    /// Performans an UNIX connection
+    // Performs a TCP TLS connection
+    #[cfg(feature = "tls")]
+    async fn connect_tcp_tls(
+        hostname: &str,
+        socket_addr: SocketAddr,
+        tls_connector: TlsConnector,
+    ) -> RedisResult<ActualConnection>;
+    /// Performs a UNIX connection
     #[cfg(unix)]
     async fn connect_unix(path: &Path) -> RedisResult<ActualConnection>;
 }
@@ -60,6 +74,9 @@ pub(crate) trait Connect {
 #[cfg(feature = "tokio-comp")]
 mod tokio_aio {
     use super::{async_trait, ActualConnection, Connect, RedisResult, SocketAddr, TcpStreamTokio};
+
+    #[cfg(feature = "tls")]
+    use super::TlsConnector;
 
     #[cfg(unix)]
     use super::{Path, UnixStreamTokio};
@@ -72,6 +89,18 @@ mod tokio_aio {
             Ok(TcpStreamTokio::connect(&socket_addr)
                 .await
                 .map(ActualConnection::TcpTokio)?)
+        }
+        #[cfg(feature = "tls")]
+        async fn connect_tcp_tls(
+            hostname: &str,
+            socket_addr: SocketAddr,
+            tls_connector: TlsConnector,
+        ) -> RedisResult<ActualConnection> {
+            let cx = tokio_tls::TlsConnector::from(tls_connector);
+            Ok(cx
+                .connect(hostname, TcpStreamTokio::connect(&socket_addr).await?)
+                .await
+                .map(ActualConnection::TcpTlsTokio)?)
         }
         #[cfg(unix)]
         async fn connect_unix(path: &Path) -> RedisResult<ActualConnection> {
@@ -87,6 +116,10 @@ pub(crate) enum ActualConnection {
     /// Represents a Tokio TCP connection.
     #[cfg(feature = "tokio-comp")]
     TcpTokio(TcpStreamTokio),
+    /// Represents a Tokio TLS encrypted TCP connection
+    #[cfg(feature = "tokio-comp")]
+    #[cfg(feature = "tls")]
+    TcpTlsTokio(TlsStreamTokio<TcpStreamTokio>),
     /// Represents a Tokio Unix connection.
     #[cfg(unix)]
     #[cfg(feature = "tokio-comp")]
@@ -109,6 +142,9 @@ impl AsyncWrite for ActualConnection {
         match &mut *self {
             #[cfg(feature = "tokio-comp")]
             ActualConnection::TcpTokio(r) => Pin::new(r).poll_write(cx, buf),
+            #[cfg(feature = "tokio-comp")]
+            #[cfg(feature = "tls")]
+            ActualConnection::TcpTlsTokio(r) => Pin::new(r).poll_write(cx, buf),
             #[cfg(unix)]
             #[cfg(feature = "tokio-comp")]
             ActualConnection::UnixTokio(r) => Pin::new(r).poll_write(cx, buf),
@@ -124,6 +160,9 @@ impl AsyncWrite for ActualConnection {
         match &mut *self {
             #[cfg(feature = "tokio-comp")]
             ActualConnection::TcpTokio(r) => Pin::new(r).poll_flush(cx),
+            #[cfg(feature = "tokio-comp")]
+            #[cfg(feature = "tls")]
+            ActualConnection::TcpTlsTokio(r) => Pin::new(r).poll_flush(cx),
             #[cfg(unix)]
             #[cfg(feature = "tokio-comp")]
             ActualConnection::UnixTokio(r) => Pin::new(r).poll_flush(cx),
@@ -139,6 +178,9 @@ impl AsyncWrite for ActualConnection {
         match &mut *self {
             #[cfg(feature = "tokio-comp")]
             ActualConnection::TcpTokio(r) => Pin::new(r).poll_shutdown(cx),
+            #[cfg(feature = "tokio-comp")]
+            #[cfg(feature = "tls")]
+            ActualConnection::TcpTlsTokio(r) => Pin::new(r).poll_shutdown(cx),
             #[cfg(unix)]
             #[cfg(feature = "tokio-comp")]
             ActualConnection::UnixTokio(r) => Pin::new(r).poll_shutdown(cx),
@@ -160,6 +202,9 @@ impl AsyncRead for ActualConnection {
         match &mut *self {
             #[cfg(feature = "tokio-comp")]
             ActualConnection::TcpTokio(r) => Pin::new(r).poll_read(cx, buf),
+            #[cfg(feature = "tokio-comp")]
+            #[cfg(feature = "tls")]
+            ActualConnection::TcpTlsTokio(r) => Pin::new(r).poll_read(cx, buf),
             #[cfg(unix)]
             #[cfg(feature = "tokio-comp")]
             ActualConnection::UnixTokio(r) => Pin::new(r).poll_read(cx, buf),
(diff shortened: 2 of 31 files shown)


src/connection.rs
```diff
@@ -111,14 +132,39 @@ impl IntoConnectionInfo for String {
...
 fn url_to_tcp_connection_info(url: url::Url) -> RedisResult<ConnectionInfo> {
+    let host = match url.host() {
+        Some(host) => host.to_string(),
+        None => fail!((ErrorKind::InvalidClientConfig, "Missing hostname")),
+    };
+    let port = url.port().unwrap_or(DEFAULT_PORT);
+    let addr = if url.scheme() == "redis+tls" {
+        #[cfg(feature = "tls")]
+        {
+            if url.fragment() == Some("insecure") {
```
Marwes, line 143: Nit: Could do

```rust
let insecure = url.fragment() == Some("insecure");
ConnectionAddr::TcpTls {
                    host,
                    port,
                    insecure,
                }
```

Marwes: Should any fragment other than `insecure` and `None` be an error?

gferon: > Should any fragment other than `insecure` and `None` be an error?

I guess it makes sense, I just pushed a new commit to address both of your comments.

gferon: @Marwes I guess I should also add something in the documentation about the new feature flag?

gferon:
@Marwes I guess I should also add something in the documentation about the new feature flag?

Marwes:
There doesn't seem to be a place where features are documented atm (though it would be nice, problably in the top level docs in src/lib.rs).

Adding some/a test for TLS, even without certificate validation would be useful.

gferon:
> There doesn't seem to be a place where features are documented atm (though it would be nice, problably in the top level docs in src/lib.rs).
> 
> Adding some/a test for TLS, even without certificate validation would be useful.

Alright, I'll figure something out later then.

gferon:
@Marwes could you try to help me figure out what's going on with the test suite? It's most probably linked to my change, but I can't figure it out right now.

src/aio.rs
```diff
@@ -52,7 +59,14 @@ use crate::aio_async_std;
 pub(crate) trait Connect {
     /// Performs a TCP connection
     async fn connect_tcp(socket_addr: SocketAddr) -> RedisResult<ActualConnection>;
-    /// Performans an UNIX connection
+    // Performs a TCP TLS connection
```
badboy, line 62:
```suggestion

    // Performs a TCP TLS connection
```

src/aio.rs
```diff
@@ -52,7 +59,14 @@ use crate::aio_async_std;
...
+    #[cfg(feature = "tls")]
+    async fn connect_tcp_tls(
+        hostname: &str,
+        socket_addr: SocketAddr,
+        insecure: bool,
+    ) -> RedisResult<ActualConnection>;
+    /// Performs a UNIX connection
```
badboy, line 69:
```suggestion

    /// Performs a UNIX connection
```

src/aio.rs
badboy: Let's unclutter this a bit by adding newlines.

src/aio.rs
```diff
@@ -73,6 +90,27 @@ mod tokio_aio {
...
         }
+        #[cfg(feature = "tls")]
+        async fn connect_tcp_tls(
+            hostname: &str,
+            socket_addr: SocketAddr,
+            insecure: bool,
+        ) -> RedisResult<ActualConnection> {
+            let tls_connector: tokio_tls::TlsConnector = if insecure {
+                TlsConnector::builder()
+                    .danger_accept_invalid_certs(true)
```
badboy, line 101: That seems ... dangerous and I'd rather not set dangerous configuration by default.
Reducing the security of the TLS connection should be up to the user, not the library.

src/aio.rs
```diff
@@ -87,13 +125,21 @@ pub(crate) enum ActualConnection {
     /// Represents a Tokio TCP connection.
     #[cfg(feature = "tokio-comp")]
     TcpTokio(TcpStreamTokio),
+    /// Represents a Tokio TLS encrypted TCP connection
+    #[cfg(feature = "tokio-comp")]
+    #[cfg(feature = "tls")]
```
badboy, line 130: Can we combine these feature selects into a single line? I assume this would be:

```suggestion
    #[cfg(any(feature = "tokio-comp", feature = "tls"))]
```

src/aio.rs
```diff
@@ -87,13 +125,21 @@ pub(crate) enum ActualConnection {
...
     /// Represents a Tokio Unix connection.
     #[cfg(unix)]
     #[cfg(feature = "tokio-comp")]
     UnixTokio(UnixStreamTokio),
     /// Represents an Async_std TCP connection.
     #[cfg(feature = "async-std-comp")]
     TcpAsyncStd(aio_async_std::TcpStreamAsyncStdWrapped),
+    /// Represents an Async_std TLS encrypted TCP connection.
+    #[cfg(feature = "async-std-comp")]
+    #[cfg(feature = "tls")]
```
badboy, line 141:
```suggestion
    #[cfg(any(feature = "async-std-comp", feature = "tls"))]
```

src/aio.rs
badboy: Wait, isn't this what we have the `tokio-tls-comp` feature for? Gosh, we're getting too many features :D

badboy: Guess that applies to all usage of similar things below, also for `async-std-tls-comp`

src/aio.rs
badboy: Oh only now saw that this is behind the insecure flag. Ignore my comment then.

src/aio.rs
```diff
@@ -390,10 +460,27 @@ async fn connect_simple<T: Connect>(
...
+            ref host,
+            port,
+            insecure,
+        } => {
+            let socket_addr = get_socket_addrs(host, port)?;
+            <T>::connect_tcp_tls(host, socket_addr, insecure).await?
+        }
+
+        #[cfg(not(feature = "tls"))]
+        ConnectionAddr::TcpTls { .. } => {
```
badboy, line 477: We shouldn't even have this variant if it's not compiled with `tls`, right?

src/aio_async_std.rs
```diff
@@ -1,5 +1,9 @@
 use crate::aio::{ActualConnection, Connect};
 use crate::types::RedisResult;
+#[cfg(feature = "tls")]
+use async_native_tls::TlsConnector;
+#[cfg(feature = "tls")]
+use async_native_tls::TlsStream;
```
badboy, line 6:
```suggestion
use async_native_tls::{TlsConnector, TlsStream};
```

src/connection.rs
```diff
@@ -22,7 +25,7 @@ static DEFAULT_PORT: u16 = 6379;
 pub fn parse_redis_url(input: &str) -> Result<url::Url, ()> {
     match url::Url::parse(input) {
         Ok(result) => match result.scheme() {
-            "redis" | "redis+unix" | "unix" => Ok(result),
+            "redis" | "redis+tls" | "redis+unix" | "unix" => Ok(result),
```
badboy, line 28: I remember that some clients did use `rediss:` for secure connections (see also [the iana spec](https://www.iana.org/assignments/uri-schemes/prov/redis)).

Did you look what other clients are currently using?
(I'm ok with `redis+tls`, but if `rediss:` is used widely we maybe should have that as well)

src/connection.rs
```diff
@@ -263,6 +318,67 @@ impl ActualConnection {
...
+                ActualConnection::TcpTls(TcpTlsConnection {
+                    reader: tls,
+                    open: true,
+                })
+            }
+            #[cfg(not(feature = "tls"))]
+            ConnectionAddr::TcpTls { .. } => {
+                fail!((
+                    ErrorKind::InvalidClientConfig,
+                    "Cannot connect to TCP with TLS without the tls feature"
```
badboy, line 379: Oh I see. We already do this for unix socket as well, so that explains why it is happening further above as well. I'm ok with that then.

src/aio.rs
gferon: Yeah, unfortunately it's a little hard to expose this `TlsConnector` to the user. We could figure it out, though.

src/aio.rs
gferon: well, it's like the `unix` one, it's mostly because we want to parse the address string and return an error if the user didn't configure the crate properly for their usage.

src/connection.rs
gferon: I merely copied `redis+unix` for consistency, but I guess the IANA spec is a much better source of truth.

Terkwood:
Curious about the build errors here.  The summary text says the were stopped, but the logs were a little bit harder to understand -- did someone intervene to stop the builds, or did they just idle for a long time and then fail?

Marwes:
travis stops the build automatically if no output has been received for 10 minutes. So there is a bug in here somewhere.

src/connection.rs
```diff
@@ -38,6 +41,22 @@ pub fn parse_redis_url(input: &str) -> Result<url::Url, ()> {
...
+        /// Port
+        port: u16,
+        /// Disable hostname verification when connecting.
+        ///
+        /// # Warning
+        ///
+        /// You should think very carefully before you use this method. If hostname
+        /// verification is not used, any valid certificate for any site will be
+        /// trusted for use from any other. This introduces a significant
+        /// vulnerability to man-in-the-middle attacks.
```
Terkwood, line 57: :+1:

Terkwood:
Potentially related, in the travis log: https://travis-ci.org/github/mitsuhiko/redis-rs/jobs/693945209#L253

`Warning: The home dir /var/run/stunnel4 you specified can't be accessed: No such file or directory`

But probably benign.  I worry that something related to travis _might_ be the issue here, not the change set.  I'll pull down the repo and try running tests on my own machine.

Terkwood:
On my local machine (a linux laptop built in approximately 1433 AD), I was able to get the tests to pass:

```sh
REDISRS_SERVER_TYPE=tcp RUST_TEST_THREADS=1 cargo test --no-default-features --tests -- --nocapture
```

This is the same spot that failed in the travis build.  So I'm not sure there's anything wrong with the change set, and I'm a little bit more suspicious of the travis build's config.  Too -- and  I  haven't dug into this -- some of the other builds attached to the last commit, here, _passed_.  

Just out of sheer masochism, I wonder if re-triggering the builds causes different results than the last run.

Terkwood:
There have been changes since then, but previous iterations of this work got 100% green-lights on Travis:

https://travis-ci.org/github/mitsuhiko/redis-rs/builds/669753751   

https://github.com/mitsuhiko/redis-rs/pull/305/commits/789b3086e9492f9a6ae36f3118d63f61641e4e3c

Terkwood:
Someone has kindly kicked off the build for the latest commit:  https://travis-ci.org/github/mitsuhiko/redis-rs/builds/693945206

Thank you!

Terkwood:
Seems to be hanging at the same spot (test_getset). Run hasn't terminated yet...

Terkwood:
Failures on all of the non-lint builds, but not in the same places. 🔬

gferon:
I'm thinking it could be due to some stunnel processes that aren't killed properly, even though I don't see why the OS in the VM that Travis launches would behave differently than a container of the same distribution (which I tried locally)... Could we try to run the CI on master multiple times in a row to see if it somehow exhibits the same symptoms of flakiness?

Terkwood:
Based on work in #319, I'm relatively confident master doesn't have this issue. The only build failures I saw were routine things like lint errors.

Terkwood:
(I also don't have access to trigger the build on master 😁) If someone else is willing to do so it certainly won't hurt. Beyond that, I'm not sure what to suggest here. This is a tricky one to debug!

Terkwood:
Oh:You can  open a trivial fork with a one line PR, as a draft, and you'll trigger a build

Terkwood:
https://github.com/mitsuhiko/redis-rs/pull/337

testing with this PR

gferon:
Thanks for all the details, it 99.98% sounds like there's something up with my change (my gut feeling would be that it's linked with the test code). I'm back in front of a computer in 48h, I will try to reproduce the problem locally! I wouldn't want anybody spending too much time figuring out what I broke 😅

Terkwood:
No worries! This is an important feature. Initial checks in #337 show that master is clean

gferon:
I managed to reproduce the problem :tada: I'll post the result of the investigation soon.

gferon:
I was excited when I saw the update to redis 6.0, and I was too fast... the build from the `ppa:chris-lea/redis-server` isn't built with TLS support :crying_cat_face:

Terkwood:
😔 Thanks for looking into it... Worth a try

gferon:
I finally figured it out.

I introduced the issue in my PR when I changed the match arm in `tests/support/mod.rs` (see the [fix that restores this code to its original state](https://github.com/mitsuhiko/redis-rs/pull/305/commits/d331d20e5a6d5bdc2638245cf2d966e2844c5d5c)).

Before my change, the `net2::TcpListener2` was dropped _before_ the redis-server is spawned. After I changed it, well... it didn't work anymore :boom:

I thought I knew enough about sockets... It turns out that `SO_REUSEADDR` on Linux does not behave exactly as I thought it would. But hey, I [learned a lot in the process!](https://stackoverflow.com/a/14388707)

@Terkwood I think it's finally ready to be merged.

examples/basic.rs
```diff
@@ -154,7 +155,12 @@ fn do_redis_code() -> redis::RedisResult<()> {
 
 fn main() {
     // at this point the errors are fatal, let's just fail hard.
-    match do_redis_code() {
+    let url = if env::args().nth(1) == Some("--tls".into()) {
+        "rediss://127.0.0.1:6380/#insecure"
```
Terkwood, line 159: 👍🏼

Terkwood:
Huge progress!! 🏁🏁

badboy:
Awesome!
I'm going to land this.

@gferon Can you add some documentation in lib.rs for this in a follow up PR?

tibbon:
How does one use the TLS feature?

tibbon:
I'm getting this error: 

```
david@MacBook-Pro-2 crane % cargo run -- -tls
   Compiling redis v0.17.0
error[E0432]: unresolved import `async_native_tls`
 --> /Users/david/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.17.0/src/aio_async_std.rs:4:5
  |
4 | use async_native_tls::{TlsConnector, TlsStream};
  |     ^^^^^^^^^^^^^^^^ use of undeclared crate or module `async_native_tls`

error[E0433]: failed to resolve: use of undeclared crate or module `tokio_tls`
   --> /Users/david/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.17.0/src/aio.rs:100:32
    |
100 |             let tls_connector: tokio_tls::TlsConnector = if insecure {
    |                                ^^^^^^^^^ use of undeclared crate or module `tokio_tls`

error[E0599]: no variant or associated item named `TcpTlsTokio` found for enum `aio::ActualConnection` in the current scope
   --> /Users/david/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.17.0/src/aio.rs:113:40
    |
113 |                 .map(ActualConnection::TcpTlsTokio)?)
    |                                        ^^^^^^^^^^^
    |                                        |
    |                                        variant or associated item not found in `aio::ActualConnection`
    |                                        help: there is a variant with a similar name: `TcpTokio`
...
125 | pub(crate) enum ActualConnection {
    | -------------------------------- variant or associated item `TcpTlsTokio` not found here

error[E0599]: no variant or associated item named `TcpTlsAsyncStd` found for enum `aio::ActualConnection` in the current scope
   --> /Users/david/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.17.0/src/aio_async_std.rs:161:42
    |
161 |             .map(|con| ActualConnection::TcpTlsAsyncStd(TlsStreamAsyncStdWrapped(con)))?)
    |                                          ^^^^^^^^^^^^^^
    |                                          |
    |                                          variant or associated item not found in `aio::ActualConnection`
    |                                          help: there is a variant with a similar name: `TcpAsyncStd`
    |
   ::: /Users/david/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.17.0/src/aio.rs:125:1
    |
125 | pub(crate) enum ActualConnection {
    | -------------------------------- variant or associated item `TcpTlsAsyncStd` not found here

error: aborting due to 4 previous errors
```

I've included these in my Cargo.toml:

```
redis = { version = "0.17.0" , features = ["tls"] }
native-tls = { version = "0.2" }
tokio-tls = { version = "0.3" }
async-native-tls = { version = "0.3" }
```

tibbon:
Seems to be that this is the magic line: `redis = { version = "0.17.0" , features = ["tls", "async-std-tls-comp", "tokio-tls-comp"] }
`

gferon:
> Seems to be that this is the magic line: `redis = { version = "0.17.0" , features = ["tls", "async-std-tls-comp", "tokio-tls-comp"] } `

You're right, I just pushed #382 to make sure those features are more discoverable.

PR #319, opened 2020-05-18
Add Redis Streams commands

### LINKED ISSUE #162: Add support for streams
Redis 5.0 will ship with streams. It's in beta now and will be stable in few months. It might be a good addition to the library and I hope it will work well with async libraries in Rust. I couldn't see any related issues or PR. Feel free to close this if it's a duplicate.

Streams reference : https://redis.io/topics/streams-intro

Thanks much for the library!

Terkwood:
# Description

This change set adds high-level wrappers for the Redis Streams commands (`XREAD`, `XADD`, etc).  This is effectively a merge of https://github.com/grippy/redis-streams-rs into `redis-rs`.
 
Adds a feature `streams` which is turned on by default (following the example of `geo`). Adds tests for the new commands.  Adds an example of usage.

Resolves #162.

_These changes deliver a significant feature set which was requested in August 2018._

## Task List 

- [x] address some [outstanding lints](https://travis-ci.org/github/mitsuhiko/redis-rs/jobs/688571409)
- [x] add examples of how to manipulate the reply types 
- [x] fix doc links https://github.com/mitsuhiko/redis-rs/pull/319#discussion_r427010711 etc 

## Attribution

This body of work was produced by @grippy and the contributors in https://github.com/grippy/redis-streams-rs.  Thanks for making this available to the community!

### DIFF
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -27,7 +27,7 @@ impl ToRedisArgs for StreamMaxlen {
             StreamMaxlen::Equals(v) => ("=", v),
             StreamMaxlen::Aprrox(v) => ("~", v),
         };
-        out.write_arg("MAXLEN".as_bytes());
+        out.write_arg(b"MAXLEN");
         out.write_arg(ch.as_bytes());
         val.write_redis_args(out);
     }
@@ -91,26 +91,30 @@ impl ToRedisArgs for StreamClaimOptions {
         W: ?Sized + RedisWrite,
     {
         if let Some(ref ms) = self.idle {
-            out.write_arg("IDLE".as_bytes());
+            out.write_arg(b"IDLE");
             out.write_arg(format!("{}", ms).as_bytes());
         }
         if let Some(ref ms_time) = self.time {
-            out.write_arg("TIME".as_bytes());
+            out.write_arg(b"TIME");
             out.write_arg(format!("{}", ms_time).as_bytes());
         }
         if let Some(ref count) = self.retry {
-            out.write_arg("RETRYCOUNT".as_bytes());
+            out.write_arg(b"RETRYCOUNT");
             out.write_arg(format!("{}", count).as_bytes());
         }
         if self.force {
-            out.write_arg("FORCE".as_bytes());
+            out.write_arg(b"FORCE");
         }
         if self.justid {
-            out.write_arg("JUSTID".as_bytes());
+            out.write_arg(b"JUSTID");
         }
     }
 }
 
+/// Argument to `StreamReadOptions`
+/// Represents the Redis GROUP <groupname> <consumername> cmd arg.
+/// This option will toggle the cmd from XREAD to XREADGROUP
+type SRGroup = Option<(Vec<Vec<u8>>, Vec<Vec<u8>>)>;
 /// Builder options for [`xread_options`] command.
 ///
 /// [`xread_options`]: ./trait.StreamCommands.html#method.xread_options
@@ -125,7 +129,7 @@ pub struct StreamReadOptions {
     noack: Option<bool>,
     /// Set the GROUP <groupname> <consumername> cmd arg.
     /// This option will toggle the cmd from XREAD to XREADGROUP.
-    group: Option<(Vec<Vec<u8>>, Vec<Vec<u8>>)>,
+    group: SRGroup,
 }
 
 impl StreamReadOptions {
@@ -175,22 +179,22 @@ impl ToRedisArgs for StreamReadOptions {
         W: ?Sized + RedisWrite,
     {
         if let Some(ref ms) = self.block {
-            out.write_arg("BLOCK".as_bytes());
+            out.write_arg(b"BLOCK");
             out.write_arg(format!("{}", ms).as_bytes());
         }
 
         if let Some(ref n) = self.count {
-            out.write_arg("COUNT".as_bytes());
+            out.write_arg(b"COUNT");
             out.write_arg(format!("{}", n).as_bytes());
         }
 
         if let Some(ref group) = self.group {
             // noack is only available w/ xreadgroup
             if let Some(true) = self.noack {
-                out.write_arg("NOACK".as_bytes());
+                out.write_arg(b"NOACK");
             }
 
-            out.write_arg("GROUP".as_bytes());
+            out.write_arg(b"GROUP");
             for i in &group.0 {
                 out.write_arg(i);
             }
@@ -429,16 +433,13 @@ impl StreamId {
     /// Converts a `Value::Bulk` into a `StreamId`.
     pub fn from_bulk_value(v: &Value) -> RedisResult<Self> {
         let mut stream_id = StreamId::default();
-        match *v {
-            Value::Bulk(ref values) => {
-                if let Some(v) = values.get(0) {
-                    stream_id.id = from_redis_value(&v)?;
-                }
-                if let Some(v) = values.get(1) {
-                    stream_id.map = from_redis_value(&v)?;
-                }
+        if let Value::Bulk(ref values) = *v {
+            if let Some(v) = values.get(0) {
+                stream_id.id = from_redis_value(&v)?;
             }
-            _ => {}
+            if let Some(v) = values.get(1) {
+                stream_id.map = from_redis_value(&v)?;
+            }
         }
 
         Ok(stream_id)
@@ -468,12 +469,17 @@ impl StreamId {
     pub fn len(&self) -> usize {
         self.map.len()
     }
+
+    /// Returns true if there are no field/value pairs in this message.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
 }
 
+type SRRows = Vec<HashMap<String, Vec<HashMap<String, HashMap<String, Value>>>>>;
 impl FromRedisValue for StreamReadReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
-        let rows: Vec<HashMap<String, Vec<HashMap<String, HashMap<String, Value>>>>> =
-            from_redis_value(v)?;
+        let rows: SRRows = from_redis_value(v)?;
         let mut reply = StreamReadReply::default();
         for row in &rows {
             for (key, entry) in row.iter() {
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -1,0 +1,1 @@
+//! Defines types to use with the streams commands.
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -417,13 +417,6 @@ pub struct StreamKey {
     pub ids: Vec<StreamId>,
 }
 
-impl StreamKey {
-    /// Return only the stream `id`'s without their data.
-    pub fn just_ids(&self) -> Vec<&String> {
-        self.ids.iter().map(|msg| &msg.id).collect::<Vec<&String>>()
-    }
-}
-
 /// Represents a stream `id` and its field/values as a `HashMap`
 #[derive(Default, Debug, Clone)]
 pub struct StreamId {
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -606,17 +606,20 @@ impl FromRedisValue for StreamPendingCountReply {
                             }
                             _ => fail!((
                                 crate::types::ErrorKind::TypeError,
-                                "Cannot parse redis data (inner)"
+                                "Cannot parse redis data (3)"
                             )),
                         },
                         _ => fail!((
                             crate::types::ErrorKind::TypeError,
-                            "Cannot parse redis data (outer)"
+                            "Cannot parse redis data (2)"
                         )),
                     }
                 }
             }
-            _ => todo!(),
+            _ => fail!((
+                crate::types::ErrorKind::TypeError,
+                "Cannot parse redis data (1)"
+            )),
         };
         Ok(reply)
     }
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -540,7 +540,7 @@ impl FromRedisValue for StreamPendingReply {
             _,
             Option<String>,
             Option<String>,
-            Vec<Vec<String>>,
+            Vec<Option<(String, String)>>,
         ) = from_redis_value(v)?;
 
         if count == 0 {
@@ -566,13 +566,15 @@ impl FromRedisValue for StreamPendingReply {
             result.start_id = start_id;
             result.end_id = end_id;
 
-            for consumer in consumer_data {
-                let mut info = StreamInfoConsumer::default();
-                info.name = consumer[0].to_owned();
-                if let Ok(v) = consumer[1].parse::<usize>() {
-                    info.pending = v;
+            for cd in consumer_data {
+                if let Some((name, pending)) = cd {
+                    let mut info = StreamInfoConsumer::default();
+                    info.name = name;
+                    if let Ok(v) = pending.parse::<usize>() {
+                        info.pending = v;
+                    }
+                    result.consumers.push(info);
                 }
-                result.consumers.push(info);
             }
 
             Ok(StreamPendingReply::Data(result))
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -452,21 +452,15 @@ impl StreamId {
     /// Fetches value of a given field and converts it to the specified
     /// type.
     pub fn get<T: FromRedisValue>(&self, key: &str) -> Option<T> {
-        match self.find(&key) {
+        match self.map.get(key) {
             Some(ref x) => from_redis_value(*x).ok(),
             None => None,
         }
     }
 
-    /// Fetches the value of a given field without performing any
-    /// type conversion.
-    pub fn find(&self, key: &&str) -> Option<&Value> {
-        self.map.get(*key)
-    }
-
     /// Does the message contain a particular field?
     pub fn contains_key(&self, key: &&str) -> bool {
-        self.find(key).is_some()
+        self.map.get(*key).is_some()
     }
 
     /// Returns how many field/value pairs exist in this message.
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -505,13 +505,10 @@ impl FromRedisValue for StreamRangeReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
         let rows: Vec<HashMap<String, HashMap<String, Value>>> = from_redis_value(v)?;
         let ids: Vec<StreamId> = rows
-            .iter()
+            .into_iter()
             .flat_map(|row| {
-                row.iter()
-                    .map(|(id, map)| StreamId {
-                        id: id.to_string(),
-                        map: map.to_owned(),
-                    })
+                row.into_iter()
+                    .map(|(id, map)| StreamId { id, map })
                     .collect::<Vec<StreamId>>()
             })
             .collect();
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -485,17 +485,17 @@ impl FromRedisValue for StreamReadReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
         let rows: SRRows = from_redis_value(v)?;
         let mut reply = StreamReadReply::default();
-        for row in &rows {
+        for row in rows {
             for (key, entry) in row.iter() {
                 let mut k = StreamKey::default();
                 k.key = key.to_owned();
-                for id_row in entry {
-                    let mut i = StreamId::default();
+                for id_row in entry[..].iter() {
                     for (id, map) in id_row.iter() {
-                        i.id = id.to_owned();
-                        i.map = map.to_owned();
+                        k.ids.push(StreamId {
+                            id: id.to_string(),
+                            map: map.to_owned(),
+                        });
                     }
-                    k.ids.push(i);
                 }
                 reply.keys.push(k);
             }
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -585,7 +585,6 @@ impl FromRedisValue for StreamPendingReply {
 
 impl FromRedisValue for StreamPendingCountReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
-        println!("spcr {:?}", v);
         let mut reply = StreamPendingCountReply::default();
         match v {
             Value::Bulk(outer_tuple) => {
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -520,13 +520,10 @@ impl FromRedisValue for StreamClaimReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
         let rows: Vec<HashMap<String, HashMap<String, Value>>> = from_redis_value(v)?;
         let ids: Vec<StreamId> = rows
-            .iter()
+            .into_iter()
             .flat_map(|row| {
-                row.iter()
-                    .map(|(id, map)| StreamId {
-                        id: id.to_string(),
-                        map: map.to_owned(),
-                    })
+                row.into_iter()
+                    .map(|(id, map)| StreamId { id, map })
                     .collect::<Vec<StreamId>>()
             })
             .collect();
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -486,15 +486,12 @@ impl FromRedisValue for StreamReadReply {
         let rows: SRRows = from_redis_value(v)?;
         let mut reply = StreamReadReply::default();
         for row in rows {
-            for (key, entry) in row.iter() {
+            for (key, entry) in row.into_iter() {
                 let mut k = StreamKey::default();
                 k.key = key.to_owned();
-                for id_row in entry[..].iter() {
-                    for (id, map) in id_row.iter() {
-                        k.ids.push(StreamId {
-                            id: id.to_string(),
-                            map: map.to_owned(),
-                        });
+                for id_row in entry.into_iter() {
+                    for (id, map) in id_row.into_iter() {
+                        k.ids.push(StreamId { id, map });
                     }
                 }
                 reply.keys.push(k);
@@ -569,7 +566,7 @@ impl FromRedisValue for StreamPendingReply {
             result.start_id = start_id;
             result.end_id = end_id;
 
-            for consumer in &consumer_data {
+            for consumer in consumer_data {
                 let mut info = StreamInfoConsumer::default();
                 info.name = consumer[0].to_owned();
                 if let Ok(v) = consumer[1].parse::<usize>() {
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -604,9 +604,15 @@ impl FromRedisValue for StreamPendingCountReply {
                                     times_delivered,
                                 });
                             }
-                            _ => todo!(),
+                            _ => fail!((
+                                crate::types::ErrorKind::TypeError,
+                                "Cannot parse redis data (inner)"
+                            )),
                         },
-                        _ => todo!(),
+                        _ => fail!((
+                            crate::types::ErrorKind::TypeError,
+                            "Cannot parse redis data (outer)"
+                        )),
                     }
                 }
             }
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -507,16 +507,18 @@ impl FromRedisValue for StreamReadReply {
 impl FromRedisValue for StreamRangeReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
         let rows: Vec<HashMap<String, HashMap<String, Value>>> = from_redis_value(v)?;
-        let mut reply = StreamRangeReply::default();
-        for row in &rows {
-            let mut i = StreamId::default();
-            for (id, map) in row.iter() {
-                i.id = id.to_owned();
-                i.map = map.to_owned();
-            }
-            reply.ids.push(i);
-        }
-        Ok(reply)
+        let ids: Vec<StreamId> = rows
+            .iter()
+            .flat_map(|row| {
+                row.iter()
+                    .map(|(id, map)| StreamId {
+                        id: id.to_string(),
+                        map: map.to_owned(),
+                    })
+                    .collect::<Vec<StreamId>>()
+            })
+            .collect();
+        Ok(StreamRangeReply { ids })
     }
 }
 
--- a/src/streams.rs
+++ b/src/streams.rs
@@ -578,13 +578,13 @@ impl FromRedisValue for StreamPendingCountReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
         let parts: Vec<Vec<(String, String, usize, usize)>> = from_redis_value(v)?;
         let mut reply = StreamPendingCountReply::default();
-        for row in &parts {
-            let mut p = StreamPendingId::default();
-            p.id = row[0].0.to_owned();
-            p.consumer = row[0].1.to_owned();
-            p.last_delivered_ms = row[0].2.to_owned();
-            p.times_delivered = row[0].3.to_owned();
-            reply.ids.push(p);
+        for row in parts {
+            reply.ids.push(StreamPendingId {
+                id: row[0].0.to_owned(),
+                consumer: row[0].1.to_owned(),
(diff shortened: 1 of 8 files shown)


grippy:
@Terkwood you rule! Thanks for porting this over. I'm taking a look now and will add comments if I see anything.

src/streams.rs
```diff
@@ -0,0 +1,659 @@
...
+        };
+        out.write_arg("MAXLEN".as_bytes());
+        out.write_arg(ch.as_bytes());
+        val.write_redis_args(out);
+    }
+}
+
+/// Builder options for [`xclaim_options`] command.
+///
+/// [`xclaim_options`]: ./trait.StreamCommands.html#method.xclaim_options
```
grippy, line 38: Change the file prefix to `trait.Commands.html`?

Terkwood: Thanks, will change this

Terkwood:
Adding an executable `streams.rs` example file sounds like a great idea, thanks!

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let parts: (usize, Option<String>, Option<String>, Vec<Vec<String>>) = from_redis_value(v)?;
+        let count = parts.0.to_owned() as usize;
+
+        if count == 0 {
+            Ok(StreamPendingReply::Empty)
+        } else {
+            let mut result = StreamPendingData::default();
+
+            let start_id = match parts.1.to_owned() {
```
Marwes, line 549: Unnecessary to_owned

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+                Some(start) => Ok(start),
+                None => Err(Error::new(
+                    ErrorKind::Other,
+                    "IllegalState: Non-zero pending expects start id",
+                )),
+            }?;
+
+            let end_id = match parts.2.to_owned() {
```
Marwes, line 557: Unnecessary to_owned

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+        let count = parts.0.to_owned() as usize;
+
+        if count == 0 {
+            Ok(StreamPendingReply::Empty)
+        } else {
+            let mut result = StreamPendingData::default();
+
+            let start_id = match parts.1.to_owned() {
+                Some(start) => Ok(start),
+                None => Err(Error::new(
```
Marwes, line 551: Could be simplified with `ok_or_else`

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+                    ErrorKind::Other,
+                    "IllegalState: Non-zero pending expects start id",
+                )),
+            }?;
+
+            let end_id = match parts.2.to_owned() {
+                Some(end) => Ok(end),
+                None => Err(Error::new(
```
Marwes, line 559: Could be simplified with `ok_or_else`

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+                    ErrorKind::Other,
+                    "IllegalState: Non-zero pending expects end id",
+                )),
+            }?;
+
+            result.count = count;
+            result.start_id = start_id;
+            result.end_id = end_id;
+
+            for consumer in &parts.3 {
```
Marwes, line 569: Could move `parts.3`

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+                let mut info = StreamInfoConsumer::default();
+                info.name = consumer[0].to_owned();
+                if let Ok(v) = consumer[1].to_owned().parse::<usize>() {
```
Marwes, line 572: Unncessary to_owned

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+                )),
+            }?;
+
+            result.count = count;
+            result.start_id = start_id;
+            result.end_id = end_id;
+
+            for consumer in &parts.3 {
+                let mut info = StreamInfoConsumer::default();
+                info.name = consumer[0].to_owned();
```
Marwes, line 571: Unncessary to_owned

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+            Ok(StreamPendingReply::Data(result))
+        }
+    }
+}
+
+impl FromRedisValue for StreamPendingCountReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let parts: Vec<Vec<(String, String, usize, usize)>> = from_redis_value(v)?;
+        let mut reply = StreamPendingCountReply::default();
+        for row in &parts {
```
Marwes, line 587: Could move and not use to_owned below

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+            let mut p = StreamPendingId::default();
```
Marwes, line 588: Could use a struct initializer, no need to default it and assign later.

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+                g.consumers = from_redis_value(v)?;
+            }
+            if let Some(v) = &map.get("last-delivered-id") {
+                g.last_delivered_id = from_redis_value(v)?;
+            }
+            reply.groups.push(g);
+        }
+        Ok(reply)
+    }
+}
```
Marwes, line 669: All these `from_redis_value` functions could use use `collect` instead of explicit for loops. Perhaps with the help of adaptors like `flat_map` in some cases

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+            }
+            reply.ids.push(i);
+        }
+        Ok(reply)
+    }
+}
+
+impl FromRedisValue for StreamPendingReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let parts: (usize, Option<String>, Option<String>, Vec<Vec<String>>) = from_redis_value(v)?;
```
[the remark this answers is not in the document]
Terkwood: Will do!

Terkwood:
Opening this for review.  Thanks, @Marwes, for taking a look.

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let rows: SRRows = from_redis_value(v)?;
+        let mut reply = StreamReadReply::default();
+        for row in rows {
+            for (key, entry) in row.iter() {
+                let mut k = StreamKey::default();
+                k.key = key.to_owned();
+                for id_row in entry[..].iter() {
+                    for (id, map) in id_row.iter() {
+                        k.ids.push(StreamId {
```
Terkwood, line 494: @grippy: I altered the structure of these loops based on @Marwes's comment in https://github.com/mitsuhiko/redis-rs/pull/319#discussion_r434025119.  Let me know if we're wrong about this, but I think we want all the `StreamId`s added to each `StreamKey`.

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+        self.len() == 0
+    }
+}
+
+type SRRows = Vec<HashMap<String, Vec<HashMap<String, HashMap<String, Value>>>>>;
+impl FromRedisValue for StreamReadReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let rows: SRRows = from_redis_value(v)?;
+        let mut reply = StreamReadReply::default();
+        for row in rows {
```
Terkwood, line 488: @Marwes: I moved the `rows` value as mentioned in https://github.com/mitsuhiko/redis-rs/pull/319#discussion_r434023996, but the subsequent `key`, `id`, `map` were still behind references.  Sorry if I'm missing something basic here, but I didn't see how to avoid the `.to_owned`s

Terkwood: Still looking at changing this into something that uses `collect` instead of `for`, which might help...

src/streams.rs
Terkwood: Got it.  I have one of these replaced in the latest commit.  Shouldn't be too bad to knock out the rest.

src/streams.rs
Terkwood: Yeah, still not seeing it even after using `collect` instead of `for` in one of these `from_redis_value` functions.  I'll continue with the rest of the review comments and circle back to the issue of `to_owned`s once you get a chance to take a look.  No rush!

src/streams.rs
Terkwood: Compile reports that it's behind a shared reference

src/streams.rs
Terkwood: Got this one resolved

src/streams.rs
```diff
@@ -485,17 +485,17 @@ impl FromRedisValue for StreamReadReply {
...
         let mut reply = StreamReadReply::default();
-        for row in &rows {
+        for row in rows {
             for (key, entry) in row.iter() {
                 let mut k = StreamKey::default();
                 k.key = key.to_owned();
-                for id_row in entry {
-                    let mut i = StreamId::default();
+                for id_row in entry[..].iter() {
                     for (id, map) in id_row.iter() {
```
Marwes, line 493: `iter()` returns references to the container, you should call `into_iter()` or just pass the iterator directly, to get values out.

Terkwood: Thank you!  🚸   Pushing fixes presently.

src/streams.rs
Terkwood: Fixed in the most recent commit.

src/streams.rs
Terkwood: Will address these tomorrow, thanks!

src/streams.rs
Marwes: If this is known to just be a two element sequence you can extract it as a tuple instead of a vec which will let you easily move it.

src/streams.rs
```diff
@@ -0,0 +1,669 @@
...
+    }
+}
+
+impl FromRedisValue for StreamPendingCountReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let parts: Vec<Vec<(String, String, usize, usize)>> = from_redis_value(v)?;
+        let mut reply = StreamPendingCountReply::default();
+        for row in parts {
+            reply.ids.push(StreamPendingId {
+                id: row[0].0.to_owned(),
```
Marwes, line 589: How come this only extracts the first element of `row`? Is it always a 1-element sequence or are the other elements just ignored?

Terkwood: Let me dig in on this, as I'm also worried about.

Terkwood: It's always a one-element sequence.  I dug into the docs and the data ends up matching what's expected to be returned from redis, it's just going through some contortions while being processed in `FromRedisValue` -- the tuple being nested in an array causes a strange shape.  

Here's a sample print-out that I captured by adding in some `println` (temporarily) to the test suite:

```text
  [("1000-1", "c99", 3, 2)]
  [("1001-0", "c99", 1, 1)]
  [("1001-1", "c99", 1, 1)]
  [("1000-1", "c99", 4, 2)]
  [("1001-0", "c99", 2, 1)]
  [("1001-1", "c99", 2, 1)]
 ```

Too, I expanded the test coverage to make sure that EVERY row processed by that function has _all four_ expected values.  I'll update the branch with that additional test, now.

This is the [relevant page from the Redis documentation](https://redis.io/commands/xpending)....

...which is actually a pretty dense bit of information.  Here's the specific part that we're interested in:

xpending_detail

File: src/streams.rs (hunk 8 above)
Comment (Terkwood) [reply src/streams.rs]: Fixed!

**Comment by Terkwood** [comment]
The build failure is just a clippy warning about the complexity of the type that I introduced during the previous commit. Should be pretty easy to fix that up. Will revisit soon. ⌚

File: src/streams.rs (hunk 16 above)
Comment (Marwes) [reply src/streams.rs:591]: Ok, in that case you should be able to use a one element tuple instead of a `Vec`, `((A, B, C, D),)` instead of `Vec<(A, B, C, D)>`.

Comment (Terkwood) [reply src/streams.rs:591]: I took your advice, but for some reason, the data structure ends up mangled.

https://github.com/Terkwood/redis-rs/blob/dc648209a72a68f85b52ee4f3cc11b9d6fba0afb/src/streams.rs#L590

Here's a temp branch.  If you run the tests against this branch, you can see that `parts` is reduced to length 2, incorrectly.  It should be length 3.

I tested using the following:
```sh
REDISRS_SERVER_TYPE=unix cargo test --all-features -- --skip test_cluster --nocapture
```

I've been trying several different minor changes and can't fathom why it's not coalescing the `Value::Bulk`  into a `Vec<((String, String, usize, usize),)>` as expected.

Terkwood: (Please ignore the ugly/wasteful `id: (row.0).0.to_owned()` in that branch -- I realize we can dispense with the call to `.to_owned`, once this is figured out -- but am focused on the fact that what looked like an innocent refactor is actually breaking tests.)

Terkwood: Here's the relevant test spam:

The initial `Value::Bulk` looks fine, just like we expect:
```text
bulk(bulk(string-data('"1000-1"'), string-data('"c99"'), int(2), int(2)), bulk(
string-data('"1001-0"'), string-data('"c99"'), int(1), int(1)), bulk(string-data('"1001
-1"'), string-data('"c99"'), int(1), int(1)))
```

But then we declare a nicely-shaped `parts`, which is `from_redis_value`d incorrectly:

`let parts: Vec<((String, String, usize, usize),)> = from_redis_value(v)?;`

This ends up turning out to be only size 2, not 3.  :warning: :fire: 

```
 [(("1000-1", "c99", 2, 2),), (("1001-0", "c99", 1, 1),)]

...SNIP...
thread 'test_assorted_2' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `3`', tests/test_streams.rs:315:5
```

Terkwood: I believe this macro is being called as part of the processing.  Not sure if it has any bearing on the situation, as I'm still very unaccomplished with rust macros.

```rust
macro_rules! from_redis_value_for_tuple 
```

Terkwood: At least by the time that macro is invoked, it only sees 2 of the 3 elements.  Anyway, please don't take this excessive spam as any sort of implied rush!  I'll dig in a bit more before I keep posting notes :grin:   Definitely something interesting going on here

Terkwood: OK. I think you will find this interesting.  I hacked the macro and got all the tests to pass.  :smiling_imp: 

Look toward the bottom:   `while offset < items.len() /* - 1 */ {  // WE HACKED IT! :grin:`

```rust

macro_rules! from_redis_value_for_tuple {
    () => ();
    ($($name:ident,)+) => (
        #[doc(hidden)]
        impl<$($name: FromRedisValue),*> FromRedisValue for ($($name,)*) {
            // we have local variables named T1 as dummies and those
            // variables are unused.
            #[allow(non_snake_case, unused_variables)]
            fn from_redis_value(v: &Value) -> RedisResult<($($name,)*)> {
                println!("I AM A TUPLE {:?}", v);
                match *v {
                    Value::Bulk(ref items) => {
                        // hacky way to count the tuple size
                        let mut n = 0;
                        $(let $name = (); n += 1;)*
                        if items.len() != n {
                            invalid_type_error!(v, "Bulk response of wrong dimension")
                        }

                        // this is pretty ugly too.  The { i += 1; i - 1} is rust's
                        // postfix increment :)
                        let mut i = 0;
                        Ok(($({let $name = (); from_redis_value(
                             &items[{ i += 1; i - 1 }])?},)*))
                    }
                    _ => invalid_type_error!(v, "Not a bulk response")
                }
            }

            #[allow(non_snake_case, unused_variables)]
            fn from_redis_values(items: &[Value]) -> RedisResult<Vec<($($name,)*)>> {
                // hacky way to count the tuple size
                let mut n = 0;
                $(let $name = (); n += 1;)*
                if items.len() % n != 0 {
                    invalid_type_error!(items, "Bulk response of wrong dimension")
                }

                // this is pretty ugly too.  The { i += 1; i - 1} is rust's
                // postfix increment :)
                let mut rv = vec![];
                if items.len() == 0 {
                    return Ok(rv)
                }
                let mut offset = 0;
                while offset < items.len() /* - 1 */ {  // WE HACKED IT! :grin:
                    rv.push(($({let $name = (); from_redis_value(
                         &items[{ offset += 1; offset - 1 }])?},)*));
                }
                Ok(rv)
            }
        }
        from_redis_value_for_tuple_peel!($($name,)*);
    )
}```

Comment (Terkwood) [reply src/streams.rs:592]: Raised a PR for that change, specifically: #328

Comment (Terkwood) [reply src/streams.rs:592]: I created a branch to show that #328 and #319 can play nicely together, by merging the changes from #328 into the changes here.

This is the relevant commit from that branch:  https://github.com/Terkwood/redis-rs/commit/7c0d5e6bfc1cc195844fc06f5a2da1f5ff097294

**Comment by Terkwood** [comment]
I put this PR into draft status, as we want to merge #328 and remove the extraneous copies mentioned in the comments, before merging this.

**Comment by Terkwood** [comment]
Looking good with #328 merged!  I'll have _this_ PR ready for final sign-off, within the next few hours!  Thanks!  :partying_face:

File: src/streams.rs (hunk 16 above)
Comment (Terkwood) [reply src/streams.rs:592]: OK, with #328 merged I'm still running into issues.  

```text
thread 'test_assorted_2' panicked at 'called `Result::unwrap()` on an `Err` value: Response was of incompatible type: "Bulk response of wrong dimension" (response was bulk(string-data('"1000-1"'), string-data('"c99"'), int(2), int(2)))', tests/test_streams.rs:314:42
```

This is due to incorrect counting in the relevant tuple macro:

```rust
// hacky way to count the tuple size
let mut n = 0;
 $(let $name = (); n += 1;)*
  if items.len() % n != 0 {
    invalid_type_error!(items, "Bulk response of wrong dimension")
 }
```

It can be proven by adding a test:

```rust

#[test]
fn test_vec_nested_tuple() {

    let v = FromRedisValue::from_redis_value(&Value::Bulk(vec![
        Value::Bulk(vec![
            Value::Data("1".into()),
            Value::Data("2".into()),
            Value::Data("hi".into()),
            Value::Data("redis".into()),
            Value::Data("99".into()),
        ]),
        Value::Bulk(vec![
            Value::Data("3".into()),
            Value::Data("4".into()),
            Value::Data("hello".into()),
            Value::Data("rust".into()),
            Value::Data("100".into()),
        ]),
    ]))
    .expect("from");

    let w = FromRedisValue::from_redis_value(&v);

    assert_eq!(
        w,
        Ok(vec![
            (1i32, 2, "hi".to_string(), "redis".to_string(), 99),
            (3i32, 4, "hello".to_string(), "rust".to_string(), 100)
        ])
    );
}
```

Within the macro itself, I can see that `n` is incorrectly counted as 2 (the length of the OUTER vec), when it should actually be counted as 5 (the size of the inner tuples).

Apparently I didn't run tests on the experimental branch (https://github.com/Terkwood/redis-rs/commit/7c0d5e6bfc1cc195844fc06f5a2da1f5ff097294), or would have caught this ... :facepalm:   Sorry about that!

The macro-writing exceeds my current level of understanding, so I'm going to need to brush up a bit more on how the macro system works before I can contribute a very useful edit.

Let me know how you'd like to proceed?

Terkwood: It's worth noting that even with the improved version of tuple handling in #328 (shallow tuples at least get their length counted correctly, which previously wasn't the case), the test I posted in my previous comment shows that you need to call `from_redis_value` *twice*. 

This is due to the tuple being nested inside a variable-length `Vec`.

So I'm not sure that calling `from_redis_value` twice is any better than the `row[0]` version that currently used in *this* PR:

```rust
reply.ids.push(StreamPendingId {
                id: row[0].0.to_owned(),
                // ...SNIP
```

Terkwood: Raised #334 so that the issue on nested tuples-inside-vecs doesn't get lost in the shuffle, regardless of what we choose to do here.

Personally, I'd vote for merging this PR with the various, sub-optimal `row[0].0.to_owned()` calls in tact, since we would gain access to high-level wrappers for Redis Streams.

We could then pursue #334 and circle back to this, to eventually remove the copies.  Getting the macro sorted out seems pretty complicated!

Terkwood: Had another idea about this PR specifically. Maybe I can write an impl for FromRedisArgs which _only_ handles the `Vec<((String, String, usize, usize),),>` type, needed here. Will give it a try...

Terkwood: OK!  I reworked this section completely.  Converted the PR back to draft status until I can fill out the remaining TODOs.  Should have those taken care of quickly.

src/streams.rs
```diff
@@ -585,16 +585,34 @@ impl FromRedisValue for StreamPendingReply {
 
 impl FromRedisValue for StreamPendingCountReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
-        let parts: Vec<Vec<(String, String, usize, usize)>> = from_redis_value(v)?;
+        println!("spcr {:?}", v);
```
Terkwood, line 588: Removed in next commit.

src/streams.rs
```diff
@@ -0,0 +1,695 @@
...
+                }
+            }
+
+            Ok(StreamPendingReply::Data(result))
+        }
+    }
+}
+
+impl FromRedisValue for StreamPendingCountReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
```
Terkwood, line 587: This is the area that was previously under discussion.  Now manually matches the `&Value`.

src/streams.rs
```diff
@@ -0,0 +1,698 @@
...
+        let mut reply = StreamPendingCountReply::default();
+        match v {
+            Value::Bulk(outer_tuple) => {
+                for outer in outer_tuple {
```
Terkwood, line 591: Willing to rewrite this as a `flat_map` if that's your preference.  The logic is, at least, working, in this version. :microscope:

Terkwood: I'm not sure it would make things any better to take a more functional approach here.  It would appear there is the potential for nested errors that then need to be traversed in order to figure out that the function failed.

Here's a draft of that approach:

```rust
    if let Value::Bulk(outer_tuple) = v {
            let fm: Vec<Result<bool, Error>> = outer_tuple
                .into_iter()
                .map(|outer| {
                    if let Value::Bulk(inner_tuple) = outer {
                        Ok(true)
                    } else {
                        Err(todo!())
                    }
                })
                .collect();
```

Terkwood:
Is there any significant problem left with this change set which should prevent the merge?

src/streams.rs
```diff
@@ -0,0 +1,698 @@
...
+pub struct StreamKey {
+    /// The stream `key`.
+    pub key: String,
+    /// The parsed stream `id`'s.
+    pub ids: Vec<StreamId>,
+}
+
+impl StreamKey {
+    /// Return only the stream `id`'s without their data.
+    pub fn just_ids(&self) -> Vec<&String> {
```
Marwes, line 422: This method seems a bit redundant. Remove it?

src/streams.rs
```diff
@@ -0,0 +1,698 @@
...
+pub struct StreamId {
+    /// The stream `id` (entry ID) of this particular message.
+    pub id: String,
+    /// All fields in this message, associated with their respective values.
+    pub map: HashMap<String, Value>,
+}
+
+impl StreamId {
+    /// Converts a `Value::Bulk` into a `StreamId`.
+    pub fn from_bulk_value(v: &Value) -> RedisResult<Self> {
```
Marwes, line 438:
```suggestion
    fn from_bulk_value(v: &Value) -> RedisResult<Self> {
```

src/streams.rs
```diff
@@ -0,0 +1,698 @@
...
+    pub fn get<T: FromRedisValue>(&self, key: &str) -> Option<T> {
+        match self.find(&key) {
+            Some(ref x) => from_redis_value(*x).ok(),
+            None => None,
+        }
+    }
+
+    /// Fetches the value of a given field without performing any
+    /// type conversion.
+    pub fn find(&self, key: &&str) -> Option<&Value> {
```
Marwes, line 463: Having both `find` and `get` seems confusing/redundant. Users can just call `self.map.get` so perhaps this method can be removed.

src/streams.rs
```diff
@@ -0,0 +1,698 @@
...
+                    .map(|(key, entry)| {
+                        let ids = entry
+                            .into_iter()
+                            .flat_map(|id_row| {
+                                id_row.into_iter().map(|(id, map)| StreamId { id, map })
+                            })
+                            .collect();
+                        StreamKey { key, ids }
+                    })
+                    .collect::<Vec<StreamKey>>()
```
Marwes, line 500:
```suggestion
```
No need to collect before returning inside a `flat_map`

src/streams.rs
```diff
@@ -0,0 +1,698 @@
...
+
+impl FromRedisValue for StreamRangeReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let rows: Vec<HashMap<String, HashMap<String, Value>>> = from_redis_value(v)?;
+        let ids: Vec<StreamId> = rows
+            .into_iter()
+            .flat_map(|row| {
+                row.into_iter()
+                    .map(|(id, map)| StreamId { id, map })
+                    .collect::<Vec<StreamId>>()
```
Marwes, line 515:
```suggestion
```

src/streams.rs
```diff
@@ -0,0 +1,698 @@
...
+
+impl FromRedisValue for StreamClaimReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let rows: Vec<HashMap<String, HashMap<String, Value>>> = from_redis_value(v)?;
+        let ids: Vec<StreamId> = rows
+            .into_iter()
+            .flat_map(|row| {
+                row.into_iter()
+                    .map(|(id, map)| StreamId { id, map })
+                    .collect::<Vec<StreamId>>()
```
Marwes, line 530:
```suggestion
```

Terkwood:
Awesome -- thanks for the feedback! I'll work through these and hopefully have them finished by EOD.  I've got a bunch of health-related junk coming up so this is my last chance to address issues before being forced to take a break!

src/streams.rs
Terkwood: This isn't redundant.  It's used to help reduce the tediousness of operating the `XCLAIM` command.  See `test_streams.rs`:

```rust
// claim and only return JUSTID
    let claimed: Vec<String> = con
        .xclaim_options(
            "k1",
            "g1",
            "c5",
            4,
            &claim_justids,
            StreamClaimOptions::default().with_force().with_justid(),
        )
        .unwrap();
```

Terkwood: That call is made less annoying by using the mentioned method:

```rust
    let claim_justids = &reply.keys[0].just_ids();
```

Terkwood: I haven't personally worked with it yet, but the `JUSTID` option is an [explicit part of Redis Streams API](
https://redis.io/commands/xclaim):

> Moreover, as a side effect, XCLAIM will increment the count of attempted deliveries of the message unless the JUSTID option has been specified (which only delivers the message ID, not the message itself). In this way messages that cannot be processed for some reason, for instance because the consumers crash attempting to process them, will start to have a larger counter and can be detected inside the system.

Terkwood: Direct link to the existing test, for your browsing pleasure: https://github.com/Terkwood/redis-rs/blob/3f6cd92cf3f3f3e91fc3e0758dced3430c0ce38d/tests/test_streams.rs#L450

Terkwood:
OK, it looks like everything has been addressed, CI is passing, etc.  

@grippy: There's some outstanding discussion on the usage of `XCLAIM` in the comment threads.  Hopefully we can keep that as part of the PR, but if you have time, you might want to weigh in there, as it's the only place where we don't have 100% written consensus.

Terkwood:
But I don't think we necessarily need additional input to move forward, if the response above satisfies you (Marwes).  Presumably grippy favors the use of `just_ids` since it's his own creation.

It'll probably come as no surprise that I'm thus voting for an immediate merge :wink:   This PR will unlock quite a bit of potential for other users to leverage the streams API from the comfort of their rust environment :crab: :national_park:

src/streams.rs
Terkwood: Just wanted to check in on this thread.  If it helps us move toward a merge, I'm willing to play the devil's advocate against some of my statements above:  the JUSTIDs option for XCLAIM is more of an internally-focused subcommand, so I don't think there's a ton of necessity to include it _right this moment_.  From the docs:

> The command has multiple options, however most are mainly for internal use in order to transfer the effects of XCLAIM or other commands to the AOF file and to propagate the same effects to the slaves, and are unlikely to be useful to normal users

https://redis.io/commands/xclaim

So I think it could make sense to take any of the following paths, from an API design standpoint:

- go with the existing change set, which includes a nice helper method for JUSTIDs
- omit the JUSTIDs option for now, and debate its merits in a separate, future changeset (leaves us the option to add it later, in case it's desired by someone in the community)
- omit ALL of the internally-focused subcommands for XCLAIM, and raise a separate change set to add them later, as needed

Moving forward with any of these options will help us progress nicely, without painting us into a corner.

Thanks again for the thorough review!

Terkwood: Thanks for bearing with me here. I do think this can be removed. Redis itself offers logic to give just the IDs back. Removing our wrapper logic from this PR doesn't hurt much. If someone does feel strongly about this we can consider adding it back in later.

I'll push an update to remove the just_ids wrapper, which should bring us to provisional consensus among the current PR participants.

Terkwood:
OK, the additional `just_ids` method is removed.

From the [redis article](https://redis.io/commands/xclaim):

> JUSTID: Return just an array of IDs of messages successfully claimed, without returning the actual message. Using this option means the retry counter is not incremented.

So if you want the IDs only, and you want that specific behavior, you can get that.  You can just pass the `just_id` option as `true` when you call our API.

```rust
pub struct StreamClaimOptions {
    /// Set IDLE <milliseconds> cmd arg.
    idle: Option<usize>,
    /// Set TIME <mstime> cmd arg.
    time: Option<usize>,
    /// Set RETRYCOUNT <count> cmd arg.
    retry: Option<usize>,
    /// Set FORCE cmd arg.
    force: bool,
    /// Set JUSTID cmd arg. Be advised: the response
    /// type changes with this option.
    justid: bool,
}
```

@Marwes was right to point out that the small helper method isn't strictly necessary, since the server can give us what we want.  Also, I don't think the current state of the PR contradicts @grippy's previous comment -- as of the current commit, we expose all `XCLAIM` options.

Thanks again for being patient here, a lot of the comments I posted in the previous section were confused!  :grin:   But it's worth it, because we're within striking distance of having a big, new chunk of functionality exposed.  Very much appreciate the input! :rainbow:

Terkwood:
I want to push at least one more test case up before we merge, so please sit tight...

Terkwood:
OK.  It looks like `just_ids` is covered under test after all:

```rust
// claim and only return JUSTID
    let claimed: Vec<String> = con
        .xclaim_options(
            "k1",
            "g1",
            "c5",
            4,
            &claim_justids,
            StreamClaimOptions::default().with_force().with_justid(),
        )
        .unwrap();
    // we just claimed the original 10 ids
    // and only returned the ids
    assert_eq!(claimed.len(), 10);
```

So I will leave this PR in the hands of the maintainers, thanks!  The review was very helpful.

Let us know if there are additional concerns that can be addressed! :sun_behind_small_cloud:

src/streams.rs
```diff
@@ -417,13 +417,6 @@ pub struct StreamKey {
     pub ids: Vec<StreamId>,
 }
 
-impl StreamKey {
-    /// Return only the stream `id`'s without their data.
-    pub fn just_ids(&self) -> Vec<&String> {
```
Marwes: I actually just pointed this out because users can do `self.ids.iter().map(|msg| &msg.id)` themselves, which is often more effecient since the `Vec` allocation can be avoided.

But hey, whatever works :)

Terkwood:
Woo!!!! 🎉 🕺🏿 🥳

PR #322, opened 2020-05-24
fix: Return UnexpectedEof if we try to decode at eof

### LINKED ISSUE #320: RemoteError instead of IoError on killed connection
When a connection is killed via the command "CLIENT ID KILL" the killed connection does not fail with an IO Error but with a Remote Error indicating that the parser failed:

```
parse error: Parse error at 0
Unexpected `end of input`
```

The error still occurs if further commands are sent over the connection.

Sample code reproducing the issue can be found here: https://gist.github.com/chridou/d59f041540d786a22f184d0f655268ec

If an IO Error is indeed the expected error , it is possible that the`connection-manager` feature is  affected by this in possibly preventing it from reconnecting.

Marwes:
The underlying read may just return 0 (indicating eof), which will
then propagate to the next parse attempt which explicitly handles end of
file and returns the error.

Had to alter the test case so it works on redis < 5

Fixes #320

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,1 +1,2 @@
 use_try_shorthand = true
+edition = "2018"
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,5 +1,5 @@
 use std::fmt;
-use std::io::Write;
+use std::io::{self, Write};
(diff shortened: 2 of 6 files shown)


badboy:
I'll fix the clippy lints and merge this.

PR #326, opened 2020-05-30
Support *SCAN methods on async connections

swilcox3:
Following the pattern set by Iter, I implemented an AsyncIter and exposed it up through AsyncCommands.  Thanks to @Marwes who helped me a lot.
Unfortunately I'm on Windows, so I don't have redis-server installed and I can't run the tests myself.  I usually use Docker to run redis.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,14 +55,14 @@ async-trait = "0.1.24"
 
 [features]
 default = ["geospatial", "tokio-comp", "async-std-comp", "script"]
-aio = ["futures", "bytes", "pin-project-lite", "futures-util", "futures-util/sink", "tokio/sync", "tokio/stream", "tokio/tcp", "tokio/uds", "tokio/io-util", "tokio-util", "tokio-util/codec", "combine/tokio-02"]
+aio = ["bytes", "pin-project-lite", "futures-util", "futures-util/sink", "tokio/sync", "tokio/stream", "tokio/tcp", "tokio/uds", "tokio/io-util", "tokio-util", "tokio-util/codec", "combine/tokio-02"]
 tokio-rt-core = ["tokio-comp", "tokio/rt-core"]
 geospatial = []
 cluster = ["crc16", "rand"]
 script = ["sha1"]
 async-std-comp = ["aio", "async-std"]
 tokio-comp = ["aio", "tokio"]
-connection-manager = ["tokio-rt-core", "arc-swap", "futures"]
+connection-manager = ["futures", "tokio-rt-core", "arc-swap", "futures"]
 
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ async-trait = "0.1.24"
 
 [features]
 default = ["geospatial", "tokio-comp", "async-std-comp", "script"]
-aio = ["bytes", "pin-project-lite", "futures-util", "futures-util/sink", "tokio/sync", "tokio/stream", "tokio/tcp", "tokio/uds", "tokio/io-util", "tokio-util", "tokio-util/codec", "combine/tokio-02"]
+aio = ["futures", "bytes", "pin-project-lite", "futures-util", "futures-util/sink", "tokio/sync", "tokio/stream", "tokio/tcp", "tokio/uds", "tokio/io-util", "tokio-util", "tokio-util/codec", "combine/tokio-02"]
 tokio-rt-core = ["tokio-comp", "tokio/rt-core"]
 geospatial = []
 cluster = ["crc16", "rand"]
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -455,8 +455,7 @@ impl ConnectionLike for Connection {
             if self.pubsub {
                 self.exit_pubsub().await?;
             }
-            self.buf.clear();
-            self.buf.copy_from_slice(cmd);
+            self.buf = Vec::from(cmd);
             self.con.write_all(&self.buf).await?;
             self.read_response().await
         })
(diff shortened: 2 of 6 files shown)


src/cmd.rs
```diff
@@ -71,6 +71,59 @@ impl<'a, T: FromRedisValue> Iterator for Iter<'a, T> {
...
+    /// ```rust,no_run
+    /// let iter = conn.sscan("set_key").await.unwrap();
+    /// let stream = iter.stream();
+    /// futures::pin_mut!(stream);
+    /// while let Some(element) = stream.next().await {
+    ///    //Do something with element
+    /// }
+    /// ```
+    #[inline]
+    pub fn stream(self) -> impl futures::Stream<Item = T> + 'a {
```
Marwes, line 98:
```suggestion
    pub fn into_stream(self) -> impl futures::Stream<Item = T> + 'a {
```

src/cmd.rs
```diff
@@ -71,6 +71,59 @@ impl<'a, T: FromRedisValue> Iterator for Iter<'a, T> {
     }
 }
 
+use crate::aio::ConnectionLike as AsyncConnection;
+
+/// Represents a redis iterator that can be used with async connections.  
+#[cfg(feature = "aio")]
+pub struct AsyncIter<'a, T: FromRedisValue + 'a> {
+    batch: Vec<T>,
```
Marwes, line 79: Instead of reversing and storing a `Vec` here you could store `Vec`s `IntoIter` and call `next` instead of `pop` below here. Same could be done for `Iterator` variant.

src/cmd.rs
```diff
@@ -71,6 +71,59 @@ impl<'a, T: FromRedisValue> Iterator for Iter<'a, T> {
...
+    ///    //Do something with element
+    /// }
+    /// ```
+    #[inline]
+    pub fn stream(self) -> impl futures::Stream<Item = T> + 'a {
+        // we need to do this in a loop until we produce at least one item
+        // or we find the actual end of the iteration.  This is necessary
+        // because with filtering an iterator it is possible that a whole
+        // chunk is not matching the pattern and thus yielding empty results.
+        futures::stream::unfold(self, |mut cur_self| async move {
```
Marwes, line 103: Could extract this async block to a `next` function on `AsyncIter`. Then the `futures` dependency is a bit more "opt-in".

src/aio.rs
```diff
@@ -427,7 +427,7 @@ fn get_socket_addrs(host: &str, port: u16) -> RedisResult<SocketAddr> {
 }
 
 /// An async abstraction over connections.
-pub trait ConnectionLike: Sized {
+pub trait ConnectionLike: Send {
```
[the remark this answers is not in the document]
swilcox3: Fixed.

src/cmd.rs
swilcox3: I'd rather not change existing functionality as a first-time contriubtor, so I think it's best to follow the pattern.

src/cmd.rs
swilcox3: Done.

swilcox3:
Oh, this doesn't work because the implementations of aio::ConnectionLike are calling `get_packed_command`, not `get_packed_command_with_cursor`.  I'll fix that.

swilcox3:
OK fixed.  It's a breaking change though because I had to change the signature of `aio::ConnectionLike` to match the signature of `ConnectionLike`, replacing the `Cmd` with `&[u8]`.

src/aio.rs
```diff
@@ -430,7 +430,7 @@ fn get_socket_addrs(host: &str, port: u16) -> RedisResult<SocketAddr> {
 pub trait ConnectionLike {
     /// Sends an already encoded (packed) command into the TCP socket and
     /// reads the single response from it.
-    fn req_packed_command<'a>(&'a mut self, cmd: &'a Cmd) -> RedisFuture<'a, Value>;
+    fn req_packed_command<'a>(&'a mut self, cmd: &'a [u8]) -> RedisFuture<'a, Value>;
```
Marwes, line 433: Instead of changing this signature you should be able to update the `cursor` inside the `Cmd`. Remove the cursor inside `AsyncIter` and update the one in `Cmd` instead.

(There are downstream crates like https://github.com/marwes/redis-cluster-rs which rely on `Cmd` being passed in.

src/aio.rs
```diff
@@ -455,8 +455,7 @@ impl ConnectionLike for Connection {
             if self.pubsub {
                 self.exit_pubsub().await?;
             }
-            self.buf.clear();
-            self.buf.copy_from_slice(cmd);
```
Marwes: There is also `extend_from_slice` which would reuse the allocation.

swilcox3:
Finally, got the tests to pass.  Man, it's crippling when you can't run them all.  The only failure now is clippy saying that we should name the `next()` method something else because it resembles Iterator.  I like that it resembles the `Stream` syntax, but it could definitely be confusing.  Should we rename the function or allow the warning?

Marwes:
I'd rename it to `next_item` or something then.

I guess you could get tests working locally if you tweak the tests to read an environment variable which contains a custom command to run instead of redis and use that start redis in a docker instance,

Cargo.toml
```diff
@@ -62,7 +62,7 @@ cluster = ["crc16", "rand"]
 script = ["sha1"]
 async-std-comp = ["aio", "async-std"]
 tokio-comp = ["aio", "tokio"]
-connection-manager = ["tokio-rt-core", "arc-swap", "futures"]
+connection-manager = ["futures", "tokio-rt-core", "arc-swap", "futures"]
```
Terkwood, line 65: Is this duplicate value somehow necessary?  Curious as I'm working on a merge into #319.  I think this can be trimmed.  

Thanks!

Marwes: Yeah, that can probably be removed.

Terkwood: Cool, I've updated the #319.  Cheers.

PR #327, opened 2020-06-02
Remove some unnecessary overhead around Value conversions

Marwes:
Two small improvements and added a pair of helper functions for working with sequences and maps inside `Value` which could be useful to avoid intermediate structures like https://github.com/mitsuhiko/redis-rs/pull/319/files#diff-6162c7f34c4c3cad920338d2f8d00903R483 (cc @Terkwood )

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -8,6 +8,22 @@ use std::hash::{BuildHasher, Hash};
 use std::io;
 use std::str::{from_utf8, Utf8Error};
 
+macro_rules! invalid_type_error {
+    ($v:expr, $det:expr) => {{
+        fail!(invalid_type_error_inner!($v, $det))
+    }};
+}
+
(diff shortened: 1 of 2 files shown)


src/types.rs
```diff
@@ -1032,17 +1067,10 @@ impl<K: FromRedisValue + Eq + Hash, V: FromRedisValue, S: BuildHasher + Default>
...
-                let mut rv = HashMap::default();
-                let mut iter = items.iter();
-                while let (Some(k), Some(v)) = (iter.next(), iter.next()) {
-                    rv.insert(from_redis_value(k)?, from_redis_value(v)?);
-                }
-                Ok(rv)
-            }
-            _ => invalid_type_error!(v, "Response type not hashmap compatible"),
-        }
+        v.as_map_iter()
```
Terkwood, line 1070: :+1:

PR #328, opened 2020-06-04
Count items correctly in tuple macro

Terkwood:
Greetings!

I _think_ this fixes the macro so that it's correct.  Found as a result of working through #319.

See this portion of the PR, specifically: https://github.com/mitsuhiko/redis-rs/pull/319#discussion_r435184148

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -179,7 +179,7 @@ macro_rules! implement_commands {
                 }
             )*
 
-            /// Incrementally iterate the keys space.  
+            /// Incrementally iterate the keys space.
             #[inline]
             fn scan<RV: FromRedisValue>(&mut self) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
                 let mut c = cmd("SCAN");
@@ -204,7 +204,7 @@ macro_rules! implement_commands {
             }
 
             /// Incrementally iterate hash fields and associated values for
-            /// field names matching a pattern. 
+            /// field names matching a pattern.
             #[inline]
             fn hscan_match<K: ToRedisArgs, P: ToRedisArgs, RV: FromRedisValue>
                     (&mut self, key: K, pattern: P) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
@@ -213,7 +213,7 @@ macro_rules! implement_commands {
                 Box::pin(async move {c.iter_async(self).await })
             }
 
-            /// Incrementally iterate set elements.  
+            /// Incrementally iterate set elements.
             #[inline]
             fn sscan<K: ToRedisArgs, RV: FromRedisValue>(&mut self, key: K) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
                 let mut c = cmd("SSCAN");
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -230,7 +230,7 @@ macro_rules! implement_commands {
                 Box::pin(async move {c.iter_async(self).await })
             }
 
-            /// Incrementally iterate sorted set elements.  
+            /// Incrementally iterate sorted set elements.
             #[inline]
             fn zscan<K: ToRedisArgs, RV: FromRedisValue>(&mut self, key: K) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
                 let mut c = cmd("ZSCAN");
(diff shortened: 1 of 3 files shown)


src/commands.rs
```diff
@@ -179,7 +179,7 @@ macro_rules! implement_commands {
                 }
             )*
 
-            /// Incrementally iterate the keys space.  
+            /// Incrementally iterate the keys space.
```
Terkwood, line 182: These spacing issues aren't related to my change, but they're preventing the build from passing.

Marwes:
Can you add a test to validate it?

Terkwood:
_Absolutely!_

src/types.rs
```diff
@@ -1154,7 +1154,7 @@ macro_rules! from_redis_value_for_tuple {
                     return Ok(rv)
                 }
                 let mut offset = 0;
-                while offset < items.len() - 1 {
+                while offset < items.len() {
```
Marwes, line 1157:
```rust
                for chunk in items.chunks_exact(n) {
                    let mut offset = 0;
                    rv.push(($({let $name = (); from_redis_value(
                         &chunk[{ offset += 1; offset - 1 }])?},)*));
                }
```

Marwes: Update the example a bit more @Terkwood

Terkwood: I had to massage the syntax on this a little bit, and it generates some pretty crazy errors.  I caught one of your previous iterations in the comments and those seemed a bit more benign 

```text
error[E0308]: mismatched types
    --> src/types.rs:1159:50
     |
1159 |                          [$($name),*] => rv.push(($(from_redis_value($name)),*),
),
     |                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
expected tuple, found enum `std::result::Result`
...
1182 | from_redis_value_for_tuple! { T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12,
 }
     | --------------------------------------------------------------------------------
-- in this macro invocation
     |
     = note: expected tuple `(T12,)`
                 found enum `std::result::Result<_, types::RedisError>`
     = note: this error originates in a macro (in Nightly builds, run with -Z macro-bac
ktrace for more info)

error: aborting due to 78 previous errors
```

Terkwood: Currently using

```rust
for chunk in items.chunks_exact(n) {
                    match chunk {
                         [$($name),*] => rv.push(($(from_redis_value($name)),*),),
                         _ => unreachable!(),
                     }
                }
```

Terkwood: `if let Ok(a)` to grab those inner values from the `$(from_redis_value($name))` might work fine, I'm just trying to tread lightly here :grin:

Marwes: I think that should be (missing `?` to unpack the results)

```
for chunk in items.chunks_exact(n) {
    match chunk {
        [$($name),*] => rv.push(($(from_redis_value($name)?),*),),
         _ => unreachable!(),
    }
}
```

Terkwood: Just got this version pushed, and the tests are running now in Travis &c.  :sunglasses:

tests/test_types.rs
```diff
@@ -73,6 +73,19 @@ fn test_vec() {
     assert_eq!(v, Ok(vec![1i32, 2, 3]));
 }
 
+#[test]
```
Marwes, line 76: This test actually passes on master, you need to wrap the result in a `Vec` to make it use from_redis_value**s**

```

#[test]
fn test_tuple() {

    let v = FromRedisValue::from_redis_value(&Value::Bulk(vec![Value::Bulk(vec![
        Value::Data("1".into()),
        Value::Data("2".into()),
        Value::Data("3".into()),
    ])]));

    assert_eq!(v, Ok(vec![(1i32, 2, 3,)]));
}

```

Terkwood: I was able to assert _two_ levels of nested tuples using `let v` as you described: `assert_eq!(v, Ok(((1i32, 2, 3,),)));`

But this assertion failed:  `assert_eq!(v, Ok(vec![(1i32, 2, 3,)]));`

I'll check this against master branch too

Terkwood: Ok, using your updated test case on master definitely fails, there:

```text
thread 'test_tuple' panicked at 'assertion failed: `(left == right)`
  left: `Err(Response was of incompatible type: "Bulk response of wrong dimension" (response was [bulk(string-data('"1"'), string-data('"2"'), string-data('"3"'))]))`,
 right: `Ok([(1, 2, 3)])`', tests/test_types.rs:86:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So it seems like we're making progress :grin:

PR #335, opened 2020-06-07
🚧 Fix macro handling of tuples nested in vecs 🚧

### LINKED ISSUE #334: FromRedisValue fails on tuples nested inside Vecs
# Problem statement 

The `FromRedisValue` macro which counts the size of tuples gets confused when tuples are nested inside vecs: https://github.com/mitsuhiko/redis-rs/blob/494767a0adcc3b8b7ab74def216c35608af04207/src/types.rs#L1137

It can be proven by adding a test:

```rust

#[test]
fn test_vec_nested_tuple() {

    let v = FromRedisValue::from_redis_value(&Value::Bulk(vec![
        Value::Bulk(vec![
            Value::Data("1".into()),
            Value::Data("2".into()),
            Value::Data("hi".into()),
            Value::Data("redis".into()),
            Value::Data("99".into()),
        ]),
        Value::Bulk(vec![
            Value::Data("3".into()),
            Value::Data("4".into()),
            Value::Data("hello".into()),
            Value::Data("rust".into()),
            Value::Data("100".into()),
        ]),
    ]))
    .expect("from");

    let w = FromRedisValue::from_redis_value(&v);

    assert_eq!(
        w,
        Ok(vec![
            (1i32, 2, "hi".to_string(), "redis".to_string(), 99),
            (3i32, 4, "hello".to_string(), "rust".to_string(), 100)
        ])
    );
}
```

```text
---- test_vec_nested_tuple stdout ----
thread 'test_vec_nested_tuple' panicked at 'assertion failed: `(left == right)`
  left: `Err(Response was of incompatible type: "Bulk response of wrong dimension" (response was [bulk(string-data('"1"'), string-data('"2"'), string-data('"hi"'), string-data('"redis"'), string-data('"99"')), bulk(string-data('"3"'), string-data('"4"'), string-data('"hello"'), string-data('"rust"'), string-data('"100"'))]))`,
 right: `Ok([(1, 2, "hi", "redis", 99), (3, 4, "hello", "rust", 100)])`', tests/test_types.rs:113:5
```

Within the macro itself, we can see that `n` is incorrectly counted as 2 (the length of the OUTER vec), when it should actually be counted as 5 (the size of the inner tuples).

## Additional considerations

It's worth noting that dealing with tuples nested inside of vecs requires *two* explicit calls to `from_redis_value`.

Terkwood:
Under construction.   Right now all we have is a test to highlight the bug.

If we can get this completed, it resolves #334.

### DIFF
--- a/tests/test_types.rs
+++ b/tests/test_types.rs
@@ -87,6 +87,26 @@ fn test_tuple() {
 }
 
 #[test]
+fn test_nested_single_tuple() {
+    use redis::{FromRedisValue, Value};
+
+    let v =
+        FromRedisValue::from_redis_value(&Value::Bulk(vec![Value::Bulk(vec![Value::Bulk(vec![
+            Value::Data("1".into()),
+            Value::Data("2".into()),
+            Value::Data("hi".into()),
+            Value::Data("redis".into()),
+            Value::Data("99".into()),
+        ])])]));
+    assert_eq!(
+        v,
+        Ok(vec![
+            ((1i32, 2, "hi".to_string(), "redis".to_string(), 99),)
+        ])
+    );
+}
+
+#[test]
 fn test_vec_nested_tuple() {
     use redis::{FromRedisValue, Value};
 
--- a/tests/test_types.rs
+++ b/tests/test_types.rs
@@ -87,6 +87,39 @@ fn test_tuple() {
 }
 
 #[test]
+fn test_vec_nested_tuple() {
+    use redis::{FromRedisValue, Value};
+
+    let v = FromRedisValue::from_redis_value(&Value::Bulk(vec![
+        Value::Bulk(vec![
+            Value::Data("1".into()),
+            Value::Data("2".into()),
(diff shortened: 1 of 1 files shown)


Terkwood:
Hey @Marwes, I threw you an invite to collaborate on my fork of the repository in case you have any ideas about what to do with the nesting considerations in this macro.  Feel free to ignore it -- absolutely no pressure!  :star:    But I figured it might be less hassle than trying to send detail back and forth via comments & markdown.

Terkwood:
Ahhhhh, maybe this can be solved by implementing a macro which specifically deals with the Vec/Tuple nesting.... And having its output `Impl` declared before the more generic, existing macro. 🤔

I'll give it a try today and see how far I can get

Terkwood:
Made an attempt but got stuck

Terkwood:
Closing the PR for now but we have the spam on record if desired. will leave branch open

Issue #330, opened 2020-06-05
Support Access Control Lists (ACL)


Terkwood:
We want to support [ACLs, now available in Redis 6.0](https://redis.io/topics/acl).

This feature is a sub-task of the overall effort to support Redis 6.0 features.  See #312. 

Please feel free to submit technical detail in the comments, and we'll work your info into the description 

## Tasks

Is there anything to do which isn't listed below? 

- [x] provide auth command #341 
- [ ] implement ACL commands

### Add these  ACL commands 

ACL CAT
ACL DELUSER
ACL GENPASS
ACL GETUSER
ACL HELP
ACL LIST
ACL LOAD
ACL LOG
ACL SAVE
ACL SETUSER
ACL USERS
ACL WHOAMI

Terkwood:
Making progress! 🏁

Terkwood:
https://github.com/keroxp/deno-redis/pull/90

Some of the ACL commands we need to implement are detailed here. (There are also some unrelated commands, but it's worth taking a look for 2 seconds)

Terkwood:
Updated the description

PR #341, opened 2020-06-13
Redis 6 AUTH <username> <password>

weihanglo:
## Implement Redis 6 `AUTH <username> <password>`

Related #330

The unix socket URL parsing is wrong (see #132). Would like to fix it but we need discussion the URL format. Seems `url` crate does not support empty host (see https://github.com/servo/rust-url/issues/25), the alternative format with queries may be acceptable:

```
redix+unix:///<path>[?db=<db>[&pass=<password>][&user=<username>]]
```

## Breaking changes

Possibly no. If an URL does not contain `username`, it would behave in the old fashion.

Any input are welcome.

### DIFF
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -380,7 +380,11 @@ where
     C: ConnectionLike,
 {
     if let Some(passwd) = &connection_info.passwd {
-        match cmd("AUTH").arg(passwd).query_async(con).await {
+        let mut command = cmd("AUTH");
+        if let Some(username) = &connection_info.username {
+            command.arg(username);
+        }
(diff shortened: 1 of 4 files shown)


Terkwood:
Super awesome! Excited to take a closer look, thanks for submitting it 🌟 ✨ 🌠

Terkwood:
Looking good 🔥

Terkwood:
Listed this as a subtask in #330, I'll mark it off the list once there's a merge 🙏

Terkwood:
Thanks so much, @weihanglo!

PR #345, opened 2020-06-23
Add Redis 6 ACL commands support

### LINKED ISSUE #330: Support Access Control Lists (ACL)

weihanglo:
## What

Resolves #330

Add high-level command of all [`ACL` commands](https://redis.io/topics/acl) introduced in Redis 6.

## Tasks

- [x] a `Rule` enum represents arbitrary ACL rules.
- [x] a `AclInfo` struct representing response from `ACL GETUSER` command.
- [x] introduce a new `acl` feature and set as a default feature.
- [x] integration tests for `acl` commands

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,8 @@ async-std = { version = "1.5.0", optional = true}
 async-trait = "0.1.24"
 
 [features]
-default = ["streams", "geospatial", "tokio-comp", "async-std-comp", "script"]
+default = ["acl", "streams", "geospatial", "tokio-comp", "async-std-comp", "script"]
+acl = []
 aio = ["bytes", "pin-project-lite", "futures-util", "futures-util/sink", "tokio/sync", "tokio/stream", "tokio/tcp", "tokio/uds", "tokio/io-util", "tokio-util", "tokio-util/codec", "combine/tokio-02"]
 tokio-rt-core = ["tokio-comp", "tokio/rt-core"]
 geospatial = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,9 @@ required-features = ["async-std-comp"]
 name = "parser"
 required-features = ["aio"]
 
+[[test]]
+name = "test_acl"
+
 [[bench]]
 name = "bench_basic"
 harness = false
--- a/src/acl.rs
+++ b/src/acl.rs
@@ -2,6 +2,16 @@
 
 use crate::types::{ErrorKind, FromRedisValue, RedisResult, RedisWrite, ToRedisArgs, Value};
 
+macro_rules! not_convertible_type_error {
+    ($v:expr, $det:expr) => {{
+        fail!((
+            ErrorKind::TypeError,
+            "Response type not convertible to Rule",
+            format!("{:?} (response was {:?})", $det, $v)
+        ))
+    }};
+}
+
 /// ACL rules are used in order to activate or remove a flag, or to perform a
 /// given change to the user ACL, which under the hood are just single words.
 #[derive(Debug, Eq, PartialEq)]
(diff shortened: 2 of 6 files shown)


weihanglo:
No idea why travis-ci was not triggered. Could it rerun manually?

Terkwood:
If nothing else, you can re-trigger it with a trivial commit (you might not like that approach)

weihanglo:
@Terkwood 
`redis-server` on travis-ci need an update. I am fixing it.

https://travis-ci.org/github/mitsuhiko/redis-rs/builds/701296340

Terkwood:
Ah, right. Yeah the clippy is a little bit strict!

Terkwood:
In the world of wishes and dreams, it might be good for us to have an Issue Template(or a section in the README) that includes instructions for how the devs can run clippy locally. I worked thru the  same type of issue myself - - it's not too hard to figure out, but it probably costs everyone a few minutes of time, and having some explicit instructions on how to run the lints locally might help alleviate that.

Apologies that this is a little off topic... I'm not yet motivated to do it myself, so it seems better to mention it here than load up the maintainers with yet another issue 😉

weihanglo:
I am thinking of these three solutions to fix CI issue (need Redis 6 support environment).

- Travis bundles Redis v5 as well as GitHub Action does. No Redis-6 spports yet.
- Our integration tests run `redis-server` command directly. And that makes running a dockerized Redis non-trivial.
- Directly download Redis 6 and compile it is also a weird approach.

Need some mentorings to fix CI issue. Thanks!

Terkwood:
This doesn't answer your question, but one workaround could be to feature-gate the ACL commands so that they're _opt-in_ rather than _opt-out_.  We could then make ACL support a  default feature set once we're able to advance the CI environment to Redis 6.

I fear that your PR might otherwise languish if we make it depend on the CI environment being updated to Redis 6...

Cargo.toml
```diff
@@ -54,7 +54,8 @@ async-std = { version = "1.5.0", optional = true}
 async-trait = "0.1.24"
 
 [features]
-default = ["streams", "geospatial", "tokio-comp", "async-std-comp", "script"]
+default = ["acl", "streams", "geospatial", "tokio-comp", "async-std-comp", "script"]
```
Terkwood, line 57: I would remove `"acl"` from this list so that we can cope with the current CI environment.  Not an awesome suggestion, but I think having the ACL changes in the main branch are worth the compromise.

src/commands.rs
```diff
@@ -10,6 +10,9 @@ use crate::geo;
 #[cfg(feature = "streams")]
 use crate::streams;
 
+#[cfg(feature = "acl")]
```
Terkwood, line 13: :+1:

src/commands.rs
```diff
@@ -905,6 +908,128 @@ implement_commands! {
...
+
+    /// Clears the ACL log.
+    #[cfg(feature = "acl")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "acl")))]
+    fn acl_log_reset<>() {
+        cmd("ACL").arg("LOG").arg("RESET")
+    }
+
+    /// Returns a helpful text describing the different subcommands.
+    #[cfg(feature = "acl")]
```
Terkwood, line 1027: Everything is feature-config'd nicely here, thanks.

tests/test_acl.rs
```diff
@@ -0,0 +1,159 @@
+#![cfg(feature = "acl")]
```
Terkwood, line 1: Given that you have the feature configuration set up correctly, the test should pass CI if we choose to make ACL a feature that's not included by default.

Terkwood:
@weihanglo No matter how you choose to move forward with this on your PR, I did raise an issue in #347.  There might be some hope for installing a newer version of `redis-server` on the debian linux environment that Travis is contained in.  We would need to use the backports version of the debian package to get access to Redis 6.

Terkwood:
Added a note in #347:  in `deno-redis`, we're using Github Actions and pulling down a docker container which has Redis 6.  So we're able to test ACL there.

Terkwood:
Resolves #347.

src/acl.rs
```diff
@@ -0,0 +1,269 @@
...
+    where
+        W: ?Sized + RedisWrite,
+    {
+        use self::Rule::*;
+
+        let rule = match self {
+            On => "on".to_owned(),
+            Off => "off".to_owned(),
+
+            AddCommand(cmd) => "+".to_owned() + cmd,
```
Marwes, line 77: Created https://github.com/mitsuhiko/redis-rs/pull/348 which would allow you to avoid these allocations

```
out.write_arg_fmt(format_args!("+{}", cmd))
```

src/acl.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Describes patterns of keys which the user can access. Represented by
+    /// [`Rule::Pattern`][1].
+    ///
+    /// [1]: ./enum.Rule.html#variant.Pattern
+    pub keys: Vec<Rule>,
+}
+
+impl FromRedisValue for AclInfo {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let values: Vec<Value> = FromRedisValue::from_redis_value(v)?;
```
Marwes, line 140: You can use `let values = v.as_sequence().ok_or_else(|| ..)?;` here.

src/acl.rs
```diff
@@ -0,0 +1,269 @@
...
+        let mut it = values.into_iter().skip(1).step_by(2);
+        let (flags, passwords, commands, keys) = match (it.next(), it.next(), it.next(), it.next())
+        {
+            (Some(flags), Some(passwords), Some(commands), Some(keys)) => {
+                // Parse flags
+                // Ref: https://git.io/JfNyb
+                let flag_values: Vec<Vec<u8>> = FromRedisValue::from_redis_value(&flags)?;
+                let mut flags = Vec::with_capacity(flag_values.len());
+                for flag in flag_values.into_iter() {
+                    let rule = match flag.as_slice() {
```
Marwes, line 150:
```suggestion
                let flag_values = flags.as_sequence().ok_or_else(|| ..)?;
                let mut flags = Vec::with_capacity(flag_values.len());
                for flag in flag_values {
                    let flag = match flag {
                        Value::Data(flag) => flag,
                        _ => return Err("Expected bytes".into()),
                    };
                    let rule = match flag {
```

Marwes: Could even use `collect::<Result<_, _>>()` instead of using `with_capacity + push`

src/acl.rs
```diff
@@ -0,0 +1,269 @@
...
+                        _ => not_convertible_type_error!("", "Expect an ACL flag"),
+                    };
+                    flags.push(rule);
+                }
+
+                let passwords: Vec<String> = FromRedisValue::from_redis_value(&passwords)?;
+                let passwords = passwords
+                    .into_iter()
+                    .map(Rule::AddHashedPass)
+                    .collect::<Vec<Rule>>();
```
Marwes, line 165:
```suggestion
                let passwords: Vec<String> = FromRedisValue::from_redis_value(&passwords)?;
                let passwords = passwords
                    .as_sequence()
                    .ok_or_else(|| ..)
                    .map(|password| Ok(Rule::AddHashedPass(String::from_redis_value(password)?)))
                    .collect::<Result<Vec<Rule>, _>>();
```

src/acl.rs
```diff
@@ -0,0 +1,269 @@
...
+                        x if x.starts_with('-') => Rule::RemoveCommand(x[1..].to_owned()),
+                        _ => not_convertible_type_error!(
+                            command,
+                            "Expect a command addition/removal"
+                        ),
+                    };
+                    commands.push(rule);
+                }
+
+                let keys: Vec<String> = FromRedisValue::from_redis_value(&keys)?;
```
Marwes, line 183: Nit: Could use `as_sequence`

weihanglo:
@Marwes 
These suggestions are awesome! Tried my best to avoid redundant heap allocation at 44241dc and 7135040. Please take a look.

gferon:
> @weihanglo No matter how you choose to move forward with this on your PR, I did raise an issue in #347. There might be some hope for installing a newer version of `redis-server` on the debian linux environment that Travis is contained in. We would need to use the backports version of the debian package to get access to Redis 6.

I've been using this Ubuntu PPA in production for years now: https://launchpad.net/~chris-lea/+archive/ubuntu/redis-server and it has `redis` 6.0

Marwes:
@weihanglo Thanks!

PR #350, opened 2020-06-27
Use XACK in streams example

Terkwood:
This change set significantly improves the value of the streams example.

Unfortunately I did not include the call to `XACK` in the first pass!  I apologize for that!

Without some use of `XACK` to [signal back to Redis that each message has been correctly processed](https://redis.io/commands/xack), the example is misleading.

### DIFF
--- a/examples/streams.rs
+++ b/examples/streams.rs
@@ -82,9 +82,30 @@ fn demo_group_reads(client: &redis::Client) {
             }
 
             for _ in 0..repeat {
-                read_group_records(&ca, *slowness).expect("group read");
+                let read_reply = read_group_records(&ca, *slowness).expect("group read");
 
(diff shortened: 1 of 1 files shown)


examples/streams.rs
```diff
@@ -235,21 +256,7 @@ fn read_group_records(client: &redis::Client, slowness: u8) -> RedisResult<()> {
         )
         .expect("records");
 
-    for StreamKey { key, ids } in srr.keys {
```
Terkwood, line 238: Arbitrarily moved this code block.  The placement doesn't matter very much since it's an example, but it lines up somewhat nicely with the fact that `XACK` expects a group of IDs tied to a particular stream key (as opposed to being able to acknowledge multiple sets of IDs across multiple keys).

PR #356, opened 2020-07-08
Use password from query string for unix instead of password param which is not supported by url::Url

oghaffarinia:
Using a password for unix protocol is not working from url, this can fix it by reading password from query string instead


weihanglo:
@badboy @Marwes 
I think this PR is ready to merge. Could you please take a look? 😃

oghaffarinia:
Any updates on this?

Marwes:
Thanks! :+1:

PR #359, opened 2020-07-10
Upgrade the dependencies

Hirrolot:


Cargo.toml
```diff
@@ -18,45 +18,45 @@ all-features = true
 # These two are generally really common simple dependencies so it does not seem
 # much of a point to optimize these, but these could in theory be removed for
 # an indirection through std::Formatter.
-dtoa = "0.4"
```
Marwes, line 21: Bumping the patch version only restricts downstream users to that patch version or up. Unless we know that the version is in fact a minimum we need shouldn't increase it but instead let users do `cargo update` if they want to update.

Cargo.toml
```diff
@@ -18,45 +18,45 @@ all-features = true
...
-percent-encoding = "2.1"
+percent-encoding = "2.1.0"
 
 # We need this for redis url parsing
-url = "2.1"
+url = "2.1.1"
 
 # We need this for script support
-sha1 = { version = ">= 0.2, < 0.7", optional = true }
+sha1 = { version = "0.6.0", optional = true }
```
Marwes, line 31: This just forces users to the 0.6 version, which could lead to multiple sha1 crates in the dependency graph. Not a big problem but I see no reason to change this to be more restrictive.

Hirrolot:
It seems that yes, but when I typed this in `Cargo.toml`:

```
redis = "0.16.0"
tokio = "0.2.21"
```

It didn't work, complaining that it cannot resolve dependencies due to incompatible versions of some crates in `redis` and `tokio`, but now I can't reproduce this.

PR #371, opened 2020-08-09
Adds a way to MONITOR a Redis connection in async

michael-grunder:
This is similar to the PubSub helper only it's for using the Redis MONITOR command.

Example usage:

```rust
let mut monitor = con.into_monitor();
monitor.monitor().await?;

let mut stream = monitor.on_message::<String>();

while let Some(msg) = stream.next().await {
    println!("Monitor: {}", msg);
}
```

### DIFF
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -47,7 +47,7 @@ use crate::connection::{ConnectionAddr, ConnectionInfo};
 
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use crate::parser::ValueCodec;
-use crate::types::{ErrorKind, RedisError, RedisFuture, RedisResult, Value};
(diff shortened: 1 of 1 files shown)


michael-grunder:
I'm not sure whether the PR is useful generally but it looks like the tests failed pulling the Redis PPA.

It adds a bit of time to CI but what we do in [hiredis](https://github.com/redis/hiredis/blob/master/.travis.yml#L19) is pull 6.0.6 from GitHub and build it with TLS support meaning we don't need stunnel.

weihanglo:
> Edit: Adds less to build time than [I would've guessed](https://travis-ci.org/github/michael-grunder/redis-rs)

It seems reasonable to build Redis from source. Could you file a pull request to update CI configs? Either update to 6.0.6 or from source is great!

PR #372, opened 2020-08-09
Boolean implementation fixes

### LINKED ISSUE #324: bool mismatched in ToRedisArgs and FromRedisValue
[ToRedisArgs](https://docs.rs/redis/0.16.0/src/redis/types.rs.html#691-698) is converting boolean arguments into `b"true"` and `b"false"`.

That's all well and good, except [FromRedisValue](https://docs.rs/redis/0.16.0/src/redis/types.rs.html#977-995) is checking for an integer of 1 or 0.

This means any bool value of true stored using the [ToRedisArgs](https://docs.rs/redis/0.16.0/src/redis/types.rs.html#691-698) trait will ultimately return `Response was of incompatible type: "Response type not bool compatible." (response was string-data('"true"'))` when attempting to retrieve the same bool value using [FromRedisValue](https://docs.rs/redis/0.16.0/src/redis/types.rs.html#977-995).

UnicodingUnicorn:
Should resolve #324. I changed the `ToRedisArgs` implementation of booleans to encode them as `1` or `0` as redis does, and their `FromRedisValue` implementation to also parse from `Value::Data`. The latter change means that calling GET on boolean values should stop dying now.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -765,7 +765,7 @@ impl ToRedisArgs for bool {
     where
         W: ?Sized + RedisWrite,
     {
-        out.write_arg(if *self { b"true" } else { b"false" })
+        out.write_arg(if *self { b"1" } else { b"0" })
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/types.rs
```diff
@@ -1057,6 +1057,16 @@ impl FromRedisValue for bool {
                     invalid_type_error!(v, "Response status not valid boolean");
                 }
             }
+            Value::Data(ref bytes) => {
+                let res = from_utf8(bytes)?.to_string();
+                if res == "1" {
+                    Ok(true)
+                } else if res == "0" {
```
weihanglo, line 1064: `Vec<u8>` can compare to byte literal `b"1"`, whose type is `&'static [u8; 1]`. No additional allocation needed 😊
```suggestion
                if bytes == b"1" {
                    Ok(true)
                } else if bytes == "0" {
```

UnicodingUnicorn:
Done!

gferon:
Since #373 was merged, this should just need a merge commit from `master` to get green!

PR #373, opened 2020-08-13
Build Redis with TLS support in CI

michael-grunder:
Build Redis 6.0.6 from source in .travis.yml

This makes it possible to also remove stunnel but that will require more substantial changes to the integration tests


weihanglo:
Thanks! It seems not hurt the build time too much. 🎉 
@Terkwood what's your opinions of these CI changes?

PR #378, opened 2020-09-08
Break the loop after processing the initial results, when cursor == None

### LINKED ISSUE #377: AsyncIter seems to produce duplicate results
```toml
[dependencies]
tokio = { version = "^0.2.22", features = ["full"]}
redis = { version = "^0.17.0" }
```

```shell
rustc 1.46.0 (04488afe3 2020-08-24)
Redis 5.0.6 + Redis 6.0.7
```

The below example is a slightly modified version of the documented AsyncIter example (https://docs.rs/redis/0.17.0/redis/struct.AsyncIter.html#method.next_item):

```rust
con.sadd("my_set", 42i32).await?;
con.sadd("my_set", 43i32).await?;
let mut iter: redis::AsyncIter<i32> = con.sscan("my_set").await?;
while let Some(element) = iter.next_item().await {
  println!("found: {}", element);
  assert!(element == 42 || element == 43);
}
```

The above code produces this output, where I was expecting to see the keys only once:

```shell
found: 42
found: 43
found: 42
found: 43
```

The same behaviour can be seen when using other asynchronous iterations. This example returns twice the number of keys as one would expect:

```rust
let mut iter: AsyncIter<String> = con.scan_match(format!("{}*", search_base)).await?;
let mut result = vec![];
while let Some(key) = iter.next_item().await {
  result.push(key)
}
```

This behaviour is identical for the `aio::MultiplexedConnection` and `aio::Connection`. I've tested with Redis 5.0.6 and Redis 6.0.7

mibes404:
Break the loop after processing the initial results, when cursor == None. This prevents duplicate results when processing small batches. Fixes #377 

Tested with various batch sizes: 10, 100, 1000, 10.000, 100.000

### DIFF
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -110,6 +110,8 @@ impl<'a, T: FromRedisValue + 'a> AsyncIter<'a, T> {
                 if cursor == 0 {
                     return None;
                 }
+            } else {
+                return None;
             }
 
             let rv = unwrap_or!(
--- a/tests/test_async.rs
+++ b/tests/test_async.rs
@@ -226,8 +226,7 @@ fn test_transaction_multiplexed_connection() {
     .unwrap();
 }
 
-#[test]
-fn test_async_scanning() {
+fn test_async_scanning(batch_size: usize) {
     let ctx = TestContext::new();
     block_on_all(async move {
         ctx.multiplexed_async_connection()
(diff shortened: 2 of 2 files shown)


Marwes:
Can you add a test for this?

mibes404:
I've modified the unit test `test_async_scanning` in `test_async.rs` slightly, such that it validates when too many items are returned:

```rust
assert!(unseen.remove(&x));
```

It now runs two variations of that test to confirm correct behaviour for small and larger batches:

```rust
#[test]
fn test_async_scanning_big_batch() {
    test_async_scanning(1000)
}

#[test]
fn test_async_scanning_small_batch() {
    test_async_scanning(2)
}
```

I've also ran these unit tests against the current code, and it confirms that the small batch fails the test.

PR #360, opened 2020-07-11
refactor: Simplify the runtime selection

Marwes:
The runtime selection has been very complicated and filled with duplicated code. This doesnt't do away with all of this complication but it does at least encapsulate it a bit better and opens a path for using a non-tcp/unix based connections (say for mocking)

PR #382, opened 2020-09-22
Explain TLS feature in README

gferon:


gferon:
I was thinking that in general, it might not be such a good idea to have both `async-std` and `tokio` as default features? I find it quite confusing to see both building in my tree if I'm only using one or the other.

weihanglo:
> I was thinking that in general, it might not be such a good idea to have both `async-std` and `tokio` as default features? I find it quite confusing to see both building in my tree if I'm only using one or the other.

Yea. I have the same feeling. #360 is tidies up the whole async stuff. And after that we can choose one async runtime and introduce some compat crate for compatibility. That's will reduce lots of duplication both in features and code.

weihanglo (approved):
Looks great! Could be better by bumping version to 0.17.0 😜

weihanglo:
Also, I found that some features lacks proper documentation (in lib.rs). May also add them back after the clean up.

gferon:
> Looks great! Could be better by bumping version to 0.17.0 stuck_out_tongue_winking_eye

Oops, I copied the version from the paragraph below (Cluster). I've also fixed it.

> Also, I found that some features lacks proper documentation (in lib.rs). May also add them back after the clean up.

Yeah, let's wait until the features are a bit cleaner so we don't have to expose the need for both `tokio` and `async-std`.

tibbon:
Thank you so much. I'm relatively new to Rust, so documentation things like this are super useful to me!

gferon:
> Thank you so much. I'm relatively new to Rust, so documentation things like this are super useful to me!

I'm very happy to help, and welcome to Rust! 🦀

Issue #385, opened 2020-09-28
Why is the r2d2 module not public? 


ibraheemdev:
Why is the [`rd2d` module](https://github.com/mitsuhiko/redis-rs/blob/master/src/r2d2.rs) not public? This means enabling the `rd2d` features provides no functionality, and the [`r2d2-redis`](https://github.com/sorccu/r2d2-redis) crate is essentially a copy of [`rd2d.rs`](https://github.com/mitsuhiko/redis-rs/blob/master/src/r2d2.rs). How are pooled connections supposed to be managed? I can't find any documentation regarding this...

ibraheemdev:
Oh, I got it now. I'll try to add an example and docs for r2d2 pooling

invis87:
Hi, @ibraheemdev , I am new to redis and writing some "cache-service" on top of it. I don't want to open new connection for every request, so connection pool is what I am looking for. Could you tell how to create it here? 
Also why not to use r2d2-redis?

ibraheemdev:
The functionality provided by `rd2d-redis` is built in to the `redis` crate through the `r2d2` module. Here is a simple example of a pooled connection:

```rust
use redis::Commands;

fn main() {
    // get redis client
    let client: redis::Client = redis::Client::open("redis://127.0.0.1/").unwrap();

    // create r2d2 pool
    let pool: r2d2::Pool<redis::Client> = r2d2::Pool::builder().build(client).unwrap();

    // get a connection from the pool
    let mut conn: r2d2::PooledConnection<redis::Client> = pool.get().unwrap();

    // use the connection as usual
    let _: () = conn.set("KEY", "VALUE").unwrap();
    let val: String = conn.get("KEY").unwrap();
    println!("{}", val)
}
```

invis87:
Thanks much!

Looks like README should be fixed, cause now there is:
`To open a connection you need to create a client and then to fetch a connection from it. In the future there will be a connection pool for those, currently each connection is separate and not pooled.`

Issue #353, opened 2020-07-02
Example/feature request: async transactions


davechallis:
I've been struggling to get transactions working with async, so an example or some docs on that would be super helpful.

E.g. an async version of: 

```
use redis::Commands;
let key = "the_key";
let (new_val,) : (isize,) = redis::transaction(&mut con, &[key], |con, pipe| {
    let old_val : isize = con.get(key)?;
    pipe
        .set(key, old_val + 1).ignore()
        .get(key).query(con)
})?;
println!("The incremented number is: {}", new_val);
```

Though I don't know if that's currently possible, or whether it needs a new `transaction_async` or similar function adding.

davechallis:
In case it helps anyone, I did this using a macro in the end, any improvements or alternative suggestions welcome though.

It requires the caller to create their own `pipe().atomic()`, but is otherwise similar to the current `transaction` function (it takes a block to execute as the 3rd argument instead of a function though).

```rust
macro_rules! async_transaction {
    ($conn:expr, $keys:expr, $body:expr) => {
        loop {
            cmd("WATCH").arg($keys).query_async($conn).await?;

            if let Some(response) = $body {
                cmd("UNWATCH").query_async($conn).await?;
                break response;
            }
        }
    };
}
```

Full example of usage:

```rust
use redis::{pipe, cmd, AsyncCommands, RedisResult};

macro_rules! async_transaction {
    ($conn:expr, $keys:expr, $body:expr) => {
        loop {
            cmd("WATCH").arg($keys).query_async($conn).await?;

            if let Some(response) = $body {
                cmd("UNWATCH").query_async($conn).await?;
                break response;
            }
        }
    };
}

#[tokio::main]
async fn main() -> RedisResult<()> {
    let client = redis::Client::open("redis://127.0.0.1/").unwrap();
    let mut async_conn = client.get_async_connection().await?;

    let key = "the_key";

    let (new_val,): (isize,) = async_transaction!(&mut async_conn, &[key], {
        let old_val: isize = async_conn.get(key).await?;
        pipe().atomic()
            .set(key, old_val + 1).ignore()
            .get(key)
            .query_async(&mut async_conn)
            .await?
    });
    println!("The incremented number is: {}", new_val);

    Ok(())
}
```

lroux-at-jellysmack:
Hello @mitsuhiko, are you open for a PR to implement a `redis::aio::transaction` async function with the same prototype as `redis::transaction` ?
[...]

Issue #254, opened 2019-11-05
Transaction helper does not work when used with `redis::Client`


nate-onesignal:
Redis version: redis:4.0.1 docker image
Redis-rs version: 0.13.0
Rust version: 1.38
OS: Ubuntu 16.04 LTS docker image

I'm having an issue with `redis::transaction`. According to the docs, it
should retry the closure passed to it as many times as required until it's
able to run without interrupt. I have constructed an example that seems to
suggest that it doesn't do this.

```rust
use redis::{Commands, PipelineCommands};

fn main() -> Result<(), failure::Error> {
    let mut conn = redis::Client::open("redis://redis")?;
    // let mut conn = conn.get_connection()?; // If this line is uncommented, it works as expected.
    conn.set("hello", 2)?;

    let mut first_time = true;

    redis::transaction(&mut conn, &["hello"], |conn, pipe| {
        println!("TRANSACTION CLOSURE");
        let value: i32 = conn.get("hello")?;
        println!("GET `hello`: {}", value);

        if first_time {
            let client = redis::Client::open("redis://redis")?;
            let mut conn = client.get_connection()?;
            conn.set("hello", 5)?;
            println!("overwrite `hello`: {}", conn.get::<_, i32>("hello")?);
            first_time = false;
        }

        let value: redis::Value = pipe.set("hello", value + 1).query(conn)?;
        if value == redis::Value::Nil {
            return Ok(None);
        }

        Ok(Some(()))
    })?;

    let value: i32 = conn.get("hello")?;
    println!("value in redis: {}", value);

    Ok(())
}
```

Roughly the steps are as follows:

1. Set `hello` to 2
2. start a transaction on `hello`
3. Get the value of `hello`
4. If this is the first time the transaction is run, create a new client and
   overwrite the value of `hello` to 5. Print out the stored value of `hello` to
   confirm the overwrite worked.
5. Increment the value of `hello` to one greater than the value we fetched in
   step 3
6. After the transaction has finished, print out the value of `hello` stored
   in redis, and the number of times the transaction closure ran.

Expected output:

```
TRANSACTION CLOSURE
GET `hello`: 2
overwrite `hello`: 5
TRANSACTION CLOSURE
GET `hello`: 5
value in redis: 6
```

Actual output:

```
TRANSACTION CLOSURE
GET `hello`: 2
overwrite `hello`: 5
value in redis: 3
```

This indicates that the transaction block is not being run multiple times as
it should be, but only once.

[...]

PR #388, opened 2020-10-01
Create an `r2d2` example.

ibraheemdev:
The `r2d2-redis` crate is pretty popular, and the functionality it provides is already included in this crate with the `r2d2` module. 

The docs are pretty fuzzy. There is currently no documentation regarding how to use r2d2 with this crate other than:

> To open a connection you need to create a client and then to fetch a connection from it. In the future there will be a connection pool for those, currently each connection is separate and not pooled.

See #385 as well.


invis87:
Is there are a way to use `use redis::AsyncCommands;` with r2d2?

ibraheemdev:
@invis87 No. r2d2 is a synchronous connection pool crate. To use `redis::AsyncCommands`, look into [`deadpool-redis`](https://crates.io/crates/deadpool-redis) or [`bb8-redis`](https://crates.io/crates/bb8-redis/0.3.1)

Marwes:
I'd recommend just using the `MultiplexedConnection` for async as it can already be cloned. Connection pooling is only necessary in some really specifica schenarios.

ibraheemdev:
@Marwes Well, considering `r2d2-redis` has over 90k downloads, and this functionality already exists within this crate, I think the added documentation and examples wouldn't hurt.

Marwes:
The PR is good, I was just commenting on using pooling for async connections, in which case it is unnecessary in most cases.

weihanglo:
Add an `[[example]]` item into Cargo.toml would be great!

```toml
[[example]]
name = "r2d2"
required-features = ["r2d2"]
```

jasongoodwin:
It's really hard to find this info, the readme specifically says that the pooling feature doesn't exist. This needs to get merges, maybe reference using pipe with deref too eg 

let res: redis::RedisResult<Vec<i32>> = pipe.query(conn.deref_mut());

ibraheemdev:
Btw `pipe.query(conn.deref_mut())` is just `pipe.query(&mut *conn)`.

ibraheemdev:
@Marwes can this be merged if you have write access?

ibraheemdev:
Also regarding the async point, I think pooling is important for async web servers, which is a common use case.

Marwes:
> Also regarding the async point, I think pooling is important for async web servers, which is a common use case.

It usually isn't necessary if you use the multiplexed connection as that will let you send multiple concurrent requests to redis anyway. (Since redis is single threaded, or at least mostly so, having all requests on the same stream makes little difference as they become serialized in the redis server anyway).

Parth:
@Marwes regarding the suitability for `MultiplexedConnection` in the context of an "async web server":

`MultiplexedConnection` doesn't support transactions, right? [StackExchange.redis](https://stackexchange.github.io/StackExchange.Redis/Transactions.html#and-in-stackexchangeredis) gives multiplexed transactions some special treatment that I'm not sure this crate does. I was going to whip up an example similar to #254, but my basic attempt didn't really get off the ground:

```rust
    let mut conn = redis::Client::open("redis://127.0.0.1:6379")?
        .get_multiplexed_tokio_connection()
        .await?;

    redis::transaction(&mut conn, &["hello"], |conn, pipe| Ok(Some(())))?;
```

returns: 

```
error[E0277]: the trait bound `MultiplexedConnection: redis::ConnectionLike` is not satisfied
    --> src/main.rs:20:24
     |
20   |     redis::transaction(&mut conn, &["hello"], |conn, pipe| Ok(Some(())))?;
     |                        ^^^^^^^^^ the trait `redis::ConnectionLike` is not implemented for `MultiplexedConnection`
     | 
    ::: /Users/parthmehrotra/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.21.4/src/connection.rs:1080:8
     |
1080 |     C: ConnectionLike,
     |        -------------- required by this bound in `transaction`

For more information about this error, try `rustc --explain E0277`.
warning: `redis-test` (bin "redis-test") generated 1 warning
error: could not compile `redis-test` due to previous error; 1 warning emitted
```

Error message a bit confusing, but after some digging [`ConnectionLike`](https://docs.rs/redis/latest/redis/aio/struct.MultiplexedConnection.html#impl-ConnectionLike) != [`ConnectionLike`](https://docs.rs/redis/latest/redis/aio/trait.ConnectionLike.html).

And so I sought out `async` ways to do transactions, and stumbled upon: #353. Dropping in that macro and watching a key in a `cloned`, and sure enough this lets me create transactions with a cloned `MultiplexedConnection`, but, naturally, they don't behave the way you'd expect them too, transactions across cloned `MultiplexedConnection`s step on each other, unwatching each-other's keys.

So ultimately I'm trying to figure out:

> Connection pooling is only necessary in some really specifica schenarios.

Is the whole family of doing `async` transactions one of these specific scenarios, or am I just doing something wrong?

Marwes:
> Is the whole family of doing async transactions one of these specific scenarios, or am I just doing something wrong?

Yes, `MultiplexedConnection` assumes that each command run on it is independentent so you do not manually want to start a transaction (but using https://docs.rs/redis/0.21.4/redis/struct.Pipeline.html works)

Parth:
I agree `Pipeline` will work as expected, but the `WATCH`s associated with competing pipelines will not work as expected. They're going to `UNWATCH` each other's keys. Which, imo, makes `pipe().atomic()` less useful for most people.

Other crates, for example [StackExchange.Redis](https://stackexchange.github.io/StackExchange.Redis/Transactions.html#and-in-stackexchangeredis) give this stuff special treatment.

I think for people trying to do the `async web-server redis` + `transactions` they need to seek out `bb8` or `deadpool` (which is fine, because both those projects have pretty well documented examples).

But I think what we should add to this crate, is an `async` variant of `redis::transaction`. I'm struggling a bit with the lifetimes associated with that stuff, but if I'm able to sort it out, would that sort of contribution be welcome in this crate?

rudolfolah:
This PR is great, I was able to get the example working with `actix-web`: https://github.com/actix/examples/issues/569

Thanks!

jaymell:
Hmm... I'm looking back on this and am unclear why this wasn't merged years ago. If I'm understanding correctly, I think the original concern was around mentioning other libraries as offering async-compatible connection pooling in the README? I personally am fine with that comment.

@ibraheemdev if you're still out there :laughing: are you able to resolve conflicts and otherwise bring this up-to-date? Would love to get it merged. :smiley:

djc:
@jaymell I forget where but I think I pushed back on adding more examples in the past. I don't really want to be on the hook for maintaining a bunch of examples over time. Maybe the right solution is to move the r2d2 support to a separate crate? We don't maintain integrations for deadpool or bb8, either.

jaymell:
@djc I think that's fine, but in that case we should probably close this and other relevant PRs if alright with you.

djc:
Right, closing this for now.

Issue #387, opened 2020-09-29
TLS feature broken


Samuel-B-D:
None of the dependancies lines suggested in the readme compile:
```
redis = { version = "0.17.0", features = ["tls"] }

# if you use tokio
redis = { version = "0.17.0", features = ["tls", "tokio-tls-comp"] }

# if you use async-std
redis = { version = "0.17.0", features = ["tls", "async-std-tls-comp"] }
```

aio_async_std.rs includes references to tokio / async_std which are not gated by features.

Edit: minimal set of features to get TLS working with tokio currently is this : 
`redis = { version = "0.17.0", features = ["tokio-comp", "tls", "tokio-tls-comp", "async-std-tls-comp"] }`

kbacha:
I've also hit errors with TLS and the dropping of the native tls error makes debugging very difficult:
https://github.com/mitsuhiko/redis-rs/blob/f60933a467013de40bdf4a40288b551e09857160/src/types.rs#L247-L254

PR #399, opened 2020-10-21
Converts native_tls::Error to WithDescriptionAndDetail

kbacha:
Previously, the tls error was consumed and dropped by the From trait
implementation. This makes debugging difficult when there is a TLS error
as the only output is "TLS error". This change will serialize the error
that's passed in as a debug detail.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -246,9 +246,13 @@ impl From<Utf8Error> for RedisError {
 
 #[cfg(feature = "tls")]
 impl From<native_tls::Error> for RedisError {
-    fn from(_: native_tls::Error) -> RedisError {
+    fn from(err: native_tls::Error) -> RedisError {
         RedisError {
-            repr: ErrorRepr::WithDescription(ErrorKind::IoError, "TLS error"),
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -246,9 +246,13 @@ impl From<Utf8Error> for RedisError {
...
 #[cfg(feature = "tls")]
 impl From<native_tls::Error> for RedisError {
-    fn from(_: native_tls::Error) -> RedisError {
+    fn from(err: native_tls::Error) -> RedisError {
         RedisError {
-            repr: ErrorRepr::WithDescription(ErrorKind::IoError, "TLS error"),
+            repr: ErrorRepr::WithDescriptionAndDetail(
+                ErrorKind::IoError,
+                "TLS error",
+                format!("{:?}", err),
```
Marwes, line 254:
```suggestion
                format!("{}", err),
```

Marwes (changes requested):

kbacha:
This partially addresses #387

PR #400, opened 2020-10-29
Add scan_match_count and scan example

cptrodgers:
This function adds the "COUNT" argument to support that Redis has too many keys.


weihanglo:
Thanks for your contribution. It's reasonable to add a `scan_match_count` command for convenience. 👍🏽

And for the scan example, there are some ways to improve it. Here is my personal thought.

First, the example generates lots of key and that took my macbook pro 5m31s to finish it. We should should give users a quick intro instead. Maybe reducing the pipeline size would be better. 

Second, the Redis documentation says `SCAN` command is [just a hint](http://redis.io/commands/scan#the-count-option) to Redis server, so the assertion may fail randomly. 

And the last is just my personal preference. Don't take it too hard. The existing examples are categorized in a more higher-level view. A `scan` example seems too specific for a single command, and does not bring any new idea for who want to know more about redis-rs. I have an idea. We can move the scan example to documentation of `fn scan_match_count` itself. That would be more preferable IMO. 😄

cptrodgers:
Hi weihanglo,
Thanks for the quickly response.
I understood your feedback (advice). I will update it soon.

nihohit:
added in https://github.com/redis-rs/redis-rs/pull/1231

PR #405, opened 2020-11-18
feat: Update tokio dependency to 0.3

### LINKED ISSUE #396: Upgrade to tokio 0.3
Would be nice to have.

Marwes:
Closes #396

BREAKING CHANGE

Tokio 0.2 support is dropped to support tokio 0.3

### DIFF
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -98,9 +98,7 @@ impl Runtime {
 
         #[cfg(all(feature = "tokio-comp", feature = "async-std-comp"))]
         {
-            // TODO How to detect tokio? Or remove detection?
-            // if ::tokio::runtime::Handle::try_current().is_ok() {
-            if true {
+            if ::tokio::runtime::Handle::try_current().is_ok() {
                 Runtime::Tokio
             } else {
                 Runtime::AsyncStd
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -98,7 +98,9 @@ impl Runtime {
 
         #[cfg(all(feature = "tokio-comp", feature = "async-std-comp"))]
         {
-            if ::tokio::runtime::Handle::try_current().is_ok() {
+            // TODO How to detect tokio? Or remove detection?
+            // if ::tokio::runtime::Handle::try_current().is_ok() {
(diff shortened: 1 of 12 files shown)


src/aio.rs
```diff
@@ -98,7 +98,9 @@ impl Runtime {
 
         #[cfg(all(feature = "tokio-comp", feature = "async-std-comp"))]
         {
-            if ::tokio::runtime::Handle::try_current().is_ok() {
+            // TODO How to detect tokio? Or remove detection?
+            // if ::tokio::runtime::Handle::try_current().is_ok() {
```
MikailBag, line 102: It seems that this line should work.

https://docs.rs/tokio/0.3.4/tokio/runtime/struct.Handle.html#method.try_current

Marwes: Looks like it was just re-added https://github.com/tokio-rs/tokio/blob/479c545c20b2cb44a8f09600733adc8c8dcb5aa0/tokio/CHANGELOG.md#added !

weihanglo (commented):
Great to see tokio 0.3 upgrade!

Jasperav:
Any idea when this PR gets merged? What is the current status?

tests/parser.rs
```diff
@@ -96,12 +99,17 @@ where
...
         match self.ops.next() {
             Some(PartialOp::Limited(n)) => {
-                let len = std::cmp::min(n, buf.len());
-                Pin::new(&mut self.inner).poll_read(cx, &mut buf[..len])
+                let len = std::cmp::min(n, buf.remaining());
+                buf.initialize_unfilled();
+                let mut sub_buf = buf.take(len);
+                ready!(Pin::new(&mut self.inner).poll_read(cx, &mut sub_buf))?;
+                let initialized = sub_buf.initialized().len();
+                buf.advance(initialized);
```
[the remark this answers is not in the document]
Marwes: Nice catch! I think it never shows up in tests because, as there isn't something else reading into the `ReadBuf` before this call it is always empty at this point so the `initialized()` return value ends up being correct.

PR #410, opened 2020-11-30
Fix bool parse errors

### LINKED ISSUE #324: bool mismatched in ToRedisArgs and FromRedisValue

Marwes:
Rebase of #372

Should resolve #324. I changed the ToRedisArgs implementation of booleans to encode them as 1 or 0 as redis does, and their FromRedisValue implementation to also parse from Value::Data. The latter change means that calling GET on boolean values should stop dying now.

Closes #372

BREAKING CHANGE

`bool` are now written as `0` and `1` instead of `true` and `false`. Parsing a `bool` still accept `true` and `false` so this should not break anything for must users however if you are reading something out that was stored as a `bool` you may see different results.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1058,10 +1058,9 @@ impl FromRedisValue for bool {
                 }
             }
             Value::Data(ref bytes) => {
-                let res = from_utf8(bytes)?.to_string();
-                if res == "1" {
+                if bytes == b"1" {
                     Ok(true)
-                } else if res == "0" {
+                } else if bytes == b"0" {
                     Ok(false)
                 } else {
                      invalid_type_error!(v, "Response type not bool compatible.");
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Nice!

PR #413, opened 2020-12-04
Follow MOVED (again)

hiratara:
The current implementation does not handle the `"MOVED"` error properly. Therefore, throwing an update query to a `readonly` connection will result in an infinite loop. A `readonly` connection throws the query to a replica, but a replica can not handle the update query and returns `"MOVED"`. Calling `refresh_slots` does not complete the process, because it doesn't assign the query to the master node.

I don't know if it's correct to perform the update query on a `readonly` connection, but it is possible to complete the query on the node referred to by `MOVED`.

#285 tried to solve the same problem but did not merge. In this PR, we call the `refresh_slots` method even if the query is performed on a `MOVED` destination. Also, I add a test to reproduce the infinite loop(Actually, it retries the `CLUSTER SLOTS` command [16 times](https://github.com/mitsuhiko/redis-rs/blob/8a53e2699d7d7bd63f222de778ed6820b0a65665/src/cluster.rs#L458), then it returns the error response to the client).

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -457,17 +457,21 @@ impl ClusterConnection {
 
         let mut retries = 16;
         let mut excludes = HashSet::new();
-        let mut asking = None::<String>;
+        let mut redirected = None::<String>;
+        let mut is_asking = false;
         loop {
             // Get target address and response.
             let (addr, rv) = {
                 let mut connections = self.connections.borrow_mut();
-                let (addr, conn) = if let Some(addr) = asking.take() {
+                let (addr, conn) = if let Some(addr) = redirected.take() {
                     let conn = self.get_connection_by_addr(&mut *connections, &addr)?;
-                    // if we are in asking mode we want to feed a single
-                    // ASKING command into the connection before what we
-                    // actually want to execute.
-                    conn.req_packed_command(&b"*1\r\n$6\r\nASKING\r\n"[..])?;
+                    if is_asking {
+                        // if we are in asking mode we want to feed a single
+                        // ASKING command into the connection before what we
+                        // actually want to execute.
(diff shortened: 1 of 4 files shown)


src/cluster.rs
```diff
@@ -489,11 +493,16 @@ impl ClusterConnection {
                         let kind = err.kind();
 
                         if kind == ErrorKind::Ask {
-                            asking = err.redirect_node().map(|x| x.0.to_string());
+                            redirected = err.redirect_node().map(|x| format!("redis://{}", x.0));
```
hiratara, line 496: It's the boy scout rule. I also fix the `ASK` redirect. Redis addresses should have the `"redis://"` prefix.

tests/support/cluster.rs
```diff
@@ -61,7 +62,36 @@ impl RedisCluster {
         let status = dbg!(cmd).status().unwrap();
         assert!(status.success());
 
-        RedisCluster { servers, folders }
+        let cluster = RedisCluster { servers, folders };
+        if replicas > 0 {
+            cluster.wait_for_replicas(replicas);
```
hiratara, line 67: If we use replicas, we need to wait for the replica information to be passed on to all nodes.

tests/support/cluster.rs
```diff
@@ -91,14 +121,26 @@ pub struct TestClusterContext {
 
 impl TestClusterContext {
     pub fn new(nodes: u16, replicas: u16) -> TestClusterContext {
+        Self::new_with_cluster_client_builder(nodes, replicas, identity)
+    }
+
+    pub fn new_with_cluster_client_builder<F>(
```
hiratara, line 127: This method allows us to have a `readonly` client.

Marwes (approved):
LGTM

Issue #406, opened 2020-11-21
AUTH command fails when username is specified in the connection URL for Redis v5 (and possibly lower)


gyfis:
Hi,

I encountered an error where I have a redis connection URL in this format: `redis://user:password@127.0.0.1:6379` and the AUTH command fails.

My redis is version 5.0.7, and this connection string doesn't go through - fails on AUTH. Since there is no notion of user in redis v5 or lower, the username should be ignored; however, the code currently always sends the username to redis.

It'd be great if this lib could recognize which redis is currently being run, and if v5 or lower, send proper AUTH command with no username (I didn't find a reference to the redis version currently, though).

This is the code snippet that currently always sends the username:
```
    if let Some(passwd) = &connection_info.passwd {
        let mut command = cmd("AUTH");
        if let Some(username) = &connection_info.username {
            command.arg(username);
        }
        match command.arg(passwd).query::<Value>(&mut rv) {
            Ok(Value::Okay) => {}
            _ => {
                fail!((
                    ErrorKind::AuthenticationFailed,
                    "Password authentication failed"
                ));
            }
        }
    }
```

Let me know if I can give more debug info; thanks!

weihanglo:
Hmm.... 🤔 I think it is fair to reject unnecessary username input to align to behavior of old version of Redis. Below is the behavior of Redis 4.0 when passing two arguments to the `AUTH` command.  

```console
127.0.0.1:6379> AUTH username password
(error) ERR wrong number of arguments for 'auth' command
```

Just curious, what's your scenario of passing username to Redis lower than v5?

gyfis:
@weihanglo Thanks for the reply! I think it'd also makes sense because `redis-cli` for Redis below 6.0 ignores the username in URL as well when passed as an argument `-u`.

As for the scenario - I'm using [dokku/dokku](https://github.com/dokku/dokku) with the [dokku/dokku-redis](https://github.com/dokku/dokku) plugin, and they fill a username incorrectly to the URL. I added a [pull request](https://github.com/dokku/dokku-redis/pull/157) with a fix there, though I'm not sure when it will be merged. I'm able to overwrite the env variable they are setting, but I wanted this library to handle it too, when it encounters this in the future.

npajkovsky:
Well, that's easy to solve. And you were almost there

```
       if let Some(username) = &connection_info.username {
            command.arg(username);
        }
```
`connection_info` defines `username` as an `Option<String>` and that means, you can omit the value.
[...]

PR #423, opened 2020-12-22
AUTH cmd fallback

npajkovsky:
When connecting to redis v5 server, with uri containing username, connection
fails with an error:

  wrong number of arguments for 'auth' command

library can do better and try AUTH again without username.

Signed-off-by: Nikola Pajkovsky <nikola@pajkovsky.cz>

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -520,6 +520,44 @@ impl ActualConnection {
     }
 }
 
+fn connect_auth(con: &mut Connection, connection_info: &ConnectionInfo) -> RedisResult<()> {
+    let mut command = cmd("AUTH");
+    if let Some(username) = &connection_info.username {
+        command.arg(username);
+    }
+    let passwd = connection_info.passwd.as_ref().unwrap();
+    let err = match command.arg(passwd).query::<Value>(con) {
+        Ok(Value::Okay) => return Ok(()),
+        Ok(_) => {
+            fail!((
+                ErrorKind::ResponseError,
+                "Redis server refused to authenticate, returns Ok() != Value::Okay"
+            ));
+        }
+        Err(e) => e,
+    };
+    let err_msg = err.detail().ok_or((
+        ErrorKind::AuthenticationFailed,
+        "Password authentication failed",
+    ))?;
+    if !err_msg.contains("wrong number of arguments for 'auth' command") {
+        fail!((
+            ErrorKind::AuthenticationFailed,
+            "Password authentication failed",
+        ));
+    }
+
+    // fallback to AUTH version <= 5
+    let mut command = cmd("AUTH");
+    match command.arg(passwd).query::<Value>(con) {
+        Ok(Value::Okay) => Ok(()),
+        _ => fail!((
+            ErrorKind::AuthenticationFailed,
+            "Password authentication failed",
+        )),
+    }
+}
+
 pub fn connect(
     connection_info: &ConnectionInfo,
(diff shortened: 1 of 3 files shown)


src/connection.rs
```diff
@@ -520,32 +520,64 @@ impl ActualConnection {
     }
 }
 
-pub fn connect(
+fn build_connection(
     connection_info: &ConnectionInfo,
     timeout: Option<Duration>,
 ) -> RedisResult<Connection> {
     let con = ActualConnection::new(&connection_info.addr, timeout)?;
```
Marwes, line 527: `ActualConnection::new` will create a new tcp/unix stream on every call so for authenticated connections the new code will open 2 or 3 connections (if it fails the first authentication attempt). Failing to authenticate shouldn't put the connection in a bad state so you should be able to use a single connection throughout.

Marwes (changes requested):
It would be great if this could also be applied to the async variant!

Marwes:
cc #406

src/connection.rs
npajkovsky: I tried using a single connection for 2nd auth, but it did not work. Are there any knobs to debug connection state?

npajkovsky:
> It would be great if this could also be applied to the async variant!

Ok.

src/connection.rs
Marwes: Seems like redis actually just shuts down the connection if a command isn't formatted correctly. I guess that makes sense in most cases since there is no way in general of knowing how to uphold the protocol...

Creating a new connection after authentication fails seems fine then, however you could pass the `Connection` as a paramter to `connect_auth` to ensure we only create a second connection if authentication fails.

src/connection.rs
```diff
@@ -520,32 +520,64 @@ impl ActualConnection {
...
+    timeout: Option<Duration>,
+) -> RedisResult<Connection> {
+    let mut rv = build_connection(connection_info, timeout)?;
+    let mut command = cmd("AUTH");
+    if let Some(username) = &connection_info.username {
+        command.arg(username);
+    }
+    let passwd = connection_info.passwd.as_ref().unwrap();
+    let err = match command.arg(passwd).query::<Value>(&mut rv) {
+        Ok(_) => return Ok(rv),
```
Marwes, line 547: Should this be `Ok(Value::Okay) => return Ok(rv)`? Same as the second AUTH below

npajkovsky: Well, I have that too in my non-public iteration of patch. However, if I use `Ok(Value::Okay) => return Ok(rv)`
non-exhaustive error shows up.

```
error[E0004]: non-exhaustive patterns: `Ok(Nil)`, `Ok(Int(_))`, `Ok(Data(_))` and 2 more not covered
```

The question is, what to do with other `Ok(...)` patterns? I have decided, that all `Ok ` values are just fine or we can do better, that `Ok(Value::Okay) => return Ok(rv)` and other Ok(...) patters will fail with message

```
fail!((
   ErrorKind::ResponseError,
   "Invalid data returned during AUTH"
)
```

src/connection.rs
npajkovsky: > Seems like redis actually just shuts down the connection if a command isn't formatted correctly. I guess that makes sense in most cases since there is no way in general of knowing how to uphold the protocol...

I don't have yet enough information about connection life cycle, but I think, that an invalid command should not drop connection.

Marwes (approved):

PR #424, opened 2020-12-22
Add support for ACL SETUSER RESET command

bblack-net:
Add support for ACL SETUSER RESET command


Marwes:
You can already do this with `acl_setuser_rules`. Since the `ACL` api is so broad I am not sure it warrants having specific methods like this https://redis.io/commands/acl-setuser 🤔

PR #425, opened 2020-12-23
Add support for UNLINK command

bblack-net:
Greetings,

This is a small PR for supporting non-blocking delete via the UNLINK command.

https://redis.io/commands/unlink


Marwes (approved):
Seems to fail on appveyor/windows. Probably to old a version of redis on there. Can you disable the test on windows (with a comment explaining).

bblack-net:
Thanks @Marwes . I just pushed an updated test.

PR #437, opened 2021-01-13
Fix #436: Read remaining responses in case of an error

### LINKED ISSUE #436: Errors in an atomic Pipeline leave unread data that leaks to the next command
When running commands in a `redis::pipe()::atomic()` transaction, if Redis returns an error for one of the commands (e.g. `READONLY` instead of `QUEUED`), the results for the remaining commands are not read and therefore erroneously remain there in the socket buffer. They will be returned by any subsequent command.

The following program illustrates the problem:

```
use redis::{Commands, RedisResult};

fn test() -> RedisResult<()> {
    let client = redis::Client::open("redis://localhost")?;
    let mut con = client.get_connection()?;

    // Ensure Redis is not currently a replicaof
    redis::cmd("replicaof").arg("no").arg("one").query(&mut con)?;

    con.set("x", "42")?;

    // Make Redis a replica of a nonexistent master, thereby making it read-only.
    redis::cmd("replicaof").arg("1.1.1.1").arg("1").query(&mut con)?;

    // Ensure that a write command fails with a READONLY error
    let result: RedisResult<Option<(String,)>> = redis::pipe().atomic()
        .set("x", "abc").ignore()
        .get("x")
        .query(&mut con);

    let err = result.unwrap_err().to_string();
    assert!(err.starts_with("READONLY"));

    // Here we get erroneous leftover results from the previous pipeline!
    for _ in 0..3 {
        let x: RedisResult<String> = con.get("x");
        println!("{:?}", x);
    }

    Ok(())
}

fn main() {
    test().unwrap();
}
```

The expected output:
```
Ok("42")
Ok("42")
Ok("42")
```

The actual output:
```
Ok("QUEUED")
Err(An error was signalled by the server: Transaction discarded because of previous errors.)
Ok("42")
```

When using a standard `Pipeline` without `atomic`, this works as intended.

The problem seems to be that the code in `req_packed_commands` returns with the `?` operator on the first error encountered, instead of first reading the rest of the results.

gavrie:
When a `MULTI/EXEC` transaction fails due to errors, the remaining responses should be read instead of returning early. This fixes #436.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -817,13 +817,32 @@ impl ConnectionLike for Connection {
         }
         self.con.send_bytes(cmd)?;
         let mut rv = vec![];
+        let mut first_err = None;
         for idx in 0..(offset + count) {
-            let item = self.read_response()?;
-            if idx >= offset {
-                rv.push(item);
+            // When processing a transaction, some responses may be errors.
+            // We need to keep processing the rest of the responses in that case,
+            // so bailing early with `?` would not be correct.
+            // See: https://github.com/mitsuhiko/redis-rs/issues/436
+            let response = self.read_response();
+            match response {
+                Ok(item) => {
+                    if idx >= offset {
+                        rv.push(item);
+                    }
+                }
+                Err(err) => {
+                    if first_err.is_none() {
+                        first_err = Some(err);
+                    }
+                }
             }
         }
-        Ok(rv)
+
+        if let Some(err) = first_err {
+            Err(err)
+        } else {
+            Ok(rv)
+        }
     }
 
     fn get_db(&self) -> i64 {
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -505,16 +505,37 @@ where
             cmd.write_packed_pipeline(&mut self.buf);
             self.con.write_all(&self.buf).await?;
 
+            let mut first_err = None;
+
             for _ in 0..offset {
-                self.read_response().await?;
+                let response = self.read_response().await;
+                if let Err(err) = response {
+                    if first_err.is_none() {
+                        first_err = Some(err);
+                    }
+                }
             }
 
             let mut rv = Vec::with_capacity(count);
             for _ in 0..count {
-                rv.push(self.read_response().await?);
+                let response = self.read_response().await;
+                match response {
+                    Ok(item) => {
+                        rv.push(item);
+                    }
+                    Err(err) => {
+                        if first_err.is_none() {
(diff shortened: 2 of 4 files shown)


src/connection.rs
```diff
@@ -818,11 +818,18 @@ impl ConnectionLike for Connection {
...
         let mut rv = vec![];
         for idx in 0..(offset + count) {
-            let item = self.read_response()?;
-            if idx >= offset {
-                rv.push(item);
+            // When processing a transaction, some responses may be errors.
+            // We need to keep processing the rest of the responses in that case,
+            // so bailing early with `?` would not be correct.
+            // See: https://github.com/mitsuhiko/redis-rs/issues/436
+            let response = self.read_response();
```
Marwes, line 825: Any errors occuring will just be lost here, I'd expect an error (probably the first?) to still be returned after consuming the remaining commands.

Marwes (changes requested):
A test would be nice to have if you can. The (basic) async connection could also use the same fix.

gavrie:
> A test would be nice to have if you can. The (basic) async connection could also use the same fix.

@Marwes the ticket (#436) includes my test code that triggers this. Would you like me to convert it to an actual test?
I'll look at the async version as well.

src/connection.rs
gavrie: @Marwes the error from the failed transaction will be returned (`Transaction discarded because of previous errors.`). I thought of returning a vector of `Results`, but that would change the API quite a bit.

Marwes: `Transaction discarded because of previous errors.` is a pretty bad error though. Chances are the first error that occured is the one with the most information returning that seems to make the most sense to me.

gavrie: Very well, I'll proceed in that direction.

gavrie: I changed the code as suggested -- the first error is now returned.

gavrie:
I've added a test that shows the bug and verifies the fix.

gavrie:
...and I've added the async test + fix as well.

gavrie:
I see that the Windows CI is broken because of the `replicaof` command; it probably uses an old Redis version. Will try `slaveof` instead.

gavrie:
Ok, that fixe it.

Marwes (approved):

Marwes:
Thanks!

gavrie:
> Thanks!

Thank you @Marwes for the quick responses! Now we won't need to use a forked version...

PR #440, opened 2021-01-16
Fix incorrect routing for XREAD/XREADGROUP command

doyshinda:
The XREAD/XREADGROUP commands can have a variable number of args,
and proper routing of the command relies on finding the key after
the "STREAMS" keyword. This was failing to properly route the if
the keyword was not uppercase. Fixes the matching logic make each
arg uppercase before comparing to the "STREAMS" keyword.

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -565,7 +565,7 @@ fn get_hashtag(key: &[u8]) -> Option<&[u8]> {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 enum RoutingInfo {
     AllNodes,
     AllMasters,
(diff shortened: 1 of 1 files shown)


src/cluster.rs
```diff
@@ -619,7 +619,7 @@ impl RoutingInfo {
             b"XGROUP" | b"XINFO" => get_arg(&args, 2).and_then(RoutingInfo::for_key),
             b"XREAD" | b"XREADGROUP" => {
                 let streams_position = args.iter().position(|a| match a {
-                    Value::Data(a) => a == b"STREAMS",
+                    Value::Data(a) => &a.to_ascii_uppercase()[..] == b"STREAMS",
```
Marwes, line 622:
```suggestion
                    Value::Data(a) => a.eq_ignore_ascii_case(b"STREAMS"),
```
https://doc.rust-lang.org/std/primitive.slice.html#method.eq_ignore_ascii_case

Marwes (changes requested):

Marwes (approved):

PR #441, opened 2021-01-19
Test fix #436 with (non-atomic) simple pipeline.

### LINKED ISSUE #436: Errors in an atomic Pipeline leave unread data that leaks to the next command

yoav-steinberg:


yoav-steinberg:
I verified this test fails (we get `"y-value"` instead of `"x-value"`) before fix #436. This means the original bug was for all pipelines not only atomic ones.

src/types.rs
```diff
@@ -385,6 +388,7 @@ impl RedisError {
             ErrorKind::IoError => "I/O error",
             ErrorKind::ExtensionError => "extension error",
             ErrorKind::ClientError => "client error",
+            ErrorKind::ReadOnly => "read-only",
```
gavrie, line 391:
```suggestion
            ErrorKind::ReadOnly => "read-only error",
```

src/parser.rs
```diff
@@ -129,6 +129,7 @@ where
                         "CLUSTERDOWN" => ErrorKind::ClusterDown,
                         "CROSSSLOT" => ErrorKind::CrossSlot,
                         "MASTERDOWN" => ErrorKind::MasterDown,
+                        "READONLY" => ErrorKind::ReadOnly,
```
gavrie, line 132: Yup, this definitely makes things cleaner.

gavrie (approved):

Marwes (approved):

PR #307, opened 2020-04-05
Efficient ClusterPipeline routing

doyshinda:
Implements efficient routing for pipeline commands with Redis Cluster. For every command in the pipeline, determines which node is responsible for it and builds a pipeline per node. Each pipeline is first sent to its respective node, then all responses are read back. Any commands that can be retried are done individually.

This leads to a huge performance gains with a pipeline of 100 commands.
Before:
```
time:   [8.3241 ms 8.3772 ms 8.4356 ms]
thrpt:  [11.854 Kelem/s 11.937 Kelem/s 12.013 Kelem/s]
```

Now:
```
time:   [383.37 us 389.66 us 396.19 us]
thrpt:  [252.41 Kelem/s 256.63 Kelem/s 260.84 Kelem/s]
```

PR #446, opened 2021-01-27
Support for requesting `Cmds` directly from a `Connection`

doyshinda:
Adds a new function to the `ConnectionLike` trait that accepts a reference to a `Cmd`. This allows different connections to handle the `Cmd` as necessary. For a single Redis instance, there is no difference, but for Redis Cluster it allows the `Cmd` to be routed without needing to first parse it back into a `Value`.

This results in a *slight* performance increase when issuing commands with a `ClusterConnection`, with the performance benefits becoming more prominent with more commands. With the upcoming work to add pipeline support with Redis Cluster (#307), I observed a 42% reduction in execution time when issuing a pipeline query with 100 elements.

For backwards compatability, a new trait `Routable` is introduced and the code that determines which node(s) should handle a
particular command is made generic over said trait.

### DIFF
--- a/src/cluster_routing.rs
+++ b/src/cluster_routing.rs
@@ -1,4 +1,7 @@
-use super::{parse_redis_value, Value};
+use std::iter::Iterator;
+
+use crate::cmd::{Arg, Cmd};
+use crate::types::Value;
 
 pub(crate) const SLOT_SIZE: usize = 16384;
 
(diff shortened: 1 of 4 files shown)


src/cluster_routing.rs
```diff
@@ -115,6 +145,32 @@ impl Slot {
...
+        let start = if idx == 0 {
+            0
+        } else {
+            match self.args[idx - 1] {
+                Arg::Simple(n) => n,
+                _ => 0,
+            }
+        };
+        let end = match self.args[idx] {
+            Arg::Simple(n) => start + (n - start),
```
Marwes, line 164:
```suggestion
            Arg::Simple(n) => n,
```

src/cmd.rs
```diff
@@ -16,9 +16,9 @@ pub enum Arg<D> {
 /// Represents redis commands.
 #[derive(Clone)]
 pub struct Cmd {
-    data: Vec<u8>,
+    pub(crate) data: Vec<u8>,
     // Arg::Simple contains the offset that marks the end of the argument
-    args: Vec<Arg<usize>>,
+    pub(crate) args: Vec<Arg<usize>>,
```
Marwes, line 21: Could just move the `arg_idx` method into this module and make the method `pub(crate)`

Marwes (changes requested):

Marwes (approved):

PR #456, opened 2021-02-20
add support for converting binary data to bytes::Bytes

tdyas:
Add an implementation of `FromRedisValue` for converting `Value::Data` to `bytes::Bytes`. This is behind a `bytes-comp` feature.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1242,7 +1242,7 @@ impl<T: FromRedisValue> FromRedisValue for Option<T> {
     }
 }
 
-#[cfg(feature = "bytes-comp")]
+#[cfg(feature = "bytes")]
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -1242,6 +1242,16 @@ impl<T: FromRedisValue> FromRedisValue for Option<T> {
     }
 }
 
+#[cfg(feature = "bytes-comp")]
```
Marwes, line 1245:
```suggestion
#[cfg(feature = "bytes")]
```
And remove the `bytes-comp` feature, it is not necessary.

tdyas: Done.

Marwes:
Thanks!

Issue #458, opened 2021-02-23
Support MINID stream length-capping strategy released in Redis 6.2


scooter-dangle:
With the release of Redis 6.2, `MINID` is now available for [`XTRIM`](https://redis.io/commands/xtrim) and [`XADD`](https://redis.io/commands/xadd).

I started to look at what it would take to add support, and it feels like it would require some backward-incompatible interface changes. The current `MAXLEN` enum is called [`StreamMaxlen`](https://docs.rs/redis/0.20.0/redis/streams/enum.StreamMaxlen.html), which would need to be renamed. (I'm dubious of adding a separate enum, since I think it would over-complicate the command interface.)

I propose modifying that enum like this:
```rust
/// Utility enum for passing `MAXLEN|MINID [= or ~] threshold|id`
/// arguments into `StreamCommands` for XADD and XTRIM.
/// The enum value represents the count.
#[derive(PartialEq, Eq, Clone, Debug)]
pub enum StreamSizeLimit {
    /// Match an exact count
    MaxlenEquals(usize),
    /// Match an approximate count
    MaxlenApprox(usize),
    /// All entries before an ID
    MinidEquals(String),
    /// All entries before a point _somewhere_ before an ID
    MinidApprox(String),
}
```
I would also move `xadd_maxlen` and `xadd_maxlen_map` to `xadd_size_limited` and `xadd_size_limited_map`. (Technically would _add_ the latter 2 and mark the former 2 as deprecated.)

#### Naming

Any thoughts on names of the enum, its variants, or the `xadd_*` functions?

#### Graceful deprecation

Do we want to provide some assistance in helping users discover the change? I think it's likely that most usages of `StreamMaxlen` are in strict creation of the variants—not in matching on them. In that case, temporarily retaining the current variants and providing a type alias should point people in the right direction with warnings prior to making it a breaking change.

```rust
/// Utility enum for passing `MAXLEN|MINID [= or ~] threshold|id`
/// arguments into `StreamCommands` for XADD and XTRIM.
/// The enum value represents the count.
#[derive(PartialEq, Eq, Clone, Debug)]
pub enum StreamSizeLimit {
    #[deprecated(
        since = "0.21.0",
        note = "Use StreamSizeLimit::MaxlenEquals instead. Changed to match increase in API surface in Redis 6.2"
    )]
    /// Match an exact count
    Equals(usize),
    #[deprecated(
        since = "0.21.0",
[...]

PR #459, opened 2021-02-23
feat: streams MINID from Redis 6.2

scooter-dangle:
Adds support for capping stream length via MINID (an alternative
to MAXLEN) when using XADD or XTRIM.

Question: Do we need additional MINID test cases?

Resolves mitsuhiko/redis-rs#458

- - -

This PR follows the proposal in #458, which might not be everyone's preference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,9 @@ before_install: |
   sudo apt-get install stunnel -y
 
 install: |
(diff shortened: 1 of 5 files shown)


nihohit:
fixed in https://github.com/redis-rs/redis-rs/pull/1361

PR #462, opened 2021-02-27
Draft: TLS support for Cluster mode

rohitjoshi:
Fixed issue to enable TLS support for Cluster mode:

fixes: https://github.com/mitsuhiko/redis-rs/issues/461

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -166,6 +166,13 @@ impl ClusterConnection {
         for info in initial_nodes.iter() {
             let addr = match *info.addr {
                 ConnectionAddr::Tcp(ref host, port) => format!("redis://{}:{}", host, port),
+                ConnectionAddr::TcpTls{ref host, port, insecure} => {
+                    if insecure {
+                      format!("rediss://{}:{}#insecure", host, port)
+                    }else {
+                      format!("rediss://{}:{}", host, port)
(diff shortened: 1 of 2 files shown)


rohitjoshi:
@doyshinda to populate redis scheme for a node based on scheme, I have hard coded logic for insecure mode. Any suggestions to retrieve insecure flag?

doyshinda:
@rohitjoshi the `ClusterConnection` struct already contains several fields tracking connection metadata (e.g., read/write timeouts, auto-reconnect, etc.), so I would suggest adding the scheme to it and passing it to `get_slots`. A nice cleanup would be to refactor the various connection metadata attributes into a new struct type (maybe something like `ClusterConnectionMetadata`).

Additionally, a change like this needs testing. I would invest effort to make the existing cluster tests run with TLS when the `REDISRS_SERVER_TYPE=tcp+tls` env is set, similar to how the other Redis [tests](https://github.com/mitsuhiko/redis-rs/blob/master/tests/support/mod.rs#L59-L65) are doing it (cluster tests default to [tcp only](https://github.com/mitsuhiko/redis-rs/blob/master/tests/support/cluster.rs#L29) ).

tasn:
Is this still an issue? That ticket seems to be closed.

PR #464, opened 2021-03-09
Replace deprecated crates

weihanglo:
Resolve these two RustSec advisories:

- [RUSTSEC-2018-0017](https://rustsec.org/advisories/RUSTSEC-2018-0017.html)
- [RUSTSEC-2020-0016](https://rustsec.org/advisories/RUSTSEC-2020-0016.html)


Marwes:
Thanks!

PR #465, opened 2021-03-16
testing: `MockConnection` helper for testing clients without a server

tdyas:
This PR adds `MockConnection` which implements `crate::connection::ConnectionLike` and `crate::aio::ConnectionLike` to appear like a client without the need to make any network requests. It is initialized with a list of `MockCmd` instances which wrap the expected commands and responses. This allows users to test their clients without much overhead.

The `testing` module is behind the `testing` feature which is non-default. Users should activate it in the `dev-dependencies` of their crate's `Cargo.toml`.


tdyas:
Open question: What would be the suggested way to test `Pipeline` with `MockConnection`?

tdyas:
I cleaned up the PR, added module docs with example, and made sure to test by enabling the `aio` feature. This should be ready for review. (I am using a variant of this code locally.)

tdyas:
cc @Marwes -- It was unclear to me how to flag a PR for attention so decided to cc'ed you here. Is there a better way to go about getting attention for a PR? Are you the right person?

tdyas:
Rebased on latest `main` and force pushed.

Cargo.toml
```diff
@@ -68,6 +68,7 @@ script = ["sha1_smol"]
 tls = ["native-tls"]
 async-std-comp = ["aio", "async-std"]
 async-std-tls-comp = ["async-std-comp", "async-native-tls", "tls"]
+testing = ["futures"]
```
djc, line 71: Is this intended to be part of the public API?

src/testing.rs
```diff
@@ -0,0 +1,406 @@
...
+use crate::connection::ConnectionLike;
+use crate::{Cmd, ErrorKind, Pipeline, RedisError, RedisResult, Value};
+
+#[cfg(feature = "aio")]
+use crate::RedisFuture;
+
+#[cfg(feature = "aio")]
+use crate::aio::ConnectionLike as AioConnectionLike;
+
+/// Helper trait for converting test values into a `redis::Value` returned from a
```
djc, line 43: How/why is `ToRedisArgs`/`FromRedisValue` different from what we need here?

Cargo.toml
tdyas: Yes, it is intended to be part of the public API so that users (like me) can incorporate these test helpers into the unit tests in our codebases. I have made it be a feature here (which would be enabled as a dev dependency), but it could also be put into a separate crate, much like Tokio has done with the [`tokio-test` crate](https://github.com/tokio-rs/tokio/tree/master/tokio-test). Thoughts?

djc: A separate crate would allow evolving the API separately from the rest of the crate's API. I suppose we could keep the feature unstable for now (as in, we don't guarantee the public API to be compatible across semver-compatible versions of the crate).

tdyas: I will restructure this PR to create a `redis-test` crate. Is that acceptable?

djc: @jaymell what do you think? I'm just worried the API might be unstable, but maybe it's not so bad?

jaymell: Hmm, I would say that a separate crate certainly does sound like an ideal solution, though the additional maintenance costs there do temper my enthusiasm somewhat :sweat_smile:

tdyas: I'll switch to separate crate then (but the code will still live in this repo).

Cargo.toml
```diff
@@ -10,6 +10,9 @@ documentation = "https://docs.rs/redis"
 license = "BSD-3-Clause"
 edition = "2018"
 
+[workspace]
```
djc, line 13: I tend to find nested crate workspaces confusing to navigate. Would you mind changing this so that the top-level `Cargo.toml` only contains the workspace, and move the redis crate into a `redis` directory?

redis-test/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "redis-test"
+version = "0.21.5"
```
djc, line 3: Let's just call this 0.1.0.

redis-test/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+description = "Testing helpers for the `redis` crate"
+homepage = "https://github.com/redis-rs/redis-rs"
+repository = "https://github.com/redis-rs/redis-rs"
+documentation = "https://docs.rs/redis-test"
+license = "BSD-3-Clause"
+
+[dependencies]
+redis = { path = ".." }
```
djc, line 12: If we're going to publish this it should also have a specific version.

redis-test/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+bytes = { version = "1", optional = true }
```
djc, line 14: Why is `bytes` optional?

Cargo.toml
tdyas: Should that be done in a separate PR?

tdyas: (to avoid confusing commit history)

redis-test/Cargo.toml
tdyas: Because there is a conversion for `bytes::Bytes` in the crate which should only exist if the user is using `bytes::Bytes` in their own crate.

Cargo.toml
djc: In my mind the ideal way to submit it would be a single PR with clean commits: one commit that moves the code around and one commit that adds the new crate. Each commit should ideally pass CI, though I usually don't test that explicitly (but it sucks if formatting changes for one commit make it into another).

tdyas: So it sounds like that PRs are not squashed before merge then. Will do as separate commits in this PR then. (Context: My company merges via squash-and-merge generally for PRs so a separate PR would be the only way to accomplish having separate commit history in that case.)

djc: I'm also okay with separate PRs if you prefer. For my own changes I usually submit clean commits "manually" (squashing with interactive rebase).

redis-test/Cargo.toml
tdyas: I added a `version`, but kept the `path` so it continues to work with in-repo builds. Tokio does something similar.

tdyas:
I rebased the PR and turned it into two commits: the first moves the existing `redis` crate sources into a `redis` subdir, then second commit introduces the `redis-test` crate.

tdyas:
The CI tests with Rust 1.51.0 fails due to `async-std` dependency `async-global-executor` using 2021 Edition.

tdyas:
Rebased and also moved the contents of `redis_test::mock_conn` into just `redis_test`.

tdyas:
Repushed after making sure subsequent updates to Cargo.toml were taken into account.

tdyas:
Note: The lint job in CI is failing on existing code, and not the new `redis-test` crate. https://github.com/redis-rs/redis-rs/runs/7183833699?check_suite_focus=true#step:6:40

djc:
@jaymell any thoughts before we merge this?

I'll take care of the new clippy issues separately.

SET001:
is redis_test published on crates.io?

PR #474, opened 2021-04-03
WIP: Add zrangestore-- to commands.

onelson:


Marwes (changes requested):
Rather than adding one function for each combination could we add a set of types to handle them and then a single function? Something like

```rust
struct ZRangeStore {
}
enum ZRangeStoreBy {
    Limit { offset : isize, count: isize },
    ByScore,
}
// etc
```

onelson:
I think that'd be a nice way to handle it, but I'd be worried about creating a split in the overall API. Currently all the symbols in there use the "one function for each permutation of params" pattern.

I need to test this so I can actually mark this as ready for review :sweat_smile:

onelson:
@Marwes are there other commands where the enum approach has already been adopted? Seems like the Builder Pattern could also work to clean up the callsites, albeit with an extra pile of boilerplate on the implementation side.

Marwes:
I think the geolocation api does? Personally I would like to move towards using types to represent options for one command rather than multiple functions, at least in big ones like this.

onelson:
Okay, I'll give it a try. There are many new commands added in 6.2 that could benefit from such a design shift, I just didn't want to bring my own design sense to this. When in Rome, as they say. I'll look for prior art in the geo APIs and try to match the style found there.

onelson:
I've pushed a new commit aiming for a single `zrangestore()` with an options parameter, based loosely on what I saw with the `RadiusOptions` and `LposOptions` types.

There are some notable differences in approach, and I hope we can bikeshed a little bit before I get busy on docs/tests. I'm still reading to get my head around how the tests in this project work.

Anyway... on with the bikeshedding.

With The `LposOptions` and `RadiusOptions` types, they largely operate as mutable builders where the caller calls methods to manage internal fields. I didn't feel like this was such a good fit in the case of zrangestore since many of the internal fields would be mutually exclusive. For example, you can't use both `BYLEX` and `BYSCORE` together so the methods on these builders would need to unset the one when updating the other. It's a little messy.

That said, it's totally possible to follow the builder-ish approach with some carefully written docs to try and explain the inner state machine. It could even simplify other parts of the impl compared to how it's written now. The matches for seeing which specific Redis commands need to be generated would involve far less matching if we just held bools for whether `REV` was enabled, etc.

I may go on to do the builder-ish approach just to see how it feels, but I hoped to check in to make sure I didn't completely misunderstand the direction from @Marwes.

onelson:
I "split the baby in half" so to speak in the 3rd commit. This version feels like a decent compromise to me.

The approach is to provide:

- command functions for each of the exclusive options (default, `BYSCORE` or `BYLEX`)...
- each relying on a common owned-builder type (based on prior art) for the `REV` and `LIMIT`.

Splitting the responsibility down the line of exclusive and non-exclusive options greatly simplifies the supporting code for the `ZRangeStoreOptions` type.

If not for the `WITHSCORES` option which is supported *only on `zrange`* and not on `zrangestore`, this `ZRangeStoreOptions` type could perhaps be used in both situations.

Food for thought, looking forward. It would be nice for all the `zrange-` commands to have higher API affinity than they do now.

I still owe tests to verify the work and some docs on the command functions themselves.

src/commands.rs
```diff
@@ -843,6 +844,24 @@ implement_commands! {
...
 
+    /// This command is like `ZRANGE`, but stores the result in `dst` destination key.
+    fn zrangestore<DST: ToRedisArgs, K: ToRedisArgs, M: ToRedisArgs, MM: ToRedisArgs>
+            (dst: DST, src: K, min: M, max: MM, opts: ZRangeStoreOptions) {
+        cmd("ZRANGESTORE").arg(dst).arg(src).arg(min).arg(max).arg(opts)
+    }
+
+    /// This command is like `ZRANGE`, but stores the result in `dst` destination key.
+    fn zrangestore_byscore<DST: ToRedisArgs, K: ToRedisArgs, M: ToRedisArgs, MM: ToRedisArgs>
+            (dst: DST, src: K, min: M, max: MM, opts: ZRangeStoreOptions) {
```
Marwes, line 855: If we add another field that takes

```rust
enum ZRangeStoreOrdering {
    Default,
    ByScore,
    ByLex,
}
```
we only need the single `zrangestore` function.

onelson: I took the long way to get here :sweat_smile:. This is better.

onelson:
Looking at the CI/test setup and it appears travis is building redis from source, but the version is too old (6.0.6). `ZRANGESTORE` was added in 6.2.0.

I'm going to briefly write some tests in a separate crate and target a docker container (so I can iterate easier). Not sure how best to update CI to target a newer redis (or if I should even do this).

onelson:
I updated the Redis version for travis to 6.2.3. I see one failure but it looks like it could be unrelated (I hope!) to the changes I've made to the crate.

```
test test_acl_getsetdel_users ... thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Ok(["user default on nopass ~* &* +@all"])`,
 right: `Ok(["user default on nopass ~* +@all"])`', tests/test_acl.rs:30:5
```
https://travis-ci.org/github/mitsuhiko/redis-rs/jobs/772422644#L2267

onelson:
As for the appveyor checks, I'm at a loss. It looks as if the "redis-64" package installed via nuget is too old, and I don't immediately see a suitable alternative.

I did notice the [appveyor docs](https://www.appveyor.com/docs/server/) talk about configuring jobs that depend on Redis by using a docker image but I'm not sure what impact that would have on the rest of the CI scripting.

Marwes:
If you can't get a newer redis on appveyor you could ignore those tests when running there with `cfg(windows)` or an explicit test feature.

Marwes:
> https://github.com/mitsuhiko/redis-rs/pull/474#issuecomment-848497532

Just updating the test here seems fine, probably just a format change in redis (you didn't change anything around the ACL so)

onelson:
Regarding getting a newer redis on appveyor, I just haven't found a path forward that doesn't involve doing a lot of study about multi-env appveyor jobs (so we can run redis in docker or something) and also refactoring the test support module handle talking to a redis that it doesn't spawn itself.

In light of this, I went the route of adding a feature to gate the new zrangestore-related symbols. It made sense to me to name the feature for the redis version, but it can be renamed as you see fit.

Since the appveyor build disables default features, the new stuff will be skipped, allowing it to finish "green."

I branched with `cfg!()` inside the acl test to *expect a different response* depending on whether or not the redis 6.2 feature is enabled, but unfortunately, it's just failing differently a few lines down from there.

This time, it's a little more scary lookin'.

```
test test_acl_getsetdel_users ... thread 'main' panicked at 'Got user: Response type not convertible: "Expect a valid ACL flag" (response was [97, 108, 108, 99, 104, 97, 110, 110, 101, 108, 115])', tests/test_acl.rs:61:52
```

I'm going to have to read up on ACL in order to debug this further as I haven't used it before. Seems like some of the existing ACL-related commands may need refactoring to support newer redis versions.

PR #475, opened 2021-04-08
docs: add "connection-manager" cfg attr to aio

Fishrock123:
It's very unclear in the docs that this is not just enabled by the "aio" feature, but rather the "connection-manager" feature.

### DIFF
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -1095,4 +1095,5 @@ mod connection_manager {
 }
 
 #[cfg(feature = "connection-manager")]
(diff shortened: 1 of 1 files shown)


Fishrock123:
Warnings: clippy & deprecations on stable which haven't been address in the repo generally
Error seems unrelated?:
```
failures:

    test_cluster_pipeline_ordering_with_improper_command

test result: FAILED. 8 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 24.44s
```

PR #481, opened 2021-04-26
Add get_connection_info method to `Client` struct

GnaneshKunal:
This allows consumers to peak into the `Client's` `connection_info` field. Here's an issue about the same: https://github.com/mitsuhiko/redis-rs/issues/480#issue-867822352

Fix: https://github.com/mitsuhiko/redis-rs/issues/480

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -57,6 +57,11 @@ impl Client {
     pub fn get_connection_with_timeout(&self, timeout: Duration) -> RedisResult<Connection> {
         Ok(connect(&self.connection_info, Some(timeout))?)
     }
+
+    /// Returns a clone of client connection info object.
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -57,6 +57,11 @@ impl Client {
     pub fn get_connection_with_timeout(&self, timeout: Duration) -> RedisResult<Connection> {
         Ok(connect(&self.connection_info, Some(timeout))?)
     }
+
+    /// Returns a clone of client connection info object.
+    pub fn get_connection_info(&self) -> ConnectionInfo {
+        self.connection_info.clone()
```
Marwes, line 63: Could just return a reference to this instead.

PR #486, opened 2021-05-09
fix: do not read pubsub responses

enzious:
## Bugfix
 - Issue: https://github.com/mitsuhiko/redis-rs/issues/310
 - Fix:
   - Stop reading responses from pubsub commands.
   - Eliminate the need to use two different forms of encoders to read from the connection for async pubsub

## Example of Issue
```
< pubsub message comes in
> subscribe request
< subscribe function reads pubsub message instead of subscribe response
[subscribe response is left hanging]
```


enzious:
@Marwes (mentioning you since you're the last active collaborator :p)

Could you look at this PR and why it's failing on the cluster test. I noticed that another PR is failing in the same place. It's only happening on the "stable" version of rust.

Marwes:
I have no idea why the test is failing (and only on stable), not sure when I will have a chance to look into it.

It would be good to have a test for this change.

djc:
@enzious are you still interested in getting this merged? Would be nice to rebase it.

Owez:
Bump, this looks like it'd be useful for some mass subscribing I'm wanting to do

Owez:
> @Owez looks like the author abandoned this in favor of #633, would that work for you?

Thanks for the reply, I think that'd work for me. Do you know if #633 might fix/help #509 (not being able to subscribe after calling `on_message`) from the top of your head?

djc:
I don't, but I guess it seems likely.

PR #498, opened 2021-06-05
Add wildcard aio::ConnectionLike implementation for pointer types

TilBlechschmidt:
Currently, it is not possible to use types like `Box<dyn aio::ConnectionLike>` or `Arc<dyn ...>` (pointer types in general) with the connections provided by this crate.

In my scenario, I've been building a factory that provides connections appropriate for the context (e.g. a multiplexed instance, or a blocked one from a connection pool). However, as I implemented my own wrappers around the `ConnectionLike` implementations to handle some error cases, it makes sense to use pointer types and just pass the "consumer" of the factory whatever I want or need in that case.

While one *could* use generics all the way for this, it becomes quite cumbersome if not impossible when having a function in a trait that returns either a multiplexed or an owned connection. 

Correct me if I'm overlooking anything obvious, but I feel like this would be a neat little QoL addition (esp. since Rust disallows one to implement traits for objects from other crates and `Box<dyn ConnectionLike>` counts into that).

image

### DIFF
--- a/src/aio.rs
+++ b/src/aio.rs
@@ -586,6 +586,31 @@ where
     }
 }
 
+impl<T> ConnectionLike for T
+where
+    T: AsRef<dyn ConnectionLike + Send + Sync> + AsMut<dyn ConnectionLike + Send + Sync>,
+{
+    fn req_packed_command<'a>(
+        &'a mut self,
+        cmd: &'a crate::Cmd,
+    ) -> crate::RedisFuture<'a, crate::Value> {
+        self.as_mut().req_packed_command(cmd)
+    }
+
+    fn req_packed_commands<'a>(
+        &'a mut self,
+        cmd: &'a crate::Pipeline,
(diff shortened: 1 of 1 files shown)


TilBlechschmidt:
The pipeline apparently had a test failure on the `stable` compiler with some redis cluster-related test. I suspect it has nothing to do with this addition and more with a flaky test, but that is just a gut feeling. Would need confirmation on that 🙂

TilBlechschmidt:
Rebased against the latest `master` again and re-ran with the GitHub Actions successfully (no surprises there as it is just a generic trait addition).

If there is anything else left to do for me here, let me know @Marwes 🙂

TilBlechschmidt:
Bump.

Is there anything that goes against merging this PR? I'd love to get this upstream so that I can move my dependencies back to the crates.io version 🙂

nihohit:

PR #499, opened 2021-06-08
Make xread_options take &StreamReadOptions

### LINKED ISSUE #496: StreamReadOptions is not duplicatable
Currently, it is required to reconstruct the `StreamReadOptions` struct for each call to `xread_options`. Especially when implementing a `futures::Stream` like interface to interact with a redis stream, this becomes really cumbersome as one can not just pass a `StreamReadOptions` to a function that repeatedly calls `xread_options` and converts that into an `impl futures::Stream`.

I'd suggest either making `xread_options` take a reference to the `StreamReadOptions` or making it cloneable. The former is definitely preferred as it does not require additional memory copies.

As far as I understand it, `ToRedisArg` does not consume the instance so it should be no problem to just use a reference.

TilBlechschmidt:
Closes #496 

Previous discussion here: https://github.com/TilBlechschmidt/redis-rs/commit/ddf6902c08b03a6bc858afea3585ecce12c0edb3#r51877701

I think it makes sense to continue discussions in the context of a PR.

### DIFF
--- a/examples/streams.rs
+++ b/examples/streams.rs
@@ -212,7 +212,7 @@ fn read_records(client: &redis::Client) -> RedisResult<()> {
     let another_form = "0";
 
     let srr: StreamReadReply = con
-        .xread_options(STREAMS, &[starting_id, another_form, starting_id], opts)
+        .xread_options(STREAMS, &[starting_id, another_form, starting_id], &opts)
         .expect("read");
 
     for StreamKey { key, ids } in srr.keys {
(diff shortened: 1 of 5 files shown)


TilBlechschmidt:
Okay then, is there anything left for me to do here? I've cleaned up a double-borrow that was previously required, so the code changes should be pretty much minimal now.

Marwes:
Travis seems to have issues so I moved over to github actions. If you rebase CI should hopefully pass.

TilBlechschmidt:
> Travis seems to have issues so I moved over to github actions. If you rebase CI should hopefully pass.

Done! Although I'm not sure GitHub Actions runs automagically on my fork, you might have to approve something.

I also rebased PR #498 in case you want to give that a look as well 🙂

TilBlechschmidt:
Fixed another broken doc-test which slipped through my fingers due to the default `fail-fast` policy of `cargo test` 🤦

PR #506, opened 2021-06-18
Introduce lpop bulk WIP

DoumanAsh:
LPOP has optional `count` parameter allowing to specify number of elements to retrieve


src/commands.rs
```diff
@@ -589,6 +589,11 @@ implement_commands! {
         cmd("LPOP").arg(key)
     }
 
+    /// Removes and returns up to `count` number of elements from the list stored at `key`
+    fn lpop_bulk<K: ToRedisArgs>(key: K, count: usize) {
```
Marwes, line 593: Could you add this as `Option<usize>` to `lpop` instead? We already have some breaking changes going out soon

DoumanAsh: Sounds good

DoumanAsh:
@Marwes
There is one problem with count parameter, according to documentation:

```
>= 6.2: Added the count argument.
```

So it is pretty new and wouldn't be available on some old versions (e.g. windows port of redis)

Marwes:
Ignore that check on windows.

DoumanAsh:
@Marwes can you please trigger githuc CI?

DoumanAsh:
Oops my bad I reverted update of redis version...

DoumanAsh:
test_acl_getsetdel_users fails with:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Ok(["user default on nopass ~* &* +@all"])`,
 right: `Ok(["user default on nopass ~* +@all"])`', tests/test_acl.rs:30:5
```

I actually have no clue what is ACL LIST, but maybe it could be because of new redis version

Marwes:
I have no idea what it is supposed to test, but it seems that the output changes between versions so I think we should just ignore that test.

DoumanAsh:
@Marwes I put this test under `ignore` then.

Issue #336, opened 2020-06-08
Variable-length mget response


jonhoo:
The current API allows you to do an `MGET` using `conn.get(keys)` where `keys` is a `&[T: ToRedisValue]`. However, it's not clear (at least to me) how to get the _values_ from such a lookup. If I _know_ the number of keys, I can do:
```rust
assert_eq!(keys.len(), 2);
let (v1, v2): (i32, i32) = conn.get(keys)
```
But this doesn't work if `keys` is of variable length.

My first instinct was to do
```rust
let values: Vec<i32> = conn.get(keys)
```

But that fails with:
```
Response was of incompatible type: "Response type not vector compatible." (response was string-data('"0"'))
```

Is there a way to achieve this with the current API (except by doing single lookups or ignoring the result entirely)?

jonhoo:
I think what's actually happening here is that I was accidentally passing a vector of length 1, and in _that_ case the command isn't an `MGET`, and so the response isn't an array. This is a pretty surprising and tricky-to-debug outcome. I wonder if the "automatic MGET" should always be put into effect when given a list of keys, even if the list length happens to be 1.

Terkwood:
I agree that the current strategy of returning a single value when passed a vector of keys is counter-intuitive.  I think that this would be worth making a breaking change to the API, as you suggest.

There's a good argument in the other direction, too, since breaking changes will disrupt users.

xiaodonng:
came across the same issue.
agree with @jonhoo that the return value should correspond to the passed vector, most of the times the length of the vector can only be determined during runtime. This should be an issue which is very easy bumping into.

dacioromero:
Ran into this issue myself and attempted to reproducible example of this issue: https://github.com/dacioromero/redis-rs-vec

I believe the culprits are the following:

https://github.com/mitsuhiko/redis-rs/blob/1d14ce077f305043bb71e029bad8f97ab8867461/src/commands.rs#L280-L282

https://github.com/mitsuhiko/redis-rs/blob/1d14ce077f305043bb71e029bad8f97ab8867461/src/types.rs#L805-L807

https://github.com/mitsuhiko/redis-rs/blob/1d14ce077f305043bb71e029bad8f97ab8867461/src/types.rs#L682-L685

It might be as simple as always treating `Vec`s and maybe even slices as multiple arguments, though it could be an annoying breaking change for people who were relying on this behavior.

For now I'm just going to use `redis::cmd("MGET")` instead.

[...]

PR #507, opened 2021-06-19
fix: Variable-length mget response

hank121314:
## Summary

This PR fixes: #336.

I try to implement `FromRedisValue::from_byte_vec` which can convert `bytes` to a single element vector.
With this change, when user is using `let values: Vec<i32> = conn.get(keys)`, no matter how long the `keys` is, user should get the correct response.

This is my first contribution, please feel free to critique and thanks for your code review 😄 .

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -989,11 +989,9 @@ pub trait FromRedisValue: Sized {
             .collect())
     }
 
-    /// This only exists internally as a workaround for the lack of
-    /// specialization.
-    #[doc(hidden)]
+    /// Convert bytes to a single element vector.
     fn from_byte_vec(_vec: &[u8]) -> Option<Vec<Self>> {
-        None
+        Self::from_redis_value(&Value::Data(_vec.into())).map(|rv| vec![rv]).ok()
     }
 }
 
@@ -1030,6 +1028,7 @@ impl FromRedisValue for u8 {
         from_redis_value_for_num_internal!(u8, v)
     }
 
+    // this hack allows us to specialize Vec<u8> to work with
     fn from_byte_vec(vec: &[u8]) -> Option<Vec<u8>> {
         Some(vec.to_vec())
     }
@@ -1092,8 +1091,7 @@ impl FromRedisValue for String {
 impl<T: FromRedisValue> FromRedisValue for Vec<T> {
     fn from_redis_value(v: &Value) -> RedisResult<Vec<T>> {
         match *v {
-            // this hack allows us to specialize Vec<u8> to work with
-            // binary data whereas all others will fail with an error.
+            // All binary data except u8 will try to parse into a single element vector.
             Value::Data(ref bytes) => match FromRedisValue::from_byte_vec(bytes) {
                 Some(x) => Ok(x),
                 None => invalid_type_error!(v, "Response type not vector compatible."),
--- a/src/types.rs
+++ b/src/types.rs
@@ -991,7 +991,9 @@ pub trait FromRedisValue: Sized {
 
     /// Convert bytes to a single element vector.
     fn from_byte_vec(_vec: &[u8]) -> Option<Vec<Self>> {
-        Self::from_redis_value(&Value::Data(_vec.into())).map(|rv| vec![rv]).ok()
+        Self::from_redis_value(&Value::Data(_vec.into()))
+            .map(|rv| vec![rv])
+            .ok()
     }
 }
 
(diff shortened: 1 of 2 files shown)


hank121314:
Hi @djc, would you be willing to review this 🙇? Thanks!

djc:
It's on my list, but I'm very busy.

hank121314:
It's ok! Many thanks for maintaining the repository 🙇 !

src/types.rs
```diff
@@ -989,11 +989,9 @@ pub trait FromRedisValue: Sized {
             .collect())
     }
 
-    /// This only exists internally as a workaround for the lack of
-    /// specialization.
-    #[doc(hidden)]
+    /// Convert bytes to a single element vector.
     fn from_byte_vec(_vec: &[u8]) -> Option<Vec<Self>> {
-        None
```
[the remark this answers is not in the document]
hank121314: Assume that we have a variable like this `let values: Vec<i32> = conn.get(keys)`.
If `keys` is a single element vector, it will use `redis.cmd("GET")` which might return `Value::Data`.
And when the return data is `Value::Data`, it will raise an error at: https://github.com/redis-rs/redis-rs/blob/470d7c58713e7c39e65ea2e84c0e5ae2abcc147d/redis/src/types.rs#L1166-L1169

This PR tried to parse a `Value::Data` to a single element vector instead of raising an error.

jaymell:
Sorry if I'm missing something here, but does this actually fix #336? I'm not sure why it would, and in any case there's not actually a test here to ensure that it does.

hank121314:
> in any case there's not actually a test here to ensure that it does.

Hi @jaymell !

I didn't add an e2e test for this case.
I only write some unit tests for [FromRedisValue::from_redis_value](https://github.com/redis-rs/redis-rs/blob/82eeb45ac71a2c178af5088b913469f9597595f4/redis/src/types.rs#L1238), because the error is generated from it.

Here is a example e2e test for this case:

```rust
#[test]
fn test_get() {
    let ctx = TestContext::new();

    let _: () = con.set(1, "1").unwrap();
    let keys = vec![1];
    assert_eq!(keys.len(), 1);
    let data: Vec<String> = con.get(&keys).unwrap();
    assert_eq!(data, vec!["1"]);
}
```

It should raise an error before applying this PR.

Thanks for your review!

jaymell:
Please add the additional test to the pr, because again, I'm skeptical that this will actually fix the problem.

jaymell:
In any case I don't think this PR as is can claim to fix #336 if it doesn't even assert through a test that it does so. So please add the test 🙏

hank121314:
> In any case I don't think this PR as is can claim to fix https://github.com/redis-rs/redis-rs/issues/336 if it doesn't even assert through a test that it does so. So please add the test 🙏

That's true! I have added the test.

jaymell:
Thanks for adding the test.

redis/src/types.rs
```diff
@@ -1211,8 +1212,7 @@ impl FromRedisValue for String {
 impl<T: FromRedisValue> FromRedisValue for Vec<T> {
     fn from_redis_value(v: &Value) -> RedisResult<Vec<T>> {
         match *v {
-            // this hack allows us to specialize Vec<u8> to work with
-            // binary data whereas all others will fail with an error.
+            // All binary data except u8 will try to parse into a single element vector.
             Value::Data(ref bytes) => match FromRedisValue::from_byte_vec(bytes) {
                 Some(x) => Ok(x),
                 None => invalid_type_error!(v, "Response type not vector compatible."),
```
jaymell, line 1218: I think the message `"Response type not vector compatible."` is no longer appropriate here, as any failure here will now be due to the failure in calling `Self::from_redis_value`

hank121314: Agree! How about `Fail to convert strint-data("any") via Self::from_redis_value`?

jaymell: As long as it's clear and spelled correctly, it's fine. I think something as simple as this would probably be ok:

```None => invalid_type_error!(v, "Conversion to Vec<T> failed."),```

djc: Sure, I think that makes sense!

hank121314: Updated! Thanks for the suggestion!

djc (approved):

jaymell:
Thanks for the PR -- maybe next time we can get it merged in under a year-and-a-half :stuck_out_tongue_winking_eye:

PR #519, opened 2021-07-29
fix: pin futures-util dependency to required version

### LINKED ISSUE #515: 0.21 fails compilation with default-features = false and tokio-comp enabled
```rust
error[E0432]: unresolved import `futures_util::Stream`
 --> /Users/djc/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.21.0/src/cmd.rs:4:16
  |
4 |     FutureExt, Stream,
  |                ^^^^^^
  |                |
  |                no `Stream` in the root
  |                help: a similar name exists in the module (notice the capitalization): `stream`
```

Maybe `tokio-comp` should also enable `streams`, or some `tokio-comp`-conditional items should also depend on `streams`?

(This broke bb8-redis.)

seanpianka:
fixes #515:
> error[E0432]: unresolved import `futures_util::Stream`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ combine = { version = "4.6", default-features = false, features = ["std"] }
 
 # Only needed for AIO
 bytes = { version = "1", optional = true }
-futures-util = { version = "0.3.0", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


djc:
@Marwes could you publish this to crates.io? Would be nice to have.

PR #529, opened 2021-08-19
"Null Bulk String" as None for Option<T>.

mstyura:
Hello. I've found out one inconvenient thing. I had custom type implemented `FromRedisValue`. But I couldn't properly deserialize it from result of `GET` query when key not found into `Option<MyCustomType>`. This happen because absence of key is returned as `bulk(nil)` and default `FromRedisValue for Option<T>` calls to underlying converter for `bulk(nil)`.
 
[According to documentation](https://redis.io/topics/protocol#bulk-string-reply) `bulk(nil)` (Null Bulk String) should be deserialized as `nil/None/NULL/nullptr`:

> This is called a Null Bulk String. The client library API should not return an empty string, but a nil object, when the server replies with a Null Bulk String. For example a Ruby library should return 'nil' while a C library should return NULL (or set a special flag in the reply object), and so forth.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1266,10 +1266,11 @@ impl FromRedisValue for InfoDict {
 
 impl<T: FromRedisValue> FromRedisValue for Option<T> {
     fn from_redis_value(v: &Value) -> RedisResult<Option<T>> {
-        if let Value::Nil = *v {
-            return Ok(None);
+        match v {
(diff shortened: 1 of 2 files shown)


mstyura:
Ah, this seems to be invalid. All the extra `bulk` things seems to be due to usage of cluster connection.

PR #542, opened 2021-09-24
remove stunnel as a dep, use redis native tls

### LINKED ISSUE #532: Update TLS tests to use redis native TLS instead of stunnel
At the time of writing the [original patch that added TLS support](https://github.com/mitsuhiko/redis-rs/pull/305) to this library the testing mechanism being used [wasn't being built with TLS support](https://github.com/mitsuhiko/redis-rs/pull/305#issuecomment-655124274) but since moving over to [building redis in CI](https://github.com/mitsuhiko/redis-rs/pull/373) there has been the possibility of pulling out `stunnel` and taking advantage of the built in support, but it wasn't pursued at the time.

Fast forward to now, I have need for TLS + Cluster support and have [a fork where that is working](https://github.com/particle-iot/redis-rs). I took inspiration from a [previous PR](https://github.com/mitsuhiko/redis-rs/pull/462) as well as adopting the requests of the code reviewer before it could land. In trying to implement the TLS tests for cluster mode, I'm finding that trying to get it working via `stunnel` means a fair bit of refactoring of both the `tests/support/cluster.rs` and `tests/support/mod.rs`. Instead I'd like to use the native TLS support to implement these tests and would be happy to port the non-cluster mode tests to use that system as well.

For my own ease, I'd like to couple these changes together solving this issue and [the request for cluster mode + TLS support](https://github.com/mitsuhiko/redis-rs/issues/461) in one PR, but if the project maintainers would prefer them as separate PRs, working toward landing the move to redis native TLS first, then landing the TLS + Cluster mode support, I'm happy to split it up. I understand that maintaining a library is no easy feat and would prefer to not increase that burden significantly.

Thanks for taking the time to read this issue and consider it. I greatly appreciate this project and the effort that has been put into it so far.

wraithan:
Fixes #532 

The `openssl` code is more involved, but it is a direct translation of the [gen-test-certs.sh](https://github.com/redis/redis/blob/8c291b97b95f2e011977b522acf77ead23e26f55/utils/gen-test-certs.sh) from the redis repo used for testing. I'm betting it could be stripped down a bit but I'm not great with `openssl` so I just used what they did.

When I put up my PR for cluster mode + TLS, I'll likely factor the `openssl` code out into a separate function that can be called from the `tests/support/cluster.rs` module, but for now I've left it all inline.

I've verified it works locally on my macos machine and on github actions. Happy to incorporate any feedback you may have, my next day I'll definitely be able to work on this is Oct 8, though I may have some time before then.

### DIFF
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -1,14 +1,6 @@
 #![allow(dead_code)]
 
-use std::{
-    env, fs,
-    io::{self, Write},
-    net::SocketAddr,
-    path::PathBuf,
-    process,
-    thread::sleep,
-    time::Duration,
-};
+use std::{env, fs, io, net::SocketAddr, path::PathBuf, process, thread::sleep, time::Duration};
 
 use futures::Future;
 use redis::Value;
@@ -51,7 +43,6 @@ enum ServerType {
 
 pub struct RedisServer {
     pub process: process::Child,
-    stunnel_process: Option<process::Child>,
     tempdir: Option<tempfile::TempDir>,
     addr: redis::ConnectionAddr,
 }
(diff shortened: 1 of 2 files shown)


tests/support/mod.rs
```diff
@@ -214,32 +254,20 @@ impl RedisServer {
...
                     process: spawner(&mut redis_cmd),
-                    stunnel_process: None,
                     tempdir: Some(tempdir),
                     addr,
                 }
             }
         }
     }
 
-    pub fn wait(&mut self) {
```
wraithan, line 225: I couldn't find anything that used this, even with different sets of features turned on, so I removed it.

PR #548, opened 2021-10-08
Adds support for TLS with Cluster mode

### LINKED ISSUE #461: Clustermode SSL enabled:  panicked at 'No reach.'
When I enable SSL by passing "rediss://127.0.0.1:10000/", it throws panic error `No reach.`.

Here is a line number: 
https://github.com/mitsuhiko/redis-rs/blob/master/src/cluster.rs#L169

wraithan:
Fixes #461

There is still some cleanup needed of some code that is copy pasta inside of `src/cluster.rs` but it is passing all tests, and getting the tests passing actually showed me that I'd missed the ASK and MOVED responses when I did my initial implementation for my own uses, yay tests!

I noticed that `make lint` doesn't run as part of CI but found some lints failing in the areas I was already working in so I've cleaned those up.

Let me know if there is anything I can do to facilitate getting this or my other PR reviewed or merged. Thanks for maintaining this library!

EDIT: I've done the clean up I was meaning to do in `src/cluster.rs` now, it technically allocates a little more, but centralizes all the TLS connection string code.

EDIT2: Removed mentions of #542 since it was merged and it no longer matters that this is a superset.

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -71,11 +71,9 @@ pub struct ClusterConnection {
     password: Option<String>,
     read_timeout: RefCell<Option<Duration>>,
     write_timeout: RefCell<Option<Duration>>,
-    #[cfg(feature = "tls")]
     tls: Option<TlsMode>,
 }
 
-#[cfg(feature = "tls")]
 #[derive(Clone, Copy)]
 enum TlsMode {
     Secure,
@@ -126,6 +124,8 @@ impl ClusterConnection {
                     }
                 }
             },
+            #[cfg(not(feature = "tls"))]
+            tls: None,
             initial_nodes,
         };
         connection.refresh_slots()?;


src/cluster.rs
```diff
@@ -71,6 +71,25 @@ pub struct ClusterConnection {
...
+}
+
+#[cfg(feature = "tls")]
+#[derive(Clone, Copy)]
+enum TlsMode {
+    Secure,
+    Insecure,
+}
+
+impl TlsMode {
```
doyshinda, line 85: This will fail to compile when the `tls` flag is not requested:
```bash
$ REDISRS_SERVER_TYPE=tcp+tls RUST_TEST_THREADS=1 cargo test --features cluster test_parse_redis_url
   Compiling redis v0.21.3-alpha.0 (~/projects/redis-rs)
error[E0433]: failed to resolve: use of undeclared type `TlsMode`
  --> src/cluster.rs:88:13
   |
88 |             TlsMode::Insecure
   |             ^^^^^^^ use of undeclared type `TlsMode`

error[E0433]: failed to resolve: use of undeclared type `TlsMode`
  --> src/cluster.rs:90:13
   |
90 |             TlsMode::Secure
   |             ^^^^^^^ use of undeclared type `TlsMode`

error[E0412]: cannot find type `TlsMode` in this scope
  --> src/cluster.rs:85:6
   |
85 | impl TlsMode {
   |      ^^^^^^^ not found in this scope
...
```

wraithan: Ah, my bad, I'll fix that up later tonight!

Marwes:
Thanks!

PR #560, opened 2021-11-05
updated README

mreinigjr:
Updated the README sections 'Async Support', 'TLS Support' and 'Cluster Support' so that version 0.21.4 of redis is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,10 @@ To enable asynchronous clients a feature for the underlying feature need to be a
 
 ```
 # if you use tokio
-redis = { version = "0.17.0", features = ["tokio-comp"] }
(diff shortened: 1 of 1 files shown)


mreinigjr:
I noticed that the versions in the README may need to be updated because cargo build reported errors specifically for `redis = { version = "0.17.0", features = ["tokio-comp"] }`, when using cargo 1.56.0 with rustc 1.56.1. I did test the bump in version for the other stated features as well using cargo build, i.e. async-std, tokio tls, async-std tls and cluster. No additional testing besides running cargo build or code dives where performed to confirm that the README version bumps in this PR are accurate.

PR #561, opened 2021-11-25
check_connection returns RedisResult<()>

rukai:
Currently, when we have code like:
```rust
     let client = redis::cluster::ClusterClient::open(nodes).unwrap();
     let mut replication_connection = client.get_connection().unwrap();
```
And we have redis configured with a password, then we will get a useless error: "It failed to check startup nodes."

With this PR it will bubble up the error and we get:
```
thread 'redis_int_tests::basic_driver_tests::test_cluster_dr_redis' panicked at 'called `Result::unwrap()` on an `Err` value: All of the connections failed: connection errors: [
    NOAUTH: Authentication required.,
    NOAUTH: Authentication required.,
    NOAUTH: Authentication required.,
    NOAUTH: Authentication required.,
    NOAUTH: Authentication required.,
    NOAUTH: Authentication required.,
]', shotover-proxy/tests/redis_int_tests/basic_driver_tests.rs:1305:62

```

As a bonus this PR also fixes a bug in create_initial_connections where only a single connection was returned.
If this was a load-bearing bug im happy to rebreak it though.


doyshinda:
Looks like you forgot to refactor the `check_connection` return type in `src/r2d2.rs`:
```
error[E0308]: mismatched types
  --> src/r2d2.rs:16:20
   |
16 |                 if conn.check_connection() {
   |                    ^^^^^^^^^^^^^^^^^^^^^^^ expected `bool`, found enum `Result`
```


>As a bonus this PR also fixes a bug in create_initial_connections where only a single connection was returned.
If this was a load-bearing bug im happy to rebreak it though.

Definitely not a load-bearing bug, since the actual connections are created in `refresh_slots()` (based on the output of the "CLUSTER SLOTS" command sent to the single connection).

doyshinda (approved):
LGTM!

src/cluster.rs
```diff
@@ -171,34 +171,25 @@ impl ClusterConnection {
     }
 
     /// Check that all connections it has are available (`PING` internally).
-    pub fn check_connection(&mut self) -> bool {
+    pub fn check_connection(&mut self) -> RedisResult<()> {
```
Marwes, line 174: This changes the public API so I would have to make a breaking release which seems overkill. Could you make the `Result` returning method private and use that internally and keep this public version with `bool`? Then I can make a semver compatible release

Marwes (changes requested):

src/cluster.rs
rukai: This is not possible because I would also have to revert the Connection trait changes as that is public.
That would be a major problem because everything goes through that trait and I cant add a private method to a trait.

nihohit:

PR #562, opened 2021-11-25
Fix clippy warnings

rukai:
I'm pretty sure the black_box's I removed had no effect due to the inner function returning `()`.
I'm happy to instead put them back and manually ignore the clippy lint if removing them is problematic.


rukai:
The failure is an unrelated intermittent failure. It also occurred on master. 
image

djc:
Fixed this up in #609.

PR #564, opened 2021-11-30
Fix intermittent ci failure

rukai:
The failure can currently be seen on the last master branch build and on https://github.com/mitsuhiko/redis-rs/pull/562
image

I cant see any solution other than retrying or bumping the sleep up.
The unsub is triggered by the connection being dropped but only when redis notices the lost connection does the unsubscription occur. There is an inherent race condition that cannot be avoided here.

I opted to add a retry with timeout of 5s instead of just bumping the sleep to avoid increasing the runtime of the test on faster machines.

Happy to just bump the sleep up if a simpler solution is preferred.


liangyuanpeng (approved):
LGTM

PR #567, opened 2021-12-16
Auto-implement ConnectionLike for Deref + DerefMut

holmesmr:
This provides an implementation of `ConnectionLike` for any type that can be mutably and immutably dereferenced to an implementation of `ConnectionLike` through `Deref` and `DerefMut`. The primary motivation is being able to use `r2d2::PoolConnection`s directly to satisfy `ConnectionLike`, which makes code using pooled connections easier
to read. A blanket impl for `r2d2::PoolConnection` would suffice for this case, but this is more generalised as not to be r2d2 specific, which would also make it usable for other storage types providing `Deref + DerefMut`.


holmesmr:
Apologies - revised this PR as I remembered immediately after opening that `DerefMut` implies `Deref` already 😅

djc (approved):
LGTM, thanks!

PR #570, opened 2022-01-02
Add new list commands

JamesPatrickGill:
This PR aims to add the following new list commands:

From v6.2.0:

1. [BLMOVE](https://redis.io/commands/blmove)
2. [LMOVE](https://redis.io/commands/lmove)

From v7.0.0:

1. [BLMPOP](https://redis.io/commands/blmpop)
2. [LMPOP](https://redis.io/commands/lmpop)

As this is my first contribution please let me know if anything needs amending about the PR. Looking for guidance in some areas as I am unfamiliar with the code but will comment on the code in these areas.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -545,6 +545,44 @@ implement_commands! {
     }
 
     // list operations
+    
+    /// Pop an element from the front of a list, push it to the front of another list
+    /// and return it; or block until one is available
+    fn blmove_left_to_left<K: ToRedisArgs>(srckey: K, dstkey: K, timeout: usize) {
+        cmd("BLMOVE").arg(srckey).arg(dstkey).arg("LEFT").arg("LEFT").arg(timeout)
+    }
+
+    /// Pop an element from the front of a list, push it to the end of another list
+    /// and return it; or block until one is available
+    fn blmove_left_to_right<K: ToRedisArgs>(srckey: K, dstkey: K, timeout: usize) {
+        cmd("BLMOVE").arg(srckey).arg(dstkey).arg("LEFT").arg("RIGHT").arg(timeout)
+    }
+
+    /// Pop an element from the end of a list, push it to the front of another list
+    /// and return it; or block until one is available
+    fn blmove_right_to_left<K: ToRedisArgs>(srckey: K, dstkey: K, timeout: usize) {
+        cmd("BLMOVE").arg(srckey).arg(dstkey).arg("RIGHT").arg("LEFT").arg(timeout)
+    }
+
+    /// Pop an element from the end of a list, push it to the end of another list
+    /// and return it; or block until one is available
+    fn blmove_right_to_right<K: ToRedisArgs>(srckey: K, dstkey: K, timeout: usize) {
+        cmd("BLMOVE").arg(srckey).arg(dstkey).arg("RIGHT").arg("RIGHT").arg(timeout)
+    }
+
+    /// Pops `count` elements from the start of the first non-empty list key from the list of 
+    /// provided key names - defaults to 1 element if count not defined. Or blocks
+    /// until one is available.
+    fn blmpop_left<K: ToRedisArgs>(timeout: usize, numkeys: usize, key: K, count: Option<core::num::NonZeroUsize>) {
+        cmd("BLMPOP").arg(timeout).arg(numkeys).arg(key).arg("LEFT").arg("COUNT").arg(count)
+    }
+
+    /// Pops `count` elements from the end of the first non-empty list key from the list of 
+    /// provided key names - defaults to 1 element if count not defined. Or blocks
+    /// until one is available.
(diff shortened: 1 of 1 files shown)


src/commands.rs
```diff
@@ -545,6 +545,44 @@ implement_commands! {
...
+    fn blmpop_left<K: ToRedisArgs>(timeout: usize, numkeys: usize, key: K, count: Option<core::num::NonZeroUsize>) {
+        cmd("BLMPOP").arg(timeout).arg(numkeys).arg(key).arg("LEFT").arg("COUNT").arg(count)
+    }
+
+    /// Pops `count` elements from the end of the first non-empty list key from the list of 
+    /// provided key names - defaults to 1 element if count not defined. Or blocks
+    /// until one is available.
+    fn blmpop_right<K: ToRedisArgs>(timeout: usize, numkeys: usize, key: K, count: Option<core::num::NonZeroUsize>) {
+        cmd("BLMPOP").arg(timeout).arg(numkeys).arg(key).arg("RIGHT").arg("COUNT").arg(count)
+    }
```
JamesPatrickGill, line 585: Not sure how arg is handling the `Option` here. If it is defaulting it to `1` then this should work but if it is converting it to just no arg then this will need tweaking as the new commands require the COUNT to be passed a number as the next arg.

Marwes: It will default to no argument on `None`. May be better to just pass a `usize`?

src/commands.rs
```diff
@@ -545,6 +545,44 @@ implement_commands! {
     }
 
     // list operations
+    
+    /// Pop an element from the front of a list, push it to the front of another list
+    /// and return it; or block until one is available
+    fn blmove_left_to_left<K: ToRedisArgs>(srckey: K, dstkey: K, timeout: usize) {
```
Marwes, line 551: I'd collapse these four functionns into 

```
enum Direction {
    Left,
    Right,
}
fn blmove(srckey: K, dstkey: K, src_dir: Direction, dst_dir: Direction timeout: usize)
```

JamesPatrickGill: Sounds good, thoughts on using this enum in the `LMPOP` and `BLMPOP` cmds and collapsing them? To keep consistency going forward when cmds use the `LEFT | RIGHT` arg.

```rs
    /// Pops `count` elements from the start of the first non-empty list key from the list of 
    /// provided key names - defaults to 1 element if count not defined
    fn lmpop_left<K: ToRedisArgs>( numkeys: usize, key: K, count: Option<core::num::NonZeroUsize>) {
        cmd("LMPOP").arg(numkeys).arg(key).arg("LEFT").arg("COUNT").arg(count)
    }

    /// Pops `count` elements from the end of the first non-empty list key from the list of 
    /// provided key names - defaults to 1 element if count not defined
    fn lmpop_right<K: ToRedisArgs>( numkeys: usize, key: K, count: Option<core::num::NonZeroUsize>) {
        cmd("LMPOP").arg(numkeys).arg(key).arg("RIGHT").arg("COUNT").arg(count)
    }
```

to 

```rs
    /// Pops `count` elements from the first non-empty list key from the list of 
    /// provided key names
    fn lmpop<K: ToRedisArgs>( numkeys: usize, key: K, dir: Direction, count: Option<core::num::NonZeroUsize>) {
        cmd("LMPOP").arg(numkeys).arg(key).arg(dir).arg("COUNT").arg(count)
    }

```

JamesPatrickGill: I used the `Direction` enum in all places I could for consistency. How does it look

src/commands.rs
```diff
@@ -2140,3 +2163,25 @@ impl ToRedisArgs for LposOptions {
...
+        W: ?Sized + RedisWrite,
+    {
+        match self {
+            Direction::Left => {
+                out.write_arg(b"LEFT")
+            },
+            Direction::Right => {
+                out.write_arg(b"RIGHT")
+            },
+        }
```
Marwes, line 2185:
```suggestion
        let s = match self {
            Direction::Left => b"LEFT",
            Direction::Right => b"RIGHT",
        };
        out.write_arg(s);
```

Marwes (approved):

src/commands.rs
```diff
@@ -2140,3 +2163,22 @@ impl ToRedisArgs for LposOptions {
...
+    Left,
+    Right,
+}
+
+impl ToRedisArgs for Direction {
+    fn write_redis_args<W>(&self, out: &mut W)
+    where
+        W: ?Sized + RedisWrite,
+    {
+        let s = match self {
```
Marwes, line 2178:
```suggestion
        let s: &[u8] = match self {
```
Seems to be some compilation errors.

JamesPatrickGill:
Hey, thanks for merging this :) I've saw that this has failed on merge into master. Not sure on the process so just wondering if there is anything I can do to help move this through to being in a published version?

Marwes:
Released as 0.21.5

PR #578, opened 2022-01-27
[auth] make cluster aware of username

verrchu:
I found it impossible to auth with user:pass combination in cluster mode

therefore I propose this naive fix
it just adds username in all places where password is mentioned in cluster context


verrchu:
@Marwes can I please geat a review on this?

djc:
I think this was fixed in #596, anything left to do here?

PR #580, opened 2022-01-31
Impl `ToRedisArgs` for integers of size 128

Aurora2500:
Implementing `ToRedisArgs` for `i128`, `u128`, `NonZeroI128` and `NonZeroU128` required changing the version of `itoa` from `0.4.3` to `1.0.*`, as the former doesn't implement its `Integer` trait for these types.

Aditionally, the interface of the `itoa` crate has changed between these two versions, making some of its uses in other parts of the code non functiona. This PR also fixes them.


nihohit:
Implemented in another pr

PR #589, opened 2022-03-06
Update Command expiration values to be an appropriate type

### LINKED ISSUE #575: usize not the right type for the expire seconds family of commands
Redis uses an `i64` to represent valid second amounts for expire arguments.

Negative values have a [defined behavior](https://redis.io/commands/expire):
> Note that calling EXPIRE/PEXPIRE with a non-positive timeout or EXPIREAT/PEXPIREAT with a time in the past will result in the key being deleted rather than expired (accordingly, the emitted key event will be del, not expired).

And [`i64::MAX`](https://doc.rust-lang.org/std/primitive.i64.html#) is likely what people want to use when attempting to create volatile keys (for an lru-cache) that will practically never expire.

Also `usize` values could cause errors that could be caught by the compiler (if we used `i64`):

```
127.0.0.1:6379> expire a 9223372036854775807
(integer) 0
127.0.0.1:6379> expire a 9223372036854775808
(error) ERR value is not an integer or out of range
127.0.0.1:6379> 
```

joshleeb:
This change updates expiration values provided to some commands in the `Command` trait to be of type `i64` or `u64` where appropriate, replacing the previously used `usize` type.

More specifically, the commands that are updated are

- SETEX, PSETEX
  - Expiration value changed from `usize` to `u64`
- EXPIRE, PEXPIRE, EXPIREAT, PEXPIREAT
  - Expiration value changed from `usize` to `i64`

For the `*SETEX` commands, there is no mention of supporting negative values. And indeed, providing a negative value through redis-cli returns an error message.

For the `*EXPIRE*` commands, I couldn't find the paragraph in the docs referenced in #575, nor any other mention that the commands accept a negative expiration argument. However, in testing with redis-cli it seems a negative expiration value is valid and in checking the code I found it to confirm these observations. See [here][code1] and [here][code2].

Note that this is a breaking change.

Closes #575 

[code1]: https://github.com/redis/redis/blob/e3ef73dc2a557232c60c732705e8e6ff2050eba9/src/expire.c#L570-L571
[code2]: https://github.com/redis/redis/blob/e3ef73dc2a557232c60c732705e8e6ff2050eba9/src/expire.c#L483-L491

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -93,12 +93,12 @@ implement_commands! {
     }
 
     /// Set the value and expiration of a key.
-    fn set_ex<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V, seconds: usize) {
+    fn set_ex<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V, seconds: u64) {
         cmd("SETEX").arg(key).arg(seconds).arg(value)
     }
 
     /// Set the value and expiration in milliseconds of a key.
-    fn pset_ex<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V, milliseconds: usize) {
+    fn pset_ex<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V, milliseconds: u64) {
         cmd("PSETEX").arg(key).arg(milliseconds).arg(value)
     }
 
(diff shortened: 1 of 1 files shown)


ServiceDog:
[There should also be an option to specify "NX", "XX", etc.](https://redis.io/commands/pexpire/)

Unless there's a way to add optional arguments that I've missed in this repo

Thanks!

djc:
This makes sense to me. Could you rebase on latest main?

joshleeb:
> Could you rebase on latest main?

Done :+1: thanks @djc

jaymell (approved):

jaymell:
Thanks for the PR / your patience. I've approved but want to hold off on merging until we figure out how the next release(s) will be handled. There are some more breaking changes coming soon, so it likely won't be too long.

tarique1988:
@jaymell will the fix to what @ServiceDog mentioned be ready anytime soon? I am experiencing the same issue https://github.com/redis-rs/redis-rs/pull/589#issuecomment-1418476574

jaymell:
@tarique1988 Do you mind opening a separate issue for this? Honestly I don't even remember seeing this b/c it's not really related to the PR.

tarique1988:
@jaymell just created #927 for the same. Thanks a lot. 🙏🏻

jaymell:
Sorry for the long delay merging. Thanks for the PR!

PR #597, opened 2022-04-19
Fix AsyncIter Stream trait implementation

Roger:
The `Stream` trait implementation was dropping the future on each `poll_next`, in some cases blocking forever if polling the Future returns `Poll::Pending`, ie. when more than 10 items are iterated

fixes: https://github.com/mitsuhiko/redis-rs/issues/583 and https://github.com/mitsuhiko/redis-rs/issues/537

It's my first time using Future's polling and streams, so reviews are more than welcome!

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -1,7 +1,8 @@
 #[cfg(feature = "aio")]
 use futures_util::{
+    future::BoxFuture,
     task::{Context, Poll},
-    FutureExt, Stream,
+    Stream, StreamExt,
 };
 #[cfg(feature = "aio")]
 use std::pin::Pin;
@@ -70,30 +71,30 @@ impl<'a, T: FromRedisValue> Iterator for Iter<'a, T> {
 #[cfg(feature = "aio")]
 use crate::aio::ConnectionLike as AsyncConnection;
 
-/// Represents a redis iterator that can be used with async connections.
+/// The inner future of AsyncIter
 #[cfg(feature = "aio")]
-pub struct AsyncIter<'a, T: FromRedisValue + 'a> {
+struct AsyncIterInner<'a, T: FromRedisValue + 'a> {
     batch: std::vec::IntoIter<T>,
     con: &'a mut (dyn AsyncConnection + Send + 'a),
     cmd: Cmd,
 }
 
+/// Represents the state of AsyncIter
 #[cfg(feature = "aio")]
-impl<'a, T: FromRedisValue + 'a> AsyncIter<'a, T> {
-    /// ```rust,no_run
-    /// # use redis::AsyncCommands;
-    /// # async fn scan_set() -> redis::RedisResult<()> {
-    /// # let client = redis::Client::open("redis://127.0.0.1/")?;
-    /// # let mut con = client.get_async_connection().await?;
-    /// con.sadd("my_set", 42i32).await?;
-    /// con.sadd("my_set", 43i32).await?;
-    /// let mut iter: redis::AsyncIter<i32> = con.sscan("my_set").await?;
-    /// while let Some(element) = iter.next_item().await {
-    ///     assert!(element == 42 || element == 43);
-    /// }
-    /// # Ok(())
-    /// # }
-    /// ```
+enum IterOrFuture<'a, T: FromRedisValue + 'a> {
+    Iter(AsyncIterInner<'a, T>),
+    Future(BoxFuture<'a, (AsyncIterInner<'a, T>, Option<T>)>),
+    Empty,
+}
+
+/// Represents a redis iterator that can be used with async connections.
+#[cfg(feature = "aio")]
+pub struct AsyncIter<'a, T: FromRedisValue + 'a> {
+    inner: IterOrFuture<'a, T>,
+}
+
+#[cfg(feature = "aio")]
+impl<'a, T: FromRedisValue + 'a> AsyncIterInner<'a, T> {
     #[inline]
     pub async fn next_item(&mut self) -> Option<T> {
         // we need to do this in a loop until we produce at least one item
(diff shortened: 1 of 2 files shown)


src/cmd.rs
```diff
@@ -70,30 +71,30 @@ impl<'a, T: FromRedisValue> Iterator for Iter<'a, T> {
...
-    /// let mut iter: redis::AsyncIter<i32> = con.sscan("my_set").await?;
-    /// while let Some(element) = iter.next_item().await {
-    ///     assert!(element == 42 || element == 43);
-    /// }
-    /// # Ok(())
-    /// # }
-    /// ```
+enum IterOrFuture<'a, T: FromRedisValue + 'a> {
+    Iter(AsyncIterInner<'a, T>),
+    Future(BoxFuture<'a, (AsyncIterInner<'a, T>, Option<T>)>),
```
thalesfragoso, line 86: It might be better to use the future directly (instead of BoxFuture) together with e.g. `pin-project-lite`.

src/cmd.rs
```diff
@@ -125,13 +126,60 @@ impl<'a, T: FromRedisValue + 'a> AsyncIter<'a, T> {
...
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[inline]
+    pub async fn next_item(&mut self) -> Option<T> {
+        let inner = if let IterOrFuture::Iter(ref mut inner) = self.inner {
+            inner
+        } else {
+            panic!("next_item shouldn't be called in the middle of poll_next");
```
thalesfragoso, line 149: This doesn't seem like a good idea. Futures (and streams) have no guarantees w.r.t being polled to completion, and are expected to work regardless (at least most of the time).
If this method must stay, then we could probably implement it in terms of `StreamExt::next` and (possibly) look into deprecating it.

thalesfragoso (commented):
Thank you for taking the time to tackle this. I left some comments and suggestions.

src/cmd.rs
Roger: Oh, that's right! I can just use `StreamExt::next` on self here.. I don't know how I missed that, thanks!. About the deprecation, I don't know who should make that decision, but imho makes sense to just expose the api as an Stream than a custom async iterator

src/cmd.rs
Roger: I'll take a look at pin-project, since I never used it before, what is the advantage of that, not having a heap allocation?

thalesfragoso: Here it would be mostly to avoid an allocation and the dynamic dispatch.

However, I just realized that this future is coming from an async block/fn, so I don't think it's possible to avoid this without TAIT (or a bigger code change), better keep it this way then, sorry for the noise.

Roger: Don't worry, I learn about TAIT thanks to your comment :)

src/cmd.rs
```diff
@@ -70,30 +71,30 @@ impl<'a, T: FromRedisValue> Iterator for Iter<'a, T> {
...
-    /// con.sadd("my_set", 42i32).await?;
-    /// con.sadd("my_set", 43i32).await?;
-    /// let mut iter: redis::AsyncIter<i32> = con.sscan("my_set").await?;
-    /// while let Some(element) = iter.next_item().await {
-    ///     assert!(element == 42 || element == 43);
-    /// }
-    /// # Ok(())
-    /// # }
-    /// ```
+enum IterOrFuture<'a, T: FromRedisValue + 'a> {
```
djc, line 84: I feel like this introduces quite a bit of complexity that doesn't seem necessary from first principles. It seems to me that we should be able to write an `AsyncIter` (I guess with a hand-written `Future` impl) that doesn't need the intermediate `AsyncIterInner` or `IterOrFuture` stuff, but mainly has the `AsyncConnection` as internal state.

Is there some reason for not going that route that I'm missing? Is there a good reason to batch stuff that I'm not seeing? (Other than, what the code currently looks like.)

Roger: I don't know why this was implemented using batching, I guess that is following the pattern of sync iterator.

From what I understanding, `ConnectionLike::req_packed_command` returns in batch already.

Then, with a hand crafted future, wouldn't be the same issue? since we need to keep the future while we poll it, until it returns `Poll::ready`, but maybe I'm missing something here since it's my first time using future's api.

Roger: Yes, that's what I mean, shouldn't be a problem internally, but since it's a public trait, then will break compatibility with other crates/binaries that implement `ConnectionLike`, right? and I don't know if there's any policy on when to break compat in redis-rs, since still doesn't hit 1.0 release, it means that we can't be explicit in the version for the breakage

jaymell: RE batching: I'm probably missing something in the discussion here, but is there any way to avoid batching in the implementation, at least without something significantly more complicated? The current (albeit broken) impl is simply following the procedure for using these commands, e.g., `SCAN`, which returns results in batches and requires you to specify the next "cursor" value in order to get the next batch. 

I'm not sure how we would intelligently pull in the batches from the server and stream them out except by doing something approximately like what's been implemented in this PR.

Sorry for being extremely late to the party here, but as the problem has come up again of late, I think we should consider reviving these changes and getting them in. I'm also not clear why there are discussions of making breaking changes or how bypassing `req_packed_command` would accomplish anything.

PhotonQuantum: I also couldn't understand why bypassing `req_packed_command` was necessary. Since the server returns results in batches, and we want an interface that yields one result at a time, we need some kind of flattening, the `Iter` variant (the buffer) is definitely required.

Or there's another solution where we stream batches of results (Item=Vec<T>), and ask the user to flatten the stream manually. This approach may eliminate the intermediate state completely, but is apparently a breaking change.

Personally I would vote up the current implementation of this PR.

jaymell:
@Roger Can we rebase this PR and resolve conflicts?

Roger:
@jaymell I went ahead with the rebase and fixed the new linting issue, also squashed everything into a single commit

jaymell:
It looks like there is a pending fix for the issue that occurred on the nightly build: https://github.com/rust-lang/rust/pull/107688

jaymell:
@Roger Can we also add an additional test based on the repro [here](https://github.com/redis-rs/redis-rs/issues/759#issue-1557436158) (or something equivalent)?

Roger:
@jaymell sure, I'll take a look tomorrow

Roger:
@jaymell let me know if anything else needs to be changed

jaymell (approved):

jaymell:
Merged. Thanks!

PR #603, opened 2022-05-05
Add Script::load function

zhiburt:
Hi there

This PR exposes a load function to not do it manually.
It also changes a repeat logic in sync version of `Script::invoke` to limit the repeat cycle to 2 attempts (The same way as is done in async version).

Take care.

### DIFF
--- a/src/script.rs
+++ b/src/script.rs
@@ -4,6 +4,7 @@ use sha1_smol::Sha1;
 use crate::cmd::cmd;
 use crate::connection::ConnectionLike;
 use crate::types::{ErrorKind, FromRedisValue, RedisResult, ToRedisArgs};
+use crate::Cmd;
 
 /// Represents a lua script.
 #[derive(Debug, Clone)]
@@ -124,23 +125,12 @@ impl<'a> ScriptInvocation<'a> {
     /// Invokes the script and returns the result.
     #[inline]
     pub fn invoke<T: FromRedisValue>(&self, con: &mut dyn ConnectionLike) -> RedisResult<T> {
-        let mut eval_cmd = cmd("EVALSHA");
-        eval_cmd
-            .arg(self.script.hash.as_bytes())
-            .arg(self.keys.len())
-            .arg(&*self.keys)
-            .arg(&*self.args);
-
+        let eval_cmd = self.eval_cmd();
         match eval_cmd.query(con) {
             Ok(val) => Ok(val),
             Err(err) => {
                 if err.kind() == ErrorKind::NoScriptError {
-                    let mut load_cmd = cmd("SCRIPT");
-                    load_cmd
-                        .arg("LOAD")
-                        .arg(self.script.code.as_bytes())
-                        .query(con)?;
-
+                    self.load_cmd().query(con)?;
                     eval_cmd.query(con)
                 } else {
                     Err(err)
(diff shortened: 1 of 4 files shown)


zhiburt:
Sorry.....
one more time, please?

*It must be it

src/script.rs
```diff
@@ -124,26 +124,26 @@ impl<'a> ScriptInvocation<'a> {
...
-                Err(err) => {
-                    if err.kind() == ErrorKind::NoScriptError {
-                        cmd("SCRIPT")
-                            .arg("LOAD")
-                            .arg(self.script.code.as_bytes())
-                            .query(con)?;
-                    } else {
-                        fail!(err);
-                    }
+        let mut eval_cmd = cmd("EVALSHA");
```
djc: Why do you think we no longer need the `loop`?

src/script.rs
```diff
@@ -204,13 +182,25 @@ impl<'a> ScriptInvocation<'a> {
...
-            .query_async(con).await?;
+        let hash: String = self.load_cmd().query_async(con).await?;
 
         debug_assert_eq!(hash, self.script.hash);
 
         Ok(hash)
     }
+
+    fn load_cmd(&self) -> Cmd {
+        let mut cmd = cmd("SCRIPT");
```
djc, line 193: Style nit: I think we could just chain this directly all the way through instead of introducing a variable binding?

src/script.rs
```diff
@@ -204,13 +182,25 @@ impl<'a> ScriptInvocation<'a> {
...
+        cmd.arg("LOAD").arg(self.script.code.as_bytes());
+        cmd
+    }
+
+    fn eval_cmd(&self) -> Cmd {
+        let mut cmd = cmd("EVALSHA");
```
djc, line 199: Same here.

tests/test_async_async_std.rs
```diff
@@ -295,14 +295,11 @@ fn test_script_load() {
     let script = redis::Script::new("return 'Hello World'");
 
     block_on_all(async move {
-        let mut con = ctx.multiplexed_async_connection_async_std().await?;
-
-        let hash = script.prepare_invoke().load_async(&mut con);
-        assert_eq!(hash, Ok(script.get_hash().to_string()));
+        let mut con = ctx.multiplexed_async_connection_async_std().await.unwrap();
```
djc, line 298: Why did you change these to `unwrap()`?

djc (commented):
Thanks, this mostly looks good!

src/script.rs
zhiburt: As you may see that a sync and async implementation is different.
I think that the async implementation with 1 retry is better. So I replace a loop implementation with it.

The loop may cause a number of retries or even "potentially" an infinite loop.
See when there's no script `LOAD`ed we load it, but then if someone `FLUSH`es redis `EVAL` will be an error an we will try to `LOAD` the script one more time.

Also it's beneficial because the unification of the code.

tests/test_async_async_std.rs
zhiburt: To fix the build :)

Do you think it matters?

src/script.rs
djc: Fair enough.

djc (approved):

PR #605, opened 2022-05-08
Add support for zmpop

gkorland:


src/commands.rs
```diff
@@ -813,6 +813,18 @@ implement_commands! {
         cmd("ZPOPMIN").arg(key).arg(count)
     }
 
+    /// Removes and returns up to count members with the highest scores, 
+    /// from the first non-empty sorted set in the provided list of key names.
+    fn zmpopmax<K: ToRedisArgs>(keys: &'a [K], count: isize) {
```
jaymell, line 818: Naming convention: `zmpop_max`

src/commands.rs
```diff
@@ -813,6 +813,18 @@ implement_commands! {
...
+        cmd("ZMPOP").arg(keys.len()).arg(keys).arg("MAX").arg(count)
+    }
+
+    /// Removes and returns up to count members with the lowest scores, 
+    /// from the first non-empty sorted set in the provided list of key names.
+    fn zmpopmin<K: ToRedisArgs>(keys: &'a [K], count: isize) {
```
jaymell, line 824: Naming convention: `zmpop_min`

jaymell: Are you sure this works as expected? I don't see `COUNT` specified

src/commands.rs
jaymell: Are you sure this works as expected? I don't see `COUNT` specified

jaymell (changes requested):
Hi -- thanks for the PR! Fixing `clippy` warnings is fine, but can we put them in a separate PR for organization's sake?

src/commands.rs
jaymell: Sorry, perhaps I'm misinterpreting here, but shouldn't the command at a minimum look like the following:

```cmd("ZMPOP").arg(keys.len()).arg(keys).arg("MAX").arg("COUNT").arg(count)```

I don't believe `count` can be specified without a preceding `COUNT` arg

Comment (djc) [reply src/commands.rs:818]: Oh, I missed the part about the explicit `COUNT`.

**Comment by gkorland** [comment]
@djc @jaymell I'm sorry for the broken PR, I guess this is what happens when you don't write tests.
I did review the the tests and I don't see any tests for the APIs.

e.g. looking at the test for `SET` operations it seems like the test doesn't really test the API, but only test the raw `cmd` API.

```rust
fn test_set_ops() {
    let ctx = TestContext::new();

    redis::cmd("SADD").arg("foo").arg(1).execute(&mut con);
    redis::cmd("SADD").arg("foo").arg(2).execute(&mut con);
    redis::cmd("SADD").arg("foo").arg(3).execute(&mut con);

    let mut s: Vec<i32> = redis::cmd("SMEMBERS").arg("foo").query(&mut con).unwrap();
    s.sort_unstable();
    assert_eq!(s.len(), 3);
    assert_eq!(&s, &[1, 2, 3]);

    let set: HashSet<i32> = redis::cmd("SMEMBERS").arg("foo").query(&mut con).unwrap();
    assert_eq!(set.len(), 3);
    assert!(set.contains(&1i32));
    assert!(set.contains(&2i32));
    assert!(set.contains(&3i32));

    let set: BTreeSet<i32> = redis::cmd("SMEMBERS").arg("foo").query(&mut con).unwrap();
    assert_eq!(set.len(), 3);
    assert!(set.contains(&1i32));
    assert!(set.contains(&2i32));
    assert!(set.contains(&3i32));
}
```

djc:
Would you mind rebasing this on main so the clippy fixes go away?

jaymell (approved):

Issue #585, opened 2022-02-11
[question]: Low level commands with RediSearch


Pepeye:
Hi guys, I am test driving this crate with Redis modules and wondering if anyone been able to get this library to work with RediSearch. I'm using the low level `redis::cmd("JSON.GET").arg(...)` commands and trying to run the following search.

Example JSON docs users:

```json
{
    "id": "ssexl7Vh3M6sX7gUQa-dd",
    "name": {
        "first": "John",
        "last": "Appleseed"
    },
    "email": "john.appleseed@email.com",
    "age": 43,
    "phones": [ "+44 1234567", "+44 2345678"],
    "ts": 1644514838189
}
```

I created an index using the following command
```sh
# create index
FT.CREATE idx:users ON JSON SCHEMA $.name.first AS first TEXT $.email AS email  TAG

# search the index using either of queries below in cli - returns a list of users
FT.SEARCH idx:users "@first:John"
FT.SEARCH idx:users "John"

#  1) (integer) 2
#  2) "user:ghiKLASKkoRjQyt-W741e"
#  3) 1) "$"
#     2) "{\"id\":\"ghiKLASKkoRjQyt-W741e\",\"name\":{\"first\":\"John\",\"last\":\"Appleseed\"},\"email\":\"john.appleseed@email.com\",\"age\":43,\"phones\":[\"+44 1234567\",\"+44 2345678\"],\"ts\":1644450185241}"
#  4) "user:yNZ3ofiH-AypMpOkKU8ub"
#  5) 1) "$"
#     2) "{\"id\":\"yNZ3ofiH-AypMpOkKU8ub\",\"name\":{\"first\":\"John\",\"last\":\"Appleseed\"},\"email\":\"john.appleseed@email.com\",\"age\":43,\"phones\":[\"+44 1234567\",\"+44 2345678\"],\"ts\":1644450234167}"
```

My rust test code is below.

```rust
use redis::AsyncCommands;
use serde::{Serialize, Deserialize};
use derive_redis_json::RedisJsonValue;
use chrono::Utc;

#[derive(RedisJsonValue)]
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
pub struct User {
    pub id: String,
    pub name: Name,
    pub email: String,
    pub age: i64,
    pub phones: Vec<String>,
    pub ts: i64,
}

#[derive(RedisJson)]
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
pub struct Name {
[...]

PR #608, opened 2022-05-16
Return errors from parsing inner items

### LINKED ISSUE #587: FromRedisValue for Vec silently ignores errors
Hello, first sorry if this has been already discussed, but I couldn't find anything about it.
I was surprised by the default behavior of FromRedisValue for Vec. Loading data from redis resulted in less than expected number of items. For example:

```
        let res = redis::Value::Bulk(vec![redis::Value::Int(12), redis::Value::Nil]); //obtained from redis, nil value is unexpected here
        let something: Vec<u32> = FromRedisValue::from_redis_value(&res).unwrap();
        assert_eq(something.len(), 2);
 
```
Looking at the library code I see the following:

```
    fn from_redis_values(items: &[Value]) -> RedisResult<Vec<Self>> {
        Ok(items
            .iter()
            .filter_map(|item| FromRedisValue::from_redis_value(item).ok())
            .collect())
    }

```
which effectively skips any invalid items, but I think that the more appropriate behavior here is to fail because that will be an indicator for an error in the program logic. If this is by design, I think at least a print warning should be added, because having a FromRedisValue for struct with several fields of Vec<T> with FromRedisValue for each T where T is actually also failry large struct makes hidden errors like this quite hard to spot and debug.

djc:
Thanks to @doyshinda for pointing this out in #585.

Fixes #587.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1014,10 +1014,10 @@ pub trait FromRedisValue: Sized {
     /// from another vector of values.  This primarily exists internally
     /// to customize the behavior for vectors of tuples.
     fn from_redis_values(items: &[Value]) -> RedisResult<Vec<Self>> {
-        Ok(items
+        items
             .iter()
-            .filter_map(|item| FromRedisValue::from_redis_value(item).ok())
-            .collect())
+            .map(FromRedisValue::from_redis_value)
+            .collect()
     }
 
     /// This only exists internally as a workaround for the lack of
--- a/tests/test_basic.rs
+++ b/tests/test_basic.rs
@@ -235,12 +235,12 @@ fn test_filtered_scanning() {
         }
     }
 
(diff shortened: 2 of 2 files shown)


valkum:
The tests currently fail because in `test_filtered_scanning` `HSCAN` actually returns a `(String, usize)`, the field and the value.

Changing 
```rust
let iter = con.hscan_match("foo", "key_0_*").unwrap();

for x in iter {
        // type inference limitations
        let x: usize = x;
        unseen.remove(&x);
    }
```
to 

```rust
let iter = con
        .hscan_match::<&str, &str, (String, usize)>("foo", "key_0_*")
        .unwrap();

    for (_, x) in iter {
        println!("should be next");
        unseen.remove(&x);
    }```

fixes the test.

djc:
(CI is not running due to the recent branch rename, see #612.)

Issue #599, opened 2022-04-26
Object Idletime


NathanielCurnick:
Are there any plans to add a high level command for the Redis command [OBJECT IDLETIME](https://redis.io/commands/object-idletime/)? 

Would be very helpful especially when working with threads 

Thanks for considering!

PR #610, opened 2022-05-16
add object commands

Roger:
ENCODING, IDLETIME, FREQ, REFCOUNT
closes: #599


djc (commented):
Looks okay, though it would surely be nice to add some tests, too.

djc (approved):

djc (approved):

PR #611, opened 2022-05-17
Fix return type of iter in test_filtered_scanning

valkum:
https://github.com/redis-rs/redis-rs/pull/608 changes the `iter` behavior makes these wrong assumed types now causing tests to fail.

The correct return type for HSCAN is (String, usize) as the field name is returned along the value of that field.


djc:
Awesome, thanks for fixing this!

PR #614, opened 2022-05-23
Refactor tests using explicit APIs

gkorland:


djc (approved):
Definitely seems like an improvement to me!

PR #617, opened 2022-05-24
use 2021 edition of Rust

tdyas:
Use the 2021 edition of Rust.


tdyas:
I've seen this CI failure even without this change:

```
error: unused import: `Deref`
 --> src/connection.rs:4:16
  |
4 | use std::ops::{Deref, DerefMut};
  |                ^^^^^
  |
  = note: `-D unused-imports` implied by `-D warnings`
```

https://github.com/redis-rs/redis-rs/runs/6566383193?check_suite_focus=true#step:6:212

djc:
I see that we don't yet check for an MSRV in CI, but unless there's a compelling reason to upgrade the edition I think it would be better to hold off a little while longer to keep the MSRV lower.

PR #618, opened 2022-05-24
update github urls to the new url

tdyas:
Update the GitHub URLs in various places to take into account the new `redis-rs` organization.


tdyas:
Should the URLs in `CHANGELOG.md` also be updated?

djc:
> Should the URLs in `CHANGELOG.md` also be updated?

Would be nice I think?

djc (approved):
Thanks for taking care of this!

PR #619, opened 2022-05-24
remove unused import of `Deref` since already covered by `DerefMut`

tdyas:
Remove the unnecessary import of `std::ops::Deref` which was triggering a clippy lint because access to the `.deref` method is already covered by the import of `std::ops::DerefMut`.


djc (approved):

djc:
Thanks!

PR #620, opened 2022-05-25
Miscellaneous Cargo improvements

djc:


Cargo.toml
```diff
@@ -1,7 +1,6 @@
 [package]
 name = "redis"
 version = "0.21.5"
-authors = ["Armin Ronacher <armin.ronacher@active-4.com>", "Jan-Erik Rediger <janerik@fnordig.de>"]
```
tasn, line 4: Maybe put:
```toml
authors = "Redis-rsAuthors"
```
I think they is useful for both copyright purposes and general metadata. I'm sure there are things that expect this.

djc: I'm not convinced -- I've removed this field from most of the crates that I help maintain and it's been fine. I don't see any value in terms of general metadata from such a generic description.

jaymell (approved):

Issue #615, opened 2022-05-23
design question: how to pass packed argument arrays to `redis::aio::ConnectionLike`


tdyas:
Question: I would like to pass pre-packed Redis argument slices to `redis::aio::ConnectionLike`. What is the preferred way to do so?

Background: The synchronous trait `redis::connection::ConnectionLike` can take pre-packed argument slices because `ConnectionLike::req_packed_command` and `ConnectionLike::req_packed_commands` take a `&[u8]`. (`ConnectionLike::req_command` takes a `Cmd` and converts to packed form and then calls `req_packed_command`.) The async trait `redis::aio::ConnectionLike` cannot though because the methods `req_packed_command` and `req_packed_commands` on `redis::aio::ConnectionLike` take `&Cmd` and `&Pipeline` respectively, and not a `&[u8]`. (The methods are also actually poorly named because they don't actually take packed arguments.)

Context:  For my company, I wrote an async client that drives N connections to Redis by pre-packing each Redis request into a `Vec<u8>` and then sends it via a MPMC channel to a set of worker tasks that each manage a single connection to Redis. (The pre-packing is necessary to send the request over the MPMC channel given Rust's ownership semantics.) 

With the synchronous trait, this is straightforward to do because each worker can just invoke `req_packed_command` with the `&[u8]` of the packed form of the command. It is not possible to do this currently with `redis::aio::ConnectionLike`, however, because `req_packed_command` takes a `&Cmd`.

Thus, my company's async client needs an upstream change to this crate in order for its worker threads to still use existing connection structs that implement `redis::aio::ConnectionLike` .

Solutions:

1. My initial version of this change was to modify `Cmd` and `Pipeline` to add constructors that take  pre-packed data. [This is the relevant commit in my fork of this crate.](https://github.com/toolchainlabs/redis-rs/commit/2306cda3298f2d3a84774539b600925abb3346cd). It adds `with_packed_data` constructors to each type. The goal with that change was to avoid changing the signatures of methods in `redis::aio::ConnectionLike` which would be a breaking change . The downside is that `Cmd` and `Pipeline` would be used for conflicting purposes: carrying pre-packed data and managing the gathering and serialization of the packed data.

[...]

PR #623, opened 2022-06-02
[WIP] update redis::aio::ConnectionLike to accept Bytes

tdyas:
This PR teaches the `redis::aio::ConnectionLike` trait to accept prepacked Redis commands. The synchronous trait `redis::connection::ConnectionLike` already accepts prepacked commands because its `req_packed_command` and `req_packed_commands` methods take a `&[u8]` slice.

The asynchronous trait, however, despite having methods with the same names, takes references `&Cmd` or &Pipeline` respectively, which are not prepacked. The prevents using the async support in this crate for use cases like the ones described in https://github.com/redis-rs/redis-rs/issues/615.

Breaking changes:
- Change the `&Cmd` argment to `req_packed_command` and `&Pipeline` argument to `req_packed_commands` to both take a `Bytes` with a prepacked command.
- Add `req_command` and `req_pipeline` methods to `redis::aio::ConnectionLike` to allow existing users to continue to be able to pass in a `&Cmd` or `&Pipeline`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,9 @@ sha1_smol = { version = "1.0", optional = true }
 
 combine = { version = "4.6", default-features = false, features = ["std"] }
 
+bytes = { version = "1" }
+
 # Only needed for AIO
-bytes = { version = "1", optional = true }
 futures-util = { version = "0.3.15", default-features = false, optional = true }
 pin-project-lite = { version = "0.2", optional = true }
 tokio-util = { version = "0.7", optional = true }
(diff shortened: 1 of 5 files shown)


djc:
I don't love how this PR duplicates a whole bunch of code, and I feel like, despite breaking compatibility, it just proliferates awkward design from the current state of the traits. I'd like to see what you think `ConnectionLike` would look like if you designed it from scratch to for optimal performance (for example, with prepacked commands).

In my mind we probably want a low-level sealed trait that can abstract over `BytesMut` and `Vec<u8>` and maybe other similar things to abstract over basically `reserve()` + `io::Write`, and then maybe wrap that in a `CommandBuilder` type that has a basic method `argument(arg: &[u8])`.

tdyas:
What about using `bytes::Buf` to abstract over `Bytes` and `Vec<u8>`?

djc:
Yeah, seems reasonable.

PR #625, opened 2022-06-02
example how to use redis r2d2 connection pool

nonirosenfeldredis:


djc:
Documenting this as a docstring on the macro probably doesn't make much sense. Did you check how this appears in the rustdoc output?

tasn:
Looks like a duplicate of #388?

djc:
Yup, it certainly looks like a duplicate.

jaymell:
@djc Shall we close this as well?

Issue #492, opened 2021-05-25
Implement pubsub for the cluster connection


jaudiger:
Hi folks,

I was reading the documentation when I saw this sentence: `Note that the cluster support currently does not provide pubsub functionality.`.

Is this limitation still present? And are there any other limitations with cluster usage? I'm writing an application that will connect on a Redis Cluster, I'm just looking if this library could be used. I only need to create a PublishSubscribe Redis Client. Of course, this Client will use SSL to connect to the cluster.

Thanks in advance for your help!
Regards,
Jérémy

extrawurst:
@jaudiger did you find out more about this. I am looking for the same info. I am also wondering how I can use cluster mode + deadpool connection pooling

jaudiger:
No sorry, not really. I change Redis over a Kafka implementation, and I didn't dig more on Redis.

extrawurst:
@Marwes care to share why you closed this with no comment?

namkai:
@Marwes Was this feature added?

Marwes:
No

marziply:
What work is required to get pubsub working in cluster mode? Happy to contribute if I'm pointed in the right direction. I personally require cluster mode with pubsub so this is acting as a blocker for what I'm working on at the moment.

nihohit:
https://github.com/redis-rs/redis-rs/pull/898

nihohit:
https://github.com/redis-rs/redis-rs/pull/1449

nihohit:
an async cluster pubsub is included as part of the 0.28.0 release.

Issue #487, opened 2021-05-10
Async pubsub hang


kowu-zf:
Using an async subscriber can lead to an infinite loop, if a pubsub message is received before the subscribe/unsubscribe response.
Luckily is already being addressed by https://github.com/mitsuhiko/redis-rs/pull/486

reproducer attached nevertheless, run ./bug stress and ./bug hang in parallel, hang might hang.
```rust
use anyhow::Result;
use redis::Commands;
use tokio_stream::StreamExt;

#[tokio::main(flavor = "current_thread")]
async fn main() -> Result<()> {
    let arg = std::env::args().nth(1);

    match arg.as_deref() {
        Some("hang") => hang().await,
        Some("stress") => stress(),
        _ => panic!("usage: ./bug [hang|stress]"),
    }
}

async fn hang() -> Result<()> {
    let conn = redis::Client::open("redis://127.0.0.1")?
        .get_async_connection()
        .await?;
    let mut pubsub = conn.into_pubsub();

    pubsub.subscribe("foo").await?;
    pubsub.subscribe("fooo").await?;

    let mut onmsg = pubsub.into_on_message();

    println!("{:?}", onmsg.next().await);

    Ok(())
}

fn stress() -> Result<()> {
    let mut conn = redis::Client::open("redis://127.0.0.1")?.get_connection()?;

    loop {
        conn.publish("foo", "bar")?;
    }
}
```

kowu-zf:
This Issue still occurrs in the 0.20.2 release.

Marwes:
Yes, because https://github.com/mitsuhiko/redis-rs/pull/486 is not merged as it lacks a test and needs to be rebased

nihohit:
If you're using Redis/Valkey v6 or newer, you can use the RESP3 pubsub in order to use a regular connection with pubsub - https://docs.rs/redis/latest/redis/#resp3-async-pubsub
If not, I hope that in the next version the async pubsub connection could be split, which will allow this behavior.
https://github.com/redis-rs/redis-rs/pull/1144

nihohit:
fixed in https://github.com/redis-rs/redis-rs/pull/1144

Issue #310, opened 2020-04-22
Pubsub loses channel messages when dynamic subscribing channels in the same pubsub


milabobo:
redis-rs version: 0.15.1
Environment: MacOS
Cargo version: 1.42

Hey, guys:
        I found pubsub might lose channel messages when subscribing another channel in handing incoming messages, for startup, i subscribe channel "foo", then i publish some messages to the channel, after receiving some messages from the channel, i try to subscribe another channel within the message handing loop, then i found some messages lost. I have written a test code below:

```Rust
use std::thread;
use std::time::Duration;
use redis::{self, cmd};

fn main() -> Result<(), Box<dyn std::error::Error>>  {
    let client = redis::Client::open(format!("redis://127.0.0.1:6379"))?;
    let clone_client = client.clone();
    // receive message from foo channel
    thread::spawn(move || {
        let mut conn = client.get_connection().unwrap();
        let mut pubsub = conn.as_pubsub();
        pubsub.subscribe("foo").unwrap();
        loop {
            let msg = pubsub.get_message().unwrap();
            let channel = msg.get_channel_name();
            let content: i32 = msg.get_payload().unwrap();
            println!("receive channel {}, message {}", channel, content);
            if content == 5 {
                // subscribe bar channel using the same pubsub
                println!("sub bar channel");
                pubsub.subscribe("bar").unwrap();
            }
        }
    });
    // send message to foo channel
    thread::sleep(Duration::from_secs(1));
    let mut clone_conn = clone_client.get_connection()?;
    let mut m = 10;
    while m > 0 {
        m -= 1;
        cmd("PUBLISH").arg("foo").arg(m).query::<i32>(&mut clone_conn)?;
    }
    cmd("PUBLISH").arg("bar").arg(0).query::<i32>(&mut clone_conn)?;
    thread::sleep(Duration::from_secs(20));
    Ok(())
}
```

Messages printed

> receive channel foo, message 9
> receive channel foo, message 8
> receive channel foo, message 7
> receive channel foo, message 6
> receive channel foo, message 5
> sub bar channel
> receive channel foo, message 3
> receive channel foo, message 2
> receive channel foo, message 1
> receive channel foo, message 0
> receive channel bar, message 0

Message **4** lost.

Terkwood:
Nicely done!  I think a good first step for whoever looks into this PR would be to encapsulate your test case within the existing test suite, so that we can get a convenient green light when we finally figure out how to solve the problem.

[...]

Issue #303, opened 2020-03-28
pubsub support for MultiplexedConnection?


GopherJ:

nihohit:
What's the need here?
MultiplexedConnection's architecture doesn't allow for simple Pub/Sub, since it's strictly send/respond. Implementing this won't be simpler, so we'll only attempt it if there's a specific need for it.

esemeniuc:
Would love to see this, I don't enjoy having to maintain reconnection logic in my loops

nihohit:
https://github.com/redis-rs/redis-rs/pull/898

nihohit:
pubsub support has been added with RESP3 support.

PR #633, opened 2022-06-12
Split stream async pubsub, and impl Stream for PubSub

enzious:
## Bugfix/Feature
 - Issue: https://github.com/mitsuhiko/redis-rs/issues/310
 - Previous fix: https://github.com/redis-rs/redis-rs/pull/486
 - Fix
   - Use Stream implementations and buffer responses for combined PubSub
 - Feature
   - Users can send commands (mainly subscribe) while reading pubsub messages
   - Allow for splitting to a Sink & Stream

## Example of Issue
```
< pubsub message comes in
> subscribe request
< subscribe function reads pubsub message instead of subscribe response
[subscribe response is left hanging]
```

### DIFF
--- a/redis/examples/async-pub-sub.rs
+++ b/redis/examples/async-pub-sub.rs
@@ -1,4 +1,4 @@
-use futures_util::StreamExt as _;
+use futures_util::StreamExt;
 use redis::AsyncCommands;
 
 #[tokio::main]
@@ -8,12 +8,20 @@ async fn main() -> redis::RedisResult<()> {
     let mut pubsub_conn = client.get_async_connection().await?.into_pubsub();
 
     pubsub_conn.subscribe("wavephone").await?;
-    let mut pubsub_stream = pubsub_conn.on_message();
-
     publish_conn.publish("wavephone", "banana").await?;
 
-    let pubsub_msg: String = pubsub_stream.next().await.unwrap().get_payload()?;
+    let pubsub_msg: String = pubsub_conn.next().await.unwrap().get_payload()?;
     assert_eq!(&pubsub_msg, "banana");
+
+    let (mut sink, mut stream) = pubsub_conn.split();
+
+    let pubsub_msg = tokio::spawn(async move { stream.next().await });
+
+    sink.subscribe("wavephone2").await?;
+    publish_conn.publish("wavephone2", "apple").await?;
+
+    let pubsub_payload: String = pubsub_msg.await.unwrap().unwrap().get_payload::<String>()?;
+    assert_eq!(&pubsub_payload, "apple");
 
     Ok(())
 }
--- a/redis/src/aio/connection.rs
+++ b/redis/src/aio/connection.rs
@@ -3,26 +3,28 @@ use super::async_std;
 use super::ConnectionLike;
 use super::{setup_connection, AsyncStream, RedisRuntime};
 use crate::cmd::{cmd, Cmd};
-use crate::connection::{ConnectionAddr, ConnectionInfo, Msg, RedisConnectionInfo};
+use crate::connection::{ConnectionAddr, ConnectionInfo, RedisConnectionInfo};
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use crate::parser::ValueCodec;
-use crate::types::{ErrorKind, FromRedisValue, RedisError, RedisFuture, RedisResult, Value};
-use crate::{from_redis_value, ToRedisArgs};
+use crate::types::{ErrorKind, RedisError, RedisFuture, RedisResult, Value};
+use crate::{from_redis_value, FromRedisValue, Msg, ToRedisArgs};
 #[cfg(all(not(feature = "tokio-comp"), feature = "async-std-comp"))]
 use ::async_std::net::ToSocketAddrs;
 use ::tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
 #[cfg(feature = "tokio-comp")]
 use ::tokio::net::lookup_host;
 use combine::{parser::combinator::AnySendSyncPartialState, stream::PointerOffset};
-use futures_util::future::select_ok;
-use futures_util::{
-    future::FutureExt,
-    stream::{Stream, StreamExt},
-};
+use futures::channel::mpsc::{self, channel};
+use futures_util::future::{select_ok, FutureExt};
+use futures_util::stream::{SplitSink, SplitStream, Stream, StreamExt};
+use futures_util::SinkExt;
+use pin_project_lite::pin_project;
+use std::collections::VecDeque;
 use std::net::SocketAddr;
 use std::pin::Pin;
+use std::task::Poll;
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
-use tokio_util::codec::Decoder;
+use tokio_util::codec::{Decoder, Framed};
 
 /// Represents a stateful redis TCP connection.
 pub struct Connection<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
@@ -256,76 +258,271 @@ where
     }
 }
 
-/// Represents a `PubSub` connection.
-pub struct PubSub<C = Pin<Box<dyn AsyncStream + Send + Sync>>>(Connection<C>);
+pin_project! {
+    /// Represents a pubsub connection.
+    #[project = PubSubProj]
+    pub struct PubSub<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
+        #[pin]
+        framed: Framed<C, ValueCodec>,
+        queue: VecDeque<Msg>,
+    }
+}
 
-/// Represents a `Monitor` connection.
-pub struct Monitor<C = Pin<Box<dyn AsyncStream + Send + Sync>>>(Connection<C>);
+fn stream_dropped_error() -> RedisError {
+    RedisError::from((
+        ErrorKind::ResponseDropped,
+        "Stream was dropped, response cannot be received",
+    ))
+}
 
 impl<C> PubSub<C>
 where
     C: Unpin + AsyncRead + AsyncWrite + Send,
 {
-    fn new(con: Connection<C>) -> Self {
-        Self(con)
+    fn new(con: Connection<C>) -> PubSub<C> {
+        let framed = ValueCodec::default().framed(con.con);
+
+        PubSub {
+            framed,
+            queue: Default::default(),
+        }
+    }
+
+    /// Splits the [`PubSub`] into a [`SplitPubSubSink`] and [`SplitPubSubStream`].
+    pub fn split(self) -> (SplitPubSubSink<C>, SplitPubSubStream<C>) {
+        SplitPubSubSink::new(self.framed, self.queue)
+    }
+
+    /// Query the connection with a command and retrieves the response.
+    pub async fn query<'a, T>(&'a mut self, cmd: &'a Cmd) -> RedisResult<T>
+    where
+        T: FromRedisValue,
+    {
+        let mut out = Vec::new();
+
+        cmd.write_packed_command(&mut out);
+        self.framed.send(out).await?;
+
+        self.next_response().await
+    }
+
+    async fn next_response<T>(&mut self) -> RedisResult<T>
+    where
+        T: FromRedisValue,
+    {
+        while let Some(item) = self.framed.next().await {
+            match item {
+                Ok(Ok(value)) => match Msg::from_value(&value) {
+                    Some(msg) => self.queue.push_front(msg),
+                    _ => return from_redis_value(&value),
+                },
+                _ => break,
+            }
+        }
+
+        Err(stream_dropped_error())
     }
 
     /// Subscribes to a new channel.
     pub async fn subscribe<T: ToRedisArgs>(&mut self, channel: T) -> RedisResult<()> {
-        cmd("SUBSCRIBE").arg(channel).query_async(&mut self.0).await
+        self.query(cmd("SUBSCRIBE").arg(channel)).await
     }
 
     /// Subscribes to a new channel with a pattern.
     pub async fn psubscribe<T: ToRedisArgs>(&mut self, pchannel: T) -> RedisResult<()> {
-        cmd("PSUBSCRIBE")
-            .arg(pchannel)
-            .query_async(&mut self.0)
-            .await
+        self.query(cmd("PSUBSCRIBE").arg(pchannel)).await
     }
 
     /// Unsubscribes from a channel.
     pub async fn unsubscribe<T: ToRedisArgs>(&mut self, channel: T) -> RedisResult<()> {
-        cmd("UNSUBSCRIBE")
-            .arg(channel)
-            .query_async(&mut self.0)
-            .await
+        self.query(cmd("UNSUBSCRIBE").arg(channel)).await
     }
 
     /// Unsubscribes from a channel with a pattern.
     pub async fn punsubscribe<T: ToRedisArgs>(&mut self, pchannel: T) -> RedisResult<()> {
-        cmd("PUNSUBSCRIBE")
-            .arg(pchannel)
-            .query_async(&mut self.0)
+        self.query(cmd("PUNSUBSCRIBE").arg(pchannel)).await
+    }
+}
+
+impl Stream for PubSub {
+    type Item = Msg;
+
+    fn poll_next(
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        let PubSubProj { mut framed, queue } = self.project();
+
+        if let Some(msg) = queue.pop_back() {
+            return Poll::Ready(Some(msg));
+        }
+
+        loop {
+            match framed.as_mut().poll_next(cx) {
+                Poll::Ready(Some(Ok(Ok(value)))) => {
+                    if let Some(msg) = Msg::from_value(&value) {
+                        return Poll::Ready(Some(msg));
+                    }
+                }
+                Poll::Ready(Some(Err(_)) | Some(Ok(Err(_)))) => {}
+                Poll::Ready(None) => return Poll::Ready(None),
+                Poll::Pending => return Poll::Pending,
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.framed.size_hint()
+    }
+}
+
+/// Represents the transmitting side of a pubsub connection.
+/// The corresponding [`SplitPubSubStream`] must be polled to receive responses.
+pub struct SplitPubSubSink<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
+    sink: SplitSink<Framed<C, ValueCodec>, Vec<u8>>,
+    resp_rx: mpsc::Receiver<Value>,
+}
+
+impl<C> SplitPubSubSink<C>
+where
+    C: Unpin + AsyncRead + AsyncWrite + Send,
+{
+    // Create a [`SplitPubSubSink`] from a [`Connection`].
+    fn new(
+        framed: Framed<C, ValueCodec>,
+        queue: VecDeque<Msg>,
+    ) -> (SplitPubSubSink<C>, SplitPubSubStream<C>) {
+        let (sink, stream) = framed.split();
+        let (resp_tx, resp_rx) = channel::<Value>(1);
+
+        (
+            Self { sink, resp_rx },
+            SplitPubSubStream::new(stream, resp_tx, queue),
+        )
+    }
+
+    // Deliver a command to this pubsub connection.
+    async fn query<'a, T>(&'a mut self, cmd: &'a Cmd) -> RedisResult<T>
+    where
+        T: FromRedisValue,
+    {
+        let mut out = Vec::new();
+
+        cmd.write_packed_command(&mut out);
+        self.sink.send(out).await?;
+
+        self.resp_rx
+            .next()
             .await
+            .ok_or(stream_dropped_error())
+            .and_then(|value| from_redis_value(&value))
     }
 
-    /// Returns [`Stream`] of [`Msg`]s from this [`PubSub`]s subscriptions.
-    ///
-    /// The message itself is still generic and can be converted into an appropriate type through
-    /// the helper methods on it.
-    pub fn on_message(&mut self) -> impl Stream<Item = Msg> + '_ {
-        ValueCodec::default()
-            .framed(&mut self.0.con)
-            .filter_map(|msg| Box::pin(async move { Msg::from_value(&msg.ok()?.ok()?) }))
+    /// Subscribes to a new channel.
+    pub async fn subscribe<T: ToRedisArgs>(&mut self, channel: T) -> RedisResult<()> {
+        self.query(cmd("SUBSCRIBE").arg(channel)).await
     }
 
-    /// Returns [`Stream`] of [`Msg`]s from this [`PubSub`]s subscriptions consuming it.
-    ///
-    /// The message itself is still generic and can be converted into an appropriate type through
-    /// the helper methods on it.
-    /// This can be useful in cases where the stream needs to be returned or held by something other
-    /// than the [`PubSub`].
-    pub fn into_on_message(self) -> impl Stream<Item = Msg> {
-        ValueCodec::default()
-            .framed(self.0.con)
-            .filter_map(|msg| Box::pin(async move { Msg::from_value(&msg.ok()?.ok()?) }))
+    /// Subscribes to a new channel with a pattern.
+    pub async fn psubscribe<T: ToRedisArgs>(&mut self, pchannel: T) -> RedisResult<()> {
+        self.query(cmd("PSUBSCRIBE").arg(pchannel)).await
     }
 
-    /// Exits from `PubSub` mode and converts [`PubSub`] into [`Connection`].
-    pub async fn into_connection(mut self) -> Connection<C> {
-        self.0.exit_pubsub().await.ok();
+    /// Unsubscribes from a channel.
+    pub async fn unsubscribe<T: ToRedisArgs>(&mut self, channel: T) -> RedisResult<()> {
+        self.query(cmd("UNSUBSCRIBE").arg(channel)).await
+    }
 
-        self.0
+    /// Unsubscribes from a channel with a pattern.
+    pub async fn punsubscribe<T: ToRedisArgs>(&mut self, pchannel: T) -> RedisResult<()> {
+        self.query(cmd("PUNSUBSCRIBE").arg(pchannel)).await
+    }
+}
+
+pin_project! {
+    /// Represents the receiving side of a pubsub connection.
+    #[project = SplitPubSubStreamProj]
+    pub struct SplitPubSubStream<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
+        #[pin]
+        stream: SplitStream<Framed<C, ValueCodec>>,
+        resp_tx: mpsc::Sender<Value>,
+        // A response in the process of be forwarded to the sink.
+        resp_waiting: Option<Value>,
+        // Handles messages left in the PubSub struct queue
+        queue: VecDeque<Msg>,
+    }
+}
+
+/// Represents a `Monitor` connection.
+pub struct Monitor<C = Pin<Box<dyn AsyncStream + Send + Sync>>>(Connection<C>);
+
+impl<C> SplitPubSubStream<C>
+where
+    C: Unpin + AsyncRead + Send,
+{
+    fn new(
+        stream: SplitStream<Framed<C, ValueCodec>>,
+        resp_tx: mpsc::Sender<Value>,
+        queue: VecDeque<Msg>,
+    ) -> Self {
+        SplitPubSubStream {
+            stream,
+            resp_tx,
+            resp_waiting: None,
+            queue,
+        }
+    }
+}
+
+impl Stream for SplitPubSubStream {
+    type Item = Msg;
+
+    fn poll_next(
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        let SplitPubSubStreamProj {
+            mut stream,
+            resp_tx,
+            resp_waiting,
+            queue,
+        } = self.project();
+
+        if let Some(msg) = queue.pop_back() {
+            return Poll::Ready(Some(msg));
+        }
+
+        loop {
+            if resp_waiting.is_some() {
+                match resp_tx.poll_ready(cx) {
+                    Poll::Ready(Ok(_)) => {
+                        if let Some(resp) = resp_waiting.take() {
+                            drop(resp_tx.start_send(resp));
+                        }
+                    }
+                    Poll::Ready(Err(_)) => drop(resp_waiting.take()),
+                    Poll::Pending => return Poll::Pending,
+                };
+            }
+
+            match stream.as_mut().poll_next(cx) {
+                Poll::Ready(Some(Ok(Ok(value)))) => match Msg::from_value(&value) {
+                    Some(msg) => return Poll::Ready(Some(msg)),
+                    _ => {
+                        // This should only be reachabled if the resp_waiting conditional above is
+                        // false.
+                        *resp_waiting = Some(value)
+                    }
+                },
+                Poll::Ready(Some(Err(_)) | Some(Ok(Err(_)))) => {}
+                Poll::Pending => return Poll::Pending,
+                Poll::Ready(None) => return Poll::Ready(None),
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.stream.size_hint()
     }
 }
 
--- a/redis/tests/test_async.rs
+++ b/redis/tests/test_async.rs
@@ -503,6 +503,8 @@ mod pub_sub {
     use std::collections::HashMap;
     use std::time::Duration;
 
+    use tokio::spawn;
+
     use super::*;
 
     #[test]
@@ -513,12 +515,67 @@ mod pub_sub {
         block_on_all(async move {
             let mut pubsub_conn = ctx.async_connection().await?.into_pubsub();
             pubsub_conn.subscribe("phonewave").await?;
-            let mut pubsub_stream = pubsub_conn.on_message();
+
             let mut publish_conn = ctx.async_connection().await?;
             publish_conn.publish("phonewave", "banana").await?;
 
-            let msg_payload: String = pubsub_stream.next().await.unwrap().get_payload()?;
+            let msg_payload: String = pubsub_conn.next().await.unwrap().get_payload()?;
+
             assert_eq!("banana".to_string(), msg_payload);
+
+            Ok::<_, RedisError>(())
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn split_pub_sub_subscription() {
+        use redis::RedisError;
+
+        let ctx = TestContext::new();
+        block_on_all(async move {
+            let (mut sink, mut stream) = ctx.async_connection().await?.into_pubsub().split();
+
+            let msg = spawn(async move { stream.next().await });
+
+            sink.subscribe("phonewave").await?;
+
+            let mut pubsub_conn = ctx.async_connection().await?;
+
+            pubsub_conn.publish("phonewave", "banana").await?;
+
+            assert_eq!(
+                "banana".to_string(),
+                msg.await.unwrap().unwrap().get_payload::<String>()?
+            );
+
+            Ok::<_, RedisError>(())
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn split_pub_sub_subscription_with_drop() {
+        use redis::RedisError;
+
+        let ctx = TestContext::new();
+        block_on_all(async move {
+            let (mut sink, mut stream) = ctx.async_connection().await?.into_pubsub().split();
+
+            let msg = spawn(async move { stream.next().await });
+
+            sink.subscribe("phonewave").await?;
+
+            drop(sink);
+
+            let mut pubsub_conn = ctx.async_connection().await?;
+
+            pubsub_conn.publish("phonewave", "banana").await?;
+
+            assert_eq!(
+                "banana".to_string(),
+                msg.await.unwrap().unwrap().get_payload::<String>()?
+            );
 
             Ok::<_, RedisError>(())
         })
@@ -536,6 +593,35 @@ mod pub_sub {
             let mut pubsub_conn = ctx.async_connection().await?.into_pubsub();
             pubsub_conn.subscribe(SUBSCRIPTION_KEY).await?;
             pubsub_conn.unsubscribe(SUBSCRIPTION_KEY).await?;
+
+            let mut conn = ctx.async_connection().await?;
+            let subscriptions_counts: HashMap<String, u32> = redis::cmd("PUBSUB")
+                .arg("NUMSUB")
+                .arg(SUBSCRIPTION_KEY)
+                .query_async(&mut conn)
+                .await?;
+            let subscription_count = *subscriptions_counts.get(SUBSCRIPTION_KEY).unwrap();
+            assert_eq!(subscription_count, 0);
+
+            Ok::<_, RedisError>(())
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn split_pub_sub_unsubscription() {
+        use redis::RedisError;
+
+        const SUBSCRIPTION_KEY: &str = "phonewave-pub-sub-unsubscription";
+
+        let ctx = TestContext::new();
+        block_on_all(async move {
+            let (mut sink, mut stream) = ctx.async_connection().await?.into_pubsub().split();
+
+            spawn(async move { while stream.next().await.is_some() {} });
+
+            sink.subscribe(SUBSCRIPTION_KEY).await?;
+            sink.unsubscribe(SUBSCRIPTION_KEY).await?;
 
             let mut conn = ctx.async_connection().await?;
             let subscriptions_counts: HashMap<String, u32> = redis::cmd("PUBSUB")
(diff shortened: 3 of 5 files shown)


enzious:
@djc I am no longer using the previous fix as I've found this refactor more ergonomic.

enzious:
Funny enough, sending commands and reading messages simultaneously was enough for this guy to write another redis library: https://old.reddit.com/r/rust/comments/ux9ry5/redust_a_new_redis_client/

enzious:
Need to fix lint and check tests locally

enzious:
Should pass tests

jaymell:
Hi: I wonder if we could add a test that asserts that the original reported problem in #310 is fixed?

dwfait:
Hi @jaymell. 

Not the author here, but interested in the feature this PR would unlock. I'm happy to try to help to get this merged. 

Looking at the issue linked, it's an issue based on a race condition and would be pretty hard to replicate reliably in a test suite, if I'm understanding correctly. Do you have a suggestion on a way to test this or what that test might look like that could bypass the nondeterministic nature of it?

nihohit:
@dwfait hi, I believe that since the issue happens in the sync pubsub, testing for it should be rather easy. I'm not sure whether this PR actually fixes the issue, though - it doesn't contain the test suggested in the issue. Have you tried testing whether that test fails or succeeds consistently
I don't know if the issue also happens in the async pubsub.

nihohit:
[Edit] - I verified, and the issue doesn't happen in async pubsub, so I don't see the relation between this PR and the bug.

dwfait:
@nihohit, are you saying with async pub sub, there is currently a way to dynamically subscribe to channels whilst listening without this PR?

dwfait:
@nihohit Sure. Have created a new PR here: https://github.com/redis-rs/redis-rs/pull/914, using this one as a reference.

enzious:
Note that my tests got stuck originally. I just checked this. Unfortunate for me I guess.

enzious:
@dwfait @nihohit Actually, I'm going to admit that I'm a bit peeved with #914 

I spent a week coming up with a solution to this issue, maybe more debugging it, over a year waiting for anyone to show any interest.

And before I even have time to check my GitHub notifications I find that my work and PR have been totally commandeered.

jaymell:
> And before I even have time to check my GitHub notifications I find that my work and PR have been totally commandeered.

Just a note that the last comment on this PR for over a year was a straightforward question about tests that went unanswered.

enzious:
I reached out to another contributor somewhere about the tests and they told me they were broken.

I'm seeing failures on the CI (not lint) 

image


> > And before I even have time to check my GitHub notifications I find that my work and PR have been totally commandeered.
> 
> Just a note that the last comment on this PR for over a year was a straightforward question about tests that went unanswered.

This is true, but I had a test failing for one version of Rust in an area unrelated to the changes I had. The person I was able to reach out to said that things were slow and they didn't know why my tests were failing.

I see a lot of PRs are failing for some builds of Rust but not for others. Could you explain this for me? Is this expected?

nihohit:
> I see a lot of PRs are failing for some builds of Rust but not for others. Could you explain this for me, is this expected?

A new feature was merged, and some of its tests were flakey. This was partially fixed here - https://github.com/redis-rs/redis-rs/pull/912 so if you rebase over recent main, you should be ok.

@enzious @dwfait I would appreciate it if you could discuss you PRs and decide on a single change (of course, I'd be happy to help). A couple of things I'd appreciate if you to consider - 
1. the main issue I have with the suggested change is that it drops anything which isn't a pub/sub message. Would the user still receive disconnection errors? Also, https://github.com/redis-rs/redis-rs/issues/639 - the user should be able to send commands  and receive their response on the connection. I don't know if this solution will work in async, but consider the fix here: https://github.com/redis-rs/redis-rs/pull/910
2. https://github.com/redis-rs/redis-rs/pull/889 - I hope to deprecate aio::Connection and allow direct creation of pubsub / monitor creation. This means that soundness on `exit_pubsub` shouldn't be a consideration, since that function will be deprecated too.
3.  There are multiple requests to support pubsub on MultiplexedConnection (for example, https://github.com/redis-rs/redis-rs/issues/303 ). By the same token, https://github.com/redis-rs/redis-rs/issues/492 - the existing pubsub doesn't represent all required pubsub capabilities. You don't need to implement any of this in this PR, but I'd appreciate it if the pubsub types were generic / type erased on the internal connection, so that additional connection support was easy to implement. This isn't a hard requirement, just a request.

and @enzious , I do apologize for not pinging you in my initial response.

enzious:
I'll put some thought on 1 and 3.

dwfait:
@enzious apologies, my intention was not to try to commandeer or take credit for any work. 

This is a feature I am in need of for a project I'm currently working on, I saw this pull request and that it was last updated over a year ago, and just asked if I could help move anything forward. 

I'll admit I'm not versed in the etiquette of open source contributions. I'm happy with whatever solution or implementation comes from here, and I'm happy to help try and move things forward in any way if I can help. I'm just looking for a working solution, in the meantime I'm using a different crate for redis pubsub. 

I've closed my pull request with a link back to this one (there was a link to this in the description as well). 

Please let me know if I can help in any way, if not, I'll leave this for you.

nihohit:
also relevant: https://github.com/redis-rs/redis-rs/issues/487

enzious:
@dwfait It's fine. I know that there may have been a possibility that I may have disappeared, etc. I was more upset that I missed the initial discussion and opportunity to learn more through this PR, and was just tired with work enough to not word that as nicely as I should have. I'm sorry as well.

dwfait:
@enzious no worries! Didn't take anything personally.

I've noticed there's a test failure for the latest run in `test_pubsub`, where the order of the messages received on the channel are in the wrong order. This appears to be for the synchronous connection.

As far as I know, the order of the messages sent in redis should be guaranteed - perhaps this error is because we are now using packed_commands for the synchronous connection here, and not waiting for the response from redis before sending the next message, meaning they could potentially be received out of order?

Can I ask what the intent is to move the synchronous commands to use packed_commands here? Would we want to preserve the ability to be able to guarantee the order of messages? It seems it would be pretty important for some use cases, and might be a subtle breaking change for some applications.

Maybe we could expose a packed vs non-packed version of these commands in the interface if it's needed?

enzious:
> As far as I know, the order of the messages sent in redis should be guaranteed - perhaps this error is because we are now using packed_commands for the synchronous connection here, and not waiting for the response from redis before sending the next message, meaning they could potentially be received out of order?
You're correct, because it exits pubsub, I believe it could expect that the next message is a response.

I reverted that change.

dwfait:
> On the async side tho, if you can req/res and remain subscribed, I'll have to think about how that should work because you could receive a published message while waiting for a response as I understand it.

On the async side, we are intentionally not blocking waiting to read the response to allow us to split the stream half of the connection away to be owned for listening, so it seems like we can't guarantee this aspect of ordered messages with a split connection. 

Maybe the library should provide a non split connection in async too, if message ordering is important and you don't need dynamic subscription?

There may be a way to provide both functionality in a split connection, but it seems it would involve some communication between the two halves and message processing for the sink to know when the response for the subscription has arrived, right?

dwfait:
How do other libraries that provide this function? Does the go library linked provide a guarantee of message ordering for example?

enzious:
I think I can do it by having a [channel](https://docs.rs/futures-preview/latest/futures/channel/mpsc/fn.channel.html) in the `PubSubSink`. Basically, somehow maybe I could intercept non-pubsub messages and send them to the PubSubSink.

dwfait:
> I think I can do it by having a [channel](https://docs.rs/futures-preview/latest/futures/channel/mpsc/fn.channel.html) in the `PubSubSink`. Basically, somehow maybe I could intercept non-pubsub messages and send them to the PubSubSink.
> 

This seems like it would have quite a performance impact if you don't care about message ordering. It feels like it should be optional to me, what do you think?


> Noted in the header, when I made this PR I checked Go's implementation and they seem to throw away the responses at the time.

Yes I saw, I was just wondering if they just acknowledge that it's not possible with their library to have guaranteed order, or if the order is preserved in some other way I'm not seeing.

enzious:
> This seems like it would have quite a performance impact if you don't care about message ordering. It feels like it should be optional to me, what do you think?

I'm kinda of on the thought that it could cause a small performance issue but it would only be incurred when sending a command.

> Yes I saw, I was just wondering if they just acknowledge that it's not possible with their library to have guaranteed order, or if the order is preserved in some other way I'm not seeing.

I believe it guarantees the order of request/responses. It's just that when pub/sub messages come in they can jump between. So if we just filter out the pubsub messages we are good, right?

nihohit:
You can also consider the solution proposed here:
https://github.com/redis-rs/redis-rs/pull/898
Since PUSH messages are supposed to allow pubsub in RESP3.

enzious:
> You can also consider the solution proposed here: #898 Since PUSH messages are supposed to allow pubsub in RESP3.

In that case it may be work it to block on #898? Seems like they're putting a lot of work into that PR.

nihohit:
that's a possible solution, but notice that that PR is merged into the resp3 branch and not main - even if the PR will be merged, there's still more work before the resp3 branch will be merged into main.
So, essentially it's a question of urgency. If you're fine waiting, then sure, do.

nihohit:
a channel sounds like the right solution to me.

enzious:
A small hitch with the split stream/sink design and req/res is the fact that futures don't run unless polled. So unless the stream is polled, the request will not be able to read the response, and be stuck. Thoughts?

nihohit:
This also raises a lifetime issue, no? If one of the sink or the stream are dropped, is the other still valid?

Some solutions, off the top of my head - 
1. Use a spawned driver (like the one used in MultiplexedConnection) that queries the TcpStream, and send its results either on a stream-side channel or a sink-side channel.
2. Avoid the split design, and instead investigate ways to make the various methods take a shared reference instead of an exclusive one.
3. Allow both the sink and the stream part poll the internal connection, and have them send the "wrong" results on a channel to the other.

enzious:
> This also raises a lifetime issue, no? If one of the sink or the stream are dropped, is the other still valid?

No, but I added an error for this case.

I'm thinking something like implementing a Sink + Stream that split into a Sink & Stream, but I'm gonna have to let that roll around in my head for a while.

shachlanAmazon:
> No, but I added an error for this case.

I can see why the Sink isn't valid without the stream, but IMO the stream should stay valid even if the sink is dropped, unless there's a technical limitation on it.

nihohit:
> No, but I added an error for this case.

I can see why the Sink isn't valid without the stream, but IMO the stream should stay valid even if the sink is dropped, unless there's a technical limitation on it.

enzious:
I've been looking at tokio codec stuff and I'm thinking that it might be easier to start off with this implementation. I don't want to stray too far from the Framed trait. I trust that we can implement it well with those interfaces. Perfect is the enemy of good.

redis/examples/async-pub-sub.rs
```diff
@@ -1,19 +1,21 @@
...
-    pubsub_conn.subscribe("wavephone").await?;
-    let mut pubsub_stream = pubsub_conn.on_message();
+    let mut pubsub = client.get_async_connection().await?.into_pubsub();
 
+    pubsub.subscribe("wavephone").await?;
     publish_conn.publish("wavephone", "banana").await?;
 
-    let pubsub_msg: String = pubsub_stream.next().await.unwrap().get_payload()?;
+    // consume the response to the subscription
+    let _ = pubsub.next_message().await;
```
nihohit, line 13: why is this needed?

redis/src/aio/connection.rs
```diff
@@ -256,108 +267,205 @@ where
...
+
+        self.next_response().await
+    }
+
+    async fn next_response(&mut self) -> RedisResult<()> {
+        while let Some(item) = self.framed.next().await {
+            match item {
+                Ok(Ok(value)) => match Msg::from_value(&value) {
+                    Some(msg) => self.queue.push_front(msg),
+                    _ => return Ok(()),
```
nihohit, line 303: please add a todo to verify this isn't Value::Error after https://github.com/redis-rs/redis-rs/pull/813 will be merged

redis/src/aio/connection.rs
```diff
@@ -256,108 +267,205 @@ where
...
+    }
+
+    /// placeholder
+    pub async fn next_message(&mut self) -> RedisResult<Msg> {
+        match self.queue.pop_back() {
+            Some(msg) => Ok(msg),
+            _ => match self.framed.next().await {
+                Some(Ok(Ok(value))) => match Msg::from_value(&value) {
+                    Some(msg) => Ok(msg),
+                    _ => Err(RedisError::from((
```
nihohit, line 319: but this could also catch a valid sub/unsub response, no?
This doesn't mean that the connection was necessarily dropped.

redis/src/aio/connection.rs
```diff
@@ -81,14 +82,24 @@ where
         Ok(rv)
     }
 
-    /// Converts this [`Connection`] into [`PubSub`].
+    /// Converts this [`Connection`] into [`PubSubSink`] and a [`Stream`] of [`Msg`].
```
nihohit, line 85: please correct the doc, unless I'm missing here something

redis/examples/async-monitor.rs
```diff
@@ -0,0 +1,21 @@
...
+async fn main() -> redis::RedisResult<()> {
+    let client = redis::Client::open("redis://127.0.0.1/").unwrap();
+    let mut publish_conn = client.get_async_connection().await?;
+    let (mut sink, mut stream) = client.get_async_connection().await?.into_monitor();
+
+    sink.monitor().await?;
+
+    publish_conn.set("key", b"value").await?;
+
+    let _ = stream.next().await;
```
nihohit, line 14: why is this needed?

redis/src/aio/connection.rs
```diff
@@ -256,108 +267,205 @@ where
...
+    }
+
+    async fn next_response(&mut self) -> RedisResult<()> {
+        while let Some(item) = self.framed.next().await {
+            match item {
+                Ok(Ok(value)) => match Msg::from_value(&value) {
+                    Some(msg) => self.queue.push_front(msg),
+                    _ => return Ok(()),
+                },
+                _ => {}
```
nihohit, line 305: so you'll never return an error?

redis/src/aio/connection.rs
```diff
@@ -256,108 +267,205 @@ where
...
+            Self { sink, resp_recv },
+            unfold((stream, resp_send), |(mut stream, mut resp_send)| {
+                Box::pin(async move {
+                    let item = stream.next().await;
+
+                    match item {
+                        Some(Ok(Ok(value))) => Some(Ok(match Msg::from_value(&value) {
+                            Some(msg) => Some(msg),
+                            _ => {
+                                let _ = resp_send.send(value).await;
```
nihohit, line 393: why not item? you'll miss errors

redis/src/aio/connection.rs
```diff
@@ -256,108 +267,205 @@ where
...
+
+                                None
```
nihohit, line 395: would this close the stream? if not, how does the stream close?

redis/src/aio/connection.rs
```diff
@@ -256,108 +267,205 @@ where
...
+        let mut out = Vec::new();
+
+        cmd.write_packed_command(&mut out);
+        self.sink.send(out).await?;
+
+        self.resp_recv
+            .next()
             .await
+            .map(|_| ())
+            .ok_or(RedisError::from((
```
nihohit, line 417: you're ignoring here errors that might be relevant - for example, SSUBSCRIBE (which isn't implemented here, but this might be the basis for a cluster pubsub) can return MOVED.

redis/src/aio/connection.rs
```diff
@@ -256,108 +267,205 @@ where
...
     pub async fn punsubscribe<T: ToRedisArgs>(&mut self, pchannel: T) -> RedisResult<()> {
-        cmd("PUNSUBSCRIBE")
-            .arg(pchannel)
-            .query_async(&mut self.0)
+        self.command(cmd("PUNSUBSCRIBE").arg(pchannel)).await
+    }
+}
+
+/// Represents a `PubSub` connection.
+pub struct SplitPubSubSink<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
```
nihohit, line 364: are there any tests for this? specifically, those that check that the stream is viable without the sink?

nihohit (commented):
I think this is a good step in the right direction.

enzious:
Thanks for the review, that was a bit WIP. I'm writing tests on a wip branch and tweaking it. I'm trying to make in as elegant as possible for what it is lol

enzious:
Still working on this, think I found an impl I like. Hopefully I'll have it done soon

redis/src/aio/connection.rs
enzious: In my opinion, if the sink drops the stream should drop.

nihohit: Can you please explain your logic?
IMO we should be consistent with other Read/Write Halves implementations, such as Unix/TcpStream. AFAIS their write halves can be dropped independently, without closing the read halves.

nihohit:
@enzious Thanks! I'm returning from vacation soon, will try to review this this week.

redis/src/aio/connection.rs
enzious: I'll look at removing the limitation, I was just being consistent with the Sink requiring the Stream to work. Plus, it's an easy way to shutdown the stream which is convenient.

enzious:
> @enzious Thanks! I'm returning from vacation soon, will try to review this this week.

No worries. Enjoy the vacation 👍

redis/src/parser.rs
```diff
@@ -154,13 +154,23 @@ where
                     })
                 };
 
+                let unknown = || {
```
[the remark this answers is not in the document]
enzious: Basically, the last clause does not consume from the connection, so if you don't consume it and throw an error, this parser enters an infinite loop and maxes out the thread until the process is killed.

redis/examples/async-monitor.rs
```diff
@@ -0,0 +1,21 @@
...
+async fn main() -> redis::RedisResult<()> {
+    let client = redis::Client::open("redis://127.0.0.1/").unwrap();
+    let mut publish_conn = client.get_async_connection().await?;
+    let mut monitor = client.get_async_connection().await?.into_monitor();
+
+    monitor.monitor().await?;
+
+    publish_conn.set("key", b"value").await?;
+
+    let _ = monitor.next().await;
```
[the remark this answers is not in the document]
enzious: The response to the monitor command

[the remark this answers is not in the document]
enzious: I can make monitor even more consistent to pubsub with and have the query return the response.

redis/src/parser.rs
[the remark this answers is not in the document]
nihohit: thanks for the clarification. This sounds like an existing bug, which wasn't added by this change, right?
If so, this also requires a test, to ensure that the bug doesn't reappear.

If you want to be extra-nice, can you please open a separate PR for this fix and its test? So we can merge it ASAP.

redis/examples/async-monitor.rs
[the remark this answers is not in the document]
nihohit: So the first message on the monitor stream is the "OK" from the monitor?
Is this also the existing behavior?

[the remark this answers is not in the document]
enzious: Oh, sorry, I got confused. I already made changes to fix this in monitor. I was going back and forth and forgot where I landed with it. This is not needed.

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.stream.size_hint()
+    }
+}
+
+pin_project! {
+    /// Represents a pubsub connection.
+    #[project = MonitorProj]
+    pub struct Monitor<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
```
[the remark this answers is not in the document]
enzious: I think that this implementation is easier to follow when you queue the responses

redis/tests/test_async.rs
```diff
@@ -546,6 +610,8 @@ mod pub_sub {
             let subscription_count = *subscriptions_counts.get(SUBSCRIPTION_KEY).unwrap();
             assert_eq!(subscription_count, 0);
 
+            drop(sink);
```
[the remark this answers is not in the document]
enzious: It's not, removed

redis/src/parser.rs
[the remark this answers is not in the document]
enzious: #942

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+
+pin_project! {
+    /// Represents the receiving side of a pubsub connection.
+    #[project = SplitPubSubStreamProj]
+    pub struct SplitPubSubStream<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
+        #[pin]
+        stream: SplitStream<Framed<C, ValueCodec>>,
+        #[pin]
+        resp_tx: mpsc::Sender<Value>,
+        resp_waiting: Option<Value>,
```
nihohit, line 449: the code isn't clear about the purpose of this field. Can you please document it?

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+    }
+}
+
+pin_project! {
+    /// Represents the receiving side of a pubsub connection.
+    #[project = SplitPubSubStreamProj]
+    pub struct SplitPubSubStream<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
+        #[pin]
+        stream: SplitStream<Framed<C, ValueCodec>>,
+        #[pin]
```
nihohit, line 447: why is the sender pinned?

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+                        }
+                    }
+                    Poll::Pending => return Poll::Pending,
+                };
+            }
+
+            match stream.as_mut().poll_next(cx) {
+                Poll::Ready(Some(Ok(Ok(value)))) => match Msg::from_value(&value) {
+                    Some(msg) => return Poll::Ready(Some(Ok(msg))),
+                    _ => *resp_waiting = Some(value),
```
nihohit, line 508: please comment that `resp_waiting` must be None, because of the previous check

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+        loop {
+            if resp_waiting.is_some() {
+                match resp_tx.as_mut().poll_ready(cx) {
+                    Poll::Ready(Ok(_)) => {
+                        if let Some(resp) = resp_waiting.take() {
+                            map_response_err(resp_tx.as_mut().start_send(resp))?;
+                        }
+                    }
+                    Poll::Ready(Err(err)) => {
+                        if !err.is_disconnected() {
```
nihohit, line 497: shouldn't there be an `else` here? how do you handle disconnects?

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+            resp_waiting: None,
+        }
+    }
+}
+
+fn map_response_err<T>(value: Result<T, SendError>) -> Result<T, RedisError> {
+    value.map_err(|_| {
+        RedisError::from((
+            ErrorKind::ResponseForwardError,
+            "Response failed to forward",
```
nihohit, line 470: what does this error mean to the user? what can the user do with it?

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+    fn new(stream: SplitStream<Framed<C, ValueCodec>>, resp_tx: mpsc::Sender<Value>) -> Self {
+        SplitPubSubStream {
+            stream,
+            resp_tx,
+            resp_waiting: None,
+        }
+    }
+}
+
+fn map_response_err<T>(value: Result<T, SendError>) -> Result<T, RedisError> {
```
nihohit, line 466: the name doesn't show that this function is used only for forwarding errors from stream to sink.

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+            mut resp_tx,
+            resp_waiting,
+        } = self.project();
+
+        loop {
+            if resp_waiting.is_some() {
+                match resp_tx.as_mut().poll_ready(cx) {
+                    Poll::Ready(Ok(_)) => {
+                        if let Some(resp) = resp_waiting.take() {
+                            map_response_err(resp_tx.as_mut().start_send(resp))?;
```
nihohit, line 493: this means that if the sink is dropped before receiving a response, the stream will be stuck?

nihohit (commented):
commented on the Pub/sub implementation.
Consider stashing the Monitor change to another PR, in order to reduce the scope of this change.

redis/src/aio/connection.rs
enzious: As I understood it, if the Sink is dropped you wanted the stream to continue receiving messages.

redis/src/aio/connection.rs
enzious: The channel shouldn't be Ready if it's dropped.

redis/src/aio/connection.rs
```diff
@@ -256,109 +258,350 @@ where
...
+
+        PubSub {
+            framed,
+            queue: Default::default(),
+        }
+    }
+
+    /// Splits the [`PubSub`] into a [`SplitPubSubSink`] and [`SplitPubSubStream`].
+    pub fn split(self) -> (SplitPubSubSink<C>, SplitPubSubStream<C>) {
+        SplitPubSubSink::new(self.framed)
```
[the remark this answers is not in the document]
enzious: I'll have to think about that.

redis/src/aio/connection.rs
nihohit: yes. `err` is a channel error, or a Redis error? looks like a Redis error to me, and a Redis disconnect should be handled.

redis/src/aio/connection.rs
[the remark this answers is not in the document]
nihohit: > Having the channel tx/rx in a single struct seems inelegant.

I agree with you, but IMO so if writing the same logic twice :)
Maybe it's possible to write some shared logic using traits, but without thinking about it very much, it sounds like you'll need to write some code to convince the compiler that can be avoided by just jamming both sink & stream in the pubsub.

redis/src/aio/connection.rs
enzious: This `err` is a SendError for the channel.

redis/src/aio/connection.rs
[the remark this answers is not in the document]
enzious: You still need to have special logic for `PubSub`. `PubSub` needs to store pubsub messages while looking for a response but `SplitPubSubStream` needs to forward the response to a channel. And you can't do a delegation to the methods of the Sink because you need to poll stream to start the forward. You would have to spawn a task that would consume messages from the stream.

IMO I think this is clean enough for now, unless you have any ideas for how to handling the combination.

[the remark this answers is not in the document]
nihohit: PubSub needs to store pubsub messages while looking for a response but SplitPubSubStream needs to forward the response to a channel.

This sounds solvable to me, but I'm fine with continuing with the current design and with broad and robust testing, and then we can try after this PR to change the internal implementation without breaking the external behavior of the pubsub.

enzious:
> commented on the Pub/sub implementation. Consider stashing the Monitor change to another PR, in order to reduce the scope of this change.

Are you sure you want to drop the Monitor change? It's pretty small and handles request responses.

redis/src/aio/connection.rs
enzious: As I see it, it is nearly impossible for the user to hit this error. This is just gracefully handling the Result<...> types involved in handling forwarding the messages. The 2 errors of SendError are for the channel being full and the sink being dropped. Since you have to `.await` on responses, you will hold a reference and only be able to await one at a time.

nihohit: Then please document what this error is, and how it potentially can happen. 
Just add something that will help people handle this error if they ever see it.

nihohit:
> Are you sure you want to drop the Monitor change? It's pretty small and handles request responses.

I prefer it, but not going to insist on it. 
It adds noise to the git diff, and I'd prefer it if after we're done with this PR we'd do the same with the monitor instead of having half of the monitor work in this PR, and leaving the rest to another PR.

redis/src/aio/connection.rs
```diff
@@ -256,76 +258,277 @@ where
...
     pub async fn punsubscribe<T: ToRedisArgs>(&mut self, pchannel: T) -> RedisResult<()> {
-        cmd("PUNSUBSCRIBE")
-            .arg(pchannel)
-            .query_async(&mut self.0)
+        self.query(cmd("PUNSUBSCRIBE").arg(pchannel)).await
+    }
+}
+
+impl Stream for PubSub {
+    type Item = Result<Msg, RedisError>;
```
nihohit, line 348: why is it `RedisResult<Msg>` and not `Msg`?
This is a breaking change -  
https://github.com/redis-rs/redis-rs/blob/5e977e93a29622fdd105b9e8776e771c88e575e3/redis/src/aio/connection.rs#L306

redis/src/aio/connection.rs
```diff
@@ -256,76 +258,277 @@ where
...
+        // not occur in the context of the request/response pattern.
+        RedisError::from((
+            ErrorKind::ResponseForwardError,
+            "Response failed to forward to sink",
+        ))
+    })
+}
+
+impl Stream for SplitPubSubStream {
+    type Item = Result<Msg, RedisError>;
```
nihohit, line 481: same - should be Msg

nihohit (commented):
Sorry, I missed the change in item in previous reviews. 
I think that once that's fixed, we're ready.

redis/src/aio/connection.rs
enzious: Should the stream end on errors? Feels odd to consume errors

nihohit: when does the server send errors over a pubsub connection, and does it terminate the connection?
If the server might send errors that aren't responses to requests from the sink, then yes - pass them on. If not, then there's no need for the result.

redis/src/aio/connection.rs
```diff
@@ -502,42 +489,34 @@ impl Stream for SplitPubSubStream {
...
                         // This should only be reachabled if the resp_waiting conditional above is
                         // false.
                         *resp_waiting = Some(value)
                     }
                 },
-                Poll::Ready(Some(Err(err)) | Some(Ok(Err(err)))) => {
-                    return Poll::Ready(Some(Err(err)));
-                }
-                Poll::Ready(None) => return Poll::Ready(None),
+                Poll::Ready(Some(Err(_)) | Some(Ok(Err(_)))) => {}
```
nihohit, line 517: making sure - will the stream terminate once the pubsub is disconnected from the server? can we add tests that drop the server and see that the stream completes?

redis/src/aio/connection.rs
```diff
@@ -256,76 +258,271 @@ where
...
+        }
+
+        loop {
+            match framed.as_mut().poll_next(cx) {
+                Poll::Ready(Some(Ok(Ok(value)))) => {
+                    if let Some(msg) = Msg::from_value(&value) {
+                        return Poll::Ready(Some(msg));
+                    }
+                }
+                Poll::Ready(Some(Err(_)) | Some(Ok(Err(_)))) => {}
```
nihohit, line 367: making sure - will the stream terminate once the pubsub is disconnected from the server? can we add tests that drop the server and see that the stream completes?

redis/src/aio/connection.rs
```diff
@@ -256,76 +258,271 @@ where
...
+        }
+    }
+
+    /// Splits the [`PubSub`] into a [`SplitPubSubSink`] and [`SplitPubSubStream`].
+    pub fn split(self) -> (SplitPubSubSink<C>, SplitPubSubStream<C>) {
+        SplitPubSubSink::new(self.framed, self.queue)
+    }
+
+    /// Query the connection with a command and retrieves the response.
+    pub async fn query<'a, T>(&'a mut self, cmd: &'a Cmd) -> RedisResult<T>
```
nihohit, line 297: AFAIS this shouldn't be pub - most messages sent will break the pubsub.

nihohit (approved):
@jaymell 
ready for your review, once the additional tests are added.

jaymell:
I think this looks great! 

@nihohit This PR is breaking. How do you want to handle merging this?

@enzious Can you please update/clean up the top-level comment into something a bit more appropriate for the squash/merge commit? :pray:

nihohit:
@jaymell I think we should release a semver breaking version, together with https://github.com/redis-rs/redis-rs/pull/889 .
In the release notes we'll explain the break, its rationale, and give examples on how to change the code (which isn't very complex).
I don't think that some glue code that will allow people to keep using the existing API is constructive, but maybe we can release a beta version, and ask people to use it and comment.

jaymell:
> @jaymell
> 
> > @enzious Can you please update/clean up the top-level comment into something a bit more appropriate for the squash/merge commit? 🙏
> 
> Done 👍

Sorry, where did you do this? I don't see it.

enzious:
image
Is this not what you meant?

jaymell:
> image Is this not what you meant? I made the first commit message more general to the PR

No, sorry, I literally meant the top-level comment on this PR. Just hoping for a nice paragraph from you summarizing this. Or alternately, squash the commits on your end into a single commit and add it there.... if you don't mind.

enzious:
> > image Is this not what you meant? I made the first commit message more general to the PR
> 
> No, sorry, I literally meant the top-level comment on this PR. Just hoping for a nice paragraph from you summarizing this. Or alternately, squash the commits on your end into a single commit and add it there.... if you don't mind.

Alright, got it

nihohit:
@enzious you didn't update the PR's description. Would you like us to take care of that, so that we could merge this?

jaymell:
> @enzious you didn't update the PR's description. Would you like us to take care of that, so that we could merge this?

I believe the commit message was updated in the commit itself, so we can probably just go ahead and use that. Since this one is semver-breaking, I think we should hold off a bit longer and plan on merging all the pending things after this next point release.

enzious:
Alright, I updated the PR.

scullionw:
excited for this!

was testing out this branch but the subscribe command is just blocking forever for me

```rust
#[tokio::main]
async fn main() {
    let pubsub = redis::Client::open("redis://127.0.0.1").unwrap();

    let (mut sink, _stream) = pubsub
        .get_async_connection()
        .await
        .unwrap()
        .into_pubsub()
        .split();

    dbg!("about to subscribe");

    sink.subscribe("channel_a").await.unwrap();

    dbg!("subscribed");
}

```

Cargo.toml
```toml
[package]
name = "pubsub_test"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
redis = { git = "https://github.com/enzious/redis-rs.git", branch = "feature/split-stream-pubsub", features = [
    "tokio-comp",
] }
tokio = { version = "1.34.0", features = ["full"] }
```

Output
Screenshot 2023-11-27 at 1 00 18 PM

nihohit:
@enzious could you please look into this?

jaymell:
I think the issue with @scullionw's example is that the stream is not being actively polled and hence the subscribe command is never actually being sent. I'm not sure if this example could be made to work.

nihohit:
so `stream.next` is what causes `sink.subscribe` to actually complete? This dependency shouldn't be there.

enzious:
> so `stream.next` is what causes `sink.subscribe` to actually complete? This dependency shouldn't be there.

The easy solution is to not split the PubSub. SplitPubSub was in my mind a way to send off the stream somewhere and use the sink to subscribe, publish, etc. The example given would work without splitting, as Stream is implemented for PubSub. I suggested earlier in the PR just removing SplitPubSub from this PR and just impl Stream for PubSub.

jaymell:
It seems like a potentially reasonable use-case to split them, but I'm not sure if the abstractions are set up correctly. Is it necessary to actually use a SplitSink/Stream under the hood? Is it possible for the sink side to handle the underlying connection itself and simply relays `Msg`s to the stream?

scullionw:
> The easy solution is to not split the PubSub. SplitPubSub was in my mind a way to send off the stream somewhere and use the sink to subscribe, publish, etc. The example given would work without splitting, as Stream is implemented for PubSub and I don't see why it needs to be split here. I suggested earlier in the PR just removing SplitPubSub from this PR and just impl Stream for PubSub. The user could just use channels to send off their messages.

are you sure? the reason i came to this branch was specifically because I couldn't read the stream and send subscribe messages because both methods need a mutable reference.

nihohit:
> I think the issue with @scullionw's example is that the stream is not being actively polled and hence the subscribe command is never actually being sent. I'm not sure if this example could be made to work.

AFAIS the PubSubSink owns the sink, and (at least seemingly) sends the request on it. It then waits on the receiver from the PubSubStream, which should've been woken up once by its internal stream , which I assume registered on the `cx` for waking once the message was sent. I'm not sure why it didn't.

jaymell:
I'm not terribly familiar with `tokio::select` or its nuances, but this line from the docs would suggest to me that once the `rx.recv()` has completed, the stream is no longer being polled:

> Waits on multiple concurrent branches, returning when the first branch completes, cancelling the remaining branches. 

Am I misunderstanding here?

nihohit:
well, it's called in a loop, so I assumed that it's re-entered. but maybe there's a cancellation issue here, and we can solve it by keeping the futures internally.

nihohit:
Yes, as I suspected - the issue was that the futures were dropped and recreated. Fixed example, now working.
Thanks [@withoutboats](https://github.com/withoutboats) @withoutboats for your blog post that informed me of this issue - https://without.boats/blog/three-problems-of-pinning/

```rs
use std::time::Duration;

use futures_util::{pin_mut, FutureExt, StreamExt};
use redis::AsyncCommands;

#[tokio::main]
    let client = redis::Client::open("redis://127.0.0.1").unwrap();
    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
    let mut publish_conn = client.get_async_connection().await?;

    let (mut sink, mut stream) = client
        .get_async_connection()
        .await
        .unwrap()
        .into_pubsub()
        .split();

    tokio::spawn(async move {
        tokio::time::sleep(Duration::from_secs(3)).await;
        tx.send(()).unwrap();
        tokio::time::sleep(Duration::from_secs(3)).await;
        let _: () = publish_conn.publish("mychannel", "banana").await.unwrap();
    });

    let subscribe_fut = async {
        rx.recv().await;
        dbg!("about to subscribe");
        sink.subscribe("mychannel").await.unwrap();
        dbg!("subscribed");
    }
    .fuse();
    let receive_fut = async {
        if let Some(message) = stream.next().await {
            dbg!(message);
        }
    }
    .fuse();
    pin_mut!(subscribe_fut, receive_fut);
    loop {
        tokio::select! {
            _ = &mut subscribe_fut => {dbg!("sub done");},
            _ = &mut receive_fut => {dbg!("receive done");},
        }
    }
}
```

enzious:
Here's a reduction of the pattern I'm using:
```rust
{
  actix_rt::spawn(async move {
    if let Err(err) = pubsub.subscribe("/").await {
      panic!("Failed to subscribe to root: {}", &err);
    }

    loop {
      futures::select! {
        // Handle local message queue
        msg = rx.recv().fuse() => {
          let _ = pubsub.subscribe("something").await;
        },

        // Handle messages from pubsub
        msg = pubsub.next().fuse() => {
          let _ = pubsub.unsubscribe("something").await;
        },

        complete => break,
      };
    }
  });
}
```

scullionw:
if the stream needs to constantly be polled, can the library take care of it with a tokio::spawn and then provide a channel receiver as the api?

nihohit:
> if the stream needs to constantly be polled for the sink to work, can the library take care of it with a tokio::spawn and then provide a channel receiver as the api?

IMO this is something simple enough for the user to implement, and that requires from the library side both a decision on the runtime and on the channel type, so it would be better to not implement it.

jaymell:
Couldn't we avoid the need to constantly poll the stream side by moving most of the logic into the sink and having the stream just receive/emit `Msg`s from the sink via a channel? IMO this makes the most sense.

scullionw:
> > if the stream needs to constantly be polled for the sink to work, can the library take care of it with a tokio::spawn and then provide a channel receiver as the api?
> 
> IMO this is something simple enough for the user to implement, and that requires from the library side both a decision on the runtime and on the channel type, so it would be better to not implement it.

it’s not that it is complex, it’s that it is easy to misuse the current api

nihohit:
> Couldn't we avoid the need to constantly poll the stream side by moving most of the logic into the sink and having the stream just receive/emit Msgs from the sink via a channel? IMO this makes the most sense.

But that would mean that if the sink is dropped, the stream will no longer function. Of course, the inverse is true right now - if the stream is dropped, the sink will no longer function, but the sink isn't really useful without the stream, so this makes sense to me.

> it’s not that it is complex, it’s that it is easy to misuse the current api

True, but sadly (as the article shows) that is true of async Rust in general. But let's consider this - we can abstract the runtime away, but can we abstract the channel away? Or will it be reasonable to force the user to use a specific channel, regardless of the user's dependencies?

scullionw:
maybe the channel doesn’t need to be exposed to the user. 

Could just be a type (which manages the channe internally) provided by the redis crate that implements AsyncIterator (basically exposing a stream like before but one that doesn’t need to be polled for the sink to work)

nihohit:
> Could just be a type (which manages the channel internally) provided by the redis crate that implements AsyncIterator 

Good idea, worth trying. IMO I would still not add the request to this PR, but implement it separately, just to get this PR moving.

jaymell:
Per @enzious's previous comment, I'm not opposed to removing the Split Sink/Stream from this PR. Regardless, can we resolve conflicts?

jaymell (approved):

nihohit:
If the preference is to remove pub/sub from this PR, then we should at least add tests that check the fix without a split pub/sub.

@enzious if you're not interested, I might re-introduce the split functionality in a later PR, based on your work. Is that fine with you?

enzious:
> If the preference is to remove pub/sub from this PR, then we should at least add tests that check the fix without a split pub/sub.
> 
> @enzious if you're not interested, I might re-introduce the split functionality in a later PR, based on your work. Is that fine with you?

Yeah, I'll look at it in the next few days, remove the split functionality, and see if there are any tests we need.

nihohit:
@enzious you'll probably want to rebase over main - https://github.com/redis-rs/redis-rs/pull/889 was merged, and it might affect this PR.

nihohit:
@enzious are you still interested in pushing this change?

felipefdl:
That will be awesome to have it merged.

nihohit:
closing in favor of https://github.com/redis-rs/redis-rs/pull/1144

PR #635, opened 2022-06-16
Cluster: Create read_from_replicas option

utkarshgupta137:
Currently, all cluster commands in readonly mode first go to the replica & then in case of errors, they're redirected to the primary.
Instead of the unnecessary replica call & then creating a connection, we can check if a command is readonly or not & then set it to the appropriate node.
This results in a benchmark improvement of ~80% for write commands in readonly mode.
IMO, the option should be renamed from `readonly` to `read_from_replicas`, but that would create a breaking change.

### DIFF
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -156,7 +156,7 @@ impl ClusterConnection {
     /// method.
     pub fn set_write_timeout(&self, dur: Option<Duration>) -> RedisResult<()> {
         // Check if duration is valid before updating local value.
-        if dur.is_some() && dur.unwrap() == Duration::default() {
+        if dur.is_some() && dur.unwrap().is_zero() {
             return Err(RedisError::from((
                 ErrorKind::InvalidClientConfig,
                 "Duration should be None or non-zero.",
@@ -179,7 +179,7 @@ impl ClusterConnection {
     /// method.
     pub fn set_read_timeout(&self, dur: Option<Duration>) -> RedisResult<()> {
         // Check if duration is valid before updating local value.
-        if dur.is_some() && dur.unwrap() == Duration::default() {
+        if dur.is_some() && dur.unwrap().is_zero() {
             return Err(RedisError::from((
                 ErrorKind::InvalidClientConfig,
                 "Duration should be None or non-zero.",
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -61,7 +61,7 @@ use crate::cluster_pipeline::UNROUTABLE_ERROR;
 pub use crate::cluster_pipeline::{cluster_pipe, ClusterPipeline};
 use crate::cluster_routing::{Routable, RoutingInfo, Slot, SLOT_SIZE};
 
-type SlotMap = BTreeMap<u16, String>;
+type SlotMap = BTreeMap<u16, [String; 2]>;
 
 /// This is a connection of Redis cluster.
 pub struct ClusterConnection {
@@ -69,7 +69,7 @@ pub struct ClusterConnection {
     connections: RefCell<HashMap<String, Connection>>,
     slots: RefCell<SlotMap>,
     auto_reconnect: RefCell<bool>,
-    readonly: bool,
+    read_from_replicas: bool,
     username: Option<String>,
     password: Option<String>,
     read_timeout: RefCell<Option<Duration>>,
@@ -96,13 +96,13 @@ impl TlsMode {
 impl ClusterConnection {
     pub(crate) fn new(
         initial_nodes: Vec<ConnectionInfo>,
-        readonly: bool,
+        read_from_replicas: bool,
         username: Option<String>,
         password: Option<String>,
     ) -> RedisResult<ClusterConnection> {
         let connections = Self::create_initial_connections(
             &initial_nodes,
-            readonly,
+            read_from_replicas,
             username.clone(),
             password.clone(),
         )?;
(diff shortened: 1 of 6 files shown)


src/cluster.rs
```diff
@@ -153,12 +154,12 @@ impl ClusterConnection {
     /// block indefinitely. It is an error to pass the zero `Duration` to this
     /// method.
     pub fn set_write_timeout(&self, dur: Option<Duration>) -> RedisResult<()> {
-        let mut t = self.write_timeout.borrow_mut();
-        *t = dur;
         let connections = self.connections.borrow();
         for conn in connections.values() {
             conn.set_write_timeout(dur)?;
         }
+        let mut t = self.write_timeout.borrow_mut();
```
djc, line 161: Why is this ordering important? Maybe add a comment here explaining this?

Cargo.toml
```diff
@@ -47,7 +47,9 @@ r2d2 = { version = "0.8.8", optional = true }
 
 # Only needed for cluster
 crc16 = { version = "0.4", optional = true }
+itertools = { version = "0.10", optional = true }
```
djc, line 50: What are we pulling in itertools for? If it's just `unique()`, I'd prefer some kind of hand-rolled version of that if it isn't too much code (say, less than 100).

djc (commented):
This sounds good, pretty simple!

src/cluster.rs
utkarshgupta137: I've added a check & custom error. I think it is self-explanatory now.

Cargo.toml
utkarshgupta137: Done.

utkarshgupta137:
@djc What do you think about renaming the parameter to `read_from_replicas` (same name as in [redis-py](https://github.com/redis/redis-py/blob/master/redis/asyncio/cluster.py#L153))? We can keep `readonly` as a deprecated option.

djc:
> @djc What do you think about renaming the parameter to `read_from_replicas` (same name as in [redis-py](https://github.com/redis/redis-py/blob/master/redis/asyncio/cluster.py#L153))? We can keep `readonly` as a deprecated option.

Sounds good to me, including the deprecation!

src/cluster.rs
utkarshgupta137: Hey, if you look at the updated code, I've moved it back. In case the user passes a zero duration, we would update the local value first, then attempt to set it for each connection which would then raise the error.
So instead of doing that, I had moved the updating local value part after updating it for each connection.
After your comment, I've moved that part back & instead added a direct check for zero duration before updating & also added an error message.

src/cluster_client.rs
```diff
@@ -50,19 +50,26 @@ impl ClusterClientBuilder {
...
+    ///
+    /// If True, then read queries will go to the replica nodes & write queries will go to the
+    /// primary nodes. If there are no replica nodes, then all queries will go to the primary nodes.
+    pub fn read_from_replicas(mut self, read_from_replicas: bool) -> ClusterClientBuilder {
+        self.read_from_replicas = read_from_replicas;
         self
     }
+
+    /// Use `read_from_replicas`.
+    #[deprecated(since = "0.22.0")]
```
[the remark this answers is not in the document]
utkarshgupta137: The documentation already mentions it, adding it in a note again seems overkill?

src/cluster_client.rs
```diff
@@ -50,19 +50,26 @@ impl ClusterClientBuilder {
...
-    /// Set read only mode for new ClusterClient (default is false).
-    /// If readonly is true, read queries will go to the replica nodes, write queries will go to
-    /// the primary nodes. If there are no replica nodes, all queries will go to the primary nodes.
-    pub fn readonly(mut self, readonly: bool) -> ClusterClientBuilder {
-        self.readonly = readonly;
+    /// Enable read from replicas for new ClusterClient (default is false).
+    ///
+    /// If True, then read queries will go to the replica nodes & write queries will go to the
+    /// primary nodes. If there are no replica nodes, then all queries will go to the primary nodes.
+    pub fn read_from_replicas(mut self, read_from_replicas: bool) -> ClusterClientBuilder {
```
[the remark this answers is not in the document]
utkarshgupta137: As in
```
    pub fn read_from_replicas(mut self) -> ClusterClientBuilder {
        self.read_from_replicas = true;
        self
    }
```
?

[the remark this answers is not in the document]
djc: Yes!

src/cluster_client.rs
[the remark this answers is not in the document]
djc: The note shows up in the compiler output, which makes life easier for the user who encounters this.

djc (approved):
This seems reasonable to me!

utkarshgupta137:
@djc Can you merge this? I will then rebase #640.

src/commands.rs
```diff
@@ -17,6 +17,37 @@ use crate::streams;
 #[cfg(feature = "acl")]
 use crate::acl;
 
+#[cfg(feature = "cluster")]
+pub(crate) fn is_readonly_cmd(cmd: &[u8]) -> bool {
+    matches!(
+        cmd,
+        // @admin
+        b"LASTSAVE" | b"PRIVILEGE" | b"PRIVILIDGE" | b"SETCLIENTADDR" |
```
jaymell, line 25: I'm not familiar with these commands and can't find a reference to them anywhere -- should these be here?
```
PRIVILEGE
PRIVILIDGE
SETCLIENTADDR
```

utkarshgupta137: I can't find the documentation either, but they exist in a redis 4.0 deployment I use.

jaymell: If they are not proper redis commands, I think we should remove them.

utkarshgupta137: Done

Issue #594, opened 2022-03-30
Next release planning


jqnatividad:
Especially with the itoa reaching v1 and dtoa 0.4.8 replaced by ryu v1. 

https://github.com/mitsuhiko/redis-rs/pull/588

djc:
I think this will need to be 0.22, since #576 changed the available features. I have not yet audited the rest of the PRs to see if there are API-incompatible changes, but I wouldn't be surprised if there were more.

djc:
#666 prepares a 0.21.6 release.

djc:
0.21.6 has been released, thanks!

PR #636, opened 2022-06-17
add support for using ahash internally & for redis values

utkarshgupta137:


jqnatividad:
Hi folks,
How do you use the "ahash" feature as its not listed as a feature in the Cargo.toml file?

djc:
Optional dependencies implicitly create a feature toggle.

utkarshgupta137:
@djc I think we need a new release for people to be able to use it.

djc:
Well, you can use it now as a Git dependency, and that seems to be orthogonal.

jqnatividad:
@djc, It seems that if you use an optional dependency that is not explicitly declared a feature, you're prevented from publishing it on crates.io:

```
$ cargo publish --dry-run --allow-dirty
    Updating crates.io index
   Packaging qsv v0.58.2 (/Users/johndoe/Documents/GitHub/qsv)
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for `redis`.
      ... required by package `qsv v0.58.2 (/Users/johndoe/Documents/GitHub/qsv)`
  versions that meet the requirements `^0.21` are: 0.21.5, 0.21.4, 0.21.3, 0.21.2, 0.21.1, 0.21.0

  the package `qsv` depends on `redis`, with features: `ahash` but `redis` does not have these features.
```

utkarshgupta137:
Should I raise a PR to create a feature for this then?

djc:
> @djc, It seems that if you use an optional dependency that is not explicitly declared a feature, you're prevented from publishing it on crates.io:

I think this is wrong. If you want to publish qsv to crates.io it's not allowed to have git dependencies but this PR has not yet been released to crates.io. So what you're seeing here just means that the released versions don't have a feature called `ahash` which is clearly expected. See #594 to track release planning.

PR #644, opened 2022-06-28
Added RedisJSON Support

d3rpp:
This PR adds support for RedisJSON.

I removed the `rustfmt` disable line and it didn't seem to affect the file but when it was on VSCode decided to delete the entire file on saving so it's probably a bug in the tooling. Feel free to add it back I guess.

This edits are hidden behind a new `json` feature so it isn't always added where it isn't needed


Cargo.toml
```diff
@@ -49,21 +49,38 @@ r2d2 = { version = "0.8.8", optional = true }
 crc16 = { version = "0.4", optional = true }
 rand = { version = "0.8", optional = true }
 # Only needed for async_std support
-async-std = { version = "= 1.8.0", optional = true}
+async-std = { version = "= 1.8.0", optional = true }
```
d3rpp, line 52: space added by formatter

Cargo.toml
```diff
@@ -49,21 +49,38 @@ r2d2 = { version = "0.8.8", optional = true }
...
 
 # Only needed for TLS
 native-tls = { version = "0.2", optional = true }
 tokio-native-tls = { version = "0.3", optional = true }
 async-native-tls = { version = "0.4", optional = true }
 
+# Only needed for JSON Support
+serde = { version = "1.0.137", optional = true }
+serde_json = { version = "1.0.81", optional = true }
+
```
d3rpp, line 63: `serde` and `serde_json` added for serialization and deserialization

Cargo.toml
```diff
@@ -49,21 +49,38 @@ r2d2 = { version = "0.8.8", optional = true }
...
+	"futures-util",
+	"futures-util/alloc",
+	"futures-util/sink",
+	"tokio/io-util",
+	"tokio-util",
+	"tokio-util/codec",
+	"tokio/sync",
+	"combine/tokio",
+	"async-trait",
+]
```
d3rpp, line 83: added by formatter, feel free to remove

Cargo.toml
```diff
@@ -49,21 +49,38 @@ r2d2 = { version = "0.8.8", optional = true }
...
+# Only needed for JSON Support
+serde = { version = "1.0.137", optional = true }
+serde_json = { version = "1.0.81", optional = true }
+
 # Optional aHash support
 ahash = { version = "0.7.6", optional = true }
 
 [features]
 default = ["acl", "streams", "geospatial", "script"]
+json = []
```
d3rpp, line 69: added new `json` feature to make sure the new stuff doesn't go where it doesn't need to go

src/commands.rs
```diff
@@ -1,13 +1,19 @@
 // can't use rustfmt here because it screws up the file.
-#![cfg_attr(rustfmt, rustfmt_skip)]
+// #![cfg_attr(rustfmt, rustfmt_skip)]
```
d3rpp, line 2: my tooling didn't like this line, again feel free to put this back

Cargo.toml
```diff
@@ -49,21 +49,38 @@ r2d2 = { version = "0.8.8", optional = true }
...
+# Only needed for JSON Support
+serde = { version = "1.0.137", optional = true }
+serde_json = { version = "1.0.81", optional = true }
+
 # Optional aHash support
 ahash = { version = "0.7.6", optional = true }
 
 [features]
 default = ["acl", "streams", "geospatial", "script"]
+json = ["serde", "serde_json"]
```
d3rpp, line 69: new feature

jaymell:
Hi, thanks for the PR! Could you please remove the unnecessary formatting changes and also add some tests for the commands added?

src/commands.rs
```diff
@@ -372,6 +378,140 @@ implement_commands! {
...
+        cmd("JSON.CLEAR").arg(key).arg(path.to_string())
+    }
+
+    /// Deletes a value at `path`.
+    #[cfg(feature = "json")]
+    fn json_del<K: ToRedisArgs, P: ToString>(key: K, path: P) {
+        cmd("JSON.DEL").arg(key).arg(path.to_string())
+    }
+
+    /// See [json_del](redis::commands::Commands::json_del)
```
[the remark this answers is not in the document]
d3rpp: I suppose I should make it clearer

but it is an alias to JSON.DEL

https://redis.io/commands/json.forget/

src/commands.rs
```diff
@@ -372,6 +378,140 @@ implement_commands! {
...
+    /// Removes and returns an element from the `index` in the array.
+    /// `index` defaults to `-1` (the end of the array)
+    #[cfg(feature = "json")]
+    fn json_arrpop<K: ToRedisArgs, P: ToString>(key: K, path: P, index: Option<isize>) {
+        cmd("JSON.ARRPOP").arg(key).arg(path.to_string()).arg(index.unwrap_or(-1))
+    }
+
+    /// Trims an array so that it contains only the specified inclusive range of elements.
+    ///
+    /// This command is extremely forgiving and using it with out-of-range indexes will not produce an error. There are a few differences between how RedisJSON v2.0 and legacy versions handle out-of-range indexes.
```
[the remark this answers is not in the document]
d3rpp: no worries

d3rpp:
thanks for the review, one of these comments has asked me to add tests so I'll do that tomorrow, and once those are done and hopefully passing ill revert the formatting changes as requested

jaymell:
I wonder if we should consider name-spacing or otherwise separating these commands further from the core redis commands given that they are part of a separate Redis module. I'm not positive offhand but don't believe we currently have any other module-specific commands currently implemented in the library.

d3rpp:
would you mean kind of like how you import the `redis::Commands` trait to access commands you'd then import `redis::json::Commands` to get these?

d3rpp:
Is there a specific port or something that I'd want my Redis server listening on? I can't seem to get it to connect for testing purposes

d3rpp:
I've needed to also update the TestContext code to add a `--loadmodule` arg to the `redis-server` which will then take the path to the RedisJSON Module from the env var `REDISRS_REDISJSON_PATH`.

Could I get some opinions on if I should be doing it differently?

d3rpp:
the redis version in the CI server doesn't have a RedisJSON version. would it be wise to add one?

The main issue is that there are no distributed binaries, only source code, so should I add a step to the CI where it installs redisjson?

d3rpp:
sorry bare with me im bad at github ci

d3rpp:
image

RedisJSON doesn't compile in rust version 1.51.0, as such I say we should just remove testing for 1.51.0 since we can't separate the json testing in the current layout

.github/workflows/rust.yml
```diff
@@ -23,7 +24,8 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.51.0
+          # RedisJSON literally will refuse to compile in this version of rust
```
jaymell, line 27: We don't want to remove Rust 1.51. There's no reason for RedisJSON to be built using the same Rust versions that the client library uses -- `stable` should be fine.

d3rpp: you make a very good point, I'll add a `+stable` to the cargo build command for building redisjson

djc:
> > would you mean kind of like how you import the `redis::Commands` trait to access commands you'd then import `redis::json::Commands` to get these?
> 
> Yes, I think you would just need to call `implement_commands!` in another module and add the following (likewise for `AsyncCommands`)
> 
> ```rust
> impl<T> Commands for T where T: ConnectionLike {}
> ```
> 
> The additional trait could be a bit cumbersome but might be an effective way to organize commands for external modules. @djc Any thoughts?

Hmm, I think that could make sense, yes.

d3rpp:
Sorry I haven't abandoned this I've just been busy.

The reason the tests are currently failing is that the test software is spawning so many Redis servers that the OS is literally running out of sockets to give to it leading to OS Errors 24, 104, and 11

This is happening on my machine as well as the CI Servers, there is a fix via `ulimit` but it is the equivalent of patching a broken bone with a band-aid.

I propose going into the make file and running every test module individually, it means the tests would take forever to run but there would never be too many connections

d3rpp:
> > > would you mean kind of like how you import the `redis::Commands` trait to access commands you'd then import `redis::json::Commands` to get these?
> > 
> > 
> > Yes, I think you would just need to call `implement_commands!` in another module and add the following (likewise for `AsyncCommands`)
> > ```rust
> > impl<T> Commands for T where T: ConnectionLike {}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The additional trait could be a bit cumbersome but might be an effective way to organize commands for external modules. @djc Any thoughts?
> 
> Hmm, I think that could make sense, yes.

I think this would also be good to add though they're still modularised with the `json` feature. As such it may not be strictly necessary but nonetheless a good idea.

d3rpp:
The tests now work

Makefile
```diff
@@ -10,22 +10,38 @@ test:
...
-	@REDISRS_SERVER_TYPE=unix cargo test --test parser --test test_basic --test test_types --all-features -- --test-threads=1
+	@REDISRS_SERVER_TYPE=unix cargo test --test parser --test test_basic --test test_types --features test_default -- --test-threads=1
 
 	@echo "===================================================================="
 	@echo "Testing Connection Type UNIX SOCKETS"
 	@echo "===================================================================="
-	@REDISRS_SERVER_TYPE=unix cargo test --all-features -- --skip test_cluster
+	@REDISRS_SERVER_TYPE=unix cargo test --features test_default -- --skip test_cluster --skip test_json
+
+test-json:
```
d3rpp, line 30: by separating the tests for this feature into a different command set entirely we allow future additions (perhaps redisearch for example) and this makefile would be ready to accommodate for them.

Cargo.toml
```diff
@@ -58,11 +58,18 @@ native-tls = { version = "0.2", optional = true }
...
+serde = { version = "1.0.137", optional = true }
+serde_json = { version = "1.0.81", optional = true }
+
 # Optional aHash support
 ahash = { version = "0.7.6", optional = true }
 
 [features]
 default = ["acl", "streams", "geospatial", "script"]
+# added to allow for separate testing of the JSON feature
+test_default = ["acl", "streams", "aio", "geospatial", "cluster", "script", "tls", "async-std-comp", "async-std-tls-comp", "tokio-comp", "tokio-native-tls-comp", "connection-manager", "streams"]
```
d3rpp, line 71: `--all-features` without the `json` feature, this prevents EVERY SINGLE SERVER INSTANCE from loading redisJSON

d3rpp:
the `implement_commands!` macro is now called twice and the `commands` module has been turned into a folder with a mod.rs, the macro itself has been moved to a separate file and the json commands have been moved to `src/commands/json.rs` with the use of 
```rust
impl<T> Commands for T where T: ConnectionLike {}

#[cfg(feature = "aio")]
impl<T> AsyncCommands for T where T: crate::aio::ConnectionLike + Send + Sized {}
```
as mentioned above by @djc 
ngl this is better

Cargo.toml
d3rpp: would it be better to instead just use the makefile and specify  them manually? or use a list in the make file?

d3rpp:
well y'all completely remade this repo more or less so I'll close this and open a new one with the new method.

PR #645, opened 2022-07-03
fix: &Vec is_single_arg check

OmmyZhang:
Without this, `con.get(&vec!["key1", "key2"]` will cause `Err(An error was signalled by the server: wrong number of arguments for 'get' command)`.

(Not sure if I put tests in the right place.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -898,6 +898,10 @@ impl<T: ToRedisArgs> ToRedisArgs for &T {
     {
         (*self).write_redis_args(out)
     }
+
+    fn is_single_arg(&self) -> bool {
+        (*self).is_single_arg()
+    }
 }
 
 /// @note: Redis cannot store empty sets so the application has to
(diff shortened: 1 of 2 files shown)


OmmyZhang:
Test in `test_is_single_arg` will not work. `is_single_arg()` for &Vec will just use `ToRedisArgs` for Vec. But if `dbg!(key.is_single_arg())` here 

https://github.com/redis-rs/redis-rs/blob/c21c2df6d8e18fbaab06de5b3d6617f72bf6c6e8/src/commands.rs#L301-L302

it will always return `true` as it goes here  https://github.com/redis-rs/redis-rs/blob/c21c2df6d8e18fbaab06de5b3d6617f72bf6c6e8/src/types.rs#L677-L679

Then it uses "GET" instead of "MGET", and cause that error.

djc:
Thanks!

Issue #579, opened 2022-01-27
`Direction` is not exposed making it impossible to use `blmove`


tasn:
`redis::commands` is a private module, so one can't access `Direction`.

rudolphfroger-ds:
Any progress on this?

tasn:
Weird that it didn't close it, thanks for the update!

PR #646, opened 2022-07-03
Make `Direction` accessible to use with the `LMOVE` command 

thorbadour:
This solves https://github.com/redis-rs/redis-rs/issues/579


djc (approved):

djc:
Thanks!

PR #650, opened 2022-07-17
Add Support for RedisJSON

d3rpp:
I took on board all of the feedback given in my first attempt at this, and this should be a much cleaner implementation, especially regarding testing.

the `TestContext` now has a new function `new_with_modules` taking a `Vec` of `RedisModules`, it is then used to load them from environment variables when necessary. and selectively.

It is not breaking change to the outside world, and the addition lays some basic foundation for further support of other Redis Modules.

I hope this is better than my bad implementation before

### DIFF
--- a/redis/src/commands.rs
+++ b/redis/src/commands.rs
@@ -849,6 +849,30 @@ implement_commands! {
         cmd("ZINTERSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("AGGREGATE").arg("MAX")
     }
 
+    /// [`Commands::zinterstore`], but with the ability to specify a
+    /// multiplication factor for each sorted set by pairing one with each key
+    /// in a tuple.
+    fn zinterstore_weights<K: ToRedisArgs, W: ToRedisArgs>(dstkey: K, keys: &'a [(K, W)]) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(key, weight)| (key, weight)).unzip();
+        cmd("ZINTERSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("WEIGHTS").arg(weights)
+    }
+
+    /// [`Commands::zinterstore_min`], but with the ability to specify a
+    /// multiplication factor for each sorted set by pairing one with each key
+    /// in a tuple.
+    fn zinterstore_min_weights<K: ToRedisArgs, W: ToRedisArgs>(dstkey: K, keys: &'a [(K, W)]) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(key, weight)| (key, weight)).unzip();
+        cmd("ZINTERSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("AGGREGATE").arg("MIN").arg("WEIGHTS").arg(weights)
+    }
+
+    /// [`Commands::zinterstore_max`], but with the ability to specify a
+    /// multiplication factor for each sorted set by pairing one with each key
+    /// in a tuple.
+    fn zinterstore_max_weights<K: ToRedisArgs, W: ToRedisArgs>(dstkey: K, keys: &'a [(K, W)]) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(key, weight)| (key, weight)).unzip();
+        cmd("ZINTERSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("AGGREGATE").arg("MAX").arg("WEIGHTS").arg(weights)
+    }
+
     /// Count the number of members in a sorted set between a given lexicographical range.
     fn zlexcount<K: ToRedisArgs, L: ToRedisArgs>(key: K, min: L, max: L) {
         cmd("ZLEXCOUNT").arg(key).arg(min).arg(max)
@@ -1034,6 +1058,30 @@ implement_commands! {
     /// a new key using MAX as aggregation function.
     fn zunionstore_max<K: ToRedisArgs>(dstkey: K, keys: &'a [K]) {
         cmd("ZUNIONSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("AGGREGATE").arg("MAX")
+    }
+
+    /// [`Commands::zunionstore`], but with the ability to specify a
+    /// multiplication factor for each sorted set by pairing one with each key
+    /// in a tuple.
+    fn zunionstore_weights<K: ToRedisArgs, W: ToRedisArgs>(dstkey: K, keys: &'a [(K, W)]) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(key, weight)| (key, weight)).unzip();
+        cmd("ZUNIONSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("WEIGHTS").arg(weights)
+    }
+
+    /// [`Commands::zunionstore_min`], but with the ability to specify a
+    /// multiplication factor for each sorted set by pairing one with each key
+    /// in a tuple.
+    fn zunionstore_min_weights<K: ToRedisArgs, W: ToRedisArgs>(dstkey: K, keys: &'a [(K, W)]) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(key, weight)| (key, weight)).unzip();
+        cmd("ZUNIONSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("AGGREGATE").arg("MIN").arg("WEIGHTS").arg(weights)
+    }
+
+    /// [`Commands::zunionstore_max`], but with the ability to specify a
+    /// multiplication factor for each sorted set by pairing one with each key
+    /// in a tuple.
+    fn zunionstore_max_weights<K: ToRedisArgs, W: ToRedisArgs>(dstkey: K, keys: &'a [(K, W)]) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(key, weight)| (key, weight)).unzip();
+        cmd("ZUNIONSTORE").arg(dstkey).arg(keys.len()).arg(keys).arg("AGGREGATE").arg("MAX").arg("WEIGHTS").arg(weights)
     }
 
     // hyperloglog commands
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -922,6 +922,119 @@ fn test_redis_server_down() {
 }
 
 #[test]
+fn test_zinterstore_weights() {
+    let ctx = TestContext::new();
+    let mut con = ctx.connection();
+
+    let _: () = con
+        .zadd_multiple("zset1", &[(1, "one"), (2, "two"), (4, "four")])
+        .unwrap();
+    let _: () = con
+        .zadd_multiple("zset2", &[(1, "one"), (2, "two"), (3, "three")])
+        .unwrap();
+
+    // zinterstore_weights
+    assert_eq!(
+        con.zinterstore_weights("out", &[("zset1", 2), ("zset2", 3)]),
+        Ok(2)
+    );
+
+    assert_eq!(
+        con.zrange_withscores("out", 0, -1),
+        Ok(vec![
+            ("one".to_string(), "5".to_string()),
+            ("two".to_string(), "10".to_string())
+        ])
+    );
+
+    // zinterstore_min_weights
+    assert_eq!(
+        con.zinterstore_min_weights("out", &[("zset1", 2), ("zset2", 3)]),
+        Ok(2)
+    );
+
+    assert_eq!(
+        con.zrange_withscores("out", 0, -1),
+        Ok(vec![
+            ("one".to_string(), "2".to_string()),
+            ("two".to_string(), "4".to_string()),
+        ])
+    );
+
+    // zinterstore_max_weights
+    assert_eq!(
+        con.zinterstore_max_weights("out", &[("zset1", 2), ("zset2", 3)]),
+        Ok(2)
+    );
+
+    assert_eq!(
+        con.zrange_withscores("out", 0, -1),
+        Ok(vec![
+            ("one".to_string(), "3".to_string()),
+            ("two".to_string(), "6".to_string()),
+        ])
+    );
+}
+
+#[test]
+fn test_zunionstore_weights() {
+    let ctx = TestContext::new();
+    let mut con = ctx.connection();
+
+    let _: () = con
+        .zadd_multiple("zset1", &[(1, "one"), (2, "two")])
+        .unwrap();
+    let _: () = con
+        .zadd_multiple("zset2", &[(1, "one"), (2, "two"), (3, "three")])
+        .unwrap();
+
+    // zunionstore_weights
+    assert_eq!(
+        con.zunionstore_weights("out", &[("zset1", 2), ("zset2", 3)]),
+        Ok(3)
+    );
+
+    assert_eq!(
+        con.zrange_withscores("out", 0, -1),
+        Ok(vec![
+            ("one".to_string(), "5".to_string()),
+            ("three".to_string(), "9".to_string()),
+            ("two".to_string(), "10".to_string())
+        ])
+    );
+
+    // zunionstore_min_weights
+    assert_eq!(
+        con.zunionstore_min_weights("out", &[("zset1", 2), ("zset2", 3)]),
+        Ok(3)
+    );
+
+    assert_eq!(
+        con.zrange_withscores("out", 0, -1),
+        Ok(vec![
+            ("one".to_string(), "2".to_string()),
+            ("two".to_string(), "4".to_string()),
+            ("three".to_string(), "9".to_string())
+        ])
+    );
+
+    // zunionstore_max_weights
+    assert_eq!(
+        con.zunionstore_max_weights("out", &[("zset1", 2), ("zset2", 3)]),
+        Ok(3)
+    );
+
+    assert_eq!(
+        con.zrange_withscores("out", 0, -1),
+        Ok(vec![
+            ("one".to_string(), "3".to_string()),
+            ("two".to_string(), "6".to_string()),
+            ("three".to_string(), "9".to_string())
+        ])
+    );
+}
+
+#[test]
 fn test_zrembylex() {
     let ctx = TestContext::new();
     let mut con = ctx.connection();


d3rpp:
Sorry about that one, as you can see, Github CI is not my strength

redis/src/commands/json.rs
```diff
@@ -0,0 +1,135 @@
...
+#[cfg(feature = "cluster")]
+use crate::commands::ClusterPipeline;
+
+use serde::ser::Serialize;
+
+implement_commands! {
+    'a
+	
+	/// Append the JSON `value` to the array at `path` after the last element in it.
+	#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
```
jaymell, line 17: Is there any reason for these attributes, given that they're all within a module that's conditional on the feature?

redis/src/commands/implementer_macro.rs
```diff
@@ -0,0 +1,275 @@
+macro_rules! implement_commands {
```
jaymell, line 1: I wonder if we should just call this file `macros.rs`?

README.md
```diff
@@ -102,8 +102,47 @@ fn fetch_an_integer() -> String {
...
+    let client = Client::open("redis://127.0.0.1").unwrap();
+    let connection = client.get_connection().unwrap();
+
+    // runs `JSON.SET {key} {path} {b}`
+    connection.json_set(key, path, b)?
+}
+
+```
+
+**NOTE:** Importing both `redis::Commands` and `redis::json::Commands` may cause you build to fail,
```
jaymell, line 130: Can we export `redis::json::Commands`, etc., from the commands module with different names so that we don't have the potentially confusing build failures?

redis/src/commands/json.rs
d3rpp: You make a good point, will fix

redis/src/commands/implementer_macro.rs
d3rpp: sure, will fix

README.md
d3rpp: In this case `redis::Commands` and `redis::json::Commands` are implementations of the same trait. I'm not certain if the trait could be renamed but I can test or perhaps I can figure something out.

the `redis::json::Commands` is imported conditionally in the `lib.rs` file https://github.com/d3rpp/redis-rs-tokio-drift/blob/da865eef905e67437c78afdf05427130899cff77/redis/src/lib.rs#L443

I could *attempt* to make it export as `redis::JsonCommands` for example but I can guarantee nothing.

d3rpp: that was surprisingly easy

redis/src/lib.rs
```diff
@@ -369,6 +369,10 @@ pub use crate::connection::{
 pub use crate::parser::{parse_redis_value, Parser};
 pub use crate::pipeline::Pipeline;
 
+#[cfg(feature = "json")]
+#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+pub use crate::commands::json::Commands as JSONCommands;
```
jaymell, line 374: Looks good, but can we do the aliasing within the `commands` module itself, then you can just `use` the alias here. Also don't forget we'll need `AsyncCommands` aliased as well.

d3rpp: > can we do the aliasing within the commands module itself?

yes, I'll get on that
> don't forget we'll need AsyncCommands aliased as well

ah whoops, will fix

Cargo.toml
```diff
@@ -1,2 +1,7 @@
 [workspace]
 members = ["redis", "redis-test"]
+
+# Exclude ci directories, giving CI the ability to build
+# Packages without them being included in the workspace even thought they aren't
+# this one line has wasted 3 hours of my time
```
jaymell, line 6: I empathize with your pain here but can we remove the comment? :stuck_out_tongue:  

Actually, do we even need this? Would a .gitignore line for the `.ci` folder accomplish what's needed?

d3rpp: No because the issue is that when I try to build redisjson for testing I do it in the .do folder and cargo goes up a couple levels and finds the workspace and complains about it being in the project but not being in the project

I'm happy to remove the comment though you're right.

I tried using a blob pattern to exclude the whole CI folder but it didn't work unfortunately

jaymell: Ok thanks for the explanation. I see the issue now. Can we not just move the directory where redis-json is checked out into a directory that isn't a subdir of the main project? 

Would really like to keep something that's purely CI-related out of Cargo.toml if possible.

d3rpp:
So when you clone something outside the GitHub workspace on the CI servers, it throws it to some arbitrary location that got a UUID in its path.

So instead of moving it I've changed it to the ci will add the exclude to cargo and then once RedisJSON is done building, it will restore it to its previous state before running tests

also apparently the `--loadmodule` needs to have an absolute path because when i used `./__ci/redis-json/target/release/librejson.so` it refused to load and the JSON tests failed, though when moved to `/tmp/librejson.so` it works just fine

d3rpp:
what

d3rpp:
yeah surely that was one time I ran the same tests on the same code in the same CI servers twice and had no issues, I do it here, and it failed on a test that I didn't go anywhere near tampering with

.github/workflows/rust.yml
```diff
@@ -54,8 +55,30 @@ jobs:
...
+    - name: Checkout RedisJSON
+      uses: actions/checkout@v2
+      with:
+        repository: "RedisJSON/RedisJSON"
+        path: "./__ci/redis-json"
+        set-safe-directory: false
+
+    - name: Compile RedisJSON
+      run: |
+        # copy current Cargo.toml
```
jaymell, line 67: Can we add a small comment above this section briefly explaining why the steps here look so crazy? I'm not too concerned about it and don't think it's necessary to wrestle with it further, but I'm hoping that we'll be able to solve the problem here a bit more elegantly in the future.

jaymell:
Thanks again for the PR. I'm about ready to approve this, provided @djc is okay w/ the restructuring in the PR and we address the other minor review comment above.

d3rpp:
As you can see there is a merge conflict, so i'll need to fox that first since in my fork it wont let me do it automatically

redis/Cargo.toml
```diff
@@ -61,11 +61,16 @@ async-native-tls = { version = "0.4", optional = true }
 # Optional aHash support
 ahash = { version = "0.7.6", optional = true }
 
+# Only needed for JSON Support
+serde_json = {version = "1.0.82", optional = true }
```
djc, line 65: Style nit: please put a space inside the opening brace here.

README.md
```diff
@@ -102,8 +102,42 @@ fn fetch_an_integer() -> String {
...
+Then you can simply include the `JSONCommands` which will add the `json` commands to all Redis Connections (not to be confused with just `Commands` which only adds the default commands)
+
+```rust
+use redis::Client;
+use redis::JSONCommands;
+use redis::RedisResult;
+
+// Result returns Ok(true) if the value was set
+// Result returns Err(e) if there was an error with the server itself 
+fn set_json_bool<P: ToString>(key:P, path:P, b: bool) -> RedisResult<bool> {
```
djc, line 120: Style nit: add some spaces after the `:` in arguments.

README.md
```diff
@@ -102,8 +102,42 @@ fn fetch_an_integer() -> String {
...
+    // It will always be a Vec, if no results were found at the path it'll be an empty array
+}
+
+```
+
 ## Development
 
+If you want to test with modules, you must:
+- Have those modules
+- Set the `REDISRS_{module-name}_PATH` environment variable
```
djc, line 137: I think we should add some underscores to these environment variable names to make them less surprising, `REDIS_RS_{module-name}_PATH` and `REDIS_RS_REDIS_JSON_PATH`.

redis/src/commands/json.rs
```diff
@@ -0,0 +1,118 @@
...
+#[cfg(feature = "cluster")]
+use crate::commands::ClusterPipeline;
+
+use serde::ser::Serialize;
+
+implement_commands! {
+    'a
+	
+	/// Append the JSON `value` to the array at `path` after the last element in it.
+	fn json_arrappend<K: ToRedisArgs, P: ToString, V: Serialize>(key: K, path: P, value: V) {
```
djc, line 17: I think `arr_append` would fit with the Rust naming convention better (same for all the other functions).

I don't like the `ToString` argument type here... why can't this just be `&str`? We don't want to require allocations if at all possible.

Also I think this would be a little easier to read if you used `impl Trait` for these arguments (so `key: impl ToRedisArgs` etc instead of the indirection of named generics).

Given that rustfmt is disabled here (I do wonder if that's really still necessary these days), you'll need to spend some effort to manually format this sanely. That means 4-space indents, and the `cmd()` calls should probably be formatted like this:

```rust
        cmd("JSON.ARRAPPEND")
            .arg(key)
            .arg(path.to_string())
            .arg(serde_json::to_string(&value).unwrap())
```

I don't think we want to `unwrap()` here, should just use the `?` operator and add an error variant for `serde_json::Error` that's conditional on the `json` feature.

Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["redis", "redis-test"]
+members = ["redis", "redis-test"]
```
djc, line 2: Nit: please add a trailing newline.

redis/src/commands/json.rs
```diff
@@ -0,0 +1,118 @@
...
+	/// Reports the type of JSON value at `path`s.
+	fn json_type<K: ToRedisArgs, P: ToString>(key: K, path: P) {
+		cmd("JSON.TYPE").arg(key).arg(path.to_string())
+	}
+}
+
+impl<T> Commands for T where T: ConnectionLike {}
+
+#[cfg(feature = "aio")]
+impl<T> AsyncCommands for T where T: crate::aio::ConnectionLike + Send + Sized {}
```
djc, line 118: Nit: trailing newline.

redis/src/commands/mod.rs
```diff
@@ -50,281 +63,7 @@ pub(crate) fn is_readonly_cmd(cmd: &[u8]) -> bool {
     )
 }
 
-macro_rules! implement_commands {
```
djc, line 53: I'd like this refactoring to be separated from the other changes in this PR, either as separate commits in this PR (with all the JSON changes squashed into a single PR) or as as a separate PR.

redis/tests/support/mod.rs
```diff
@@ -45,6 +45,10 @@ enum ServerType {
     Unix,
 }
 
+pub enum RedisModules {
+    RedisJSON,
```
djc, line 49: The enum should be called `RedisModule` (or just `Module`, @jaymell what do you think?); the variant should be called `RedisJson`.

redis/tests/support/mod.rs
```diff
@@ -98,18 +102,80 @@ impl RedisServer {
...
+            },
+            Some(modules),
+        )
     }
 
     pub fn new_with_addr<F: FnOnce(&mut process::Command) -> process::Child>(
         addr: redis::ConnectionAddr,
         tls_paths: Option<TlsFilePaths>,
         spawner: F,
+        modules: Option<Vec<RedisModules>>,
```
djc, line 160: I think this can just take a `&[RedisModules]` instead. If there are no modules, it can be an empty slice, so no need for the `Option` wrapper.

redis/tests/support/mod.rs
```diff
@@ -98,18 +102,80 @@ impl RedisServer {
...
+            None,
+            |cmd| {
+                cmd.spawn()
+                    .unwrap_or_else(|err| panic!("Failed to run {:?}: {}", cmd, err))
+            },
+            None,
+        )
+    }
+
+    pub fn new_with_modules(modules: Vec<RedisModules>) -> RedisServer {
```
djc, line 116: This can take a `&[RedisModules]` instead, since we presumably don't need to mutate this? Also I think naming this `with_modules()` (omitting the `new_` prefix) is more idiomatic.

Also we don't want to duplicate `new()` here, `new()` should just call `with_modules()` with an empty `modules` argument.

redis/tests/support/mod.rs
```diff
@@ -98,18 +102,80 @@ impl RedisServer {
...
         let mut redis_cmd = process::Command::new("redis-server");
+
+        // Load Redis Modules
+        // This will take the enum variants and load them from Environment Variables
+        // This method allows for implementation of other Redis Modules in the future
+        if let Some(mods) = modules {
+            for m in mods {
+                match m {
+                    RedisModules::RedisJSON => {
+                        redis_cmd
```
djc, line 171: I don't think we should just `unwrap()` here. Would it be better to pass `modules` in as a `HashMap` or `&[(RedisModule, &OsStr)]` or whatever that has the path in it?

redis/tests/support/mod.rs
```diff
@@ -239,6 +305,42 @@ impl TestContext {
         TestContext { server, client }
     }
 
+    pub fn new_with_modules(modules: Vec<RedisModules>) -> TestContext {
```
djc, line 308: Same here with the duplication.

redis/src/commands/mod.rs
```diff
@@ -17,6 +20,16 @@ use crate::streams;
 #[cfg(feature = "acl")]
 use crate::acl;
 
+#[cfg(feature = "json")]
+#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+mod json;
+
+#[cfg(feature = "json")]
+pub use json::Commands as JSONCommands;
```
djc, line 28: These should be called `JsonCommands` and `AsyncJsonCommands`.

djc (commented):
I think this is in a decent shape, though I have a lot of little things to fix up.

d3rpp:
Given that this PR is stale because of an annoying merge request and the request to move a bunch of commits around, I'm going to remake this again sorted hopefully exactly how you'd like it.

I thank you for being so patient with my apparent inability to code properly, and I'll see you in a bit with what is hopefully a final attempt at getting this feature in 😄

djc:
I think it'd be good to keep this PR open, so we can track what happens with the PR feedback you've gotten. Just open a separate PR that only contains the macro move, we'll merge that and that will avoid most of the merge conflicts down the line. Then you can squash all the commits here into a single one (let me know if you need help with that) and rebase it onto the post-macro-move main branch.

README.md
d3rpp: the GitHub repo for RedisJSON is [RedisJSON/RedisJSON](https://github.com/RedisJSON/RedisJSON) as opposed to Redis-Json or redis_json which gives the idea that it meant to be read as one word

Might I suggest something like `REDIS_RS_REDISJSON_PATH` which also follows the naming of RediSearch being one word, so it'd be `REDIS_RS_REDISEARCH_PATH`.

redis/src/commands/json.rs
d3rpp: > We don't want to require allocations if at all possible.

Understandable, having a `&str` would do the job better

.github/workflows/rust.yml
d3rpp: Yeah this is something that `cargo` would need to change, telling it *not* to go up a level or 2 to check for workspaces would be nice to be able to disable

README.md
djc: The way I understand the SCREAMING_CASE and snake_case conventions is that internal capitals (except in acronyms) will introduce a new word and thus require a `_` separator.

I recently had a discussion with a coworker about this and asked Twitter what people thought, see also https://twitter.com/djco/status/1513935084281290763. So I would still consider `REDIS_JSON` more intuitive for an environment variable name if the project is called RedisJSON.

redis/tests/support/mod.rs
d3rpp: It is loaded from an environment variable, would it be wise to add a few lines in the `Makefile` that check that the env vars are set?

Otherwise I could also set it to use `expect` with an actually helpful message other that "Error: Not Present"

README.md
d3rpp: You make a very good point

d3rpp:
> I think it'd be good to keep this PR open, so we can track what happens with the PR feedback you've gotten.

No worries I'll keep this one over it and link them together once I make the new one

> Just open a separate PR that only contains the macro move

No worries, I'll put it at the top of my todo list tomorrow

> Then you can squash all the commits here into a single one (let me know if you need help with that) and rebase it onto the post-macro-move main branch.

I have 50% of an idea on how to squash commits but I will do it after the macro move gets accepted

redis/tests/support/mod.rs
djc: Yeah, I'm questioning whether it makes sense to do this as an environment variable as opposed to in the code.

I prefer to avoid panics in most cases, so maybe this method should be made fallible, returning a `Result` with `Err` in the case of a missing environment variable?

djc:
> No worries I'll keep this one over it and link them together once I make the new one

That would still make it much harder to navigate.

> I have 50% of an idea on how to squash commits but I will do it after the macro move gets accepted

I usually use `git rebase -i/--interactive`, then you just replace all the `pick`s with `f` (for `fixup`) except the first one in the rebase action list.

redis/tests/support/mod.rs
d3rpp: To hard code it may make life difficult for testing outside of the CI Environment so I definitely think we should keep the Env Var stage since the location of the libs may vary, for example on the CI it is stored at `/tmp/librejson.so` but on my machine it is stored at `/opt/librejson.so`, but on other machines it may be stored at `/usr/lib/librejson.so`, to wrap the `TestContext` generator with a result would make this better but it'd mean shifting around all of the tests to include `.unwrap()`'s in the tests themselves, or we could do this

```rust
pub fn new() {
    with_modules(&[]).unwrap()
}
pub fn with_modules(modules: &[RedisModule]) -> Result<TestContext, Error> {
    // ...
}
```
either way, at some point, an error is being unwrapped

djc: Okay, let's keep the environment variable and call `expect()` on it with a nice explanatory message.

d3rpp:
I have an copy of my changes and access to this, I'll open a new one with just the macro move than I'll merge your changes + my changes into a second PR

redis/src/commands/json.rs
d3rpp: > I don't think we want to `unwrap()` here, should just use the `?` operator and add an error variant for `serde_json::Error` that's conditional on the `json` feature.

Given the current state of the macro, that likely won't be possible and I'll need to make a new macro to do it. I'll get started on that with heavy inspiration from the first macro

d3rpp: It would be easier and faster to leave the `unwrap()` in there since the failure only happens if the person writing it doesn't know how json works

https://docs.rs/serde_json/latest/serde_json/fn.to_string.html#errors

d3rpp: Progress: I wrote a new macro for json commands. that takes a `RedisResult<Cmd>` in the `$block` as opposed to `&mut Cmd` (needed to add an arg function that takes/returns `self` and returns `self` as opposed to `&mut self`).

All tests pass, and the unwrapping of the results from `$block` is done with `?` before the query which means it still just returns 1 `RedisResult` with 2 chances to fail (serialize error and query error).

I'd recommend reviewing my macro once I open the PR since it's my first time making a macro this complex

PR #651, opened 2022-07-19
fixed to get the real ip in redis cluster when refresh slots, this wi…

wxmayifei:
Under redis 7.0, redis will return 127.0.0.1 in the 'cluster nodes' or 'cluster slots' commands with the same conneciton, this fix will replace 127.0.0.1 to the real host which can be used to establish connection

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,9 @@ fn fetch_an_integer() -> redis::RedisResult<isize> {
 
 ## Async support
 
-To enable asynchronous clients a feature for the underlying feature need to be activated.
+To enable asynchronous clients, enable the relevant feature in your Cargo.toml,
+`tokio-comp` for tokio users or `async-std-comp` for async-std users.
+
 
 ```
 # if you use tokio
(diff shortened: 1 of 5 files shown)


djc:
Can you describe in more detail the behavior that we're trying to paper over here? Is this a "regression" (at least from your perspective) in Redis 7?

wxmayifei:
My description is wrong, sorry ,this problem is not in Redis 7，It's in the version Redis 6 and belows.
In redis cluster mode,redis-rs will use "cluster info " or "cluster slots" command to find out all the master nodes.
For example , If there are three redis master node such as : 192.168.1.1:6379 , 192.168.1.2:6379 , 192.168.1.3:6379 .
when we execute the "cluster slots" command on 192.168.1.1:6379 to find out all the master nodes, the server will return:
127.0.0.1:6379 , 192.168.1.2:6379 , 192.168.1.3:6379, In this scene, if we not replace 127.0.0.1 to 192.168.1.1 , 
there will produce error if we want to access the 192.168.1.1:6379 node , it will access 127.0.0.1:6379 in fact

wxmayifei:
> Can you describe in more detail the behavior that we're trying to paper over here? Is this a "regression" (at least from your perspective) in Redis 7?

My description is wrong, sorry ,this problem is not in Redis 7，It's in the version Redis 6 and belows.
In redis cluster mode,redis-rs will use "cluster info " or "cluster slots" command to find out all the master nodes.
For example , If there are three redis master node such as : 192.168.1.1:6379 , 192.168.1.2:6379 , 192.168.1.3:6379 .
when we execute the "cluster slots" command on 192.168.1.1:6379 to find out all the master nodes, the server will return:
127.0.0.1:6379 , 192.168.1.2:6379 , 192.168.1.3:6379, In this scene, if we not replace 127.0.0.1 to 192.168.1.1 ,
there will produce error if we want to access the 192.168.1.1:6379 node , it will access 127.0.0.1:6379 in fact

djc:
@jaymell I don't understand Redis clustering well enough to make sense of this. Do you have an opinion?

utkarshgupta137:
> > Can you describe in more detail the behavior that we're trying to paper over here? Is this a "regression" (at least from your perspective) in Redis 7?
> 
> My description is wrong, sorry ,this problem is not in Redis 7，It's in the version Redis 6 and belows. In redis cluster mode,redis-rs will use "cluster info " or "cluster slots" command to find out all the master nodes. For example , If there are three redis master node such as : 192.168.1.1:6379 , 192.168.1.2:6379 , 192.168.1.3:6379 . when we execute the "cluster slots" command on 192.168.1.1:6379 to find out all the master nodes, the server will return: 127.0.0.1:6379 , 192.168.1.2:6379 , 192.168.1.3:6379, In this scene, if we not replace 127.0.0.1 to 192.168.1.1 , there will produce error if we want to access the 192.168.1.1:6379 node , it will access 127.0.0.1:6379 in fact

I'm unable to reproduce this. I'm getting the correct IPs (172.20.x in our case) for all nodes when running `cluster slots` on redis-server 4.0.14 using redis-cli. We've setup redis using Azure service.

wxmayifei:
@utkarshgupta137 
This may have been configured cluster-announce-ip option.

PR #640, opened 2022-06-24
[5/5] Add Async MultiplexedClusterConnection

utkarshgupta137:
Fixes #421.

This is the final PR for adding async connection support for cluster.

TODO:
- [ ] Add tests
- [ ] Add retry support
- [ ] Add reconnect support
- [ ] Profile & improve performance
- [ ] Benchmark against other implementations (redis_cluster_async & fred.rs)
- [ ] Add/improve docs

P.S. I've not added support for `simple` async connections, only `multiplexed` connections. I'm unaware of any benefits of the `simple` async connections.

PR #653, opened 2022-07-24
[3/5] cluster: refactor/optimize for Async Cluster

utkarshgupta137:
Continuing from the previous PR, this PR refactors & optimizes ClusterConnection impl for Async Cluster.

Main changes:
- Removed random connection logic
  - Since we only operate in 100% slot coverage scenarios, there is no point in maintaining random connection logic, which is supposed to help us find missing nodes in the cluster topology
  - This also allows us to remove RefCell around connections & slots
- Removed RefCell from connections/slot in the struct
  - RefCells aren't needed anymore & removing them gives a minor performance bump
- Consolidated logic for query routing into the Routable trait
  - This improves code reuse within the impl as well as later in Async impl
- Consolidated logic for create_initial_connections & refresh_slots
  - This improves code reuse within the impl as well as later in Async impl
  - Also renamed addr -> node where it is used as a string
- Directly exposed Slot fields instead of bare methods
  - There is no benefit of accessing the fields via functions
- Merge send_all_commands & recv_all_commands
  - There is no point in having 2 different methods

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -75,23 +75,7 @@ pub struct ClusterConnection {
     password: Option<String>,
     read_timeout: RefCell<Option<Duration>>,
     write_timeout: RefCell<Option<Duration>>,
-    tls: Option<TlsMode>,
-}
-
-#[derive(Clone, Copy)]
-enum TlsMode {
-    Secure,
-    Insecure,
-}
-
-impl TlsMode {
-    fn from_insecure_flag(insecure: bool) -> TlsMode {
-        if insecure {
-            TlsMode::Insecure
-        } else {
-            TlsMode::Secure
-        }
-    }
+    tls_insecure: Option<bool>,
 }
 
 impl ClusterConnection {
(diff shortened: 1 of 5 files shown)


djc:
Okay, I skimmed through the commits in this PR. Although many of the changes seem like the right direction (builder pattern, removal of `RefCell` stuff), I found it quite hard to understand the purpose of each change -- commit messages sort of mention what the change is, but not why these changes are needed. Additionally, while it's nice that there are small commits here, these changes are still quite hard to review, for example because you moved code around while also changing it substantially. I think this should probably be split into closer to 3 PRs:

* One for the `ClusterClient` changes
* One for the builder setup
* One for the routing/slots stuff

Please, in your commit messages tie the changes in that commit to the PR's goal (for example, making sync and async code more similar or reduce redundant checks.)

jaymell:
I strongly agree with the request from @djc  above -- breaking this into additional PRs will facilitate getting this changes in more smoothly. I've looked over the code in some depth, and broadly speaking it looks great but it's difficult to keep the entirety of this in context while reviewing.

Also, re the removal of the random connection logic: I'll admit I'm not entirely clear if it's needed, but it's probably the most substantive change you've made to the existing code. It currently exists both in this repo and in the [redis-cluster-async](https://github.com/redis-rs/redis-cluster-async), and I think should either be kept in place for now or given its own PR. There is a strong potential to cause clients to panic if it's taken out erroneously.

utkarshgupta137:
Depends on #670

nihohit:
@utkarshgupta137, are these PRs still relevant, after https://github.com/redis-rs/redis-rs/pull/919 ? If so, I'd like it if you rebased your PRs (this, https://github.com/redis-rs/redis-rs/pull/654, & https://github.com/redis-rs/redis-rs/pull/640 ) over https://github.com/redis-rs/redis-rs/pull/919.

Also, smaller changes, like "Directly exposed Slot fields instead of bare methods" can be opened as their own PRs, for faster merging.

jaymell:
Closing, longer relevant given subsequent work.

PR #657, opened 2022-08-02
Add support for RedisJSON

d3rpp:
Part 2 of #650 

# Changes over previous editions:

To take on board some of the feedback in #650 (specifically the result variant), I needed to make a few changes

## Custom implementer macro:
This was done because the redisjson functions in implementation now return results that get unwrapped before the query, following the following structure

- Did Serialise Correctly?
    - YES: Go on with Query -> Return Result of Query
    - NO: Return `Result::Err` with `ErrorKind::Serialize` (only active with `json` feature) and a description given by the `serde_json` error messages

This change will only affect some RedisJson Commands, the core commands are untouched by this

There will likely be debate over the removed call to `::std::mem::replace`; however, this is only because since the items are now wrapped in `Result`'s, the value is limited to the scope of the initial function, and when I try to use the main macro `rustc` complains that the value doesn't live long enough.

This also called for the introduction of the `arg_take` function on `Cmd`, which added a function that takes the value by value and returns the value (instead of `&mut Cmd` from `arg`.

## Tests

It has all the same tests as the previous one except with an extra one to test that the `serde_json` errors are being handled correctly (at the top of `test_json.rs`); this attempts to serialise one of the few cases where `serde_json` fails to serialise a struct. (When a HashMap does NOT have a `string-ifiable` key (Strings and Numbers only)

---

I thought I added more but upon review I did not


redis/src/cmd.rs
```diff
@@ -303,6 +303,16 @@ impl Cmd {
...
+    /// Used by the RedisJSON to allow the returning of an error when serialisation fails on some commands
+    /// **This should NOT be used by anything else**
+    #[inline]
+    #[doc(hidden)]
+    #[cfg(feature = "json")]
+    pub(crate) fn arg_take<T: ToRedisArgs>(mut self, arg: T) -> Cmd {
+        arg.write_redis_args(&mut self);
+        self
+    }
+
```
d3rpp, line 315: no other use for this other than this specific use case so this is an internal function hidden from the docs and only allowed in the crate

README.md
```diff
@@ -103,8 +103,43 @@ fn fetch_an_integer() -> String {
 }
 ```
 
+## JSON Support
+
+Support for the RedisJSON Module can be enabled by specifying "json" as a feature in your Cargo.toml.
+
+`redis = { version = "0.17.0", features = ["json"] }`
+
+Then you can simply include the `JSONCommands` which will add the `json` commands to all Redis Connections (not to be confused with just `Commands` which only adds the default commands)
```
djc, line 112: Should update this and the example below to use `JsonCommands`.

redis/src/commands/json.rs
```diff
@@ -0,0 +1,231 @@
...
+	($err: expr) => {
+		Err(
+			RedisError::from((
+				ErrorKind::Serialize,
+				"Serialisation Error",
+				format!("{}", $err)
+			))
+		)
+	}
+}
```
djc, line 24: If you make an `impl From<serde_json::Error> for RedisError` you can just apply `?` to any `serde_json` functions within a method that returns `RedisResult`.

redis/src/commands/json.rs
```diff
@@ -0,0 +1,231 @@
...
+use crate::commands::ClusterPipeline;
+
+use serde::ser::Serialize;
+
+macro_rules! json_serialize_fail {
+	($err: expr) => {
+		Err(
+			RedisError::from((
+				ErrorKind::Serialize,
+				"Serialisation Error",
```
djc, line 19: Nit: please avoid UK spelling and use "serialization"/"serialize" everywhere.

redis/src/commands/json.rs
```diff
@@ -0,0 +1,231 @@
...
+	fn json_get<K: ToRedisArgs, P: ToRedisArgs>(key: K, path: P) {
+		Ok::<Cmd, RedisError>(
+			cmd(if key.is_single_arg() { "JSON.GET" } else { "JSON.MGET" })
+				.arg_take(key)
+				.arg_take(path)
+		)
+	}
+
+	/// Increments the number value stored at `path` by `number`.
+	fn json_numincrby<K: ToRedisArgs, P: ToRedisArgs>(key: K, path: P, value: i64) {
```
djc, line 146: Nit: should be named `json_num_incr_by`. Same for all the other method names, so `objkeys` -> `obj_keys` and `strlen` -> `str_len`.

djc (commented):
Here's an initial round of review, this is looking much better.

redis/src/commands/json.rs
d3rpp: I didn't think of that, but this looks so much better
```rust
Ok::<Cmd, RedisError>(
	cmd("JSON.ARRAPPEND")
		.arg_take(key)
		.arg_take(path)
		.arg_take(serde_json::to_string(value)?)
)
```

d3rpp: Btw the `Ok::<Cmd, RedisError>` is necessary, without it the macro complains about not being able to infer types, which is strange but this is a fairly *clean-ish* workaround

redis/src/commands/json.rs
d3rpp: the word Serialise vs Serialize makes me dislike English even more.

It's my first language and it's just a pain lol

redis/src/commands/json.rs
```diff
@@ -0,0 +1,201 @@
...
+	fn json_arr_len<K: ToRedisArgs, P: ToRedisArgs>(key: K, path: P) {
+		Ok::<Cmd, RedisError>(
+			cmd("JSON.ARRLEN")
+				.arg_take(key)
+				.arg_take(path)
+		)
+	}
+
+	/// Removes and returns an element from the `index` in the array.
+	/// `index` defaults to `-1` (the end of the array)
```
djc, line 70:
```suggestion
	/// Removes and returns an element from the `index` in the array.
	///
	/// `index` defaults to `-1` (the end of the array).
```

redis/src/commands/json.rs
```diff
@@ -0,0 +1,201 @@
...
+		Ok::<Cmd, RedisError>(
+			cmd("JSON.ARRINDEX")
+				.arg_take(key)
+				.arg_take(path)
+				.arg_take(serde_json::to_string(value)?)
+		)
+	}
+
+	/// Inserts the JSON `value` in the array at `path` before the `index` (shifts to the right)
+	/// `index` must be withing the array's range.
```
djc, line 49:
```suggestion
	/// Inserts the JSON `value` in the array at `path` before the `index` (shifts to the right).
	///
	/// `index` must be withing the array's range.
```

redis/src/commands/json.rs
```diff
@@ -0,0 +1,201 @@
...
+		Ok::<Cmd, RedisError>(
+			cmd("JSON.ARRAPPEND")
+				.arg_take(key)
+				.arg_take(path)
+				.arg_take(serde_json::to_string(value)?)
+		)
+	}
+
+	/// Index array at `path`, returns first occurance of `value`
+	/// Pass `None` to `start_stop` to default it to `0, 0` (the redis default in this command)
```
djc, line 38:
```suggestion
	/// Index array at `path`, returns first occurance of `value`
	///
	/// Pass `None` to `start_stop` to default it to `0, 0` (the Redis default in this command).
```

redis/src/cmd.rs
```diff
@@ -303,6 +303,16 @@ impl Cmd {
         self
     }
 
+    /// Used by the RedisJSON to allow the returning of an error when serialization fails on some commands
```
djc, line 306: From this comment it's still not clear to me exactly why we need this?

d3rpp: This is the function I added because `rustc` was complaining about the `&mut Cmd` not living long enough, so instead of returning a mutable reference, it returns the command by value.

I'll update the comment to be more specific about its purpose

redis/src/commands/json.rs
```diff
@@ -0,0 +1,201 @@
...
+			format!("{}", serde_err)
+		))
+	}
+}
+
+implement_json_commands! {
+    'a
+	
+	/// Append the JSON `value` to the array at `path` after the last element in it.
+	fn json_arr_append<K: ToRedisArgs, P: ToRedisArgs, V: Serialize>(key: K, path: P, value: &'a V) {
```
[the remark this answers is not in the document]
d3rpp: due to the formation of the implementation macro, if I don't add any named generics, the macro panics, I could remove it from the macro and replace everything here with `impl ...` though.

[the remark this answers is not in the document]
djc: Okay, never mind then.

redis/src/commands/json.rs
d3rpp: I just realised I forgot to add `start` and `stop` to this command, will fix

README.md
```diff
@@ -103,8 +103,43 @@ fn fetch_an_integer() -> String {
...
+Then you can simply include the `JsonCommands` which will add the `json` commands to all Redis Connections (not to be confused with just `Commands` which only adds the default commands)
+
+```rust
+use redis::Client;
+use redis::JsonCommands;
+use redis::RedisResult;
+
+// Result returns Ok(true) if the value was set
+// Result returns Err(e) if there was an error with the server itself OR serde_json was unable to serialize the boolean
+fn set_json_bool<P: ToString>(key: P, path: P, b: bool) -> RedisResult<bool> {
```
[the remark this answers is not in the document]
d3rpp: You make a very good point, I'll update it to say `ToRedisArgs` since I keep forgetting that in theory the entire Shrek movie as an mp4 file could be a redis key (it's binary safe)

redis/tests/support/mod.rs
```diff
@@ -98,18 +106,38 @@ impl RedisServer {
...
+            },
+            modules,
+        )
     }
 
     pub fn new_with_addr<F: FnOnce(&mut process::Command) -> process::Child>(
         addr: redis::ConnectionAddr,
         tls_paths: Option<TlsFilePaths>,
         spawner: F,
+        modules: &[RedisModule],
```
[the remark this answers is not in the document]
d3rpp: I'll make the change on my end instead of updating it here since I'll need to update some of the invocations of the function to match

d3rpp:
lol whoops

d3rpp:
Still todo: add `start` and `stop` to the `json_arr_index` function

d3rpp:
I will admit it is not the worlds cleanest solution but its cleaner than
```rust
json_arr_index("key", "path", &json!({"hello": "world"}), None, None)

// or

json_arr_index("key", "path", &json!({"hello": "world"}), 0, 0)
```

redis/src/commands/json.rs
```diff
@@ -0,0 +1,218 @@
...
+		Ok::<Cmd, RedisError>(
```
djc, line 29: If you rewrite it like this:

```rust
let mut cmd = cmd("JSON.ARRAPPEND");
cmd.arg(key)
    .arg(path)
    .arg(serde_json::to_string(value)?);
Ok::<_, RedisError>(cmd)
```

I think you no longer need `arg_take()`? (Also note that we don't need to specify the `Cmd` type in the `Ok` turbofish.)

Also while looking this I noticed all the indentation in this file is tabs instead of spaces, should fix that up (VS Code has a `Convert Indentation to Spaces` thing.)

djc (commented):
I don't remember, why did we want to have the JSON commands in separate traits (`JsonCommands`, `AsyncJsonCommands`) anyway?

d3rpp:
> I don't remember, why did we want to have the JSON commands in separate traits (`JsonCommands`, `AsyncJsonCommands`) anyway?

If we don't it's 2 Separate traits with the same name so If you import both of them the compiler complains unless you alias one of them.

djc:
I meant, why can't we make these available as part of the normal command traits (but subject to the Cargo feature flag)?

d3rpp:
> I meant, why can't we make these available as part of the normal command traits (but subject to the Cargo feature flag)?

That's what I did in the First PR (#644)

Screen Shot 2022-08-19 at 10 16 04

which lead to the creation of separate modules, and to import them at the same time, it needs a different name.

redis/src/commands/json.rs
d3rpp: I'll do some investigation into this, I think I tried something like this, but I may be wrong

djc: I tried it for one command and it seemed to work? But maybe I got it wrong...

djc:
Okay, let's keep the additional trait for now.

redis/src/commands/json.rs
d3rpp: > I tried it for one command and it seemed to work? But maybe I got it wrong...

sorry, I've been busy with a hackathon over the weekend, will look into this asap

djc: No hurry from my side! Hope you had fun at the hackathon.

d3rpp: > If you rewrite it like this:
> 
> ```rust
> let mut cmd = cmd("JSON.ARRAPPEND");
> cmd.arg(key)
>     .arg(path)
>     .arg(serde_json::to_string(value)?);
> Ok::<_, RedisError>(cmd)
> ```
> 
> I think you no longer need `arg_take()`? (Also note that we don't need to specify the `Cmd` type in the `Ok` turbofish.)
> 
> Also while looking this I noticed all the indentation in this file is tabs instead of spaces, should fix that up (VS Code has a `Convert Indentation to Spaces` thing.)

```rust
let mut cmd = cmd("JSON.ARRAPPEND");

cmd.arg(key)
	.arg(path)
	.arg(serde_json::to_string(value)?);

Ok::<_, RedisError>(cmd)
```

Does indeed seem to work, ill update the rest of them

I've run the command to use spaces instead of tabs

redis/src/commands/json.rs
```diff
@@ -0,0 +1,237 @@
...
+
+impl From<serde_json::Error> for RedisError {
+    fn from(serde_err: serde_json::Error) -> RedisError {
+        RedisError::from((
+            ErrorKind::Serialize,
+            "Serialization Error",
+            format!("{}", serde_err)
+        ))
+    }
+}
```
djc, line 22: Let's move this below the `RedisError` definition, guarded with whatever feature flags are necessary.

redis/src/commands/json_macro.rs
```diff
@@ -0,0 +1,145 @@
+macro_rules! implement_json_commands {
```
djc, line 1: Let's move this into the `json` module, somewhere near the bottom.

redis/tests/support/mod.rs
```diff
@@ -45,6 +45,10 @@ enum ServerType {
     Unix,
 }
 
+pub enum RedisModule {
```
djc, line 48: Naming nit: I think the `Redis` in both `RedisJson` and `RedisModule` is superfluous, so let's ditch that?

djc (commented):
Some spit and polish, but I think this is almost ready to merge!

redis/src/commands/json_macro.rs
d3rpp: > Let's move this into the `json` module, somewhere near the bottom.

I can put it at the top, but it complains if I put it at the bottom since it's *technically* being declared after use

djc: Ah, fair. At the top it is!

djc (approved):
Alright, nice work! @jaymell did you want to take another look at this?

jaymell:
It looks good to me, though @d3rpp I'm going to request that if possible you squash this down to one commit and give it a nice commit message to save me a bit of work merging :grin:

d3rpp:
> It looks good to me, though @d3rpp I'm going to request that if possible you squash this down to one commit and give it a nice commit message to save me a bit of work merging 😁 (and of course resolve conflicts)

oooooh last time I tried that it was absolutely painful, is there not a squash checkbox next to the merge button?

d3rpp:
It'll take me a hot minute but I'll get it done

d3rpp:
image

umm, idk
is there a squash button in the drop-down next to the merge button?

djc:
Yeah, I just squashed it at merge. @d3rpp thanks for persisting in getting this merged!

PR #660, opened 2022-08-14
Add support to casting RedisResult to CString.

nihohit:


nihohit:
The linter seems to fail on code that isn't a part of this PR.

djc:
#663 cleans up the unrelated clippy failures.

djc:
This seems reasonable to me. @jaymell what do you think?

redis/src/types.rs
```diff
@@ -1147,13 +1160,24 @@ impl FromRedisValue for bool {
...
+}
+
 impl FromRedisValue for String {
     fn from_redis_value(v: &Value) -> RedisResult<String> {
         match *v {
             Value::Data(ref bytes) => Ok(from_utf8(bytes)?.to_string()),
             Value::Okay => Ok("OK".to_string()),
             Value::Status(ref val) => Ok(val.to_string()),
-            _ => invalid_type_error!(v, "Response type not string compatible."),
+            _ => invalid_type_error!(v, "Response type not CString compatible."),
```
djc, line 1180: This is the wrong way around, you have `CString` in the error message for the `String` impl and `string` in the `CString` impl.

nihohit: 🤦
Fixed

jaymell:
Looks good. Should we add a couple simple tests?

nihohit:
@jaymell I'll be happy to add some, but I don't see equivalent `String` tests to copy from.
What tests would you like me to add?

jaymell:
> @jaymell I'll be happy to add some, but I don't see equivalent `String` tests to copy from. What tests would you like me to add?

Nothing special, there are some similar examples if you look at the [other type tests](https://github.com/redis-rs/redis-rs/blob/main/redis/tests/test_types.rs), e.g., make sure that a null byte in a value returns the error we expect.

nihohit:
@jaymell added tests.

djc (approved):

djc:
Thanks!

Issue #631, opened 2022-06-09
All key-values from InfoDict with typing


abrunner94:
👋   I have a use case for retrieving all key-values via the `INFO` command. However, currently there doesn't seem to be a straightforward way of doing that unless you know which key you're looking for. I am also looking to convert the values to their respective types as opposed to having all values as strings.

PR #661, opened 2022-08-14
add derive Clone and as_map method to InfoDict

danni-m:


danni-m:
This would also help to fix #631

djc:
I'm fine with the `Clone`. I'm a little on the fence about exposing the implementation detail that we're using a `HashMap` here. What about instead implementing an iterator over (key, value) items?

(Clippy failures can be ignored, see #663.)

danni-m:
@djc first of all, it is possible, and I can do it to move with this PR (Should we implement a new `Iterator` or just expose a slice?). 
My use case is more specific for orchestrating Redis instances where we need a lot more than `get` for Redis info. We preferred not to implement our own parser and use the one provided by the library, but at some point, we were limited by the API. 
My reasoning for exposing Hashmap:
1. Hashmap already provides a lot of useful APIs (like `get`, `len` and `keys` for example)
2. We do use hashmap internally. There's not a lot of value (IMHO) in recreating the same APIs that hashmap already supports.

djc (approved):

djc (commented):
Actually, I think we should probably implement `Deref<Target = &HashMap<String, Value>>` instead of adding `as_map()`. That would also allow us to get rid of some of the simple proxy methods.

danni-m:
@djc removed as_map in favour of Deref

djc:
Thanks! @jaymell want to take a look?

jaymell:
Looks good!

PR #662, opened 2022-08-14
Support setting custom params on TlsConnectorBuilder

glebpom:


djc:
What's the use case here? What kind of custom parameters are you trying to set?

I think adding an argument to a public function based on a Cargo feature violates the additivity constraint for Cargo features, so I don't think the API design works well here.

nihohit:

PR #663, opened 2022-08-15
Fix clippy suggestions

djc:


redis/src/geo.rs
```diff
@@ -52,7 +52,7 @@ impl ToRedisArgs for Unit {
 ///
 /// * You may want to use either `f64` or `f32` if you want to perform mathematical operations.
 /// * To keep the raw value from Redis, use `String`.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
```
nihohit, line 55: IIRC this will prevent you from using floats as `T`, since floats don't derive `Eq`. If this is intentional, you might want to fix the documentation to not reference floats.

djc: Good call, fixed that up. Thanks for the review!

jaymell (approved):

nihohit:
@djc @jaymell BTW, I noticed that you're running clippy without the `--all-features` flag, which means that all of your code which is under a feature flag isn't checked. Is this something you want fixed?

djc:
@nihohit yes, that would be nice to fix! Thanks!

nihohit:
@djc done
https://github.com/redis-rs/redis-rs/pull/667

PR #664, opened 2022-08-20
impl Debug for MultiplexedConnection & Pipeline

### LINKED ISSUE #649: Debug impl for `MultiplexedConnection` & `Pipeline`
Hello,
First of all, thanks for the amazing work on the `redis` crate!
I'm trying to use the `tracing` crate for an async application with tokio & tokio-console, however, I'm unable to use the `#[instrument]` attribute on async function as there is no Debug impl for `MultiplexedConnection` & `Pipeline`

Would you consider such addition to the code?
`aio::Pipeline` seems the most difficult part to implement. What information and other considerations should be taken into account for implementing Debug for Pipeline?

elpiel:
Resolves #649 
This allows the usage of `#[instrument]` when using the `tracing` crate.
All generics of Pipeline used in MultiplexedConnection impl Debug this is why I've added an impl of Pipeline only when all of them also impl Debug:

```rust
pub struct MultiplexedConnection {
    pipeline: Pipeline<Vec<u8>, Value, RedisError>,
    db: i64,
}
```

cc @djc

### DIFF
--- a/redis/src/aio.rs
+++ b/redis/src/aio.rs
@@ -1,6 +1,8 @@
 //! Adds experimental async IO support to redis.
 use async_trait::async_trait;
 use std::collections::VecDeque;
+use std::fmt;
+use std::fmt::Debug;
 use std::io;
 use std::mem;
 use std::net::SocketAddr;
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

elpiel:
Whoop whoop, Thank you as well!

PR #669, opened 2022-08-23
[1/5] cluster: implement proper builder pattern & minor refactor

utkarshgupta137:
Most of the crates in the Rust ecosystem use `builder()` & `build()` for client & builder as well as `new()` for both of them. Currently, the redis crate uses the unintuitive `open()` methods for the same.
With this PR, I've changed the cluster client to follow the proper builder pattern and redirect the old methods to the new ones.

Also refactored the cluster client in 2 main ways:
- Moved the `build()` method from the client to the builder & made it public
- Create a crate-level struct `ClusterParams` to hold all the params inside a single struct instead of multiple fields in the main struct
  - This struct will be used in the later PRs to pass cluster params to & inside the ClusterConnection impl

In addition, improved the documentation as I saw fit.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -12,7 +12,7 @@
 //! use redis::cluster::ClusterClient;
 //!
 //! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/", "redis://127.0.0.1:6377/"];
-//! let client = ClusterClient::open(nodes).unwrap();
+//! let client = ClusterClient::new(nodes).unwrap();
 //! let mut connection = client.get_connection().unwrap();
 //!
 //! let _: () = connection.set("test", "test_data").unwrap();
@@ -27,7 +27,7 @@
 //! use redis::cluster::{cluster_pipe, ClusterClient};
 //!
 //! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/", "redis://127.0.0.1:6377/"];
-//! let client = ClusterClient::open(nodes).unwrap();
+//! let client = ClusterClient::new(nodes).unwrap();
 //! let mut connection = client.get_connection().unwrap();
 //!
 //! let key = "test";
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -55,6 +55,7 @@ use super::{
     Cmd, Connection, ConnectionAddr, ConnectionInfo, ConnectionLike, ErrorKind, IntoConnectionInfo,
     RedisError, RedisResult, Value,
 };
+use crate::cluster_client::ClusterParams;
 
 pub use crate::cluster_client::{ClusterClient, ClusterClientBuilder};
 use crate::cluster_pipeline::UNROUTABLE_ERROR;
@@ -95,25 +96,23 @@ impl TlsMode {
 
 impl ClusterConnection {
     pub(crate) fn new(
+        cluster_params: ClusterParams,
         initial_nodes: Vec<ConnectionInfo>,
-        read_from_replicas: bool,
-        username: Option<String>,
-        password: Option<String>,
     ) -> RedisResult<ClusterConnection> {
         let connections = Self::create_initial_connections(
             &initial_nodes,
-            read_from_replicas,
-            username.clone(),
-            password.clone(),
+            cluster_params.read_from_replicas,
+            cluster_params.username.clone(),
+            cluster_params.password.clone(),
         )?;
 
         let connection = ClusterConnection {
             connections: RefCell::new(connections),
             slots: RefCell::new(SlotMap::new()),
             auto_reconnect: RefCell::new(true),
-            read_from_replicas,
-            username,
-            password,
+            read_from_replicas: cluster_params.read_from_replicas,
+            username: cluster_params.username,
+            password: cluster_params.password,
             read_timeout: RefCell::new(None),
             write_timeout: RefCell::new(None),
             #[cfg(feature = "tls")]
@@ -135,7 +134,7 @@ impl ClusterConnection {
             },
             #[cfg(not(feature = "tls"))]
             tls: None,
-            initial_nodes,
+            initial_nodes: initial_nodes.to_vec(),
         };
         connection.refresh_slots()?;
 
--- a/redis/src/cluster_client.rs
+++ b/redis/src/cluster_client.rs
@@ -4,7 +4,7 @@ use super::{
     ConnectionAddr, ConnectionInfo, ErrorKind, IntoConnectionInfo, RedisError, RedisResult,
 };
 
-/// Used to configure and build a [ClusterClient](ClusterClient).
+/// Used to configure and build a [`ClusterClient`].
 pub struct ClusterClientBuilder {
     initial_nodes: RedisResult<Vec<ConnectionInfo>>,
     read_from_replicas: bool,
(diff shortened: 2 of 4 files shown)


redis/src/cluster_client.rs
```diff
@@ -1,164 +1,186 @@
...
-    pub fn open(self) -> RedisResult<ClusterClient> {
-        ClusterClient::build(self)
+    /// Upon failure to parse initial nodes or if the initial nodes have different passwords or
+    /// usernames, an error is returned.
+    pub fn build(self) -> RedisResult<ClusterClient> {
+        let mut initial_nodes = self.initial_nodes?;
+
+        let mut cluster_params = self.cluster_params;
+        let password = if cluster_params.password.is_none() {
+            cluster_params.password = initial_nodes[0].redis.password.clone();
```
jaymell, line 50: This will cause a panic if `initial_nodes` is empty for some reason.

redis/src/cluster.rs
```diff
@@ -95,25 +96,23 @@ impl TlsMode {
 
 impl ClusterConnection {
     pub(crate) fn new(
-        initial_nodes: Vec<ConnectionInfo>,
-        read_from_replicas: bool,
-        username: Option<String>,
-        password: Option<String>,
+        cluster_params: &ClusterParams,
```
jaymell, line 99: I'm not sure there's a reason to use references for the params here, given that you're just cloning anyway. I would let the caller do the cloning if needed.

utkarshgupta137: I prefer to pass references as much as possible so that if I don't need a clone later on, I don't have to modify the whole call chain. Also if I only need a part of the struct, I don't have to clone the whole struct.
This is a matter of personal opinion. If you prefer, I can modify the code to pass cloned struct.

jaymell (approved):

redis/src/cluster_client.rs
```diff
@@ -44,6 +44,12 @@ impl ClusterClientBuilder {
     /// usernames, an error is returned.
     pub fn build(self) -> RedisResult<ClusterClient> {
         let mut initial_nodes = self.initial_nodes?;
+        if initial_nodes.is_empty() {
```
djc, line 47: Instead of doing it this way, I'd prefer to anchor this in the type system, something like this:

```rust
let first_node = match initial_nodes.first() {
    Some(node) => node,
    None => return Err(..),
};
```

then use `first_node` instead of `initial_nodes[0]` below.

redis/src/cluster_client.rs
```diff
@@ -41,69 +43,72 @@ impl ClusterClientBuilder {
...
+        } else {
+            &None
+        };
 
-        for (index, info) in initial_nodes.into_iter().enumerate() {
-            if let ConnectionAddr::Unix(_) = info.addr {
-                return Err(RedisError::from((ErrorKind::InvalidClientConfig,
-                                         "This library cannot use unix socket because Redis's cluster command returns only cluster's IP and port.")));
-            }
+        initial_nodes = initial_nodes
```
djc, line 68: What's the impetus here for moving from a `for`-loop to iterator combinators? IMO in this case it doesn't much improve the code (and this is another change that is unrelated to the stated goal of the commit) -- in fact, it looks like it adds another level of indentation which makes the code harder to read.

utkarshgupta137: I initially thought that this function is called each time we need a new connection, so I changed it to an iterator pattern which would be (slightly) faster.

utkarshgupta137:
@jaymell @djc It's failing on older versions due to an unrelated crate error:
```
error: package `async-global-executor v2.3.0` cannot be built because it requires rustc 1.59 or newer, while the currently active rustc version is 1.57.0
```

redis/src/cluster_client.rs
```diff
@@ -104,6 +104,12 @@ impl ClusterClientBuilder {
         self
     }
 
+    /// Use `build()`.
```
djc, line 122: This should be in the same place as the previous `open()` method so the diff is more clear.

redis/src/cluster_client.rs
```diff
@@ -147,11 +159,17 @@ impl ClusterClient {
             self.password.clone(),
         )
     }
+
+    /// Use `new()`.
+    #[deprecated(since = "0.22.0", note = "Use new()")]
+    pub fn open<T: IntoConnectionInfo>(initial_nodes: Vec<T>) -> RedisResult<ClusterClient> {
```
djc, line 174: Please keep this in the same place as the previous `open()` method at least for this commit, so it's obvious what changed.

redis/src/cluster.rs
```diff
@@ -96,11 +96,11 @@ impl TlsMode {
 
 impl ClusterConnection {
     pub(crate) fn new(
-        cluster_params: &ClusterParams,
-        initial_nodes: &[ConnectionInfo],
+        cluster_params: ClusterParams,
```
djc, line 99: These changes should be part of the "use ClusterParams struct to pass params" commit.

redis/src/cluster_client.rs
djc: You apparently applied these changes in some other place in an earlier commit, but not here.

djc:
This is much better! A few more minor changes. Please add a separate commit that bumps the MSRV to 1.59.

redis/src/cluster_client.rs
djc: How/why do you think it would be faster? I don't think that's true.

redis/src/cluster_client.rs
utkarshgupta137: I've deliberately moved deprecated methods to the bottom of the struct to keep them out of the way.

redis/src/cluster_client.rs
utkarshgupta137: I've deliberately moved deprecated methods to the bottom of the struct to keep them out of the way.

redis/src/cluster_client.rs
utkarshgupta137: I'm relatively new to rust, but from what I've read, iterators can be parallelised better than for-loops & also we don't have to check & increase the capacity of the vector that we collect into. It is a very slight improvement.

Ref: https://users.rust-lang.org/t/we-all-know-iter-is-faster-than-loop-but-why/51486

utkarshgupta137:
> This is much better! A few more minor changes. Please add a separate commit that bumps the MSRV to 1.59.

MSRV change is entirely unrelated to this PR. I've created a separate PR: https://github.com/redis-rs/redis-rs/pull/673

redis/src/cluster_client.rs
djc: I'd prefer that that is done in a separate commit to faciliate review.

redis/src/cluster_client.rs
djc: I'd prefer that that is done in a separate commit to faciliate review.

utkarshgupta137:
@djc I've added the bump MSRV commit to this PR for testing purposes, but please merge #673 first.

redis/src/cluster_client.rs
djc: Okay, please don't resolve my review comments unless you've done as I suggested.

There will be no meaningful difference in performance here, especially if you allocate the new `Vec` by using `with_capacity()` up front. The `for`-loop is more readable in this case.

redis/src/cluster_client.rs
```diff
@@ -39,68 +41,69 @@ impl ClusterClientBuilder {
     /// usernames, an error is returned.
     pub fn build(self) -> RedisResult<ClusterClient> {
         let initial_nodes = self.initial_nodes?;
-
         let mut nodes = Vec::with_capacity(initial_nodes.len());
```
djc, line 44: Let's move this to directly before the loop.

redis/src/cluster_client.rs
djc: So this invariant is not checked here. Is it checked somewhere else? If yes, then we should add a comment pointing that out. If not, should check it here.

djc: Ok.

djc (approved):

djc:
Okay, so please rebase and self-review part 2 to have small commits and align with the other feedback I've given so far, this has taken a lot of time.

PR #675, opened 2022-09-02
Make Direction a public enum to use with Commands like `blmove` (#646)

rudolphfroger-ds:
Cherry-picked from the main branch


djc:
Looks like CI is not running for this branch. Will need to merge #676 and then push this branch again.

djc:
Can you push again trigger CI?

jaymell:
I just realized the workflow changes weren't actually brought into this branch :neutral_face:

djc:
> I just realized the workflow changes weren't actually brought into this branch :neutral_face: 

I was under the impression they wouldn't need to be? But might be wrong about that.

Issue #470, opened 2021-03-24
Support ipv6 link-local addresses


socs:
URL parsing fails on ipv6 link-local addresses, eg

```
fn main() -> Result<(), Box<dyn std::error::Error>> {
    let _ = redis::Client::open("[fe80::1%foo]:6379")?; // -> Error: Redis URL did not parse
    Ok(())
}
```

This works with redis-cli
```
# redis-cli -h fe80::1%foo -p 6379
[fe80::1%foo]:6379> ping
PONG
```

The limitation stems from rust-url library which follows a url spec that [omits zone identifier](https://url.spec.whatwg.org/#host-representation) on ipv6 ll addresses.
Perhaps a custom parser could be used by redis-rs, such as the one in redis-cli, so that ipv6 ll addresses is accepted.

djc:
Should be fixed with https://github.com/redis-rs/redis-rs/pull/679.

Issue #332, opened 2020-06-05
support ipv6 connect


sallyyu0:
There are places that connect() still uses string url stored in slots(ex: "redis://fd00:10:0:5:5861:50ff:fe00:85:6379") instead of ConnectionInfo. Please update those to use ConnectionInfo instead to avoid parse error.  Below is an example for converting string to ConnectionInfo. 

```
fn get_conn_info(redis_url: &str) -> Option<ConnectionInfo> {
    let v: Vec<&str> = redis_url.split("://").collect();
    if v.len() == 2 {
        let h: Vec<&str> = v[1].split(":").collect();
        let port: u16 = h[h.len() - 1].parse().unwrap();
        let (host, _) = v[1].split_at(v[1].len() - port.to_string().len() - 1);
        return Some(ConnectionInfo {
            addr: Box::new(ConnectionAddr::Tcp(host.to_string(), port)),
            db: 0,
            passwd: None,
        });
    } else {
        return None;
    }
}

```

sallyyu0:
Maybe it's better to fix format "redis://{}:{}" to "redis://[{}]:{}" for IPv6 address as root cause.

sallyyu0:
```
diff --git a/src/cluster.rs b/src/cluster.rs
index 0d74feb..5b67384 100644
--- a/src/cluster.rs
+++ b/src/cluster.rs
@@ -41,6 +41,8 @@
 use std::cell::RefCell;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::iter::Iterator;
+use std::net::Ipv6Addr;
+use std::str::FromStr;
 use std::thread;
 use std::time::Duration;
 
@@ -262,7 +264,10 @@ impl ClusterConnection {
 
         for info in initial_nodes.iter() {
             let addr = match *info.addr {
-                ConnectionAddr::Tcp(ref host, port) => format!("redis://{}:{}", host, port),
+                ConnectionAddr::Tcp(ref host, port) => match Ipv6Addr::from_str(host) {
+                    Ok(_) => format!("redis://[{}]:{}", host, port),
+                    Err(_) => format!("redis://{}:{}", host, port),
+                },
                 _ => panic!("No reach."),
             };
 
@@ -314,12 +319,11 @@ impl ClusterConnection {
                             continue;
                         }
                     }
-
-                    if let Ok(mut conn) =
-                        connect(addr.as_ref(), self.readonly, self.password.clone())
-                    {
[...]

Issue #293, opened 2020-03-05
Connect to IPv6 with interface fails


nils-van-zuijlen:
I'm trying to connect to an IPv6 redis server the same way as with the `redis-cli` command below:

```
$ redis-cli -h fe80::cafe:beef%eno1
[fe80::cafe:beef%eno1]:6379> 
```

I'm trying to do it with the rust code
```rs
let client = redis::Client::open("redis://fe80::cafe:beef%eno1/").expect("Can't connect to redis");
```
But when I run it, I get the following error:
```
thread 'main' panicked at 'Can't connect to redis: Redis URL did not parse', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I also tried this alternative syntax, but I got the same result.
```rs
let client = redis::Client::open("redis://fe80::cafe:beef%eno1/").expect("Can't connect to redis");
```

badboy:
In general IPv6 addresses in URLs need to be wrapped in `[` and `]` to be parsed correctly. Looks like the `%eno1` suffix however still won't work. I'd need to check how and if that's valid in URLs.

Can you try and build a [`ConnectionInfo`](https://docs.rs/redis/0.15.1/redis/struct.ConnectionInfo.html) directly?

badboy:
With [this patch](https://github.com/mitsuhiko/redis-rs/commit/411601c64c0d1a93db8b8ac0a5d1d04d37ecd4b8) it could be possible to do it like this:

```rust
redis::Client::open(("fe80::cafe:beef%eno1", 6379))
```

sallyyu0:
I used ConnectionInfo with ipv6 address wrapped in [] but failed with following error.
```
234:failed to get connection for cluster ConnectionInfo { addr: Tcp("[fd00:10:0:5:5861:50ff:fe00:85]", 6379), db: 0, passwd: None } : failed to lookup address information: Name or service not known
234:failed to get connection for cluster ConnectionInfo { addr: Tcp("[fd00:10:0:5:5861:50ff:fe00:a2]", 6379), db: 0, passwd: None } : failed to lookup address information: Name or service not known
234:failed to get connection for cluster ConnectionInfo { addr: Tcp("[fd00:10:0:4:5861:50ff:fe00:8e]", 6379), db: 0, passwd: None } : failed to lookup address information: Name or service not known
[...]

Issue #290, opened 2020-02-21
connect to ipv6 server failed because of SocketAddrs invalid


Right2Left:
https://github.com/mitsuhiko/redis-rs/blob/462e41af9a05ba515eeb208b87ceace1c2017758/src/connection.rs#L212
when the redis server listen on a ipv6 address, i write the client demo code to connect like this:
```rust
    let redis_url = "redis://[cdcd:910a:2222:5498:8475:1111:3900:2020]:6379/";
    let client = match redis::Client::open(redis_url ) {
        Ok(client) => client,
        Err(e) => {
            panic!("Redis open error.");
        }
    };
    let connection = match client .get_connection(){
        Ok(con) => con,
        Err(e) => {
            panic!("Redis get connection error.");
        }
    };
```
i found it will panic at "Redis get connection error."
To view the code in connection.rs#L212, maybe it's not support a ipv6 address, i am new to rust, had not good idea, just write a demo function(terrible code) to convert the address to SocketAddrV6 then connect, like this:
```rust
fn get_ipv6_add_from_string(ipv6_str: &str) -> Ipv6Addr {
    let format_str = ipv6_str
        .replace('[', "")
        .replace(']', "")
        .replace("::", ":0:");
    let str_vec: Vec<&str> = format_str.split(':').collect();
    let mut v6_vec: Vec<u16> = Vec::new();
    for iter in str_vec.iter() {
        let value = match u16::from_str_radix(iter.clone(), 16) {
            Ok(v) => v,
            Err(e) => {
                println!("Can not parse ipv6 to u16, err:{:?}.", e);
                0u16
            }
        };
        v6_vec.push(value)
    }
    if v6_vec.len() != 8 {
        return Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1);
    }
    let addr6 = Ipv6Addr::new(
        v6_vec[0], v6_vec[1], v6_vec[2], v6_vec[3], v6_vec[4], v6_vec[5], v6_vec[6], v6_vec[7],
    );
    addr6
}


let addr_v6 = get_ipv6_add_from_string(host);
let sock_addr_v6 = SocketAddrV6::new(addr_v6, *port, 0, 0);

let tcp = TcpStream::connect(sock_addr_v6 )?
```
or to use to_socket_addrs method, but also need convert  string, just like:
```rust
let host_v6 = host .replace('[', "").replace(']', "");
let socket_addr = match (&host_v6 [..], port).to_socket_addrs() {
[...]

PR #679, opened 2022-09-09
Support ipv6 connect for #290 #293 #332 #470

buaazp:
Connect to ipv6 addr failed because url.host().to_string() got a `[fdbd:dc02:ff:1:1:174:229:45]` String.

Which cannot parse to IPv6Addr: 
https://doc.rust-lang.org/src/std/net/addr.rs.html#963-980

Just use IPv6Addr's to_string() aka Display trait can fix it.
https://doc.rust-lang.org/src/std/net/ip.rs.html#1758-1850

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -167,7 +167,24 @@ impl IntoConnectionInfo for String {
 
 fn url_to_tcp_connection_info(url: url::Url) -> RedisResult<ConnectionInfo> {
     let host = match url.host() {
-        Some(host) => host.to_string(),
+        Some(host) => {
+            // Here we manually match host's enum arms and call their to_string().
+            // Because url.host().to_string() will add `[` and `]` for ipv6:
+            // https://docs.rs/url/latest/src/url/host.rs.html#170
+            // And these brackets will break host.parse::<Ipv6Addr>() when
+            // `client.open()` - `ActualConnection::new()` - `addr.to_socket_addrs()`:
+            // https://doc.rust-lang.org/src/std/net/addr.rs.html#963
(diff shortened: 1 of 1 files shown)


redis/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "redis"
-version = "0.21.5"
+version = "0.21.6"
```
djc, line 3: This shouldn't be a part of your PR, probably some rebase/merge issue?

redis/src/connection.rs
```diff
@@ -167,7 +167,11 @@ impl IntoConnectionInfo for String {
 
 fn url_to_tcp_connection_info(url: url::Url) -> RedisResult<ConnectionInfo> {
     let host = match url.host() {
-        Some(host) => host.to_string(),
+        Some(host) => match host {
+            url::Host::Domain(path) => path.to_string(),
+            url::Host::Ipv4(v4) => v4.to_string(),
+            url::Host::Ipv6(v6) => v6.to_string(),
+        },
```
djc, line 174: Maybe add a comment here on why we can't use `host.to_string()` directly, to preserve your research?

djc (approved):
Nice, thanks!

redis/Cargo.toml
buaazp: Got it. Removed.

redis/src/connection.rs
buaazp: Sure, this issue is quite convoluted and should indeed be explained with enough comments.

buaazp:
Updated follow the review comments. And added some ipv6 test cases to make sure this change is OK.

djc (approved):

jaymell (approved):

Issue #682, opened 2022-09-16
AclInfo: Expect a valid ACL flag


garyhai:
Implementation of `FromRedisValue` for `AclInfo` misses "allchannels" flag and therefore fail to call `conn.acl_getuser` with `not_convertible_error`

```rust
use redis::{acl::AclInfo, Client, Commands};

const CONN_INFO: &str = "redis://127.0.0.1";

fn main() {
  let client = Client::open(CONN_INFO).unwrap();
  let mut conn = client.get_connection().unwrap();
  let info: AclInfo = match conn.acl_getuser("default") {
    Err(e) => {
      dbg!(e);
      return;
    }
    Ok(info) => info,
  };
  dbg!(info);
}
```
output:
```sh
[src/main.rs:10] e = Response type not convertible: "Expect a valid ACL flag" (response was [97, 108, 108, 99, 104, 97, 110, 110, 101, 108, 115])
```

To fix the issue, in acl.rs:
```rust
let flags = flags
                    .as_sequence()
                    .ok_or_else(|| {
                        not_convertible_error!(flags, "Expect a bulk response of ACL flags")
                    })?
                    .iter()
                    .map(|flag| match flag {
                        Value::Data(flag) => match flag.as_slice() {
                            b"on" => Ok(Rule::On),
                            b"off" => Ok(Rule::Off),
                            b"allkeys" => Ok(Rule::AllKeys),
                            b"allcommands" => Ok(Rule::AllCommands),
                            b"nopass" => Ok(Rule::NoPass),
                            b"allchannels" => Ok(Rule::Pattern("&*".into()),
                            _ => Err(not_convertible_error!(flag, "Expect a valid ACL flag")),
                        },
```

djc:
Thanks for the report! Would you mind submitting a PR for this?

garyhai:
I'd be happy to submit a PR for the bug. But a more complete change involves structural changes to the Rule, causing compatibility issues. And there are more flags to add with latest version of redis. It is best to be carefully designed by professional team.

djc:
I guess as the volunteer maintainer of this crate I would qualify as professional, but I'm also very busy with maintaining a number of other crates in addition to working at my day job. So this is unlikely to get done unless you (or someone else motivated to work on this) can make some progress on it.

garyhai:
Thank you for your selfless contribution to this project. Let me try to fix this bug without breaking compatibility.

PR #685, opened 2022-09-20
Add Rule::Other to cover newly defined flags. #682

garyhai:
Fix bug of unhandled newly defined flags when call `acl_getuser`.

### DIFF
--- a/redis/src/acl.rs
+++ b/redis/src/acl.rs
@@ -101,7 +101,7 @@ impl ToRedisArgs for Rule {
 
             Reset => out.write_arg(b"reset"),
 
-            Other(rule) => out.write_arg_fmt(format_args!("{}", rule)),
+            Other(rule) => out.write_arg(rule.as_bytes()),
         };
     }
 }
@@ -169,7 +169,7 @@ impl FromRedisValue for AclInfo {
                             b"allkeys" => Ok(Rule::AllKeys),
                             b"allcommands" => Ok(Rule::AllCommands),
                             b"nopass" => Ok(Rule::NoPass),
-                            other => Ok(Rule::Other(String::from_utf8_lossy(other).to_string())),
+                            other => Ok(Rule::Other(String::from_utf8_lossy(other).into_owned())),
                         },
                         _ => Err(not_convertible_error!(
                             flag,
--- a/redis/src/acl.rs
+++ b/redis/src/acl.rs
@@ -64,8 +64,9 @@ pub enum Rule {
     /// The user returns to the same state it has immediately after its creation.
     Reset,
 
-    /// Raw text of ACL rule that not enumerated above.
-    /// Ref: https://redis.io/docs/manual/security/acl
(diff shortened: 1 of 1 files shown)


djc:
Thanks for helping out!

This technically still breaks compatibility, since it adds a variant on an enum that is not marked as non-exhaustive. That's fine, though, since the main branch already broke compatibility. Given that this is the case, I think it would make sense to also add variants for any newly defined flags that we know about at this point.

garyhai:
> This technically still breaks compatibility, since it adds a variant on an enum that is not marked as non-exhaustive. That's fine, though, since the main branch already broke compatibility. Given that this is the case, I think it would make sense to also add variants for any newly defined flags that we know about at this point.

Instead of causing compatibility issues every time a new flag is added, Rule::Other is used as a transitional solution to ease compatibility. Rule may enumerate explicitly a list of commonly used flags.

redis/src/acl.rs
```diff
@@ -95,6 +100,8 @@ impl ToRedisArgs for Rule {
             ResetKeys => out.write_arg(b"resetkeys"),
 
             Reset => out.write_arg(b"reset"),
+
+            Other(rule) => out.write_arg_fmt(format_args!("{}", rule)),
```
djc, line 104: Instead of this, please use `out.write_arg(rule.as_bytes())`

redis/src/acl.rs
```diff
@@ -162,7 +169,7 @@ impl FromRedisValue for AclInfo {
                             b"allkeys" => Ok(Rule::AllKeys),
                             b"allcommands" => Ok(Rule::AllCommands),
                             b"nopass" => Ok(Rule::NoPass),
-                            _ => Err(not_convertible_error!(flag, "Expect a valid ACL flag")),
+                            other => Ok(Rule::Other(String::from_utf8_lossy(other).to_string())),
```
djc, line 172: Please use `into_owned()` instead of `to_string()` here.

garyhai: Thanks! This is what I call professionalism.

djc (approved):
This looks good to me. @jaymell any thoughts?

jaymell:
I can be persuaded here, but I'm not thrilled with the change. The core issue, of course, is the existing API has not left us with great options. But I wonder if, given that we're going to make a backward-incompatible change here, if we should consider marking the enum as non-exhaustive and adding in the new variants as appropriate?

garyhai:
> I can be persuaded here, but I'm not thrilled with the change. The core issue, of course, is the existing API has not left us with great options. But I wonder if, given that we're going to make a backward-incompatible change here, if we should consider marking the enum as non-exhaustive and adding in the new variants as appropriate?

The non-exhaustion structure has too many limitations and imposes too much mental burden. Rule::Other is just for exhaustion and ergonomics. I suggest putting this break change into the next non-compatibility release.

jaymell:
Ok, merging this. Thanks for the contribution.

PR #601, opened 2022-04-27
[Feat]: Implement support for rustls

LeoRowan:
This PR will add support for tls connections using rustls under the "rustls" feature flag. It also adds support for tokio rustls under the "tokio-rustls-comp" flag.

Any feedback is greatly appreciated!
Resolves #574.

PR #686, opened 2022-09-22
feat(tls): support passing ca cert for tls handshake

jgramoll:


redis/examples/tls.rs
```diff
@@ -0,0 +1,44 @@
+use redis::{Commands, ConnectionAddr, ConnectionInfo, RedisConnectionInfo};
```
djc, line 1: I'd prefer not to add yet another example for this.

redis/src/aio/tokio.rs
```diff
@@ -108,7 +109,12 @@ impl RedisRuntime for Tokio {
                 .use_sni(false)
                 .build()?
         } else {
-            TlsConnector::new()?
+            let mut builder = TlsConnector::builder();
+            if let Some(ca_cert) = ca_cert {
```
djc, line 113: In order to make this work with rustls as well (#601), I think it would be better to take `&[u8]` here and use `Certificate::from_der()`.

redis/src/aio.rs
```diff
@@ -64,6 +64,7 @@ pub(crate) trait RedisRuntime: AsyncStream + Send + Sync + Sized + 'static {
         hostname: &str,
         socket_addr: SocketAddr,
         insecure: bool,
+        ca_cert: &Option<String>,
```
djc, line 67: This should take an `Option<&[u8]>` (`&` on the inside) instead.

djc (commented):
Thanks!

redis/src/aio.rs
jgramoll: upated

redis/src/aio/tokio.rs
jgramoll: updated

redis/examples/tls.rs
jgramoll: Removed. Is there some documentation you would like me to update?

djc: Yeah, whatever the public entry point is (which I don't quite recall) would probably be good to mention what this is, what it expects to get, that sort of thing.

redis/src/aio/async_std.rs
```diff
@@ -162,7 +163,12 @@ impl RedisRuntime for AsyncStd {
                 .danger_accept_invalid_hostnames(true)
                 .use_sni(false)
         } else {
-            TlsConnector::new()
+            let mut tls_connector = TlsConnector::new();
+            if let Some(ca_cert) = ca_cert {
+                let cert = native_tls::Certificate::from_der(ca_cert)?;
```
jgramoll, line 168: Should this be 'from_der' or 'from_pem'?

tredmon: Perhaps `ca_cert` should be an `Option<Certificate>`, since I don't think the redis library needs to care about the cert format.

jgramoll: I updated to `Option<Certificate>` I had to wrap the `native_tls::Certificate` because it didn't implement Eq, PartialEq, Debug required by `ConnectionAddr`

redis/src/aio/tokio.rs
```diff
@@ -100,6 +102,7 @@ impl RedisRuntime for Tokio {
         hostname: &str,
         socket_addr: SocketAddr,
         insecure: bool,
+        ca_cert: Option<&Certificate>,
```
djc, line 105: I think we should take an `Option<Certificate>` here, so that cloning is ultimately left to the caller.

redis/src/tls.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /// Returns the DER-encoded representation of this certificate.
+    pub fn to_der(&self) -> Result<Vec<u8>> {
+        let der = self.0.to_der()?;
+        Ok(der)
+    }
+}
+
+impl From<&Certificate> for NativeCertificate {
```
djc, line 29: I think we we should get rid of this `impl` and make it explicit at the caller sites.

djc:
I think defining our own `Certificate` type is a good way to go, should be able to get this merged soon!

redis/src/tls.rs
jgramoll: Done

redis/src/aio/tokio.rs
jgramoll: Done

nonirosenfeldredis:
Hi, I really need this functionality as well . how can we promote its merge?

djc:
Looks like the question about documentation is still open.

nihohit:

PR #690, opened 2022-09-29
cluster: simplify TlsMode logic, add param to builder, & refactor create_initial_connections/connect

utkarshgupta137:
- Minor doc changes in cluster_client
- Moved related imports/structs/functions in cluster.rs together to improve code readability
- Convert create_initial_connections to a method to reduce code duplication
- Convert connect to a method to reduce passing arguments over & over
- Create a connection inside the connect method directly instead of creating a client first
- Simplified logic around TlsMode
  - Currently, we check if the connection is a Tls connection or not as well as the value of the insecure param each time we get a connection. The value of the insecure param is also added to the node string, which is used as a key to get connections.
  - Instead, it is safe to assume that the connection type (Tls or not) & insecure flag is the same for all the nodes in the cluster. Thus, we can just parse the TlsMode once within the build() function in the cluster client builder & only use it when we need to create the connection
  - This also simplifies creating the node string from the ConnectionAddr

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -54,7 +54,7 @@ use crate::cluster_pipeline::UNROUTABLE_ERROR;
 use crate::cluster_routing::{Routable, RoutingInfo, Slot, SLOT_SIZE};
 use crate::cmd::{cmd, Cmd};
 use crate::connection::{
-    Connection, ConnectionAddr, ConnectionInfo, ConnectionLike, IntoConnectionInfo,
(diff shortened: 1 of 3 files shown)


0xWOF:
As https://github.com/redis-rs/redis-rs/pull/718 is closed, every changes need to merge on this pull request is merged. I think this pull request can be closed. I'll continue to create pull request with https://github.com/redis-rs/redis-rs/pull/670 to merge necessary changes.

PR #694, opened 2022-10-15
Update licenses and READMEs

jaymell:
Copy/update License files into individual crates;
copy root README to redis crate and add a basic
README to the testing crate.


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2013-2019 by Armin Ronacher, Jan-Erik Rediger.
+Copyright (c) 2013-2022 by Armin Ronacher, Jan-Erik Rediger, and redis-rs contributors
```
djc, line 1: I think we should just change this to `redis-rs contributors` only.

redis-test/LICENSE
```diff
@@ -0,0 +1,31 @@
+Copyright (c) 2022 by redis-rs contributors
```
djc, line 1: I usually do this by adding a symlink to the parent-dir `LICENSE`, which avoids the need to keeping this in sync.

redis/LICENSE
```diff
@@ -0,0 +1,33 @@
+Copyright (c) 2013-2022 by Armin Ronacher, Jan-Erik Rediger, and redis-rs contributors
```
djc, line 1: Symlink?

redis/README.md
```diff
@@ -0,0 +1,175 @@
+# redis-rs
```
djc, line 1: Instead of adding a copy (which we then have to keep in sync manually) you can use the `package.readme` key in the Cargo metadata to refer to `../README.md`.

djc (approved):

djc:
These symlinks are a very common practice, no worries!

Might want to add a version bump commit?

jaymell:
Thanks, want to take a quick look again?

djc (approved):

Issue #658, opened 2022-08-05
Async, connection-pooled Cluster client


zombiezen:
Perhaps I have missed something, but is there not an equivalent of `redis::aio::ConnectionManager` for Redis Cluster?

djc:
Are you aware of https://github.com/redis-rs/redis-cluster-async already?

zombiezen:
Yes, but it looks like it does not pool connections? https://docs.rs/redis_cluster_async/latest/redis_cluster_async/struct.Client.html#method.get_connection seems to open a new connection each time.

djc:
I don't think `redis::aio::ConnectionManager` is really a pool -- as I understand it, it's just a connection that knows how to reconnect when its inner connection drops. For pooling, I recommend looking at deadpool or bb8 (though I don't know if a connection manager for the cluster thing already exists).

jaymell:
There is an example using bb8 with the `redis-cluster-async` lib here: https://github.com/svix/svix-webhooks/blob/main/server/svix-server/src/redis/cluster.rs

Owez:
Small note to anyone attempting this: pubsub is hard with async + connection pooling with deadpool

casret:
@djc Do you happen to know why they are maintained as 2 seperate crates?

djc:
They were historically developed separately/by different people. No good reason now...

nihohit:
Closing since cluster connection has been merged.

PR #696, opened 2022-10-19
Add Async Cluster Support

jaymell:
Bring in code from https://github.com/redis-rs/redis-cluster-async
with minimal changes in order to allow for easier maintenance of
both projects and to better promote code reuse.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -177,6 +177,7 @@ struct Pipeline<C> {
     connections: ConnectionMap<C>,
     slots: SlotMap,
     state: ConnectionState<C>,
+    #[allow(clippy::complexity)]
     in_flight_requests: stream::FuturesUnordered<
         Pin<Box<Request<BoxFuture<'static, (String, RedisResult<Response>)>, Response, C>>>,
     >,
@@ -237,7 +238,7 @@ impl<C> CmdArg<C> {
                             .and_then(|key_count_str| key_count_str.parse::<usize>().ok());
                         key_count_res.and_then(|key_count| {
                             if key_count > 0 {
-                                get_cmd_arg(cmd, 3).map(|key| slot_for_key(key))
+                                get_cmd_arg(cmd, 3).map(slot_for_key)
                             } else {
                                 // TODO need to handle sending to all masters
                                 None
@@ -245,7 +246,7 @@ impl<C> CmdArg<C> {
                         })
                     })
                 }
-                Some(b"XGROUP") => get_cmd_arg(cmd, 2).map(|key| slot_for_key(key)),
+                Some(b"XGROUP") => get_cmd_arg(cmd, 2).map(slot_for_key),
                 Some(b"XREAD") | Some(b"XREADGROUP") => {
                     let pos = position(cmd, b"STREAMS")?;
                     get_cmd_arg(cmd, pos + 1).map(slot_for_key)
@@ -254,7 +255,7 @@ impl<C> CmdArg<C> {
                     // TODO need to handle sending to all masters
                     None
                 }
-                _ => get_cmd_arg(cmd, 1).map(|key| slot_for_key(key)),
+                _ => get_cmd_arg(cmd, 1).map(slot_for_key),
             }
         }
         match self {
@@ -368,11 +369,10 @@ where
         let future = match this.future.as_mut().project() {
             RequestStateProj::Future { future } => future,
             RequestStateProj::Sleep { sleep } => {
-                return match ready!(sleep.poll(cx)) {
-                    () => Next::TryNewConnection {
-                        request: self.project().request.take().unwrap(),
-                        error: None,
-                    },
+                ready!(sleep.poll(cx));
+                return Next::TryNewConnection {
+                    request: self.project().request.take().unwrap(),
+                    error: None,
                 }
                 .into();
             }
@@ -453,10 +453,9 @@ where
     C: ConnectionLike + Connect + Clone + Send + Sync + 'static,
 {
     async fn new(initial_nodes: &[ConnectionInfo], retries: Option<u32>) -> RedisResult<Self> {
-        let tls = initial_nodes.iter().all(|c| match c.addr {
-            ConnectionAddr::TcpTls { .. } => true,
-            _ => false,
-        });
+        let tls = initial_nodes
+            .iter()
+            .all(|c| matches!(c.addr, ConnectionAddr::TcpTls { .. }));
         let connections = Self::create_initial_connections(initial_nodes).await?;
         let mut connection = Pipeline {
             connections,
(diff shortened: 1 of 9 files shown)


jaymell:
@djc Wanted to get your feedback on the approach here. Still a work in progress, but I'm thinking it's the best way at this point to move forward with getting native async clustered support in redis-rs.

djc:
Seems reasonable.

utkarshgupta137:
The mpsc used in this implementation is unnecessary and adds an extra level of indirection. I ran some basic benchmarks and found that sending the message directly to the underlying pipeline in MultiplexedConnection is faster.

redis/src/cluster_async/mod.rs
```diff
@@ -0,0 +1,1239 @@
...
+}
+
+// Get slot data from connection.
+async fn get_slots<C>(addr: &str, connection: &mut C, use_tls: bool) -> RedisResult<Vec<Slot>>
+where
+    C: ConnectionLike,
+{
+    trace!("get_slots");
+    let mut cmd = Cmd::new();
+    cmd.arg("CLUSTER").arg("SLOTS");
```
shachlanAmazon, line 1100: should use `CLUSTER SHARDS` on redis 7 - `CLUSTER SLOTS` is deprecated.
https://redis.io/commands/cluster-slots/

redis/src/cluster_async/mod.rs
```diff
@@ -0,0 +1,1239 @@
...
+            Connection(tx)
+        })
+    }
+}
+
+type SlotMap = BTreeMap<u16, String>;
+type ConnectionFuture<C> = future::Shared<BoxFuture<'static, C>>;
+type ConnectionMap<C> = HashMap<String, ConnectionFuture<C>>;
+
+struct Pipeline<C> {
```
shachlanAmazon, line 170: I know this maintains the naming from the `aio` module, but `Pipeline` has two meanings here - both this and `CmdArg::Pipeline` (and `crate::Pipeline`). IMO this duality should be avoided, and the struct renamed to something like `ClusterConnection`, or just `Connection` and rename `Connection` to `NodeConnection`.

redis/src/cluster_async/mod.rs
```diff
@@ -0,0 +1,1239 @@
...
+    }
+
+    fn try_request(
+        &mut self,
+        info: &RequestInfo<C>,
+    ) -> impl Future<Output = (String, RedisResult<Response>)> {
+        // TODO remove clone by changing the ConnectionLike trait
+        let cmd = info.cmd.clone();
+        let (addr, conn) = if info.excludes.len() > 0 || info.slot.is_none() {
+            get_random_connection(&self.connections, Some(&info.excludes))
```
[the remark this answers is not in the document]
jaymell: Hey, I think your request here is a reasonable one, but this specific PR is probably not the place to do it. Do you mind creating a separate issue for this?

jaymell:
@djc Hoping you can start reviewing this whenever you get a chance. I'd like to hold off on merging until after the next release is cut, but otherwise I think it's almost in a merge-able (which is not to say releasable) state.

I've tried to keep the commits as straightforward as possible but can certainly break things up further if helpful.

djc:
Want to mark it as ready for review or would you prefer to hold off until after the next release?

jaymell:
Marked it ready, would still prefer to hold off on merging until next release is created though. Also, I don't intend to release this as-is, I have a list of items [here](https://github.com/redis-rs/redis-rs/issues/745) I would to add in subsequent PRs before a release happens.

trungda:
Thanks for doing this! redis-cluster-async definitely needs more support. What's the plan with open PRs in the redis-cluster-async? I also have one improvement (refresh-slot command doesn't handle username/password very well) that I'd like to open PR. What should I do? Should I still open PR against redis-cluster-async? Thanks.

0xWOF:
Just for curiosity. Why `cluster async connection` is provided by `cluster_async`  module instead `cluster_aio.rs` file? I think it is bit weired considering naming consistency. Because `normal async connection` is provided by `aio.rs` file and `cluster sync connection` is provided by `cluster.rs` file.

redis/src/cluster_async/mod.rs
jaymell: Agreed that the naming is very confusing. Planning to address the confusion in subsequent PRs.

Soufian20:
Is it also possible to adapt the redis::aio::ConnectionManager to also support the cluster_async::client?

-> https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html

jaymell:
> Is it also possible to adapt the redis::aio::ConnectionManager to also support the cluster_async::client?
> 
> -> https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html

The clustered client itself should more or less perform the same functions as the connection manager by handling reconnections and retries when necessary. Is there something else I'm missing about `ConnectionManager` (which I personally have never worked with so am not terribly familiar)

Soufian20:
> The clustered client itself should more or less perform the same functions as the connection manager by handling reconnections and retries when necessary. Is there something else I'm missing about `ConnectionManager` (which I personally have never worked with so am not terribly familiar)?

You're right it has more or less the same functionality. It just wraps the client and the connections as a proxy and try also an initial connection. I can see someone created already an issue for that -> https://github.com/redis-rs/redis-rs/issues/658

Soufian20:
i tested your branch with this minimial cluster setup -> https://github.com/Soufian20/simple-redis-async-cluster
It works :+1:

redis/src/cluster_async/mod.rs
```diff
@@ -0,0 +1,1239 @@
+//! This is a rust implementation for Redis cluster library.
+//!
+//! This library extends redis-rs library to be able to use cluster.
+//! Client impletemts traits of ConnectionLike and Commands.
```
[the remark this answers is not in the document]
jaymell: Will be addressed in future PR prior to next release.

redis/src/cluster_async/mod.rs
jaymell: Thanks, raised as separate issue.

redis/src/cluster_async/mod.rs
```diff
@@ -0,0 +1,1239 @@
...
+            if let ConnectionAddr::Unix(_) = info.addr {
+                return Err(RedisError::from((ErrorKind::InvalidClientConfig,
+                                             "This library cannot use unix socket because Redis's cluster command returns only cluster's IP and port.")));
+            }
+            nodes.push(info);
+        }
+
+        Ok(Client {
+            initial_nodes: nodes,
+            retries: Some(DEFAULT_RETRIES),
```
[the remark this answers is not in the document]
jaymell: Yes, will be addressed when clients are consolidated, in upcoming PR.

PR #699, opened 2022-10-23
Improve MultiplexedConnection Error Handling

### LINKED ISSUE #698: ConnectionManager misaligns subsequent commands when `pipe()` has wrongtype commands
# Description
## Observed
When you use a ConnectionManager and have a WRONGTYPE-destined command in the pipe, the following commands in the pipe get vended out to the following commands using the connection.

## Expected
When you use a ConnectionManager and have a `pipe()` that has any result combination, the commands in that pipe are drained before the expression is completed - or at least the previous failed commands' results are not observable by subsequent commands.

## A little exposition
While obviously this repro is naive, sometimes values in the wild are not what you expect them to be. However, by leaving the connection in a broken/misaligned state the following commands are always wrong which is potentially catastrophic. Consider:
* fetch a configuration blob and then a thumbnail for user A in a pipe
* then fetch just a thumbnail for user B in a separate usage of the ConnectionManager

If User A's configuration blob isn't exactly right then user B gets user A's thumbnail!

### What I'm doing
I'm using `bb8`'s connection manager. It (by default) does a PING and looks for a PONG on connection checkout (tbh i'd rather pay that cost on connection return and might pr a configuration for it), which will resist (though not necessarily eliminate) the misalignment risk.

### Other stuff I experimented with
By editing the macros in `aio` I was able to tease the repro to work, though I only tried reconnecting more aggressively than is likely to be intended.
```rust
/// Handle a command result. If the connection was dropped, reconnect.
macro_rules! reconnect_if_dropped {
    ($self:expr, $result:expr, $current:expr) => {
        if let Err(ref e) = $result {
            if e.is_connection_dropped() {
                $self.reconnect($current);
            }
        }
    };
}

/// Handle a connection result. If there's an I/O error, reconnect.
/// Propagate any error.
macro_rules! reconnect_if_io_error {
    ($self:expr, $result:expr, $current:expr) => {
        if let Err(e) = $result {
            $self.reconnect($current);
            return Err(e);
        }
    };
}
```

# How to repro
Start redis listening on localhost:6379 like `docker run -p 6379:6379 --rm redis`

`cargo.toml`
```toml
[package]
name = "bugreport"
version = "0.1.0"
edition = "2021"

[dependencies]
tokio = { version = "1.19.2", features = ["full"] }
redis = { version = "0.22.1", features = ["tokio-comp", "connection-manager"] }
```

`main.rs`
```rust
use redis::AsyncCommands;

#[tokio::main]
async fn main() {
    let mut connection_manager = redis::aio::ConnectionManager::new(
        redis::Client::open("redis://localhost:6379").unwrap()
    ).await.unwrap();

    // Put a list in redis
    let _: () = connection_manager.lpush("key", "value").await.unwrap();

    // Execute a pipe with a wrongtype at the beginning
    let _: Result<(String, usize), redis::RedisError> = redis::pipe()
        .get("key") // WRONGTYPE
        .llen("key")
        .query_async(&mut connection_manager)
        .await;

    // And now the connection is in a broken state
    let _: Vec<String> = connection_manager.lrange("key", 0, -1)
        .await
        .expect("should be able to get a list but instead I get the llen");
}
```

Observe:
```
thread 'main' panicked at 'should be able to get a list but instead I get the llen:
 Response was of incompatible type: "Response type not vector compatible."
 (response was int(1))', src/main.rs:20:72
```

jaymell:
Improve MultiplexedConnection Error Handling

It was recently discovered that `aio::MultiplexedConnection`
does not properly handle errors. If a command in the pipe fails,
responses from subsequent commands in the pipe are not processed
and will be returned to the next request issued on the connection,
with potentially catastrophic results. This fix ensures that
all pending commands in the pipe are processed before returning.

Fixes #698.

### DIFF
--- a/redis/src/aio.rs
+++ b/redis/src/aio.rs
@@ -4,7 +4,6 @@ use std::collections::VecDeque;
 use std::fmt;
 use std::fmt::Debug;
 use std::io;
-use std::mem;
 use std::net::SocketAddr;
 #[cfg(unix)]
 use std::path::Path;
@@ -602,8 +601,22 @@ type PipelineOutput<O, E> = oneshot::Sender<Result<Vec<O>, E>>;
 
 struct InFlight<O, E> {
     output: PipelineOutput<O, E>,
-    response_count: usize,
+    expected_response_count: usize,
+    current_response_count: usize,
     buffer: Vec<O>,
+    first_err: Option<E>,
+}
+
+impl<O, E> InFlight<O, E> {
+    fn new(output: PipelineOutput<O, E>, expected_response_count: usize) -> Self {
+        Self {
+            output,
+            expected_response_count,
+            current_response_count: 0,
+            buffer: Vec::new(),
+            first_err: None,
+        }
+    }
 }
 
 // A single message sent through the pipeline
@@ -679,26 +692,37 @@ where
 
     fn send_result(self: Pin<&mut Self>, result: Result<I, E>) {
         let self_ = self.project();
-        let response = {
+
+        {
             let entry = match self_.in_flight.front_mut() {
                 Some(entry) => entry,
                 None => return,
             };
+
             match result {
                 Ok(item) => {
                     entry.buffer.push(item);
-                    if entry.response_count > entry.buffer.len() {
-                        // Need to gather more response values
-                        return;
-                    }
-                    Ok(mem::take(&mut entry.buffer))
                 }
-                // If we fail we must respond immediately
-                Err(err) => Err(err),
+                Err(err) => {
+                    if entry.first_err.is_none() {
+                        entry.first_err = Some(err);
+                    }
+                }
             }
-        };
+
+            entry.current_response_count += 1;
+            if entry.current_response_count < entry.expected_response_count {
+                // Need to gather more response values
+                return;
+            }
+        }
 
         let entry = self_.in_flight.pop_front().unwrap();
+        let response = match entry.first_err {
+            Some(err) => Err(err),
+            None => Ok(entry.buffer),
+        };
+
         // `Err` means that the receiver was dropped in which case it does not
         // care about the output and we can continue by just dropping the value
         // and sender
@@ -750,11 +774,9 @@ where
 
         match self_.sink_stream.start_send(input) {
             Ok(()) => {
-                self_.in_flight.push_back(InFlight {
-                    output,
-                    response_count,
-                    buffer: Vec::new(),
-                });
+                self_
+                    .in_flight
+                    .push_back(InFlight::new(output, response_count));
                 Ok(())
             }
             Err(err) => {
--- a/redis/tests/test_async.rs
+++ b/redis/tests/test_async.rs
@@ -569,4 +569,31 @@ mod pub_sub {
         })
         .unwrap();
     }
+
+    #[test]
+    fn pipe_errors_do_not_affect_subsequent_commands() {
+        use redis::RedisError;
+
+        let ctx = TestContext::new();
+        block_on_all(async move {
+            let mut conn = ctx.multiplexed_async_connection().await?;
+
+            conn.lpush::<&str, &str, ()>("key", "value").await?;
+
+            let res: Result<(String, usize), redis::RedisError> = redis::pipe()
+                .get("key") // WRONGTYPE
+                .llen("key")
+                .query_async(&mut conn)
+                .await;
+
+            assert!(res.is_err());
+
+            let list: Vec<String> = conn.lrange("key", 0, -1).await?;
+
+            assert_eq!(list, vec!["value".to_owned()]);
+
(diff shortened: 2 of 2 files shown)


redis/src/aio.rs
```diff
@@ -642,6 +642,7 @@ pin_project! {
         sink_stream: T,
         in_flight: VecDeque<InFlight<I, E>>,
         error: Option<E>,
+        redis_errors: VecDeque<E>,
```
djc, line 645: What's the distinction between `error` and `redis_errors` here? Having both of those seems... weird.

redis/src/aio.rs
```diff
@@ -677,32 +679,41 @@ where
...
-        // `Err` means that the receiver was dropped in which case it does not
-        // care about the output and we can continue by just dropping the value
-        // and sender
-        entry.output.send(response).ok();
+        let mut redis_errors = mem::take(self_.redis_errors);
+        if !redis_errors.is_empty() {
+            // Send the first redis_error found:
+            entry
+                .output
+                .send(Err(redis_errors.pop_front().unwrap()))
```
djc, line 712: So we just send the first error here. Should we change the API so that we can send all errors, instead? Why not?

redis/src/aio.rs
jaymell: Agreed. `error` basically indicates that `Sink` has failed, so it's basically a terminal state. I could rename it to `sink_error` or `terminal_error` to indicate that.

redis/src/aio.rs
jaymell: I agree that would be a better API, but I'm not sure we can easily accommodate without making changes to the `ConnectionLike` trait. 

If possible, I would like to get the minimal approach in this PR added because it mimics existing behavior -- i.e., returning the `Err` from the first failure in the pipe -- minus the data corruption. I think it can easily be added to a 21.x release as well.

jaymell: What if I consolidate all the errors into an aggregated `RedisError`? Not sure how much work that is offhand, but I'm sure it'd be doable.

redis/src/aio.rs
jaymell: I renamed both fields and made `redis_errors` an Option<E> named `first_err` to match the other Connection types. Both are needed though, as they serve different purposes.

redis/src/aio.rs
jaymell: I've simplified it to only store the first error, so that it now matches the behavior of the other Connection types.

redis/src/aio.rs
jaymell: Moved mutable fields onto the Inflight entry, which is more appropriate here.

redis/src/aio.rs
jaymell: Moved mutable fields onto the Inflight entry, which is more appropriate here.

redis/src/aio.rs
```diff
@@ -679,26 +692,37 @@ where
...
             };
+
             match result {
                 Ok(item) => {
                     entry.buffer.push(item);
-                    if entry.response_count > entry.buffer.len() {
-                        // Need to gather more response values
-                        return;
+                }
+                Err(err) => {
```
djc, line 706: So do we typically expect other return items to come in even after an error?

jaymell: Yes. I will admit that intuitively it's a bit scary, but this is how the other `Connection`s (e.g., for [aio::Connection](https://github.com/redis-rs/redis-rs/blob/d7d2f2a6b868ba6e4527110c1dd170d3675cb5c1/redis/src/aio.rs#L572), and for [the sync Connection](https://github.com/redis-rs/redis-rs/blob/d7d2f2a6b868ba6e4527110c1dd170d3675cb5c1/redis/src/connection.rs#L872)) and indeed the other redis libraries I've looked at handle the situation. Unless there's an issue parsing the commands properly, or a larger infrastructure issue, the commands will all be sent and responses will be returned.

Again, I think the API itself here is poor. We really should return `Result`s for individual commands rather than all-or-nothing but this fix needs to happen irrespective of that.

djc (approved):

redis/src/aio.rs
kvcache: > We really should return Results for individual commands rather than all-or-nothing but this fix needs to happen irrespective of that.

I actually was looking for a FromRedisValue implementation for Result for this reason 😄 

You might consider adding that as an option rather than pushing Result onto every consumer of your pipe api? Fwiw it's nice to be able to assert that the whole thing worked _else_ it is an error. But for some use cases a partial success is important and may be distinct from an overall pipe failure - hence, a FromRedisValue<Result...> impl which would even let someone decide if 1 part of the pipeline is resulty distinct from the overall pipe!

djc: PR welcome? 😄

PR #702, opened 2022-10-30
add test case for atomic pipeline

CNLHC:
`redis-test`  is a convenient tool to write UT, but the correct way to write mock responses for some cases is not obvious.

This PR contains a test case about `pipeline` with `atomic`.

This test case may help people find that they should use `Value::Bulk` as a mock response to test the pipeline with atomic instead of a plain `vec`.


jaymell:
Thanks!

jaymell (approved):

PR #704, opened 2022-11-03
fix: solve lint issues

0xWOF:
On rust 1.65.0, main branch of redis-rs can not pass lint workflow.
Solved issues is
- explicit-auto-deref
- bool-to-int-with-if
- needless-borrow

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -451,7 +451,7 @@ impl ClusterConnection {
             let (addr, rv) = {
                 let mut connections = self.connections.borrow_mut();
                 let (addr, conn) = if let Some(addr) = redirected.take() {
-                    let conn = self.get_connection_by_addr(&mut *connections, &addr)?;
+                    let conn = self.get_connection_by_addr(&mut connections, &addr)?;
                     if is_asking {
                         // if we are in asking mode we want to feed a single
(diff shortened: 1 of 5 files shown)


djc:
Wow, that was quick! Thanks for cleaning these up!

djc (approved):

0xWOF:
it is weird that this PR is pass (7.0.0, nightly) test on my repository, but not on here.
https://github.com/0xWOF/redis-rs/actions/runs/3387016797/jobs/5627179494.

I think that it is timing issue, if `con.set` is executed slowly on ci machine, it can do not pass the test.
How about your thoughts?
```rust
    let ctx = TestContext::new();

    let _: () = con.set("object_key_str", "object_value_str").unwrap();
    let _: () = con.set("object_key_int", 42).unwrap();

    assert_eq!(
        con.object_encoding::<_, String>("object_key_str").unwrap(),
        "embstr"
    );

    assert_eq!(
        con.object_encoding::<_, String>("object_key_int").unwrap(),
        "int"
    );

    assert_eq!(con.object_idletime::<_, i32>("object_key_str").unwrap(), 0); // do not pass here!
```

djc:
Looks like that test failed intermittently, since it worked on the second try.

PR #705, opened 2022-11-03
[2/5] [1] cluster_client: doc changes

0xWOF:
## Background

Hello. I want to support @utkarshgupta137's PRs. (part 2, 3, 4, 5)
I think I can help with separating PRs. And I want to start with this PR (https://github.com/redis-rs/redis-rs/pull/690).
Firstly, I create 3 PRs. When there are something I should do more, feel free to notify to me!

## Changes

- change some comments on cluster_client

## Effect

- there no changes on code and not changes on binary

## Node

- this PR is dependent on https://github.com/redis-rs/redis-rs/pull/704
- you can check actual file changes on https://github.com/0xWOF/redis-rs/pull/13

PR #706, opened 2022-11-03
[2/5] [2] cluster, cluster_client: minor refactorings (change order of code, use Self)

0xWOF:
## Changes

- cluster: change orders of declarations
- cluster_client: add `Self::`

## Effect

- no public api is changed
- cluster: implementation is not changed only order of declarations is changed
- cluster_client: add `Self::` it should do not change behavior

## Note

- this PR is depends on https://github.com/redis-rs/redis-rs/pull/705
- you can check actual file changes on https://github.com/0xWOF/redis-rs/pull/14

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -194,17 +194,6 @@ impl ClusterConnection {
         Ok(())
     }
 
-    /// Check that all connections it has are available (`PING` internally).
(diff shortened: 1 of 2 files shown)


djc:
Please rebase onto main so the first three commits disappear.

0xWOF:
I rebase it!

djc (approved):

PR #707, opened 2022-11-03
[2/5] [3] cluster: move connect to main impl

0xWOF:
## Changes

- cluster: move connect function to ClusterConnection impl
- cluster: follow up usage code of connect function

## Effect

- cluster:
    - no public api is changed
    - code is changed but behavior will not be changed

## Note

- this PR is depends on https://github.com/redis-rs/redis-rs/pull/706
- you can check actual file changes on https://github.com/0xWOF/redis-rs/pull/15

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -227,7 +227,7 @@ impl ClusterConnection {
                 _ => panic!("No reach."),
             };
 
-            if let Ok(mut conn) = connect(
+            if let Ok(mut conn) = Self::connect(
                 info.clone(),
                 read_from_replicas,
                 username.clone(),
@@ -281,7 +281,7 @@ impl ClusterConnection {
                     }
                 }
 
-                if let Ok(mut conn) = connect(
+                if let Ok(mut conn) = Self::connect(
                     addr.as_ref(),
                     self.read_from_replicas,
                     self.username.clone(),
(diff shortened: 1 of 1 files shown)


redis/src/cluster.rs
```diff
@@ -227,7 +227,7 @@ impl ClusterConnection {
                 _ => panic!("No reach."),
             };
 
-            if let Ok(mut conn) = connect(
+            if let Ok(mut conn) = Self::connect(
```
djc, line 230: This needs to be part of the previous commit; each commit should compile and pass tests.

djc (commented):
This mostly looks good to me.

redis/src/cluster.rs
0xWOF: I squash this commit with previous commit, I'll apply this rule to other PRs. Thanks!

djc:
Please rebase. I recommend changing your PRs so each PR targets the previous one's branch. GitHub will update as the original PR gets merged, and the commit view in this PR is actually useful.

0xWOF:
I didn't know that. Thank you for notify to me. I'll rebase it as soon as possible. And I'll apply this rule to other PRs.

0xWOF:
@djc I finish rebase. But I can't satisfy this recommend because I have no permission to create branch on redis-rs/redis-rs repository. And PR of `0xWOF:cluster_client_refactor_2_4` to `0xWOF:cluster_client_refactor_2_3` is not PR of redis-rs/redis-rs repository is PR of 0xWOF/redis-rs repository.
> I recommend changing your PRs so each PR targets the previous one's branch instead of main.


Instead, I'll create PR one by one after PR 5. If you know other better way, please share to me.
Thanks!

djc (approved):

djc:
Creating them one by one seems reasonable to me, and is actually easier on my GitHub notifications.

PR #709, opened 2022-11-04
[2/5] [4] cluster: change order of codes

0xWOF:
## Changes

- cluster: move ClusterConnection struct declaration to above of impl
- cluster: move set_auto_reconnect

## Effect

- cluster:
    - no public api is changed
    - code is changed but behavior will not be changed

## Note

- this PR is depends on https://github.com/redis-rs/redis-rs/pull/707
- you can check actual file changes on https://github.com/0xWOF/redis-rs/pull/16

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -62,8 +62,6 @@ use crate::types::{ErrorKind, HashMap, HashSet, RedisError, RedisResult, Value};
 pub use crate::cluster_client::{ClusterClient, ClusterClientBuilder};
 pub use crate::cluster_pipeline::{cluster_pipe, ClusterPipeline};
 
-type SlotMap = BTreeMap<u16, [String; 2]>;
-
 /// This is a connection of Redis cluster.
 pub struct ClusterConnection {
     initial_nodes: Vec<ConnectionInfo>,
@@ -76,22 +74,6 @@ pub struct ClusterConnection {
     read_timeout: RefCell<Option<Duration>>,
     write_timeout: RefCell<Option<Duration>>,
     tls: Option<TlsMode>,
-}
-
-#[derive(Clone, Copy)]
-enum TlsMode {
-    Secure,
-    Insecure,
-}
-
-impl TlsMode {
-    fn from_insecure_flag(insecure: bool) -> TlsMode {
-        if insecure {
-            TlsMode::Insecure
-        } else {
-            TlsMode::Secure
-        }
-    }
 }
 
 impl ClusterConnection {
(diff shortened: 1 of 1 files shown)


0xWOF:
I rebase it because of PR 3 is merged. After this PR, I'll create PR one by one.

redis/src/cluster.rs
```diff
@@ -94,6 +80,20 @@ impl TlsMode {
     }
 }
 
+/// This is a connection of Redis cluster.
```
djc, line 83: This is an improvement, but I'd prefer to instead move `TlsMode` and its `impl` closer to the bottom of the module.

redis/src/cluster.rs
```diff
@@ -194,6 +187,13 @@ impl ClusterConnection {
         Ok(())
     }
 
+    /// Set an auto reconnect attribute.
```
djc, line 190: I don't see the point? This was already grouped with other `pub fn ..(&self)` methods, so not much point in moving it around within that group?

0xWOF: I agree with you. There are no reason to move set_auto_reconnect. I restore it.

redis/src/cluster.rs
djc: Yes, moving `SlotMap` to the bottom makes sense to me as well.

0xWOF: 👍

redis/src/cluster.rs
0xWOF: Sure. I drop the commit.

djc (approved):

0xWOF:
ci problem is same with previous intermittent one. <https://github.com/redis-rs/redis-rs/pull/704#issuecomment-1302313412>

utkarshgupta137:
IMO, it is better to define everything before it is used in the code, to the extent possible. The purpose of the original commit was to bring the struct & impl of ClusterConnection next to each other. You're of course, free to follow your convention.

redis/src/cluster.rs
```diff
@@ -694,6 +678,22 @@ impl NodeCmd {
...
+    fn from_insecure_flag(insecure: bool) -> TlsMode {
+        if insecure {
+            TlsMode::Insecure
+        } else {
+            TlsMode::Secure
+        }
+    }
+}
+
 impl ConnectionLike for ClusterConnection {
```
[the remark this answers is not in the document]
0xWOF: Sorry for I misunderstand your purpose. I agree with your way.
@djc how about move `impl ConnectionLike for ClusterConnection` ? if you okay with this way, I want to apply this too.

utkarshgupta137:
Not to mention that the more changes you make early on, the hard it will be to apply future changes. There are a lot of changes in this file, so it would probably be best not to mess with the order of code if it doesn't make a lot of difference.

redis/src/cluster.rs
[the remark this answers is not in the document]
djc: Yes, I agree (I think I already mentioned that?).

0xWOF:
@utkarshgupta137 I think some changes are can't be avoidable and rebase difficulty is still bearable. So I'll be careful more not to do mis-rebase for now. But your concern is reasonable too. I'll find other better way to satisfy ease-of-rebase and code-change-ability when I can't bearable more.

redis/src/cluster.rs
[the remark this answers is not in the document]
0xWOF: I apply it. Thanks!

djc (approved):

PR #711, opened 2022-11-04
Add `Script::invoke_async` method

r-bk:
This PR adds the `Script::invoke_async` method and corresponding tests.

The aim is to have an asynchronous equivalent to the synchronous `Script::invoke` method.


djc (approved):
Makes sense to me, thanks! @jaymell any thoughts?

PR #718, opened 2022-11-12
[2/5] [7] cluster: make ClusterConnection use same tls_mode on every connections using ClusterParams

0xWOF:
## Changes

- cluster
    - rename tls to tls_mode
    - make ClusterConnection use same tls_mode on every connections
    - determine tls_mode from ClusterParams
- cluster_client
    - add tls_mode to ClusterParams
    - add tls_mode to builder
- connection
    - add comment

## Effect

- cluster:
    - ClusterConnection do not use other tls_mode for each connections any more.
- cluster_client
    - ** public api is changed **
    - user can determine tls_mode of ClusterConnection from builder
    - when user do not specify tls_mode, tls_mode is determined from first item of initial_nodes
- connection
    - no public api and binary is changed

## Note

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -463,9 +463,9 @@ impl ClusterConnection {
                         let kind = err.kind();
 
                         if kind == ErrorKind::Ask {
-                            redirected = err
-                                .redirect_node()
-                                .map(|(node, _slot)| build_connection_string(node, None, self.tls_mode));
+                            redirected = err.redirect_node().map(|(node, _slot)| {
+                                build_connection_string(node, None, self.tls_mode)
+                            });
                             is_asking = true;
                         } else if kind == ErrorKind::Moved {
                             // Refresh slots.
@@ -473,9 +473,9 @@ impl ClusterConnection {
                             excludes.clear();
 
                             // Request again.
-                            redirected = err
-                                .redirect_node()
-                                .map(|(node, _slot)| build_connection_string(node, None, self.tls_mode));
+                            redirected = err.redirect_node().map(|(node, _slot)| {
+                                build_connection_string(node, None, self.tls_mode)
+                            });
                             is_asking = false;
                             continue;
                         } else if kind == ErrorKind::TryAgain || kind == ErrorKind::ClusterDown {
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -768,18 +768,13 @@ fn get_slots(connection: &mut Connection, tls_mode: Option<TlsMode>) -> RedisRes
 // - Returned from redis via the ASK/MOVED response
 fn get_connection_info(node: &str, cluster_params: ClusterParams) -> RedisResult<ConnectionInfo> {
     let mut split = node.split(':');
-    let host = match split.next() {
-        Some(val) => val,
-        None => fail!((ErrorKind::InvalidClientConfig, "Invalid node string")),
-    };
-    let port_parse = match split.next() {
-        Some(val) => u16::from_str(val),
-        None => fail!((ErrorKind::InvalidClientConfig, "Invalid node string")),
-    };
-    let port = match port_parse {
-        Ok(val) => val,
-        Err(_) => fail!((ErrorKind::InvalidClientConfig, "Invalid node string")),
-    };
+    let invalid_error = || (ErrorKind::InvalidClientConfig, "Invalid node string");
+
+    let host = split.next().ok_or_else(invalid_error)?;
+    let port = split
+        .next()
+        .and_then(|string| u16::from_str(string).ok())
+        .ok_or_else(invalid_error)?;
 
     Ok(ConnectionInfo {
         addr: get_connection_addr(host.to_string(), port, cluster_params.tls_mode),
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -692,9 +692,13 @@ impl NodeCmd {
     }
 }
 
+/// TlsMode indicates use or do not use verification of certification.
+/// Check [ConnectionAddr](ConnectionAddr::TcpTls::insecure) for more.
 #[derive(Clone, Copy)]
-enum TlsMode {
+pub enum TlsMode {
+    /// Secure verify certification.
     Secure,
+    /// Insecure do not verify certification.
     Insecure,
 }
 
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -191,20 +191,9 @@ impl ClusterConnection {
         let mut connections = HashMap::with_capacity(self.initial_nodes.len());
 
         for info in self.initial_nodes.iter() {
-            let addr = match info.addr {
-                ConnectionAddr::Tcp(ref host, port) => format!("redis://{host}:{port}"),
-                ConnectionAddr::TcpTls {
-                    ref host,
-                    port,
-                    insecure,
-                } => {
-                    let tls_mode = TlsMode::from_insecure_flag(insecure);
-                    build_connection_string(host, Some(port), Some(tls_mode))
-                }
-                _ => panic!("No reach."),
-            };
+            let addr = info.addr.to_string();
 
-            if let Ok(mut conn) = self.connect(info.clone()) {
+            if let Ok(mut conn) = self.connect(&addr) {
                 if conn.check_connection() {
                     connections.insert(addr, conn);
                     break;
@@ -256,7 +245,7 @@ impl ClusterConnection {
                     }
                 }
 
-                if let Ok(mut conn) = self.connect(addr.as_ref()) {
+                if let Ok(mut conn) = self.connect(addr) {
                     if conn.check_connection() {
                         conn.set_read_timeout(*self.read_timeout.borrow()).unwrap();
                         conn.set_write_timeout(*self.write_timeout.borrow())
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -463,9 +463,7 @@ impl ClusterConnection {
                         let kind = err.kind();
 
                         if kind == ErrorKind::Ask {
-                            redirected = err.redirect_node().map(|(node, _slot)| {
-                                build_connection_string(node, None, self.tls_mode)
-                            });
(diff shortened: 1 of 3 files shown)


redis/src/cluster_client.rs
```diff
@@ -8,6 +8,7 @@ pub(crate) struct ClusterParams {
     pub(crate) password: Option<String>,
     pub(crate) username: Option<String>,
     pub(crate) read_from_replicas: bool,
+    pub(crate) tls_mode: Option<TlsMode>,
```
djc, line 11: It's not clear to me what this means. If `tls_mode` is `None`, does that mean, we don't use TLS to connect at all? Could use a comment here to explain the semantics.

redis/src/cluster.rs
```diff
@@ -686,9 +686,12 @@ impl NodeCmd {
     }
 }
 
+/// TlsMode indicate tls mode of every connections of ClusterConnection.
 #[derive(Clone, Copy)]
-enum TlsMode {
+pub enum TlsMode {
+    /// Secure means every connections of ClusterConnection are connected over tls.
```
djc, line 664: I don't think these comments represent what these were originally meant to mean -- at least I was under the impression this was about certificate verification being enabled when using TLS. Which is it? Is this different from non-cluster parts of the library?

utkarshgupta137: You can see the correct meaning of both in the original PR: https://github.com/redis-rs/redis-rs/pull/670/files#diff-3081b035ea7b3a6c6fbf36bdcc857b7d1405e36fbe98df550dc7ba187275a638R66.

redis/src/cluster_client.rs
utkarshgupta137: Yep. For `Tcp` connections it will be None. For `TcpTls` connections it will be Some.

redis/src/cluster.rs
0xWOF: Yes, you are correct. It represents do not check certificate verification. I'll change comment.
https://github.com/redis-rs/redis-rs/blob/8eda667d087967d049d238f538c640b13ac69c6e/redis/src/connection.rs#L54-L62

redis/src/cluster_client.rs
djc: In that case I think the field should be called `tls` instead of `tls_mode` (but would still like to have a comment here, too).

0xWOF: Below is semantics. I'll add this informations to comments of `pub enum TlsMode`.
- when `tls_mode` is `Some(Secure)`, then ClusterConnection use TLS to connect and verify certificate.
- when `tls_mode` is `Some(Insecure)`, then ClusterConnection use TLS to connect and do not verify certificate.
- when `tls_mode` is `None`, then ClusterConnection do not use TLS to connect.

0xWOF: @djc I finish add the comment. Do you think `tls_mode` should be replaced by `tls`? And `tls_mode` of ClusterParams should be replaced by `tls` too?

0xWOF: Personally, I prefer `tls_mode` because type is `Option<TlsMode>` and `tls` is little weird because type (`Option<TlsMode>`) is based on `TlsMode`. But I'm okay with `tls` because type is not just `TlsMode`, is `Option<TlsMode>`

djc:
Okay, please squash the added commit into the appropriate earlier commits.

0xWOF:
I rebase to squash comment commit

djc:
Yes, it's still on my list.

jaymell:
Do we actually want to expose `tls` or `tls_mode` publicly? I realize it's not currently supported, but is it possible in the future we'd want to support per-node TLS configurations? If so, the cluster-wide parameter wouldn't be beneficial. 

Apologies for the late reply here, but I'm on the fence whether the changes here are that valuable.

0xWOF:
Hello, @jaymell thank you for reply to this pull request.
As my understand current version of `redis-rs` support `per node tls configuration` on create initial connection to node only. And it is **not** support `per node tls configuration` on redirect connection to node or create connection to slot. (on this case, every connection's tls configuration is determined by first node's tls configuration)
And I think I can close this pull request and create new pull request to support `per node tls configuration`.

Hello, @utkarshgupta137. is there anything I'm missing? Rest of part2, 3, 4, 5 pull requests can be used without this pull request? (I'll modify some commits to make those work well without this pull request) Or this pull request is essential part of rest of part 2, 3, 4, 5 pull requests?

utkarshgupta137:
It does a lot more than exposing the API to the public. I would suggest not including the changes to the builder, but keeping the rest.

0xWOF:
@utkarshgupta137 Thank you for your clarification!

0xWOF:
Reopen pull request because I misread @utkarshgupta137 's comment. I'll change this pull request to satisfy (jaymell and utkarshgupta137) 's suggestions.

jaymell:
Ok, I think I'm off the fence now and would like to get these changes merged :stuck_out_tongue: 

I think it's impractical to try and support per-node configurations for things like TLS (though please let me know if you think otherwise), and in any case, we already have build methods for username/password. I'm going to add a couple small comments to the PR but otherwise think these changes do a good job simplifying maintaining connections.

redis/src/cluster_client.rs
jaymell: Can we make the change that @djc requested and change this back to `tls` in all places where it was changed?

redis/src/cluster.rs
```diff
@@ -791,16 +757,46 @@ fn get_slots(connection: &mut Connection, tls_mode: Option<TlsMode>) -> RedisRes
...
 }
 
-fn build_connection_string(host: &str, port: Option<u16>, tls_mode: Option<TlsMode>) -> String {
-    let host_port = match port {
-        Some(port) => format!("{}:{}", host, port),
-        None => host.to_string(),
+// The node string passed to this function will always be in the format host:port as it is either:
+// - Created by calling ConnectionAddr::to_string (unix connections are not supported in cluster mode)
+// - Returned from redis via the ASK/MOVED response
+fn get_connection_info(node: &str, cluster_params: ClusterParams) -> RedisResult<ConnectionInfo> {
```
jaymell, line 763: Can we reduce the repetition here a bit? Lots of potential ways to slice it, but one possibility
```
      let host = split
          .next()
          .ok_or_else(|| (ErrorKind::InvalidClientConfig, "Invalid node string"))?;
      let port = split
          .next()
          .and_then(|val| u16::from_str(val).ok())
          .ok_or_else(|| (ErrorKind::InvalidClientConfig, "Invalid node string"))?;
```

0xWOF: I refactored `get_connection_info` method to satisfy your comment with your suggestion and then add my little improvement.

0xWOF:
Thank you for your reply. I wanted to find way to support "per note configuration", but I still do not have any idea to archive it. (recently I have less time to contribute to open source project) When this pull request is merged, I'll continue to create pull request to merge remain things. And I'll continue to find way to support "per note configuration"

redis/src/cluster_client.rs
jaymell: @0xWOF Please address this one as well.

jaymell: Pretty please? :smile:

0xWOF: @jaymell I replace `tls_mode` to `tls` on changes of this pull request. (Exceptionally `tls_mode` on `get_slots` method is replaced which is not on changes of this pull request for consistency) And how about replacing `TlsMode` enum to `Tls` enum for consistency?

jaymell: I personally think it's fine as is. Goal here is just to minimize changes to existing code. So this looks good, thanks!

jaymell:
@0xWOF Can you resolve conflicts and I'll re-run the checks?

0xWOF:
@jaymell I rebased it.

jaymell (approved):

jaymell:
Thanks @utkarshgupta137 @0xWOF! I think this looks good and would like to get it merged.

It's worth noting (because it's not really mentioned in the description) that the primary benefit we get from these changes is simplifying how the connection map (`ClusterConnection::connections`) is keyed, going from a fake/partial connection string (which includes tls info, but not user/password info) to a simple 'host:port' string, which matches the format that redis itself uses to store nodes in the cluster. These changes will will make standardizing the async cluster code much easier as well, as the current redis-cluster-async fork has its own implementation for the connection map that has issues and can/should be made to match this.

My only complaint (and it's worth the library itself more than with this PR) that it's not entirely clear what's going on just looking at the changes. The connection map itself, or at least its key, probably deserves its own type for improved clarity. But there are a lot of potential improvements to be made in this module, so it will be good to get these changes in so we can continue moving forward with bringing in `cluster-async`.

utkarshgupta137:
> Thanks @utkarshgupta137 @0xWOF! I think this looks good and would like to get it merged.
> 
> It's worth noting (because it's not really mentioned in the description) that the primary benefit we get from these changes is simplifying how the connection map (`ClusterConnection::connections`) is keyed, going from a fake/partial connection string (which includes tls info, but not user/password info) to a simple 'host:port' string, which matches the format that redis itself uses to store nodes in the cluster. These changes will will make standardizing the async cluster code much easier as well, as the current redis-cluster-async fork has its own implementation for the connection map that has issues and can/should be made to match this.
> 
> My only complaint (and it's worth the library itself more than with this PR) that it's not entirely clear what's going on just looking at the changes. The connection map itself, or at least its key, probably deserves its own type for improved clarity. But there are a lot of potential improvements to be made in this module, so it will be good to get these changes in so we can continue moving forward with bringing in `cluster-async`.
> 
> 

That is why I wasn't in favour of splitting the changes so much in the first place. I understand that smaller changes make reviewing easier & I try to keep changes as small as possible. But in this case, I had basically rewritten the cluster client/connection. It was much more influenced by the python implementation than the rust implementation, so it never made much sense to compare the changes incrementally. (Similar to how you're merging the async client in one large PR instead of copying the sync client & changing it bit by bit).

djc (approved):

PR #722, opened 2022-11-16
Add hashmap to redis args

### LINKED ISSUE #444: HashMap doesn't implement ToRedisArgs
Hi,

Thanks for this awesome library.

The only problem I have with it currently is that it doesn't implement `ToRedisArgs` for `std::collections::HashMap`. Is there any reason it can't be implemented the same way it is done for `BTreeMap`? The code below seems to work:
```
impl<T: ToRedisArgs + Hash + Eq + Ord, V: ToRedisArgs> ToRedisArgs
    for HashMap<T, V>
{
    fn write_redis_args<W>(&self, out: &mut W)
    where
        W: ?Sized + RedisWrite,
    {
        for (key, value) in self.iter() {
            // otherwise things like HMSET will simply NOT work
            assert!(key.is_single_arg() && value.is_single_arg());

            key.write_redis_args(out);
            value.write_redis_args(out);
        }
    }

    fn is_single_arg(&self) -> bool {
        self.len() <= 1
    }
}
```

gibranamparan:
This is adding the std::collections::HashMap as a ToRedisArgs so that now would be possible to run HSET commands using a simple HashMap (as it already supports for BTreeMap)

Fixes #444.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1025,7 +1025,6 @@ impl<T: ToRedisArgs + Hash + Eq + Ord, V: ToRedisArgs> ToRedisArgs
         W: ?Sized + RedisWrite,
     {
         for (key, value) in self.iter() {
-            // otherwise things like HMSET will simply NOT work
             assert!(key.is_single_arg() && value.is_single_arg());
 
             key.write_redis_args(out);
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1017,24 +1017,22 @@ impl<T: ToRedisArgs + Hash + Eq + Ord, V: ToRedisArgs> ToRedisArgs for BTreeMap<
     }
 }
 
-impl<T: ToRedisArgs + Hash + Eq + Ord, V: ToRedisArgs> ToRedisArgs
-    for HashMap<T, V>
-{
-    fn write_redis_args<W>(&self, out: &mut W)
-    where
-        W: ?Sized + RedisWrite,
-    {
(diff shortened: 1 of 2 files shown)


shachlanAmazon:
Can we allocate capacity ahead of time? in large maps this might have significant perf cost.

https://morestina.net/blog/1843/the-stable-hashmap-trap

djc (approved):
Seems reasonable. @jaymell any thoughts?

djc:
> Can we allocate capacity ahead of time? in large maps this might have significant perf cost.
> 
> https://morestina.net/blog/1843/the-stable-hashmap-trap

This is only a `ToRedisArgs` impl, writing out the `HashMap`s items into Redis args. Seems to me that the capacity issue you're referring to would only be relevant for a `FromRedisValue` impl.

shachlanAmazon:
> > Can we allocate capacity ahead of time? in large maps this might have significant perf cost.
> > https://morestina.net/blog/1843/the-stable-hashmap-trap
> 
> This is only a `ToRedisArgs` impl, writing out the `HashMap`s items into Redis args. Seems to me that the capacity issue you're referring to would only be relevant for a `FromRedisValue` impl.

Oy, you're right 😵‍💫

jaymell:
The PR does not compile....

gibranamparan:
> The PR does not compile....



> The PR does not compile....



> The PR does not compile....

This should compile now.

jaymell:
Thanks!

PR #723, opened 2022-11-16
Add json_set_nx and json_set_xx

### LINKED ISSUE #721: Missing json command: json_set_nx
For the sake of completeness, it would be handy to set the json value of a key, only if the key does not exist.

d3rpp:
This is a PR to add 
- `json_set_nx` set the json value only if it is *NOT* already set
- `json_set_xx` set the json value only if it is already set

This closes #721


redis/Cargo.toml
```diff
@@ -71,7 +71,7 @@ default = ["acl", "streams", "geospatial", "script"]
 acl = []
 aio = ["bytes", "pin-project-lite", "futures-util", "futures-util/alloc", "futures-util/sink", "tokio/io-util", "tokio-util", "tokio-util/codec", "tokio/sync", "combine/tokio", "async-trait"]
 geospatial = []
-json = ["serde", "serde_json"]
+json = ["serde", "serde_json", "serde/derive"]
```
d3rpp, line 74: my bad, i messed this one up the first time around

this is needed othersise the json tests dont seem to want to run at all

image

This is the error that this fixes

d3rpp: adding `serde/derive` only imports some proc macros which don't end up in the compiled binary anyway, so this shouldn't cause issues

d3rpp:
Alright it works now, this can be reviewed :smile:

redis/tests/test_json.rs
```diff
@@ -340,19 +364,19 @@ fn test_json_numincrby() {
 
     assert_eq!(set_initial, Ok(true));
 
-    let json_numincrby_a: RedisResult<String> = con.json_numincrby(TEST_KEY, "$.a", 2);
+    let json_numincrby_a: RedisResult<String> = con.json_num_incr_by(TEST_KEY, "$.a", 2);
```
jaymell, line 367: Wait what? :joy:

d3rpp: shiiii
probably should've read all the code I replaced lmao
will fix in bug/existing tests PR
ngl I wrote this on 1 hour of sleep

PR #724, opened 2022-11-23
Limit Parser Recursion

jaymell:
Eliminate possibility of stack overflow when
parsing server responses by limiting recursion 
depth.


jaymell:
This could probably be more declarative/elegant, but this seems to at least meet the basic need.

redis/src/parser.rs
```diff
@@ -174,7 +186,7 @@ mod aio_support {
                 let buffer = &bytes[..];
                 let mut stream =
                     combine::easy::Stream(combine::stream::MaybePartialStream(buffer, !eof));
-                match combine::stream::decode_tokio(value(), &mut stream, &mut self.state) {
+                match combine::stream::decode_tokio(value(None), &mut stream, &mut self.state) {
```
djc, line 191: Just a question: I notice that there is a `decode_tokio()`. Is there another variant for other runtimes that we're not compiling in the default feature set/in CI?

djc (approved):
Seems fine to me!

redis/src/parser.rs
jaymell: I'm not sure but it does look like `combine` may have an alternative decode macro for `async-std`.

PR #725, opened 2022-11-24
Implement support for Rustls - v2

### LINKED ISSUE #574: feautre request: support rustls
Sometimes I have to cross-compile my project which uses redis with tls feature enabled and it causes a lot of pain and suffering due to openssl issues

It seemd that using rustls instead of native-tls would solve those issues

rharish101:
This PR takes over #601.

I've rebased @LeoRowan's PR onto the latest `main`. Then, I took @djc's suggestions about the feature flags, and did the following:
- Rustls features:
  - Rustls support is added under the `tls-rustls` feature flag.
  - Insecure TLS connections are permitted with Rustls using the `tls-rustls-insecure` feature flag.
  - Rustls can be instructed to use Mozilla's root certificates using the `tls-rustls-webpki-roots` feature flag.
- Rustls support for async I/O:
  - Tokio Rustls support is added under the `tokio-rustls-comp` feature flag.
  - async-std Rustls support is added under the `async-std-rustls-comp` feature flag.
- Renames for native-tls (**deprecation of older features**):
  - The native-tls feature flag has been renamed from `tls` to `tls-native-tls` (for lack of a better naming scheme that's consistent with Rustls).
  - The async-std + native-tls feature flag has been renamed from `async-std-tls-comp` to `async-std-native-tls-comp` (for consistency).
  - Both of the older flags are retained for backwards-compatibility, with deprecation notices in the README and in `Cargo.toml`.

I've also documented how to use the Rustls feature flags in the README.

Closes #601 and resolves #574.


djc:
Thanks for picking this up! Hope to be able to review this soon.

rharish101:
I just fixed a bug (a misconfiguration in the x509v3 extensions file for generating TLS certs used in testing) causing the Redis cluster tests to hang forever. Could you run the tests again?

0xAlcibiades:
This is exciting. Would like to use here: https://github.com/Alcibiades-Capital/quay/pull/41 - just saw this PR, serendipitous, because I'm trying to converge everything on tokio/rustls for sqlx, redis, http, grpc, etc.

arnaudpoullet-dkt:
Nice! I hope to see this merged soon as well to remove openssl completely from our work project :rocket:

gd87429:
Is there an ETA on this merge?

djc:
Nope, sorry. @rharish101 would you mind rebasing this once more?

rharish101:
@djc Done 👍🏽

stevethedev:
I'm checking in to see if there's any movement on rustls support.

PeterGrace:
Hi @djc , @rharish101 , what can I do to help get this PR merged?  I'd like to use this crate in a project that cannot use openssl.

djc:
I'll take a look at it tomorrow -- sorry for the slow response.

rharish101:
> Hi @djc , @rharish101 , what can I do to help get this PR merged? I'd like to use this crate in a project that cannot use openssl.

From my side, I don't think there's anything that needs to be done before code review, apart from rebasing the PR when asked. Let me know if I'm wrong, though.

redis/Cargo.toml
```diff
@@ -74,11 +82,16 @@ geospatial = []
 json = ["serde", "serde_json"]
 cluster = ["crc16", "rand"]
 script = ["sha1_smol"]
-tls = ["native-tls"]
+tls-native-tls = ["native-tls"]
+tls-rustls = ["rustls", "webpki-roots"]
```
djc, line 86: I think we should use `rustls-native-certs` as the default, and make `webpki-roots` the configurable option.

djc (approved):
This is looking pretty good to me -- apologies again for the slow review!

redis/Cargo.toml
rharish101: Just did it 👍🏽.

rharish101:
Just rebased it and made Rustls use native certs by default. The tests pass on my machine, so if the CI says it's good, feel free to merge it.

rharish101:
The failing tests seem odd. The error seems like a correctness bug, not a TLS connection issue (as you might expect from this PR).

jaymell:
Likely test flakiness, re-running

rharish101:
Great, they work now. Thanks!

henrysun007:
Will this PR be merged soon? Eager to use the rustls support.

redis/Cargo.toml
```diff
@@ -76,11 +84,16 @@ geospatial = []
 json = ["serde", "serde/derive", "serde_json"]
 cluster = ["crc16", "rand"]
 script = ["sha1_smol"]
-tls = ["native-tls"]
+tls-native-tls = ["native-tls"]
```
jaymell, line 87: Can we not rename the `tls` feature?

jaymell: Is this the only breaking change?

rharish101: This, along with the async-std + native-tls feature, are the only breaking changes.

I renamed them simply because the naming convention wouldn't be consistent with Rustls. Also, `async-std-tls-comp` and `tokio-native-tls-comp` are two features with inconsistent names (since both depend on `native-tls`). Further, using a feature called `tls` implies a default TLS implementation, which IMHO should be left up to the user (since both have their pros and cons).

jaymell: Your points are totally reasonable. I'm just a bit weary of breaking users unnecessarily just for the name change. Could we potentially "deprecate" the features, e.g., change `tls` to
```
tls = ["tls-native-tls"]
```

Likewise for `async-std-tls-comp`, add some comments nothing such, and we can remove them later?

rharish101: Okay, sure. I'll make the changes and add deprecation warnings. In which file would you prefer to add them? `lib.rs`?

jaymell: Thanks! Wherever we already mention TLS is fine (which offhand I think is just the README?) as well as above the "deprecated" features in Cargo.toml.

I don't think there is any actual way to deprecate a feature, but if there is, I guess we should do that too :laughing:.

rharish101: I added the `tls` and `async-std-tls-comp` features back, with a deprecation notice both in the README and in `Cargo.toml`.

As for adding compile-time deprecation notices, I don't think there's an official way (since https://github.com/rust-lang/cargo/issues/7130 exists), unless you want to manually add a compile-time warning with the help of a third-party crate (like [compile-warning](https://crates.io/crates/compile_warning)).

jaymell:
@rharish101 Do you mind resolving the conflict? :pray:

djc:
(Might be worth noting that rustls 0.21 has come out, but there are no matching tokio-rustls/webpki-roots versions quite yet.)

jaymell:
I'm a bit worried that, as is, our test runs will not be hitting native-tls at all with the current configuration, correct? I hate to keep adding more jobs, but I think it's probably worth it in this case.

rharish101:
Oh yeah, good catch. I'll work on it.

jaymell:
Thanks. And sorry "jobs" was not the right terminology. I was thinking in terms of Github CI for some reason, but obviously I think the solution is to add yet another run to the Makefile :sweat:

rharish101:
Test flakiness, perhaps, since the failing test is for RedisJSON, which wasn't changed at all.

jaymell:
Merged. Thanks for bringing this to the finish line!

JonasMS:
Folks, thank you for this work AND please update the _features_ section of the docs when you add features!

PR #726, opened 2022-11-25
RedisJSON related CI / Test Changes

d3rpp:
When `cargo` starts running tests, it checks if the required features are enabled.

the json tests require `serde/derive`, which is not enabled by default
so I added an entry to `dev-dependencies`, which enables the required features and allows the json tests to run in CI.

additionally, I updated the names of the tests and their internal variable names to match for consistency (e.g. `json_numincrby` => `json_num_incr_by` (this time checked)).


d3rpp:
Alright, Fixes made, ready for review.

I've got two more PRs after this:

- update docs in the `lib.rs` file to showcase the `json` feature
- #721

jaymell:
This looks good, but now that the json tests are actually running, I wonder if we're posing an unnecessary burden on other contributors by requiring the `RedisJSON` module to be present locally even for completely unrelated tests. 

Should we consider something like renaming the `test_json` commands to `test_module_json_*`, excluding them from the test runs in the Makefile, and putting them instead in a separate `test-json` or `test-module` command so they can be run separately?

djc:
Yeah, I think this change isn't quite what we want. We should not add a serde dev-dependency just to force some tests to run. If we want to run the RedisJSON tests in CI, we should pass `--features json` (or whatever) to a job in CI and make sure that this enables everything that it needs.

Would be nice if the commits here were a bit cleaner (that is, each commit does one thing and describes it in the commit message).

jaymell:
Closing this PR, as other CI issues have forced forced me to address the issues here asap. Thanks for the PR, though.

PR #730, opened 2022-11-28
Adds support for NOMKSTREAM and MINID options in xadd.

sgarciac:
The 'LIMIT' option would be still missing but this PR takes xadd closer to completeness.

### DIFF
--- a/redis/src/streams.rs
+++ b/redis/src/streams.rs
@@ -34,6 +34,59 @@ impl ToRedisArgs for StreamMaxlen {
     }
 }
 
+/// Builder options for [`xadd_options`] command.
+///
+/// [`xadd_options`]: ../trait.Commands.html#method.xadd_options
+///
+#[derive(Default, Debug)]
+pub struct StreamAddOptions {
+    /// Set MAXLEN <threshold> cmd arg.
+    maxlen: Option<StreamMaxlen>,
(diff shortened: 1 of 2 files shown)


redis/src/streams.rs
```diff
@@ -34,6 +34,59 @@ impl ToRedisArgs for StreamMaxlen {
...
 }
 
+/// Builder options for [`xadd_options`] command.
+///
+/// [`xadd_options`]: ../trait.Commands.html#method.xadd_options
+///
+#[derive(Default, Debug)]
+pub struct StreamAddOptions {
+    /// Set MAXLEN <threshold> cmd arg.
+    maxlen: Option<StreamMaxlen>,
```
djc, line 44: I think we should use snake_case names (`max_len`, `min_id`, `no_mk_stream`) for both the fields and the methods here.

djc (commented):
Looks pretty good to me, modulo some naming nits. Thanks!

jaymell:
Can we add some tests?

sgarciac:
ok :+1:

nihohit:
@sgarciac can you rebase and add tests?

nihohit:
fixed in https://github.com/redis-rs/redis-rs/pull/1361

Issue #325, opened 2020-05-28
Corrupt connections


importcjj:
Is it possible to have a such situation on an asynchronous connection?

1. `PING` query sent a `PING` to the server via a certain connection, and the server returned a `PONG` response.
2. Before reading the response,  an error (timeout, etc) occurred, and the query was interrupted, but the connection was still alive.
3. The next `GET` query sent some bytes and used the `PONG` remained above as a response from the server.

Or maybe you can look at those:

https://github.com/importcjj/mobc/issues/28
https://github.com/importcjj/mobc/issues/29

Marwes:
I can see how that would happen with `aio::Connection`. If the query_async call is interrupted before it can read the response then the next call must completely read any interrupted request before proceeding.

importcjj:
What about adding a new `req_failed` flag to every async connection. when an error occurs in the ` req_packed_command` function(either fail to write or read), this flag will be set to `True`, and then users can know the prev req of this connection was failed.

importcjj:
@Marwes for a connection pool, this flag is important for it to determine if a connection should be dropped when check-in.

tailhook:
> What about adding a new `req_failed` flag to every async connection. when an error occurs in the ` req_packed_command` function(either fail to write or read), this flag will be set to `True`, and then users can know the prev req of this connection was failed.

Better approach would be a flag `in_progress` which is raised when starting request and is cleared after response is fully received. Then in progress connections should be dropped when reach a connection pool (because future that is able to finish the request correctly is dropped).

Marwes:
This isn't because of an error though, the problem is that a timeout interrupts the connection in the middle of reading or writing  on the socket which leaves the connection in an undeterminate state.

I can see two solutions here

* Mark the connection as `request_started` before every read/write and clear the flag after the response has been fully read. If another request starts and it finds that `request_started == true` it knows that there is a request that never finished completely and there is no way to recover, just return an error. 
This is easy to implement but makes `Connection` pretty bad to work with if timeouts are used as timeouts will almost always mean that the connection is unrecoverable and a full reconnect is needed.

* Store all the state needed to continue executing a previous request inside `Connection` itself (not in the future). This includes how many bytes were written so far in a request etc so that an earlier, interrupted request can be fully sent and the response received before the next request is sent.
A bit tricker since we can't even use helpers like `write_all` but it would probably be more user friendly.

tailhook:
[...]

Issue #533, opened 2021-08-28
Feature request: Add PubSub to MultiplexedConnection


Bajix:
According to the Redis [documentation](https://redis.io/topics/client-side-caching) on server-assisted client side caching as of Redis 6 the RESP3 protocol can run queries while receiving invalidation messages on the same connection. This implies that a multiplexed connection can now be used for PubSub, the currently only Connection structs can currently receive messages.

yurenji:
+1 

I think this issue is related to  https://github.com/mitsuhiko/redis-rs/issues/509

Bajix:
Yea I suppose what that other issue really would want is PubSub on Multiplexed and not a change to regular connection PubSub. It's also possible to add support for invalidation without adding PubSub to Multiplexed, though ideally both would be supported and the latter enables the former by way of the compatibility layer designed in the Redis spec should first-class support for invalidation not take precedence.

yurenji:
I have a use case which is related to this issue: The program needs to subscribe 100-300 channels (I suppose spawn a task for each subscribe) and write received data in a HashMap (or [dashmap](https://docs.rs/dashmap/4.0.2/dashmap/)) to share state, and in the main thread another task will do some calculation based on the hashmap data.  

Is there a suggested way to implement this? Or can I implement this without a pubsub MultiplexedConnection?

sarendsen:
+1

nihohit:
https://github.com/redis-rs/redis-rs/pull/898

nihohit:
multiplexed connection now supports RESP3 pubsub. closing.

Issue #329, opened 2020-06-05
Support RESP3 protocol


Terkwood:
[Basic usage: try out the HELLO command](https://redis.io/commands/hello).

[Deeper dive: read the RESP3 repository](https://github.com/antirez/resp3).

This issue is a subset of fulfilling overall Redis 6.0 support.  See #312.

This ticket is a stub, for now.  If anyone wants to throw technical specification into the comments, we'll gladly edit the description to include the relevant details.

nihohit:
RESP3 is now mostly supported, the rest of the tasks are tracked in https://github.com/redis-rs/redis-rs/issues/749

PR #733, opened 2022-12-03
Enable discovery of aborted Connections

kvcache:
This will allow connection managers like bb8 to drop async Connections
when the Future that owned them was dropped.
See: https://github.com/djc/bb8/blob/main/redis/src/lib.rs#L78-L80

Futures drop synchronously, so there's no practical way to block them
until their response from Redis has been fully hydrated. Instead,
this change _only_ seeks to make such a situation knowable to a
ConnectionManager so it can make a choice as to what to do with the
connection.

The impact of reusing a connection in this state has no upper bound;
it violates privacy of unrelated values. This change addresses this
by adding an IO error in the Connection under ConnectionLike such
that if the connection is in some partial & unknown state it will
refuse to execute the requested commands.

By using IoError as the ErrorKind, it is believed that other pools
will be able to make a good choice about what to do when they find
a Connection in this state, no matter how it got there.

Relates to: https://github.com/redis-rs/redis-rs/issues/325

### DIFF
--- a/redis/src/aio.rs
+++ b/redis/src/aio.rs
@@ -576,9 +576,11 @@ where
             }
 
             if self.send_state != ConnectionSendState::Quiescent {
-                return Err(RedisError::from((ErrorKind::IoError, "Your connection was leaked. You might read previous commands' output.")))
+                return Err(RedisError::from((
+                    ErrorKind::IoError,
+                    "Your connection was leaked. You might read previous commands' output.",
+                )));
             }
-
 
             self.buf.clear();
             self.send_state = ConnectionSendState::Sending;
@@ -609,7 +611,10 @@ where
             }
 
             if self.send_state != ConnectionSendState::Quiescent {
-                return Err(RedisError::from((ErrorKind::IoError, "Your connection was leaked. You might read previous commands' output.")))
+                return Err(RedisError::from((
+                    ErrorKind::IoError,
+                    "Your connection was leaked. You might read previous commands' output.",
+                )));
             }
 
             self.buf.clear();
--- a/redis/src/aio.rs
+++ b/redis/src/aio.rs
@@ -248,6 +248,25 @@ pub struct Connection<C = Pin<Box<dyn AsyncStream + Send + Sync>>> {
     // This flag is checked when attempting to send a command, and if it's raised, we attempt to
     // exit the pubsub state before executing the new request.
     pubsub: bool,
+
+    // A description of this connection's state.
+    //
+    // This is important for aborted Futures, which can leave the connection in a Sending
+    // or Receiving state. If you reuse a connection in such a state you may receive responses
+    // from previous interactions!
+    send_state: ConnectionSendState,
+}
+
+#[derive(PartialEq)]
+enum ConnectionSendState {
+    /// It's not doing anything of note. You can make requests to Redis with it.
+    Quiescent,
+    /// It's busy sending to Redis. You should let it completely drain before interacting with it.
+    /// Note that after sending, Redis is likely to reply, so draining a send queue is insufficient
+    /// to return this connection to a quiescent state.
+    Sending,
+    /// It's busy receiving responses from Redis. You should let it completely drain before interacting with it.
+    Receiving,
 }
 
 fn assert_sync<T: Sync>() {}
@@ -265,6 +284,7 @@ impl<C> Connection<C> {
             decoder,
             db,
             pubsub,
+            send_state,
         } = self;
         Connection {
             con: f(con),
@@ -272,7 +292,20 @@ impl<C> Connection<C> {
             decoder,
             db,
             pubsub,
+            send_state,
         }
+    }
+
+    /// Is this connection currently locked in an exchange with Redis?
+    ///
+    /// When a Future surrounding a Connection is dropped before completion,
+    /// the Connection may still be busy. If you are reading some large reply
+    /// and time out, the connection is not immediately reusable.
+    /// This is intended to help facilitate connection managers like bb8 know
+    /// when it is safe to reuse a connection and when it is better to
+    /// reconnect.
+    pub fn is_busy(&self) -> bool {
+        self.send_state != ConnectionSendState::Quiescent
     }
 }
 
@@ -289,6 +322,7 @@ where
             decoder: combine::stream::Decoder::new(),
             db: connection_info.db,
             pubsub: false,
+            send_state: ConnectionSendState::Quiescent,
         };
         authenticate(connection_info, &mut rv).await?;
         Ok(rv)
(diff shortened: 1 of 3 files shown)


jaymell:
Thanks for this. I'll try to provide a more thorough review soon.

kvcache:
Sorry about that messing up the first approval run @jaymell. Should be good now.
```
➜  redis-rs git:(325) cargo fmt -- --check && cargo clippy --all-targets --all-features -- -D warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
```
`make test` also passes on my laptop; it looks like that should cover the most predictable approval gotchas.

jaymell:
I'm a bit nervous about the implementation changes here, in that it will in most cases just break connections once an error occurs. It seems like this has the possibility to break existing users of this impl without an obvious recourse for what to do when things break.

On the other hand, I don't know why you would want a possibly corrupt connection under any circumstances. But I personally don't use this Connection type, so am not terribly familiar with the use cases, other than it being the only one that supports pub-sub. 

Elegant solutions are not always my strong suit :laughing:, but I'm almost tempted to just new-type the existing `Connection` in a `SimpleConnection` struct that has the steps you've defined here and possibly deprecate the original one later, or possibly introduce a `ConnectionManager` for non-multiplexed connections.

kvcache:
> I'm a bit nervous about the implementation changes here, in that it will in most cases just break connections once an error occurs. It seems like this has the possibility to break existing users of this impl without an obvious recourse for what to do when things break.

Right, but if they were using this connection during this time their command would not be receiving the responses they asked for. When the connection is reused **while its state** is in the middle of a communication with Redis the Connection cannot honor its API semantic: Function A will return Function B's results 😬

> On the other hand, I don't know why you would want a possibly corrupt connection under any circumstances. But I personally don't use this Connection type, so am not terribly familiar with the use cases, other than it being the only one that supports pub-sub.

This is the connection type that `bb8-redis` uses. Anyone who uses bb8-redis is vulnerable to this. This is around the fringes of my Rust expertise but I think the issue here is that `Connection` is not "unwind-safe." From the Rust memory perspective it's fine, but from a redisrs-redis communication perspective it's not unwind-safe.

I'm not making a new Error type, and I'm not panic-ing. Every user of this library had to consider `RedisError#ErrorKind#IoError` at some point, right? This PR only models something that should have always been an `IoError` as an `IoError`. No function signature changes, no incompatibility - just better assurances that when you call `get("key1")` you don't receive the result of `get("key2")`. I think this is a bug in Connection that should be fixed, rather than allowing users to choose `KnownBrokenConnection` versus `PatchedConnection` 😜 

> Elegant solutions are not always my strong suit 😆, but I'm almost tempted to just new-type the existing `Connection` in a `SimpleConnection` struct that has the steps you've defined here and possibly deprecate the original one later, or possibly introduce a `ConnectionManager` for non-multiplexed connections.

`bb8` and `bb8-redis` are, afaik, the right `ConnectionManager`-like thing to use for sharing this `Connection`. `bb8-redis` just can't maintain valid-state connections in its pool because it doesn't have enough information.

I had to move us back over to `MultiplexedConnection` over the weekend. Due to [the other bug that caused broken MultiplexedConnection state](https://github.com/redis-rs/redis-rs/issues/698) (another variant of logical unwind-unsafety, which I recall you are quite familiar with 😉 thanks again) and the age of the current 0.22.1 release I had to use a git hash dependency like `redis = { version = "0.22.1", features = ["tokio-comp"], git = "https://github.com/redis-rs/redis-rs", rev = "8e7e01efe4e6897752110e4866c9356416557b0e" }` =)

djc:
So I forget, why do we have the `Connection`/`MultiplexedConnection` difference? It seems like we should offer one connection type that does the right thing.

kvcache:
> So I forget, why do we have the `Connection`/`MultiplexedConnection` difference? It seems like we should offer one connection type that does the right thing.

=) I don't know the answer to that! Fwiw MultiplexedConnection is so far solid while Connection is pretty brittle. https://github.com/djc/bb8/issues/95 Looks like bb8 has had issues with it too. That said, this PR would let the bb8 binding trivially return `true` for their is_broken() method to avoid reusing chopped Connections.

I have not reviewed MultiplexedConnection's Drop-safety or its test coverage. I switched around the regression test in this PR with MultiplexedConnection and it _does not reproduce_. If it's not drop-safe, it's at least much harder to tease it into corrupting the Redis connection. I'm so far quite happy with `HEAD`'s MultiplexedConnection.

jaymell:
The biggest difference between `Connection` and `MultiplexedConnection` I believe is support for pub-sub, which has been requested previously (i.e., #533 ). My understanding is that we can't do this without support for RESP3 (requested in #329). 

I definitely agree that we should aim for one implementation here, but in the absence of that, I think using `MultiplexedConnection` is the right choice for most use cases.

jaymell:
If we release these changes without a new type, in my view it should potentially be treated as a breaking change, and we should ensure that we clearly document proper handling of a broken connection. 

Again, I don't know why anyone would _not_ want these changes, but they seem to have the potential to break existing users without an obvious process of remediating a connection that breaks.

djc:
I feel like the way forward here is to do a 0.23 that drops the current `Connection`, renaming `MultiplexedConnection` to `Connection`. I guess we could `#[deprecate]` `Connection` in 0.22?

jaymell:
I don't think we should remove `Connection`, though, without having pub-sub support on `MultiplexedConnection`. In my view it's reasonable to merge these changes, provided the means for detecting and responding to a broken connection is clearly documented. And we should probably consider it breaking and do it on 0.23.

djc:
I really don't think the changes here are the right approach, because they rely on the user to take care of cleaning up after an aborted future. That basically means every `Connection` request should check the state and clean up before issuing the actual request to be correct, right?

I think it's better to get rid of `Connection` sooner rather than later. Leaving `Connection` in a broken state after cancelled futures can easily happen and should never lead to issues.

So we "just" have to implement pub-sub support on `MultiplexedConnection`?

jaymell:
> So we "just" have to implement pub-sub support on MultiplexedConnection? 

`Connection` also has support for using `MONITOR`, which functions similarly to pub-sub.

kvcache:
I missed this conversation in my github notifications 😄 sorry for the high latency in volunteering some response.

> That basically means every Connection request should check the state and clean up before issuing the actual request to be correct, right?

Well, they can just use the connection and take the `IoError`. It's not great, but it's better than getting who-knows-what back from the Connection, and `IoError` can already happen today. If it happens to a user who is using a bare Connection, probably that user already has some kind of reconnect handler up the `?` chain?

> I think it's better to get rid of Connection sooner rather than later.

Strongly agree with this. If `Connection` were subsumed by `MultiplexedConnection` we/you could remove this Drop-unsafe self-corrupting type.

For that potential future, I'd suggest as a user my preferred naming would be to make MultiplexedConnection -> Connection, and add a type alias MultiplexedConnection = Connection annotated with `#[deprecated(since = "1.2.3", note = "MultiplexedConnection was renamed to Connection. Please use Connection instead.")]`

> but they seem to have the potential to break existing users without an obvious process of remediating a connection that breaks.

In the meantime, users of Connection (so, all of your `bb8-redis` users) are vulnerable to logical connection corruption where responses are silently coerced into malaligned api calls - which is much harder to recover from imo.
This PR will allow a 1-line interim fix in bb8-redis here: https://github.com/djc/bb8/blob/main/redis/src/lib.rs#L78-L80 replacing
```rust
    fn has_broken(&self, _: &mut Self::Connection) -> bool {
        false
    }
```
with
```rust
    fn has_broken(&self, connection: &mut Self::Connection) -> bool {
        !connection.is_busy()
    }
```
because bb8's connection wrapper reference is drop-safe and this function will definitely be called. This change will allow bb8 to make sure `Connection`s are not in a chop-dropped state where they've sent some stuff and have not consumed the response. Chopped connections will not be returned to the bb8 connection pool with this change and the connection pool will establish a new connection to replace the broken one.

bb8-redis tries to ping/pong on check out (which sometimes people disable for latency reasons) but if you happened to have a ping in your pending chopped buffer you'd pass that check too and the broken Connection is returned to the user as "valid" today.

nihohit:
closing, since `aio::Connection` was deprecated.

PR #750, opened 2022-12-28
fix typo in `redis/src/lib.rs`

geometryolife:


jaymell:
I don't think this is a typo.

PR #754, opened 2023-01-19
Add API to get whole batches from cursors

jwilm:
This solves a limitation where aggregating the results of a cursor into a vector required a O(n) move between the Vec owned by redis-rs and a Vec owned by the application.

The new API is exposed similarly to the existing async_iter() method on `Cmd` with the suffix `_batched`. The new batched cursor is used internally by the orignal AsyncIter API to reduce code duplication and also increase test coverage of the new iterator type.


jaymell:
Thanks for the PR. Do you mind rebasing this? We published a recent (and slightly breaking) change to related code that might affect this.

jwilm:
Hey @jaymell, I just rebased the work on top of main.

redis/src/cmd.rs
```diff
@@ -475,9 +528,25 @@ impl Cmd {
     #[cfg(feature = "aio")]
     #[inline]
     pub async fn iter_async<'a, T: FromRedisValue + 'a>(
-        mut self,
+        self,
```
jaymell, line 531: I guess this is technically a breaking change, yes? :sweat_smile:

jaymell:
Can we get this to pass the linter?

Issue #749, opened 2022-12-24
RESP3 support


nihohit:
Hi, the `Value` enum supports ATM only RESP2 types. I would like to also add RESP3 support - would you prefer to add that as more enum types to Value (which would be a breaking change), or using some other way?

jaymell:
I'm not clear enough on the differences to know what's appropriate at this point. But I would proceed as you think appropriate and we can iterate on what you have. Thanks for tackling this!

djc:
The bigint dependency should probably be guarded by a Cargo feature, not quite sure if we should default to having it on or or not. Perhaps we should get rid of Eq for Value instead of pulling in ordered-float?

nihohit:
Would you like RESP3 support to be hidden behind a feature flag? That will prevent build breakages, and ensure that users will opt-in to it, at least in the short term.

djc:
Hmm, not sure what would be appropriate here. Is RESP3 better for existing use cases than RESP2? How do the client and server negotiate which gets used? (Sorry, not that familiar yet with this part of the project.)

nihohit:
The negotiation happens during the [HELLO handshake](https://redis.io/commands/hello/). 
> Clients that want to handshake using the RESP3 mode need to call the HELLO command and specify the value "3" as the protover argument, like so:

djc:
I think for now RESP3 support should be opt-in on the side of the user, but I don't think it should be a Cargo feature.

djc:
Put it in the `RedisConnectionInfo` somehow?

nihohit:
That's the question on how the user determines whether to use RESP3 or not.
The question is whether the `ConnectionLike` traits need to change - 
`fn req_packed_command(&mut self, cmd: &[u8]) -> RedisResult<Value>` - if we modify `Value` and break exhaustive `match` on it, that's not a problem. If we don't want to break it, then we need to some other mechanism of returning RESP3 values.

PR #757, opened 2023-01-22
Initial implementation of RESP3

altanozlu:
Hi! 
This PR is initial implementation of RESP3, there will be more PR to support RESP3 better.
I've implemented all RESP3 types (excluding streamed types), RESP3 can be enabled per connection via `?resp3=true` addition to connection uri. 
It currently supports PubSub as RESP2 PubSub support in library but in future PRs it'll support handling normal commands and PubSub messages in one connection. #533 
Only `num-bigint` is added as dependency to support `BigNumber` type.

Changes in support of #329 and #749


redis/src/types.rs
```diff
@@ -121,9 +120,25 @@ pub enum Value {
     /// to express nested structures.
     Bulk(Vec<Value>),
     /// A status response.
-    Status(String),
+    Status(String), // maybe change to SimpleString ?
```
shachlanAmazon, line 123: +1

nihohit:
If we're making breaking changes here, we should also consider renaming the `Bulk` and `Data` enum variants. 
What is called `Bulk` in `Value` maps to what the Redis RESP protocol calls an array, and `Data` ironically is called Bulk string in RESP. IMO `Value::Bulk` should be called `Value::Array` and `Value::Data` should be called `Value::Bulk`, in order to match the Redis terminology.

https://redis.io/docs/reference/protocol-spec/

jaymell:
I'm unsure about the breaking changes. I think it's probably reasonable given the plainly wrong names for some of the the data types, but we should probably have a new name for the variant entirely rather than make confusion worse for users of this lib by changing the meaning of `Value::Bulk`, for example.

jaymell:
Thanks for the PR, btw. This will take some time to review as I come up to speed on RESP3 but eager to see how this turns out as we definitely need the functionality.

altanozlu:
hello @jaymell, now it passes all tests for both resp2 and resp3.

redis/tests/test_basic.rs
```diff
@@ -1100,11 +1139,11 @@ fn test_zrandmember() {
     let results: Vec<String> = con.zrandmember(setname, Some(-5)).unwrap();
     assert_eq!(results.len(), 5);
 
-    let results: Vec<String> = con.zrandmember_withscores(setname, 5).unwrap();
```
altanozlu, line 1103: old Vec<String> test works on resp2 but, doesn't work on resp3.

jaymell:
@nihohit Since you also have interest in RESP3, would you mind reviewing these changes? I'm tied up with trying to get the async-cluster changes merged so would appreciate another pair of eyes if possible here. :pray:

nihohit:
@jaymell will do!

redis/tests/support/mod.rs
```diff
@@ -336,6 +339,40 @@ where
         }
         Value::Okay => write!(writer, "+OK\r\n"),
         Value::Status(ref s) => write!(writer, "+{s}\r\n"),
+        Value::Map(ref values) => {
+            write!(writer, "%{}\r\n", values.len())?;
```
nihohit, line 343: notice that Bulk, Map, Set, and Push all use the same logic, with the difference of a single char in the beginning. Maybe try reducing the repetition?

redis/src/types.rs
```diff
@@ -121,9 +120,23 @@ pub enum Value {
...
+    /// Ordered map value from the server. use `as_map_iter` function.
+    Map(Vec<Value>),
+    /// Unordered set value from the server.
+    Set(Vec<Value>),
+    /// A floating number response from the server.
+    Double(f64),
+    /// A boolean response from the server.
+    Boolean(bool),
+    /// First String is format and other is the string
+    VerbatimString(String, String),
```
nihohit, line 135: instead of the first string, maybe use an enum, since there are only 2 legal values - text and markdown.

redis/src/types.rs
```diff
@@ -121,9 +120,23 @@ pub enum Value {
...
+    /// Very large number that out of the range of the signed 64 bit numbers
+    BigNumber(String),
```
nihohit, line 137: IMO [BigInt](https://crates.io/crates/num-bigint) is more fitting here.

redis/src/types.rs
```diff
@@ -121,9 +120,23 @@ pub enum Value {
...
+    /// Push data from the server.
+    Push(Vec<Value>),
```
nihohit, line 139: I'm trying to verify this, but it seems like the first value in the push data can be an enum too - the current wording implies that the only valid types are PubSub and Monitor. Not sure whether we want to use this in the type or not.

redis/src/parser.rs
```diff
@@ -152,13 +159,95 @@ where
...
+                        _ => Err(StreamErrorFor::<I>::message_static_message(
+                            "Expected boolean, got garbage",
+                        )),
+                    })
+                };
+                let blob_error = || {
+                    blob().map(|line| {
+                        let desc = "An error was signalled by the server";
+                        let mut pieces = line.splitn(2, ' ');
+                        let kind = match pieces.next().unwrap() {
```
nihohit, line 205: please reuse the logic in `error()`

redis/src/parser.rs
```diff
@@ -152,13 +159,95 @@ where
                         }
                     })
                 };
-
+                let map = || {
+                    int().then_partial(move |&mut length| {
+                        let length = length as usize * 2;
```
nihohit, line 164: please reuse this logic instead of copying it.

nihohit (commented):
Very impressive work all around! And as far as I see, the parsing logic is great :)

Most of the comments involved code duplication, that I believe can be easily be reduced. From what I see, the main issues are:

-  the async clients are missing the handling of push messages
- In general, the `execute_push_messages` are empty.
- The case where the user requests RESP 3 but the server doesn't provide it needs to be handled.

Please ping me here or via my email if I can help with advancing this PR.

redis/src/types.rs
nihohit: If not an enum, at least separate it as a string. Consider adding field names, for clarity.

redis/src/types.rs
nihohit: Consider adding field names, for clarity.

redis/src/types.rs
```diff
@@ -1372,20 +1428,36 @@ macro_rules! from_redis_value_for_tuple {
...
-                    invalid_type_error!(items, "Bulk response of wrong dimension")
-                }
-
-                // this is pretty ugly too.  The { i += 1; i - 1} is rust's
-                // postfix increment :)
                 let mut rv = vec![];
                 if items.len() == 0 {
                     return Ok(rv)
                 }
-                for chunk in items.chunks_exact(n) {
```
nihohit, line 1385: can you please explain in words the changes you've made here and above? I don't understand them.

redis/tests/test_basic.rs
nihohit: please also keep the RESP2 test.

redis/src/connection.rs
```diff
@@ -998,11 +1051,22 @@ impl<'a> PubSub<'a> {
     /// The message itself is still generic and can be converted into an
     /// appropriate type through the helper methods on it.
     pub fn get_message(&mut self) -> RedisResult<Msg> {
-        loop {
-            if let Some(msg) = Msg::from_value(&self.con.recv_response()?) {
-                return Ok(msg);
-            } else {
-                continue;
+        if self.con.resp3 {
```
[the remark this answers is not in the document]
altanozlu: yeah they are same now, i was going to ask do we need to separate pub/sub connection and normal cmd connection ?

redis/src/types.rs
altanozlu: as an example test_zunionstore_weights will fail without these lines, because old code expects
`(bulk(string-data('"one"'), string-data('"5"'), string-data('"three"'), string-data('"9"'), string-data('"two"'), string-data('"10"'))` 
resp3 gives 
`bulk(bulk(string-data('"one"'), double(5.0)), bulk(string-data('"three"'), double(9.0)), bulk(string-data('"two"'), double(10.0))))`
so even items.len() % n clause will fail here.
btw i'm not good at writing macros

redis/src/types.rs
altanozlu: I think using resp3 names is better

redis/src/types.rs
altanozlu: Sure, i think using enum for push type is better

redis/src/connection.rs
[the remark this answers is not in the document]
nihohit: are you asking whether we need to keep the struct separation between pubsub and regular connection?
If I understood the question correctly, then I think that for now, yes - let's keep the separation. We need to maintain backwards compatibility with RESP2.

redis/src/types.rs
nihohit: Sorry, I wasn't clear - I meant that instead of returning a string, you should return a BigInt type, from the BigInt crate.

redis/src/types.rs
nihohit: I discussed it with a maintainer of OSS Redis, and they said that this shouldn't be an enum, but `VerbatimString` can use an enum. The first field is likely to be used by modules, so we can't limit it to a closed set of values. 
IMO the type should be `Push(String, Vec<Value>)`

redis/src/types.rs
nihohit: Well, I'm not good at reading macros, so we'll just have to learn :)

I see, so you're flattening the list and then chunking it? It seems like you have an assumption that the list either contains only `Bulk` values, or no `Bulk` values. If a list is mixed, you'll only return the flattened bulking.
If this assumption is what you want to work with, then please assert it, so that it would be documented clearly.

shachlanAmazon:
relevant issues:
https://github.com/redis-rs/redis-rs/issues/329
https://github.com/redis-rs/redis-rs/issues/749

redis-test/src/lib.rs
```diff
@@ -257,6 +257,8 @@ impl ConnectionLike for MockRedisConnection {
     fn is_open(&self) -> bool {
         true
     }
+
+    fn execute_push_messages(&mut self, _messages: Vec<Value>) {}
```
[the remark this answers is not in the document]
altanozlu: I don't plan to implement in this PR, TODO has been added.

redis/src/connection.rs
```diff
@@ -870,15 +895,20 @@ impl ConnectionLike for Connection {
         self.con.send_bytes(cmd)?;
         let mut rv = vec![];
         let mut first_err = None;
-        for idx in 0..(offset + count) {
+        let mut count = count;
```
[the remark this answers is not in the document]
altanozlu: Yea it's implemented now.

redis/src/connection.rs
```diff
@@ -897,12 +928,31 @@ impl ConnectionLike for Connection {
...
+
     fn is_open(&self) -> bool {
         self.con.is_open()
     }
 
-    fn check_connection(&mut self) -> bool {
-        cmd("PING").query::<String>(self).is_ok()
+    /// Executes received push messages from server.
+    fn execute_push_messages(&mut self, _messages: Vec<Value>) {
+        // for message in messages {
```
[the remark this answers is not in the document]
altanozlu: yep

redis/src/types.rs
```diff
@@ -216,8 +232,28 @@ impl fmt::Debug for Value {
                 }
                 write!(fmt, ")")
             }
+            Value::Push(ref values) => {
+                write!(fmt, "push(")?;
+                let mut is_first = true;
```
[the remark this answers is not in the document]
altanozlu: I've removed custom loop, i think it's cleaner now.

redis/tests/support/mod.rs
altanozlu: Now they all use same function except Push which requires special care now...

redis/src/types.rs
altanozlu: I also have added Unknown just in case

redis/src/types.rs
altanozlu: It's Vector of tuples so i don't expect to get Set or Map

altanozlu:
@nihohit I think it's ready for another round of review, take your time.

redis/tests/test_basic.rs
```diff
@@ -1139,6 +1139,14 @@ fn test_zrandmember() {
     let results: Vec<String> = con.zrandmember(setname, Some(-5)).unwrap();
     assert_eq!(results.len(), 5);
 
+    if !ctx.use_resp3 {
```
nihohit, line 1142: does this ever happen?
I meant to add a test where the context is created with resp3.

redis/src/aio.rs
```diff
@@ -598,6 +648,10 @@ where
     fn get_db(&self) -> i64 {
         self.db
     }
+
+    fn execute_push_message(&mut self, _kind: String, _data: Vec<Value>) {
+        // TODO - implement handling RESP3 push messages
```
nihohit, line 653: now that I think of it, maybe we want to use unimplemented, in order to be sure that users are aware?
`unimplemented!("RESP3 pubsub isn't implemented yet");`

what do you think?

redis/src/connection.rs
```diff
@@ -332,6 +345,8 @@ pub struct Connection {
     /// This flag is checked when attempting to send a command, and if it's raised, we attempt to
     /// exit the pubsub state before executing the new request.
     pubsub: bool,
+
+    resp3: bool,
```
nihohit, line 349: please add documentation

redis/src/connection.rs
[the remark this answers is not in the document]
nihohit: it's hard to compare via the diff page - is the logic the same in aio.rs and here? if so, please move it to a function.

redis/src/connection.rs
```diff
@@ -1027,10 +1095,18 @@ impl<'a> Drop for PubSub<'a> {
 /// connection.  It only contains actual message data.
 impl Msg {
     /// Tries to convert provided [`Value`] into [`Msg`].
+    #[allow(clippy::unnecessary_to_owned)]
```
nihohit, line 1098: why is this necessary?

nihohit (commented):
final comments. great work IMO.

redis/tests/test_basic.rs
altanozlu: i've added RESP3 to `make test` so it's valid

redis/src/aio.rs
altanozlu: then it'll panic when any push message received 🤔

redis/src/connection.rs
altanozlu: It's a false positive recommended change won't build or work.

redis/src/aio.rs
nihohit: right - that way people will be aware of the fact that pubsub isn't supported, instead of trying to understand why no message was received.
@jaymell, what's your opinion? silently ignoring, or loudly panicing?

redis/src/aio.rs
```diff
@@ -347,25 +353,42 @@ where
...
+                    match kind.bytes().next() {
+                        Some(b'u') => received_unsub = true,
+                        Some(b'p') => received_punsub = true,
+                        _ => (),
+                    }
+                    if let Value::Int(num) = data[1] {
+                        if received_unsub && received_punsub && num == 0 {
+                            break;
+                        }
+                    } else {
```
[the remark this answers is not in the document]
altanozlu: It was my bad assumption, it should be fine now

altanozlu:
@jaymell what about changing Value::Bulk to Value::Array, Value::Data to Value::BlobString, Value::Status to Value::SimpleString, Value::Int to Value::Number not sure about last one.

nihohit:
already opened a PR for that :)
https://github.com/redis-rs/redis-rs/pull/779

nihohit (approved):
Bravo!
@jaymell, reviewed and approved the PR. please notice that there are a couple of comments where you're tagged for your opinion on the matter.

altanozlu:
@jaymell @nihohit I've forgot about Attribute Type(https://github.com/antirez/RESP3/blob/master/spec.md#attribute-type), If we are gonna have Attribute Type all enums in `Value` should be like below: 
```rust
Bulk {
    data: Vec<Value>,
    attributes: Option<AttributesType>,
}
```
It could be embedded in every single Value according to spec.
Or we can simply ignore attribute messages.

nihohit:
@altanozlu 
Oh, good catch. I think that instead of adding an `attribute` field to all the enum types and increasing the complexity of this change, a simpler solution is to add tp `Value` an enum type 
```rs
AttributeValue {
    data: Value,
    attributes: HashMap<Value, Value>,
}
```

And sorry that I missed it before, but why does the `Map` type return a vec and not a HashMap?

altanozlu:
> @altanozlu Oh, good catch. I think that instead of adding an `attribute` field to all the enum types and increasing the complexity of this change, a simpler solution is to add to `Value` an enum type
> 
> ```rust
> AttributeValue {
>     data: Value,
>     attributes: HashMap<Value, Value>, // or Vec<(Value, Value)>
> }
> ```
> 
> And sorry that I missed it before, but why does the `Map` type return a vec and not a `HashMap<Value, Value>` or `Vec<(Value, Value)>`?

I didn't use HashMap<Value, Value> because it'll require deriving Eq, Hash but Vec<(Value, Value)> looks nice. Using AttributeValue may also increase complexity too but it's better than using attributes field in all enum types.

nihohit:
@altanozlu notice that you also didn't implement streamed types, but I think that's very sensible in order to limit the size of this PR. let's add that in another PR.

nihohit:
right, @altanozlu, so I think we'll add the `Attribute` enum type, change `Map` to use `Vec<(Value, Value)>`, and we'll call this PR done?

altanozlu:
@nihohit i didn't return Attribute instead it's parsed and only data is returned. The reason ise we expect Value::Int and it may return Attribute{data: Value::Int} so it'll raise an error. I also have changed Map to be Vec<(Value, Value)>.

altanozlu:
@nihohit i didn't return Attribute instead it's parsed and only data is returned. The reason ise we expect Value::Int and it may return Attribute{data: Value::Int} so it'll raise an error. I also have changed Map to be Vec<(Value, Value)>.

nihohit:
>  The reason ise we expect Value::Int and it may return Attribute{data: Value::Int}

I'm sorry, I don't understand what you mean here. Can you give me an example of a valid attribute that will cause an error?

altanozlu:
> > The reason ise we expect Value::Int and it may return Attribute{data: Value::Int}
> 
> I'm sorry, I don't understand what you mean here. Can you give me an example of a valid attribute that will cause an error?

In `test_attributes` we expect array of integers, if we give real value inside Value::Attribute as user of this library i'll get error because i expect Value::Int not Value::Attribute. Afaik we can't be sure our command replies will have attribute nested in it or not. To fix this we can make `from_redis_value` implementations recursive which calling with inner data but we will lose attribute information.

nihohit:
But `test_attribute` is an artificial case, isn't it? I might as well say that the test is incorrect because it isn't expecting the attribute (it definitely is lacking in documentation, explaining why the attribute is ignored).  
I would prefer to let the user opt-in to lose information, rather than making the decision for them. Maybe if we parse attributes, but add a helper function that recursively removes attributes.
```rs
fn remove_attribute(value: Value) -> Value {
  match value {
    Attribute(data, _attribute) => data,
    Data(data) => Data(data),
    Map(vec) => vec.iter().map((key, value) => (remove_attribute(key), remove_attribute(value)),
    ...
  }
}
```

Although this can be done easily by the user, it doesn't have to be in-library.

altanozlu:
> But `test_attribute` is an artificial case, isn't it? I might as well say that the test is incorrect because it isn't expecting the attribute (it definitely is lacking in documentation, explaining why the attribute is ignored). I would prefer to let the user opt-in to lose information, rather than making the decision for them. Maybe if we parse attributes, but add a helper function that recursively removes attributes.
> 
> ```rust
> fn remove_attribute(value: Value) -> Value {
>   match value {
>     Attribute(data, _attribute) => data,
>     Data(data) => Data(data),
>     Map(vec) => vec.iter().map((key, value) => (remove_attribute(key), remove_attribute(value)),
>     ...
>   }
> }
> ```
> 
> Although this can be done easily by the user, it doesn't have to be in-library. Another, more effecient choice would be to pass a flag to the parser to ignore attributes and behave as you implemented, but that sounds like a large change, deserving of its own PR. I recommend to full parse attributes, maybe add `remove_attribute`, and let the user decide how to handle the additional choice, instead of making the choice for them.

I've changed test_attributes now which covers both usage types and we don't need any flag. I think pr is ready now if we are not going to use float_eq. Also i will post benchmark results for RESP2 with this branch and main.

altanozlu:
```
group                                                 main                                    resp3
-----                                                 ----                                    -----
decode/decode                                         1.00    657.1±4.74ns        ? ?/sec     1.11    726.9±5.88ns        ? ?/sec
encode/integer                                        1.03   150.4±20.13µs        ? ?/sec     1.00    146.1±2.46µs        ? ?/sec
encode/pipeline                                       1.08   153.5±32.26µs        ? ?/sec     1.00   142.7±13.91µs        ? ?/sec
encode/pipeline_nested                                1.03     59.4±3.93µs        ? ?/sec     1.00     57.8±0.78µs        ? ?/sec
encode/small                                          1.00    180.2±1.57ns        ? ?/sec     1.01    181.9±2.75ns        ? ?/sec
query/simple_getsetdel                                1.00     49.4±1.44µs        ? ?/sec     1.10    54.4±30.43µs        ? ?/sec
query/simple_getsetdel_async                          1.00    57.1±10.68µs        ? ?/sec     1.12    64.2±27.25µs        ? ?/sec
query/simple_getsetdel_pipeline                       1.04     18.9±3.43µs        ? ?/sec     1.00     18.2±0.48µs        ? ?/sec
query/simple_getsetdel_pipeline_precreated            1.00     17.6±0.44µs        ? ?/sec     1.00     17.6±0.32µs        ? ?/sec
query_pipeline/async_long_pipeline                    1.00  1952.5±380.75µs        ? ?/sec    1.34      2.6±0.45ms        ? ?/sec
query_pipeline/long_pipeline                          1.00      2.4±0.46ms        ? ?/sec     1.12      2.6±0.44ms        ? ?/sec
query_pipeline/multiplexed_async_implicit_pipeline    1.00      2.1±0.45ms        ? ?/sec     1.35      2.9±0.32ms        ? ?/sec
query_pipeline/multiplexed_async_long_pipeline        1.00      2.2±0.49ms        ? ?/sec     1.21      2.7±0.44ms        ? ?/sec
```
bringing RESP3 slowed down redis-rs

nihohit:
> I've changed test_attributes now which covers both usage types and we don't need any flag.

This is an excellent compromise. Well done.

>I think pr is ready now

I agree. @jaymell @djc I think this PR is ready.

> bringing RESP3 slowed down redis-rs

that's an interesting result, and it should be throughly investigated. Are you comparing the result to current `main`, or to the commit that you branched from ([6d0ec7cd714f417353d6bf536c2f28abae66e61a](https://github.com/altanozlu/redis-rs/commit/6d0ec7cd714f417353d6bf536c2f28abae66e61a))?

altanozlu:
> 

It was the commit i've branched from.
I've rerun both resp3 and main(it's redis-rs/redis-rs now)
```
group                                                 main                                    resp3
-----                                                 ----                                    -----
decode/decode                                         1.00  668.9±146.22ns        ? ?/sec     1.08   725.3±23.19ns        ? ?/sec
encode/integer                                        1.00   148.3±12.62µs        ? ?/sec     1.00    147.8±2.24µs        ? ?/sec
encode/pipeline                                       1.02   142.9±21.51µs        ? ?/sec     1.00    140.3±1.18µs        ? ?/sec
encode/pipeline_nested                                1.02     59.1±2.28µs        ? ?/sec     1.00     57.9±0.83µs        ? ?/sec
encode/small                                          1.00    176.4±3.33ns        ? ?/sec     1.02    180.0±2.24ns        ? ?/sec
query/simple_getsetdel                                1.02     49.9±0.81µs        ? ?/sec     1.00     48.8±7.85µs        ? ?/sec
query/simple_getsetdel_async                          1.00     54.5±1.69µs        ? ?/sec     1.00     54.5±4.69µs        ? ?/sec
query/simple_getsetdel_pipeline                       1.00     18.1±0.67µs        ? ?/sec     1.00     18.1±0.30µs        ? ?/sec
query/simple_getsetdel_pipeline_precreated            1.00     17.6±1.07µs        ? ?/sec     1.00     17.5±0.33µs        ? ?/sec
query_pipeline/async_long_pipeline                    1.04      2.1±0.50ms        ? ?/sec     1.00      2.0±0.44ms        ? ?/sec
query_pipeline/long_pipeline                          1.00  1746.3±400.49µs        ? ?/sec    1.14  1986.0±463.41µs        ? ?/sec
query_pipeline/multiplexed_async_implicit_pipeline    1.00  1418.1±835.03µs        ? ?/sec    1.42      2.0±0.91ms        ? ?/sec
query_pipeline/multiplexed_async_long_pipeline        1.00  1117.0±731.50µs        ? ?/sec    1.75  1956.8±382.46µs        ? ?/sec
```

btw it's M1 Macbook Pro

altanozlu:
It seems it was my macbook!
I've tried both main and resp3 on server and it seems there is only slowdown between %3-9.

server info: AMD EPYC 7B13 Linux 5.4.0-1100-gcp
```
group                                                 main                                    resp3
-----                                                 ----                                    -----
decode/decode                                         1.00  1329.9±161.58ns        ? ?/sec    1.03  1370.9±79.60ns        ? ?/sec
encode/integer                                        1.00   252.2±20.99µs        ? ?/sec     1.09   275.6±28.94µs        ? ?/sec
encode/pipeline                                       1.00   246.6±42.76µs        ? ?/sec     1.00   246.7±18.25µs        ? ?/sec
encode/pipeline_nested                                1.00    93.8±10.26µs        ? ?/sec     1.03    96.9±10.73µs        ? ?/sec
encode/small                                          1.00   235.6±24.06ns        ? ?/sec     1.05   246.7±21.70ns        ? ?/sec
query/simple_getsetdel                                1.00   104.3±13.90µs        ? ?/sec     1.03   107.8±12.89µs        ? ?/sec
query/simple_getsetdel_async                          1.01   115.6±19.00µs        ? ?/sec     1.00   114.1±12.14µs        ? ?/sec
query/simple_getsetdel_pipeline                       1.00     37.1±4.56µs        ? ?/sec     1.02     37.8±4.90µs        ? ?/sec
query/simple_getsetdel_pipeline_precreated            1.00     37.7±5.00µs        ? ?/sec     1.01     38.0±4.02µs        ? ?/sec
query_pipeline/async_long_pipeline                    1.01   818.4±38.66µs        ? ?/sec     1.00   814.1±47.39µs        ? ?/sec
query_pipeline/long_pipeline                          1.00   817.7±56.78µs        ? ?/sec     1.01   828.0±69.84µs        ? ?/sec
query_pipeline/multiplexed_async_implicit_pipeline    1.00  1283.9±125.10µs        ? ?/sec    1.00  1284.6±112.18µs        ? ?/sec
query_pipeline/multiplexed_async_long_pipeline        1.00   833.8±41.45µs        ? ?/sec     1.02   850.5±54.70µs        ? ?/sec
```

altanozlu:
@jaymell can you re-run failed tests ?

redis/src/commands/mod.rs
```diff
@@ -2057,3 +2058,20 @@ impl ToRedisArgs for Direction {
         out.write_arg(s);
     }
 }
+
+/// Creates HELLO command for RESP3 with RedisConnectionInfo
+pub fn create_hello_command(connection_info: &RedisConnectionInfo) -> Cmd{
```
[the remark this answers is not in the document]
jaymell: Is there a reason this was resolved w/out a change? I was thinking `resp3_hello_command` or just `resp3_hello` would be preferable.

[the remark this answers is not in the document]
altanozlu: @jaymell it's changed to resp3_hello now, i forgot this.

redis/src/aio.rs
```diff
@@ -572,18 +605,28 @@ where
             }
 
             let mut rv = Vec::with_capacity(count);
-            for _ in 0..count {
+            let mut count = count;
```
jaymell, line 608: Is it ever possible we'd need the same push-command handling logic on `req_packed_command`?

altanozlu: Yep and now it's added to both aio & sync connections also tests been updated.

redis/src/aio.rs
```diff
@@ -573,7 +573,12 @@ where
             self.buf.clear();
             cmd.write_packed_command(&mut self.buf);
             self.con.write_all(&self.buf).await?;
-            self.read_response().await
+            loop {
```
nihohit, line 576: please also implement for MultiplexedConnection

jaymell: I'm ok if we leave this as a `TODO` for now (adding explicit comments for such would be helpful) and tackle in a separate PR since `MultiplexedConnection` doesn't currently support pub/sub. I'd like to keep this PR about where it is in terms of changes (though subject to change, I haven't finished reviewing :laughing:)

jaymell:
@nihohit @altanozlu I wanted to ask y'all what if anything you're using as a reference for assessing the correctness of the RESP3 implementation. The documentation [here](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md) is helpful but is also inaccurate in places. I've thus far tried to avoid the redis source code and have mostly been resorting to looking at the data on the wire, but wondering if y'all had other suggestions.

redis/src/aio.rs
nihohit: While you're right that MultiplexedConnection doesn't implement pub/sub, that isn't the only situation where PUSH might be used - for example I believe that client-side caching (thanks @altanozlu for starting the conversation) is implemented using that.

nihohit:
@jaymell I'm waiting for this PR to implement features, so can't evaluate correctness ATM.

altanozlu:
> @nihohit @altanozlu I wanted to ask y'all what if anything you're using as a reference for assessing the correctness of the RESP3 implementation. The documentation [here](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md) is helpful but is also inaccurate in places. I've thus far tried to avoid the redis source code and have mostly been resorting to looking at the data on the wire, but wondering if y'all had other suggestions.

Most of the time i use https://github.com/rueian/rueidis as a reference which we have been using in production for months.

altanozlu:
The reason to test fails is pub/sub waiting for `subscribe` command's response but RESP3 sends it as push message.

nihohit:
Sounds like we can't use resp3 in pubsub yet, so the test shouldn't use
resp3.

altanozlu:
We can use resp3 in pubsub with newest commits and tests should pass now sorry i didn't test all before pushing commits. @jaymell If we can't test correctness of it better, should we use a branch for resp3 in redis-rs so we can open pr for it but also it'll mature better without messing main branch.

jaymell:
I'm hoping to avoid a separate branch for now, though it's definitely a possibility. My main concern at the moment is knowing what features are needed/missing in order to release something that actually meets the spec. It might be good to track that list on an issue.

redis/src/aio.rs
jaymell: Makes sense. We'll need additional PRs before releasing these changes but think this particular PR is right sized for now.

altanozlu:
> I'm hoping to avoid a separate branch for now, though it's definitely a possibility. My main concern at the moment is knowing what features are needed/missing in order to release something that actually meets the spec. It might be good to track that list on an issue.

https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md
- Streamed aggregated data types
- Streamed strings
```
1.4, 8 Dec 2022, Normalize NaN to a single representation "nan" and forbid "-nan" (Effective since Redis 7.2).
1.5, 5 Jan 2023, Mark the map type as being unordered rather than ordered (Effective since Redis 6.2)
1.6, 9 March 2023, Allow doubles in scientific E notation (Effective since Redis 6.0).
```
I didn't test or implement those.

nihohit:
Added checklist to https://github.com/redis-rs/redis-rs/issues/749. 
@altanozlu can you please fix the documentation of Map? It currently says that the map is ordered, should be unordered. I believe that it's the only thing to do to check `Mark the map type as being unordered rather than ordered` as done.

nihohit:
Thanks @altanozlu. This PR leaves some additional tasks, no? push messages need handling, and what else?

altanozlu:
I think this is it.

redis/src/connection.rs
```diff
@@ -810,17 +817,34 @@ impl Connection {
         // messages are received until the _subscription count_ in the responses reach zero.
         let mut received_unsub = false;
         let mut received_punsub = false;
-        loop {
-            let res: (Vec<u8>, (), isize) = from_redis_value(&self.recv_response()?)?;
+        if self.resp3 {
```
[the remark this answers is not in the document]
jaymell: Why was this resolved?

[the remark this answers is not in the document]
nihohit: repeated logic was moved to `is_pub_sub_state_cleared`. Since the async logic requires using async functions, the whole block couldn't be implemented in the same way.

redis/src/parser.rs
```diff
@@ -351,6 +494,79 @@ mod tests {
         assert_eq!(codec.decode_eof(&mut bytes), Ok(None));
     }
 
+    #[test]
+    fn decode_resp3_double() {
+        let val = parse_redis_value(b",1.23\r\n").unwrap();
```
jaymell, line 499: Could we add a couple additional tests here, e.g., scientific notation / infinity, just for my paranoia's sake? :sweat_smile:

nihohit: also NaN / -NaN, please.

altanozlu: @nihohit can u update #749 for scientific E notation ?

redis/src/parser.rs
```diff
@@ -351,6 +495,114 @@ mod tests {
...
+        //Allow doubles in scientific E notation
+        let val = parse_redis_value(b",2.67923e+8\r\n").unwrap();
+        assert_eq!(val, Value::Double(267923000.0));
+        let val = parse_redis_value(b",2.67923E+8\r\n").unwrap();
+        assert_eq!(val, Value::Double(267923000.0));
+        let val = parse_redis_value(b",-2.67923E+8\r\n").unwrap();
+        assert_eq!(val, Value::Double(-267923000.0));
+        let val = parse_redis_value(b",2.1E-2\r\n").unwrap();
+        assert_eq!(val, Value::Double(0.021));
+        // infinity support is not specified by redis, but we can support it!
```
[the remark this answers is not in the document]
altanozlu: ohh didn't see!, i'll remove

redis/src/types.rs
```diff
@@ -104,9 +104,8 @@ pub enum ErrorKind {
     /// Error Serializing a struct to JSON form
     Serialize,
 }
-
 /// Internal low-level redis value enum.
-#[derive(PartialEq, Eq, Clone)]
+#[derive(PartialEq, Clone)]
```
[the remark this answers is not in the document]
jaymell: I think we should just remove it for now. I'm not very familiar with `float_eq` and am not sure what benefit it would provide for us at this point. We can add later if we have a use case.

redis/src/types.rs
```diff
@@ -124,20 +128,83 @@ pub enum Value {
...
+    Push {
+        /// Push Kind
+        kind: String,
+        /// Remaining data from push message
+        data: Vec<Value>,
+    },
+}
+
+#[derive(PartialEq, Clone, Debug)]
+pub enum VerbatimFormat {
```
jaymell, line 165: I've seen where markdown and txt were specifically mentioned in the RESP3 spec, but I wasn't clear (and haven't as of yet gone and looked at the redis source) whether that was intended as a comprehensive list or if they were just examples. Either way, I wanted to ask whether we should consider leaving this as a `String` just to avoid unnecessary breaking changes down the line if more variants are needed.

altanozlu: txt is already in redis code and mkd is mentioned. I think this style of unknown and Txt, Mkd is okay.

nihohit: I discussed this with @madolson (a redis core member), and she said that these can be treated as a closed set. IMO the solution @altanozlu implemented has the best of both worlds - it's both an easy to use closed-set enum for most cases, and an open String for the edge cases of new formats, which forces the user to be aware of the fact that they received a non-standard format string.

jaymell: How about the `kind` field for `Push` types? I've found even less clarity in the docs/source as relates to this one but wanted to hear yall's thoughts on whether this might merit an enum?

nihohit: https://github.com/redis-rs/redis-rs/pull/757#discussion_r1119886733

I raised that too with @madolson. She said that the `Push` messages are more likely to be used by modules, so using a String as an open-ended format is more fitting. IIRC while modules might add new values to `VerbatimFormat`, the likelihood of using a new textual format is lesser.
IMO if you want consistency between `VerbatimString` and `Push`, I'd prefer that both be enums rather than strings. It means that matching on the common case is easier, even though it does mean that if there will be multiple modules giving multiple values for push or format the user will need to use nested match statements.

altanozlu: I think `kind` could be an enum with Other(String) variant.

madolson: Enum with Other seems pretty sane for both types to me fwiw. We really aren't sure what the "kind" of push notifications will end up existing eventually. I mentioned modules might be able to send arbitrary kind messages, but it's possible we might force modules to declare themselves as kind module + their name to be more standardized. If you have any specific thoughts I can make a note of it for future conversations.

jaymell: Thanks for the clarifications. If we were to make Push `kind` an enum as well, what would the variants be? I'm trying to remember, but I feel like I've never actually seen anything but "message" (I believe the RESP3 spec had previously mentioned others, but I see it's been updated recently). 

I'm not a huge fan of enums with a catch-all variant, but I'm willing to be overruled there, and I think it definitely seems reasonable for `VerbatimString`.

altanozlu: Isn't adding new variant is major breaking change ? If it's then we need Other variant otherwise we need to bump major version every time a Redis module brings new kind to the table.

jaymell: > Isn't adding new variant is major breaking change ?

Yes, and wasn't intending for enums without a catch-all. If anything I think `String` may be preferable for `kind`. We could also consider annotating the enums with `non_exhaustive`, though I'm not sure that's desirable either.

altanozlu: I'd like to implement it as enum with other kind. Btw can you give an estimation for review progress ?

jaymell: If you want to change it, go for it. If @madolson thinks the enum approach is reasonable, it works for me.

I'm almost done with the review, and overall it looks good and I don't think any major changes will be required. I think you're safe to stack additional PRs on top of this without having to worry about significant conflicts down the line. However, I'm not ready to merge this yet and want to finish up some other things before bringing in breaking changes. Thanks again for your patience -- this is definitely important work!

redis/tests/test_async.rs
```diff
@@ -97,6 +97,29 @@ fn test_pipeline_transaction() {
     .unwrap();
 }
 
+#[test]
+fn test_push() {
```
jaymell, line 101: Could we give this and the other `test_push` test more descriptive names? I'm only 30 confident sure I know what this test is testing :sweat_smile:

altanozlu: :D u are right, is it ok now ? i don't think parse tests needs more descriptive names.

jaymell:
@altanozlu With renewed apologies for the continued delay, I'm about ready to approve this, but my bigger concern is what comes next. I'm thinking we probably shouldn't release this as-is without addressing the other items on #749, so as mentioned previously I am reluctant to put this on main branch at this point. Any thoughts on the remaining items? 

I'm thinking (without having given it enough thought, honestly) that we could _potentially_ defer the streamed types to a later release, but I think at a minimum we should have functional Push handling in place before this is releasable.

altanozlu:
> @altanozlu With renewed apologies for the continued delay, I'm about ready to approve this, but my bigger concern is what comes next. I'm thinking we probably shouldn't release this as-is without addressing the other items on #749, so as mentioned previously I am reluctant to put this on main branch at this point. Any thoughts on the remaining items?
> 
> I'm thinking (without having given it enough thought, honestly) that we could _potentially_ defer the streamed types to a later release, but I think at a minimum we should have functional Push handling in place before this is releasable.

I think it's ok to wait until dealing with push messages is done. @nihohit i think to remove execute_push_message and make caching handler more exposed. Because currently other than invalidate message it's not going to bu used.

nihohit:
> I think it's ok to wait until dealing with push messages is done. @nihohit i think to remove execute_push_message and make caching handler more exposed. Because currently other than invalidate message it's not going to bu used.

what's the caching handler?
IMO the first question to ask is what do we want to do with PUSH messages? do we allow the user to set a channel and receive updates over it? a callback? Should every connection type handle PUSH messages, or do we limit it to explicit pub/sub types?

altanozlu:
> > I think it's ok to wait until dealing with push messages is done. @nihohit i think to remove execute_push_message and make caching handler more exposed. Because currently other than invalidate message it's not going to bu used.
> 
> what's the caching handler? IMO the first question to ask is what's the UX we want for PUSH messages - how will users consume them? do we allow the user to set a channel and receive updates over it? a callback? Should every connection type handle PUSH messages, or do we limit it to explicit pub/sub types?

Caching Handler is just a trait with get/key/remove,I think users don't need to handle push messages explicitly but a channel could be added later on. There is only `invalidate` message type which is not pub/sub, so i don't think it needs `execute_push_message`.

nihohit:
So how will the user receive and use the caching handler? Will they implement the trait and then set the object somewhere?

altanozlu:
> So how will the user receive and use the caching handler? Will they implement the trait and then set the object somewhere?

only thing they can do is provide cache key(which decides cache it or don't) and cache handler 'll handle it. The library can offer some default implementations(e.g. cached crate).

nihohit:
Oh, I'm sorry, I completely misunderstood the context. You're talking about local cache :)
Local cache is one situation in which we get PUSH commands, but there might be others - specifically, modules leave room for many pushes. We need some mechanism that will handle any PUSH response, not only cache invalidation.

altanozlu:
> Oh, I'm sorry, I completely misunderstood the context. You're talking about client side caching, right? :) Local cache is one situation in which we get PUSH commands, but there might be others - specifically, modules leave room for many pushes. We need some mechanism that will handle any PUSH response, not only cache invalidation.

I agree and execute_push_message is not a solution for this, a channel with size would be better for this. What do you think @jaymell ? If user provides an option, library can send push messages to it.

jaymell:
I'm not clear on anything related to caching at this point so might be missing something, but a few thoughts:

* As far as UX, I like the idea of (on the async side, anyway) returning a `Stream` as `aio::Connection` currently does. Obviously behind the scenes it can be based on channels but it would be nice to leave that as an implementation detail if possible. 

* I think getting rid of the separate `PubSub` types would be ideal if possible. Are there still cases where we'd need to "exit" pubsub mode? If not, it would be great to just use the Connection directly.

* I'm not clear why we would remove `execute_push_message`? Maybe something like `route_push_message` would be more apt, but we'd still want some method here to route the commands to various subscribers (which could just be implemented as a map of channels?), right?

altanozlu:
> @altanozlu please retarget the PR to the `resp3` branch (https://github.com/redis-rs/redis-rs/tree/resp3), so that we can merge it there, and allow other PRs to rebase over this.

sure i think it's ready to merge into resp3

jaymell:
@altanozlu One final request before I merge. Can you clean up the top-level comment and provide something better for the commit message? :pray: Thanks!

altanozlu:
> @altanozlu One final request before I merge. Can you clean up the top-level comment and provide something better for the commit message? 🙏 Thanks!

Is it ok now ? I think `Basic RESP3 Support` is ok

jaymell:
Thanks @altanozlu. I took the liberty of removing the "resolves" line from the description -- I think we should keep these open until we merge this to to main, ideally with most/all of the remaining RESP3 TODO items.

Thanks again for the great work!

PR #758, opened 2023-01-23
Restore inherent `ClusterConnection::check_connection` method

robjtede:
Reverts breaking change made in 85e92c8ce8c24fc9dd057f39e48304a263e902d3.

Cannot be deprecated since it shares it's name with the trait method and the inherent method will always be preferred.

`doc(hidden)` as a somewhat opinionated addition

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The crate is called `redis` and you can depend on it via cargo:
 
 ```ini
 [dependencies]
-redis = "0.22.2"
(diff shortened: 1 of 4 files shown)


djc (approved):

djc:
This has been published, thanks @jaymell! In this case because of the apparently noticeable semver breakage, I've yanked 0.22.2.

PR #761, opened 2023-01-27
Allow disabling `default` user in tests

jaymell:
The goal is to strengthen tests of authentication
by ensuring that auth is actually required during
the test run, else the test could inadvertently pass 
if credentials aren't actually passed for some reason.

Also factor out some repeated test code into common method.

### DIFF
--- a/redis/tests/support/mod.rs
+++ b/redis/tests/support/mod.rs
@@ -201,13 +201,13 @@ impl RedisServer {
         }
     }
 
-    pub fn get_client_addr(&self) -> &redis::ConnectionAddr {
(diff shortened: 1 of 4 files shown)


redis/tests/support/mod.rs
```diff
@@ -205,6 +205,13 @@ impl RedisServer {
         &self.addr
     }
 
+    pub fn get_conn_info(&self) -> redis::ConnectionInfo {
```
djc, line 208: I think we should drop the `get_()` prefix (per the API guidelines). For a public method I would probably expand `conn` to `connection`, in line with the type name.

djc (approved):

Issue #763, opened 2023-01-30
set_multiple not obviously == MSET


randomairborne:
There is no way to MSET override using this library- the Commands mset_nx cannot replace, and the args() method on redis::Cmd does not support lists of tuples, which would be much nicer.

djc:
What about [`set_multiple()`](https://docs.rs/redis/latest/redis/trait.Commands.html#method.set_multiple)?

randomairborne:
Oh, I completely missed that. Would be cool if it was documented as mset, so it was easier to search for. Sorry.

djc:
If you'd be able to submit a documentation improvement as a PR, that would be great!

djc:
I'd be on board with adding `mset()` and deprecating `set_multiple()` in favor of it.

randomairborne:
I’d ditto this. I feel like the functions should follow redis naming.

djc:
@randomairborne want to change up your PR? Sorry for the churn!

randomairborne:
PR changed up

jaymell:
Fixed by #766

PR #766, opened 2023-01-31
Rename set_multiple to mset

randomairborne:
This PR renames set_multiple to mset, as per #763.


redis/src/commands/mod.rs
```diff
@@ -87,11 +87,21 @@ implement_commands! {
         cmd("SET").arg(key).arg(value)
     }
 
+<<<<<<< HEAD
```
jaymell, line 90: Can we fix the conflict?

And maybe use active instead of passive voice in the deprecation note :smile: ?

redis/src/commands/mod.rs
```diff
@@ -87,11 +87,21 @@ implement_commands! {
...
     /// Sets multiple keys to their values.
+    #[deprecated(since = "0.24.0", note = "set_multiple was determined to be a poor name. Please use mset instead")]
```
djc, line 92: I'd leave the opinionating out of here ("poor name")`, maybe something like `"Renamed to mset() to more closely match the Redis command", or something.

redis/src/commands/mod.rs
```diff
@@ -88,10 +88,16 @@ implement_commands! {
     }
 
     /// Sets multiple keys to their values.
+    #[deprecated(since = "0.24.0", note = "Renamed to mset() to reflect Redis name")]
```
djc, line 91: Sorry, one more nit: the initial deprecated release is likely to be 0.22.4, not 0.24.0.

djc (approved):

djc:
Would you mind rebasing this, and ideally also squashing the commits back into a single one?

djc (approved):

randomairborne:
Can the test dependency error be changed somehow?

randomairborne:
Do i need to manually backdate CSV?

PR #772, opened 2023-02-07
[2/5] [8] cluster: keep all params in ClusterParams struct & use it to pass all cluster related params

0xWOF:
This is pull request make user can control every parameters of ClusterConnection with builder pattern

## Changes

- cluster
    - remove every parameters from ClusterConnection
    - instead, handle parameters by CusterParams
    - move validate_duration logic from ClusterConnection to CusterParams
- cluster_client
    - add parameters from ClusterConnection to ClusterClientBuilder

## Effect

- cluster:
    - breaking change: ClusterConnection's struct member is changed
        - It is breaking change but I think it will not break user's code because almost user will create ClusterConnection with new function
- cluster_client
    - public api is added: user can control ClusterConnection's parameters with builder

## Note

- this PR is from https://github.com/redis-rs/redis-rs/pull/670

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -41,7 +41,6 @@
 use std::cell::RefCell;
 use std::collections::BTreeMap;
 use std::iter::Iterator;
-use std::str::FromStr;
 use std::thread;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


jaymell:
Just a note that I want to hold off on merging any additional changes to the cluster code until we get the async-cluster branches merged. I think these changes are potentially fine but don't want to have additional conflicts until after that's done.

0xWOF:
Hello @jaymell , if you prefer me to wait until your async cluster support pull request is merged, please feel free to comment to this pull request.

0xWOF:
@jaymell I totally agree with you. I'll wait until your async cluster support pull request is merged, and then continue to create pull request. thank you for your comment.

jaymell:
We can keep the PR open if you'd like, no issue either way, so up to you.

PR #778, opened 2023-02-09
feat(commands): support xautoclaim

### LINKED ISSUE #775: Support XAUTOCLAIM
https://redis.io/commands/xautoclaim

kkharji:
closes #775 

- [ ] add tests

cc @djc

### DIFF
--- a/redis/src/streams.rs
+++ b/redis/src/streams.rs
@@ -296,10 +296,14 @@ pub struct StreamClaimReply {
 ///
 #[derive(Default, Debug, Clone)]
 pub struct StreamAutoClaimReply {
+    /// A stream ID to be used as the <start> argument for
+    /// the next call to [`xautoclaim`](../trait.Commands.html#method.xautoclaim).
+    pub start_stream_id: String,
     /// Complex data structure containing a payload for each ID in this array
     pub ids: Vec<StreamId>,
-    /// Last id
-    pub last_id: String,
+    /// An array containing message IDs that no longer exist in the stream,
+    /// and were deleted from the PEL in which they were found.
+    pub deleted_ids: Vec<String>,
 }
 
 /// Reply type used with [`xpending`] command.
@@ -568,7 +572,7 @@ impl FromRedisValue for StreamClaimReply {
 
 impl FromRedisValue for StreamAutoClaimReply {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
-        let (last_id, rows, _unknown): (
+        let (start_stream_id, rows, deleted_ids): (
             String,
             Vec<HashMap<String, HashMap<String, Value>>>,
             Vec<String>,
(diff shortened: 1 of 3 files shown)


redis/src/streams.rs
```diff
@@ -510,6 +570,27 @@ impl FromRedisValue for StreamClaimReply {
     }
 }
 
+impl FromRedisValue for StreamAutoClaimReply {
+    fn from_redis_value(v: &Value) -> RedisResult<Self> {
+        let (start_stream_id, rows, deleted_ids): (
+            String,
+            Vec<HashMap<String, HashMap<String, Value>>>,
+            Vec<String>,
+        ) = from_redis_value(v)?;
```
kkharji, line 579: When just_id option is provided, this fails. Should we:

1. Provide empty map
	1. Parse rows as `Vec<Value>`,
	2. Check if Value can be parsed as `HashMap<_, _>`
	3. Parse as `HashMap<>` if so, else parse as `Vec<string>` then create `Vec<StreamId>` 
2. Add warning to `StreamAutoClaimReply` that it would fail when just ids is provided

option 1 seems what I would expect as API consumer

kkharji:
NOTE: on v6.2.8, deleted_ids is not returned, and would cause the parsing of xautoclaim reply to fail. Not sure what should be done in the redis version is less 7.

saiintbrisson:
Hi @kkharji, would you be able to sync this branch with `main` and fix the failing CI?

JMTamayo:
Hello everyone! Is there a reason why we don't continue with this implementation? Can we check it again? Is there any reason why the integration with main was not carried out?

jaymell:
@kkharji Hi, I'm not sure if you're still around. Are you able to address this PR again?

> NOTE: on v6.2.8, deleted_ids is not returned, and would cause the parsing of xautoclaim reply to fail. Not sure what should be done in that case.

It definitely needs to be backward-compatible if possible, but I think that can easily be accommodated in the `FromRedisValue` implementation.

redis/src/streams.rs
jaymell: I think empty hash maps sounds less than ideal. Maybe we should just add an additional type to be used for passing `JUSTID`?

kkharji:
Hey @jaymell, sure, I will sync branch and address failing CI in few days.
cc @JMTamayo @saiintbrisson

JMTamayo:
Can I help you with something? Just tell me! 😊

redis/src/streams.rs
urkle: In my PR #1169 (didn't know this one existed at the time) I took the approach of having the FromRedisValue method handle both kinds of responses and using an empty hash.   I had originally done a separate type, but that can get confusing to the user.  Thus a single type was overall simpler and abstracts away the differences so they can just use one contract to access the response.  IMHO this makes coding easier.  And a helper function can be added to the StreamAutoClaimReply struct to be able to easily get to the IDs if desired.

Also my PR handles differences between Redis 6 and redis 7 responses from this  command. (they are different)

urkle:
@kkharji can you look at my PR (#1169) and see how we can join these two together?

As right now mine has handling og redis 6, and handling of the JUSTID  response (with tests). There is just slight differences in the contract. (my response uses claimed, your's uses ids, and I do not have the simple xautoclaim, ONLY the xautoclaim_options)

kkharji:
Hey @urkle please continue further the feature on you're PR. I think with @jaymell it should be matching or better then currently offered here. Thanks.

jaymell:
Thanks @kkharji. Apologies for letting this PR hang in limbo.

JMTamayo:
thank you all ♥️

PR #779, opened 2023-02-11
Rename Value enum types in order to match Redis RESP names.

nihohit:
See the naming convention here:
https://redis.io/docs/reference/protocol-spec/


jaymell:
Thanks for this. At first glance, I think this looks good and would like to get it in the next release, which will be a breaking one anyway. I'll try to go back through this with a fine-tooth comb as much as possible after getting some other changes in, but I think we should be a go here.

nihohit:
thanks @jaymell! 
I'll rebase this PR once https://github.com/redis-rs/redis-rs/pull/757 will be merged, unless you don't think it will make it to the next release.

Also, is there some way to contact you directly? I'd like to help with the maintenance load, if you're interested.

jaymell (approved):

PR #782, opened 2023-02-13
Tls with auth

nonirosenfeldredis:
added support to connecting to Redis with secure TLS with mutual auth ( client & server) . both for simple client and cluster client. now u can set your CA certificate, certificate and private key when creating the client


nonirosenfeldredis:
absolete - see tls_sec PR

PR #785, opened 2023-02-16
Bump MSRV to 1.60

jaymell:
Required due to `csv` crate, dependency of
`criterion`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -24,7 +24,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.59.0
+          - 1.60.0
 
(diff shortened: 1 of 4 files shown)


djc:
Yeah, I started looking into this, but I didn't quite figure it out. It seems annoying to bump our library MSRV for a dev-dependency only, and then it seemed, when looking at the Criterion `Cargo.toml`, that csv was an optional dependency but I couldn't figure out why it was being enabled.

jaymell:
Ah, thanks for not being lazy there. It is an optional dependency in the latest version of criterion but not in the version we currently depend on. Let me see if I can easily update criterion to avoid this.

PR #794, opened 2023-03-01
Support ipv6 link-local urls (#470)

socs:
URL parsing fails when the host is a ipv6 link-local address due to zone identifiers being omitted by the url spec.
https://url.spec.whatwg.org/#host-representation

This fix resolves the issue by stripping the zone id from the url but adding it back into the host field during conversion to ConnectionInfo.

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::fmt;
 use std::io::{self, Write};
 use std::net::{self, TcpStream, ToSocketAddrs};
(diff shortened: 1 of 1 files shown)


socs:
Ipv6 url support was added in #679 however it did not allow for link-local zone id which I've attempted to rectify here. The reason to omit it in a general url library may be reasonable but for redis urls this ought to be supported.

jaymell:
I must profess ignorance to the nuances of IPV6, but are there no url-parsing libs that support this functionality? This just seems like a strange problem for a Redis client to have to solve. How do other libraries handle this situation?

socs:
I don't think you will find many url libs that handle link-local ipv6 zones, see this [post](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27234#c2) if you're wondering why.
I think we can close this one though, since redis-rs doesn't use redis url strings internally anymore.
If one wants to connect to a ipv6 link-local address, use eg (host, port) tuple where the host part with zone id is accepted.

The more crucial fix is in [796](https://github.com/redis-rs/redis-rs/pull/796). Without it redis-rs does not work on a ipv6 link-local redis cluster.

jaymell:
Ok, that's good to hear -- glad we finally got the simplified connection-map stuff merged, as it simplified a lot of things. I'll take a look at #796 soon. Thanks!

PR #795, opened 2023-03-01
feat(commands): introduce new generic for each key arg

### LINKED ISSUE #125: Allow multiple key arguments in certain high level commands (ex. SUNIONSTORE)
Example: [SUNIONSTORE](https://github.com/mitsuhiko/redis-rs/blob/master/src/commands.rs#L513).

The two arguments should implement ToRedisArgs, but should not be the same type:
```
    |               .sunionstore(String::from("something"), lists.iter()
    |  _____________________________________________________^ starting here...
    | |                 .map(|item| format!("{}{}", item, "test"))
    | |                 .collect::<Vec<String>>()
    | |_________________________________________^ ...ending here: expected struct
 `std::string::String`, found struct `std::vec::Vec`
    |
    = note: expected type `std::string::String`
               found type `std::vec::Vec<std::string::String>`
```
```rust
fn sunionstore<K: ToRedisArgs>(dstkey: K, keys: K) {...}
```
Here, both `String` and `Vec<String>` implement `ToRedisArgs`, so they should satisfy the function generics, except that the function requires that they be the same type `K: ToRedisArgs`. Since the first argument is `String`, the function expects the second argument to be `String` as well.

There are a few other such functions, such as `sdiffstore`, which have the same problem.

MaxOhn:
Some commands needlessly restricted users by requiring passed arguments to have the same type for redis keys.

closes #125

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -191,12 +191,12 @@ implement_commands! {
     }
 
     /// Rename a key.
-    fn rename<K: ToRedisArgs>(key: K, new_key: K) {
+    fn rename<K: ToRedisArgs, N: ToRedisArgs>(key: K, new_key: N) {
         cmd("RENAME").arg(key).arg(new_key)
     }
 
     /// Rename a key, only if the new key does not exist.
-    fn rename_nx<K: ToRedisArgs>(key: K, new_key: K) {
+    fn rename_nx<K: ToRedisArgs, N: ToRedisArgs>(key: K, new_key: N) {
         cmd("RENAMENX").arg(key).arg(new_key)
     }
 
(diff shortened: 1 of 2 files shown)


nihohit (approved):
LGTM @jaymell

redis/src/commands/mod.rs
```diff
@@ -191,12 +191,12 @@ implement_commands! {
     }
 
     /// Rename a key.
-    fn rename<K: ToRedisArgs>(key: K, new_key: K) {
+    fn rename<K: ToRedisArgs, N: ToRedisArgs>(key: K, new_key: N) {
```
jaymell, line 194: Is there any specific reason to change this? 

If not, I think we should limit the changes to where we actually need them, which from what I can tell would be a few commands like `SUNIONSTORE` that doesn't specify the 2nd argument as either a different generic or a &[K].

redis/src/commands/mod.rs
```diff
@@ -575,48 +575,48 @@ implement_commands! {
 
     /// Intersect multiple sorted sets and store the resulting sorted set in
     /// a new key using SUM as aggregation function.
-    fn zinterstore<K: ToRedisArgs>(dstkey: K, keys: &'a [K]) {
+    fn zinterstore<D: ToRedisArgs, K: ToRedisArgs>(dstkey: D, keys: &'a [K]) {
```
jaymell, line 578: As mentioned above, this command already specifies &[K] for the 2nd type. Doesn't that already get us what we need?

jaymell:
Thanks for the PR. Added some comments -- please correct let me know if you think I'm missing something. Either way, happy to get in a change for this _long-standing_ bug!

redis/src/commands/mod.rs
nihohit: No, because the K must be the same for all arguments , so you can't call the command with `dstkey: String` and `keys: &[&str] / &[u8]`.

jaymell:
Please also add some test coverage asserting that the broken commands identified in the issue are fixed. :pray:

jaymell (approved):

PR #796, opened 2023-03-02
cluster: fix parsing ipv6 cluster nodes

socs:
This function is unable to parse ipv6 hosts due to expecting <host:port> format. Fixed by using rsplit instead which may also be a bit faster.


jaymell:
Thanks for the PR. Can we add some unit tests of this function?

jaymell:
Thanks. Can we consolidate the tests down to multiple assertions in a single test?

jaymell (approved):

jaymell:
Thanks!

PR #797, opened 2023-03-03
Added missing command for SMISMEMBER

HHogg:
Added a missing fn for the SMISMEMBER cmd

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -510,6 +510,11 @@ implement_commands! {
         cmd("SMEMBERS").arg(key)
     }
 
+    /// Determine if multiple given values are members of a set.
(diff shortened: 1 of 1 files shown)


jaymell:
Hi, thanks for the PR. Can we add a test or two for the command?

HHogg:
Absolutely, happy to. I did have a quick scan of the tests but couldn't find a place where all the commands were being tested. I'll take another look for an appropriate place, if not I'll start something off with a few for the others too.

jaymell:
[test_basic.rs](https://github.com/redis-rs/redis-rs/blob/main/redis/tests/test_basic.rs) is a reasonable place to put them. Thanks!

nihohit:
@HHogg do you think you can add a test?

nihohit:
implemented in https://github.com/redis-rs/redis-rs/pull/1002

PR #804, opened 2023-03-16
connection_manager: retry and backoff on reconnect

nihohit:
This change allows the connection manager to make multiple attempts to reconnect to the server, in case the server is temporarily unavailable during the first reconnect attempt.

### DIFF
--- a/redis/src/aio.rs
+++ b/redis/src/aio.rs
@@ -1114,7 +1114,10 @@ mod connection_manager {
             // Create a MultiplexedConnection and wait for it to be established
 
             let runtime = Runtime::locate();
-            let connection = client.get_multiplexed_async_connection().await?;
+            let retry_strategy = ExponentialBackoff::from_millis(exponent_base).factor(factor);
+            let connection =
+                Self::new_connection(client.clone(), retry_strategy.clone(), number_of_retries)
+                    .await?;
 
             // Wrap the connection in an `ArcSwap` instance for fast atomic access
             Ok(Self {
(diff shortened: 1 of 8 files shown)


redis/src/aio.rs
```diff
@@ -1079,7 +1083,25 @@ mod connection_manager {
         ///
         /// This requires the `connection-manager` feature, which will also pull in
         /// the Tokio executor.
+        #[deprecated = "use new_with_backoff"]
```
jaymell, line 1086: Do we need to deprecate here? Can we just supply a reasonable default and keep the method?

nihohit: We can do that, but then we need to either give a good default, which means that we're changing the behavior of the class without informing the users, or we give a default that matches the current behavior (that's what I did), which is a bad default. My preference is to make sure that users are aware of the new function without changing the behavior under their feet. If you prefer to give a good default, that's also fine.

jaymell: I think it's fine to provide a "good" default -- I think retries are the sort of thing we'd expect a `ConnectionManager` to do, so I think we should do that rather than deprecate. Everyone reads the Changelog, right? :grinning:

redis/tests/support/mod.rs
```diff
@@ -14,7 +14,7 @@ pub fn current_thread_runtime() -> tokio::runtime::Runtime {
     let mut builder = tokio::runtime::Builder::new_current_thread();
 
     #[cfg(feature = "aio")]
-    builder.enable_io();
+    builder.enable_all();
```
jaymell, line 17: Is there a reason to change this? It's effectively the same as what we already have, isn't it?

nihohit: I think that `enable_time` wasn't there when I originally wrote this. reverting.

redis/src/aio.rs
nihohit: done

nihohit:
@jaymell fixed comments and rebased.

redis/tests/test_async.rs
```diff
@@ -638,3 +638,52 @@ mod pub_sub {
...
+            Ok(mut con) => {
+                let _: RedisResult<()> = redis::cmd("FLUSHDB").query_async(&mut con).await;
+                break;
+            }
+        }
+    }
+}
+
+#[test]
+#[cfg(feature = "connection-manager")] // TODO - investigate why the test doesn't work with rust-tls
```
jaymell, line 668: I'm confused by the TODO here. What's actually failing?

nihohit: My bad, it was fixed the comment should be removed

jaymell (approved):

PR #811, opened 2023-03-22
Expose more efficient async interfaces.

nihohit:
Since the `async` keyword can't be used in traits, the ConnectionLike trait requires that the returned futures be boxed. This is necessary for the trait, but not for usage of the concrete connections. This change allows users to use the core logic without boxing the futures, thus reducing allocations. For example, the ConnectionManager now uses MultiplexedConnection's `send_packed_command`, thus removing the unnecessary boxing and according to DHAT, reducing allocations by half.

### DIFF
--- a/redis/src/aio.rs
+++ b/redis/src/aio.rs
@@ -939,23 +939,45 @@ impl MultiplexedConnection {
         };
         Ok((con, driver))
     }
+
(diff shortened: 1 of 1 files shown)


jaymell:
Thanks, this looks great!

jaymell (approved):

Issue #805, opened 2023-03-17
Proposal: Allow Value to contain errors


philippgl:
I am sending multiple commands in a pipeline and some of them might return an error. Because `Value::Bulk(Vec<Value>)` cannot contain errors, I get the first error in `Bulk` and all other values are gone.

Example
As response to 
```
PING
SET foo bar
EVALSHA aaaaaa
```
i get something like:
```
NOSCRIPT No matching script. Please use EVAL.
```
but I want to get something like:
```
PONG
Ok
NOSCRIPT No matching script. Please use EVAL.
```

I currently cannot get `Value::Bulk`-Replies with errors and the RedisError currently contains server/connection errors as well as errors sent by the server.

I have issue #746 as well (see also https://github.com/redis-rs/redis-rs/issues/366#issuecomment-673630807) and IMHO the cleanest solution would be to extend the Value enum, so it can contain an error from the server and add something like ~~`.wrap()` (or `.ok()`) to "split" the `Value` as it is now into a `RedisResult<Value>`. In addition to pipelines this might also occur in lua scripts. I also provide `.first_error()` and `map_value_error()` in an extension trait for RedisResult. Calling `.first_error` converts an `Ok(Value)` to an Err(RedisError), if it contains a nested `Value::Error` somewhere inside (like in the current implementation) and `map_value_error()` only maps an `Ok(Value::Error)` to `Err(RedisError)`, but keeps nested errors.

In the lua script case, it might even make sense to complete the script and still return a Value::Bulk, which contains an error (among other things).

The new enum variant would look like Value::Error(ErrorKind, String) with String containing the current detail.

I can help with this, because I need to see the errors in my pipeline on the client side.

Maybe I could automatically apply this `wrap` after the parsing. If the Error is not inside a Bulk, I do not loose Information.

Transactions in redis usually return `ok` for the `multi` and `queued` for each command in the pipeline, unless there is an error in the queued commands. Then the error is returned instead of the queued and the result for the exec becomes something ExecAbortError.  In the current implementation, the returned values for the normal commands are ignored and the result of the exec is returned, unless there is an error and then the first error is returned instead. As i now ignore all errors ignored by the ` req_packed_commands`, I only get the `ExecAbortError` instead of the reason, why it aborted.

jaymell:
[...]

Issue #746, opened 2022-12-14
There should be an option to process the results of pipeline batch on a caller side.


prokazov:
This is already somewhat [known](https://github.com/redis-rs/redis-rs/issues/366#issuecomment-673630807)
but opening a separate issue to track it specifically.
Consider the following scenario: an application is populating the cache during an initial sync in big batches and one of  the following happens:
1. A batch contains 5000 hsets and a couple of the keys already exist in Redis but belong to a different data type(s). 
- Good luck with finding an offender among those 5000 items because all you get is `WRONGTYPE: Operation against a key holding the wrong kind of value` without any reference to a specific command in the pipelined batch (see the issue above).
2. A batch was accepted by Redis and `req_packed_commands`  received 4783 out of 5000 responses and then something happens (like dropped connection ). There is really no reason to re-send these 4783 entries again.

And in general - it's much more flexible if a batch result processing is deferred to a caller. 

Would it be possible to add to `ConnectionLike` something like
```
    fn req_packed_commands_with_responses(
        &mut self,
        cmd: &[u8],
        batch_results: &mut Vec<RedisResult<Value>>,
        offset: usize,
        count: usize,
    ) -> RedisResult<Value>;
```
which returns an error only if it wasn't able to pass the batch to Redis and the `batch_results` contains individual command responses?
Though there might be a better implementation.

Returning just a first error without any context is not very intuitive and doesn't match to the other redis client's behavior (hiredis/redis-py).

djc:
IMO the natural API woukd be to return some variant of `AsyncIter` for these values.

zalmane:
I think there are two cases:
1. We are sending a batch in a TRANSACTION - in this case, a single error may be enough since the transaction should have rolled back.
2. We are sending a batch without a TRANSACTION. In this case, we must return all errors as some have been processed and some were not.

The main challenge in #2 is backwards compatibility since the existing method returns only a single Error.

So we can either define a new API for getting all errors, or modify existing.
Thoughts?

djc:
I think correctness will almost always trump backwards compatibility. If the current API is wrong, we should change it.

nihohit:
https://github.com/redis-rs/redis-rs/pull/1093

nihohit:
I'm trying to decide what's the best design for this.
1. Keep failing pipelines and transactions whenever any part of them fail, and return a RedisError variant that contains `Vec<(usize, RedisError)>` with the index and error of each failing command.
[...]

PR #813, opened 2023-03-24
Add Value::Error and some compatibility functions

philippgl:
:warning:  Breaking change. (Especially ignoring Values in a pipeline, now also ignores Errors)

This implements #805.

Currently this library does not allow handling of nested errors (see also #746). This was the simplest and least breaking solution, that I could come up with. This will only break (i hope), if nested errors are returned (the currently broken case) and this can be mostly fixed by calling `.first_error()` on the RedisResult, which returns the nested Error as before. Calling `.first_error()` on the `Value` will return a `Option<RedisError>`, if you want to handle a nested `Value::Error` inside a `Value::Bulk` and calling `.wrap()`  will just handle the flat case.

I needed to adapt the two tests with errors in the pipeline. The big change is, that ignored Errors are now really ignored.

PS: My next PR will be tiny and not break stuff...

### DIFF
--- a/redis/src/aio.rs
+++ b/redis/src/aio.rs
@@ -36,7 +36,9 @@ use crate::connection::{ConnectionAddr, ConnectionInfo, Msg, RedisConnectionInfo
 
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use crate::parser::ValueCodec;
-use crate::types::{ErrorKind, FromRedisValue, RedisError, RedisFuture, RedisResult, Value};
+use crate::types::{
+    ErrorKind, FromRedisValue, RedisError, RedisFuture, RedisResult, Value, ValueResult,
+};
 use crate::{from_redis_value, ToRedisArgs};
 
 /// Enables the async_std compatibility
@@ -540,7 +542,7 @@ where
             self.buf.clear();
             cmd.write_packed_command(&mut self.buf);
             self.con.write_all(&self.buf).await?;
-            self.read_response().await
+            self.read_response().await.map_value_error()
         })
         .boxed()
     }
@@ -953,7 +955,7 @@ impl ConnectionLike for MultiplexedConnection {
                         RedisError::from(io::Error::from(io::ErrorKind::BrokenPipe))
                     })
                 })?;
-            Ok(value)
+            Ok(value).map_value_error()
         })
         .boxed()
     }
@@ -1127,7 +1129,7 @@ mod connection_manager {
                 reconnect_if_io_error!(self, connection_result, guard);
                 let result = connection_result?.req_packed_command(cmd).await;
                 reconnect_if_dropped!(self, &result, guard);
-                result
+                result.map_value_error()
             })
             .boxed()
         }
--- a/redis/src/client.rs
+++ b/redis/src/client.rs
@@ -6,6 +6,7 @@ use std::pin::Pin;
 use crate::{
     connection::{connect, Connection, ConnectionInfo, ConnectionLike, IntoConnectionInfo},
     types::{RedisResult, Value},
+    ValueResult,
 };
 
 /// The client type.
@@ -260,7 +261,9 @@ use crate::aio::Runtime;
 
 impl ConnectionLike for Client {
     fn req_packed_command(&mut self, cmd: &[u8]) -> RedisResult<Value> {
-        self.get_connection()?.req_packed_command(cmd)
+        self.get_connection()?
+            .req_packed_command(cmd)
+            .map_value_error()
     }
 
     fn req_packed_commands(
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -495,7 +495,8 @@ where
                         // if we are in asking mode we want to feed a single
                         // ASKING command into the connection before what we
                         // actually want to execute.
-                        conn.req_packed_command(&b"*1\r\n$6\r\nASKING\r\n"[..])?;
+                        conn.req_packed_command(&b"*1\r\n$6\r\nASKING\r\n"[..])
+                            .map_value_error()?;
                         is_asking = false;
                     }
                     (addr.to_string(), conn)
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -8,9 +8,9 @@ use futures_util::{
 use std::pin::Pin;
 use std::{fmt, io};
 
-use crate::connection::ConnectionLike;
 use crate::pipeline::Pipeline;
 use crate::types::{from_redis_value, FromRedisValue, RedisResult, RedisWrite, ToRedisArgs};
+use crate::{connection::ConnectionLike, Value, ValueResult};
 
 /// An argument to a redis command
 #[derive(Clone)]
@@ -114,7 +114,11 @@ impl<'a, T: FromRedisValue + 'a> AsyncIterInner<'a, T> {
             }
 
             let rv = unwrap_or!(
-                self.con.req_packed_command(&self.cmd).await.ok(),
+                self.con
+                    .req_packed_command(&self.cmd)
+                    .await
+                    .map_value_error()
+                    .ok(),
                 return None
             );
             let (cur, batch): (u64, Vec<T>) = unwrap_or!(from_redis_value(&rv).ok(), return None);
@@ -421,7 +425,7 @@ impl Cmd {
     where
         C: crate::aio::ConnectionLike,
     {
-        let val = con.req_packed_command(self).await?;
+        let val = con.req_packed_command(self).await.map_value_error()?;
         from_redis_value(&val)
     }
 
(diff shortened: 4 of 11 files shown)


nihohit:
@jaymell if you decide to take the route of adding error to `Value`, this should probably be reviewed after it is rebased over https://github.com/redis-rs/redis-rs/pull/757, since it adds multiple additional collection options to `Value`, and thus complicates this change.

nihohit:
My take on the questions:

> I think, the ErrorKind::ExecAbortError should become ErrorKind::ExecAbortError(Value), so it can actually contain all the errors.

I agree that this is the correct course. IMO it should be `ErrorKind::ExecAbortError(Vec<Value>)`, to properly represent the fact that there are multiple return values there.

> Should this be rebased on https://github.com/redis-rs/redis-rs/pull/757?

Since that PR has been [approved but not merged](https://github.com/redis-rs/redis-rs/pull/757#issuecomment-1558593966), I think that this will be merged in the same version, especially since we don't want to release 2 versions that break `Value`. That being said, since that PR has already been approved, I think the polite thing would be to rebase this PR over #757, in order not to incur more work on its contributor.

> The offset parameter of the req_packed_commands will become unused

I agree. Mark as deprecated. 

> What to do with .ignore() in general.

IMO if the user provided `ignore` flags, then they should be respected, and the user won't receive the relevant errors. If the pipeline only erred on ignored commands, the user will receive `Ok`. If I understand the situation correctly this is only relevant if the transaction didn't receive `ExecAbortError` - if the transaction was aborted, the user will receive an error, with a full list of replies.

nihohit:
> What to do in the req_packed_cmds? 

I'm sorry, I'm not sure what's the question. why have a new error, especially when we have an existing error that already describes the situation? And when would the new error be returned?

> Just return the Ok(Bulk) with errors inside and ignored values removed, I guess.

just to be sure, you mean the `[Ok, NoScriptFound, Ok]`? Then yes, that sounds correct.

> So to really deprecate the parameter I might need a new function req_packed_pipeline or similar.

Yes, but let's call it `req_pipeline`, since the pipeline isn't packed :).  But IMO that's a problem for later. Let's start by having an implementation that works, and then see what we can/should remove from the existing functions.

nihohit:
> But that PR still looks like a moving target.

I hope it will be merged to the `resp3` branch within the next couple of days. Please rebase over it and target the PR to the same branch. That branch is meant to allow development of all `Value` breaking changes.

philippgl:
> > What to do in the req_packed_cmds?
> 
> I'm sorry, I'm not sure what's the question. why have a new error, especially when we have an existing error that already describes the situation? And when would the new error be returned?

(all the strange stuff following is about transactional pipelines)

Currently the `req_packed_cmds` returns the first error, if there is an error anywhere in the response, but otherwise returns the response as is, i.e. it only strips the queued messages, if requested using `offset`. If someone would use it on main, it would return [Ok, Queued, ..., [..]] (with n times Queued, followed by a single Bulk with the n real replies). 

If I want to keep this behavior (return the real redis reply and just change the error handling), I would need a way to keep this structure and hand it over like this to the pipeline struct ([Ok, Queued, ..., SomeError, ExecAbortError]), which needs an empty ErrorKind (and the pipeline struct could then transform the error to a new error with `Vec<Value>` inside).

Or I could completely change this low level interface altogether (which seems like breaking a lot more than needed) and assume it is only used by the pipeline struct and than I would be free to repurpose the ExecAbortError to include the other errors reported by the redis. (but this again would handle the pipeline partially in the low level `req_packed_cmds`) (This is my least preferred option as I think the low level interface should be close to the wire)

I think. the nicest approach would be: `req_packed_cmds` returns the redis reply as it is on the wire (with the empty ExecAbortError) and the Pipeline struct returns a new error, which includes the `Vec<Value>`. So in the good case  `req_packed_cmds` returns the [Ok, Queued, ..., [..]] and in the error case it returns [Ok, Queued, ..., SomeError, ..., ExecAbortError] and this is transformed by the Pipeline struct to become Err(RedisError{ kind:TransactionError([Ok, ... SomeError, ...])}) (also replacing Queued with Ok).

Another really radically simple approach would be to completely remove all the transactional pipeiine stuff and convince users to used lua scripts instead (which is, what the redis docs is suggesting as well). I think anyone, who has read about the [error handling](https://redis.io/docs/manual/transactions/#errors-inside-a-transaction) or [missing rollbacks](https://redis.io/docs/manual/transactions/#what-about-rollbacks) might prefer scripts already (where you can implement your own error handling).

As scripts are supported since 2.6, the docs recommend scripting over transactional pipelines and the rust driver is newer than the script support, there is a chance, that only very few people actually use transactional pipelines with the rust driver (those who did not read to much into the redis docs).

If all of this sounds really complicated, i think the removal of transactional pipeline support would be the way to go even if it is more breaking. Transactional pipelines are pretty broken in redis and as scripts exists, i do not think they will ever be fixed and are only kept for backwards compatibility.

nihohit:
The question of whether and how much to reduce transaction logic in redis-rs is a question is for @jaymell.
My personal opinion is that we removing `Pipeline`/`ClusterPipeline` is too much of a breaking change, but we might consider either removing the `offset` argument from `req_packed_cmds`, or deprecating the function entirely from the public API.
My hunch is that removing `offset` is the right way to go, because a non-transactional pipeline makes a lot of sense, especially in sync APIs.

nihohit:
> the docs recommend scripting over transactional pipelines and the rust driver is newer than the script support

Discussed this with @matolson, a maintainer of Redis, and she says that she believes the docs are inaccurate, and that scripting isn't faster than transactions. I wouldn't drop support entirely.

madolson:
> > the docs recommend scripting over transactional pipelines and the rust driver is newer than the script support
> 
> Discussed this with @madolson, a maintainer of Redis, and she says that she believes the docs are inaccurate, and that scripting isn't faster than transactions. I wouldn't drop support entirely.

Yeah. There is some consensus that scripting is faster for "read/modify/write" operations as they avoid a round trip operation. Multi-exec is still much faster that scripts though for executing a small set of a sequential commands. Omitting the support for watch is perhaps reasonable though, we don't see significant usage of it.

philippgl:
> ... Omitting the support for watch is perhaps reasonable though, we don't see significant usage of it.

Watch is not supported at all right now, so you need to manually call `.cmd("watch")` to use it.

> ... My personal opinion is that we removing Pipeline/ClusterPipeline (or their atomic mode) is too much of a breaking change,...

I did not want to remove `Pipeline`. In fact, I only send pipelines to the redis server. Some of them might only contain a single command, if the requests do not come in fast enough inside my application.

The Pipeline struct is currently for two things: 

- The easy way to do transactions by calling `.atomic()`, but this makes the whole pipeline transactional
- Sending multiple commands at once to optimize performance (with the option to either make the whole pipeline transactional or manually manage the transactions like below)

For performance reasons it is usually beneficial to send longer pipelines, but calling `.atomic()` makes the whole pipeline transactional, so if you want performance (long pipelines) and transactions (usually short cmd sequences), you already need to do this manually by calling `.cmd("multi") and `.cmd("exec")` (probably even more than once) inside a pipeline. [^1]

The only support for transactions, that is currently offered by this driver, is `.atomic()` in the pipeline struct, which makes the whole Pipeline atomic, and I was just talking about removing this `.atomic()`, which is rather inflexible. As the handling of this `.atomic()` is currently split between the Pipeline and the `req_packed_cmds`, removing it would simplify things for the implementation, but would make creating simple transactions more complicated for the user.

One of the benefits of this PR is, that it allow the user to see all errors if a `Pipeline` fails. With this implementation it will no longer be possible, that `req_packed_cmds` just returns the first error it finds and all the error handling will be done in the `Pipeline`.

I can implement it either way (with or without transactions), but I would not like to maintain that implementation (if it would be my job) as it is (imho) a lot more complicated and the benefit (imho again) is small: if you want to maximize performance, you would need to do it manually already, e.g. if you want to have have multiple transactional sequences in one pipeline. The strange error handling at the moment makes the opportunities to use it even rarer.

Just some random thoughts about alternatives, which would remove the transaction responsibility from the Pipeline struct (which imho would allow a simpler implementation and more flexible usage):

- Add in some `.multi()` and `.exec()`, which would allow arbitrary atomic command sequences. The user could than read the result of the `.exec()` instead of the result of `.query()`. Then it would be closer to the redis interaction, which would simplify the implementation. This would probably make handling `.ignore` even more complicated. (ignoring `multi` and the queued replies will be simple, of course, but doing it in exec is complicated)
- Use a completely different struct for transactions, which could then just be appended to a pipeline. This might be combined with the above and might make it easier to handle the `ignore` in this struct

[^1] I do not do this, but use scripts (which are also transactions) inside pipelines. I can find many more situations, where I can easily use a script vs. the rare opportunities, where I could benefit from a transactional pipeline (or a transactional sequence of commands in general) and even then it is usually simple to write a script as well. And i can even add this script to the next pipeline and i would need to send a separate pipeline otherwise (or do it completely manual as described above).

nihohit:
Sorry it took a while to answer. I think I understand your position better now.
No, I don't think that this PR should include changes to the behavior of `pipeline.atomic()`, or to remove it. Let's keep this separate - if you (or somebody else) will want to propose an API change, it should be in another PR. Here let's just fix the error return behavior to be correct.

nihohit:
implemented in https://github.com/redis-rs/redis-rs/pull/1093

PR #814, opened 2023-03-24
Use static dispatch if possible

philippgl:
As most users will likely only use one kind of connection, the usual downside (increased binary size) will not affect many.

This might provide a tiny performance boost, but I have done this mostly, because using `#[inline]` with dynamic dispatch feels wrong (as there is not much to inline, if the connection type is determined at runtime).

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -407,7 +407,7 @@ impl Cmd {
     /// result to the target redis value.  This is the general way how
     /// you can retrieve data.
     #[inline]
-    pub fn query<T: FromRedisValue>(&self, con: &mut dyn ConnectionLike) -> RedisResult<T> {
+    pub fn query<T: FromRedisValue>(&self, con: &mut impl ConnectionLike) -> RedisResult<T> {
         match con.req_command(self) {
             Ok(val) => from_redis_value(&val),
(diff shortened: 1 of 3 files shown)


philippgl:
Oops, maybe this is caused by ignorance on my part. Are there closed source crates? Would this not just compile to static dispatch (with the increase in performance, but possible increase in binary size)?

I was just a bit confused by the `#[inline]` and did not think, that there could be a way to break this.

nihohit:
@philippgl I might be wrong here, I didn't try to compile this and frankly, dynamic dispatch in Rust isn't within my expertise :)
If the code still accepts `Box<dyn ConnectionLike>` then my reservation is incorrect. I'm not sure I'll have time this week to verify this, but I'll make the effort.

nihohit:
Also, I'm pretty sure that anyone who uses a single connection type is better off using [these functions](https://github.com/redis-rs/redis-rs/pull/811), but that's not a reason to reject this change :)

philippgl:
Ok, @nihohit  is right. This breaks dynamic dispatch and I only got like 0.5%  (Median of 3 runs each). And that was ping-only. If there is more work per request, it will likely be less.

So I think, this can be closed.

PR #820, opened 2023-04-01
Update Rustls to v0.21.0

rharish101:
This also removes an explicit dependency on `webpki` (or `rustls-webpki`), since `RootCertStore::add` raises a `rustls::Error` instead of `webpki::Error`.


jaymell (approved):

jaymell:
Thanks!

PR #821, opened 2023-04-03
Split aio.rs to separate files.

nihohit:
This change doesn't modify any code, except for `use` and `mod` statements.


jaymell:
If you can rebase/update this, I'd like to get it in :pray:

nihohit:
@jaymell rebased

redis/src/lib.rs
```diff
@@ -415,6 +415,7 @@ mod pipeline;
 #[cfg_attr(docsrs, doc(cfg(feature = "acl")))]
 pub mod acl;
 
+/// module for async support.
```
jaymell, line 418: Can we put the existing doc comment at the top of aio/mod.rs? (Though maybe the line "experimental" doesn't need to be in there anymore)

nihohit:
Works to like me to move async_cluster also to the aio module?

jaymell:
> Works to like me to move async_cluster also to the aio module?

I think for now just breaking up `aio.rs` is good. Maybe the various cluster modules should be consolidated in some way, but honestly I'm not that concerned about it.

redis/src/lib.rs
nihohit: done

jaymell: I think we need to remove this comment though, else it will break the module description in the top-level docs.

nihohit: done

jaymell (approved):

jaymell:
Thanks!

PR #829, opened 2023-04-15
Proposal for aio::ConnectionLike changes

nihohit:
This draft contains 2 proposals for changes / extensions to the aio::ConnectionLike API. These proposals aren't mutually exclusive, so if there's interest, both can be integrated.
The aim of these proposals is to mitigate the increasing memory cost of using `Cmd`, since that struct requires multiple allocations for creation and usage.

1. Add a method that allows the user to pass a `Bytes` object. This puts the onus on the user to create a correct command. `Bytes` was chosen instead of `Vec` since `Bytes` is cheaply clonable, and so it allows the user to easily reuse the command, even though it requires more memory than `Vec`.
2. Added another `Cmd`-equivalent object, `ThinCmd`, which is constructed using `ThinCmdBuilder` with an API similar to `Cmd`. This allows the user to easily create a Redis command using an API provided by the library. `ThinCmd` uses a `Bytes` struct internally, so it is also cheaply clonable.

The goal of this draft is to get initial thoughts and comments on the suggestion. The draft doesn't cover pipeline or using the `Cmd::cursor_arg` commands, in order to limit its scope.


nihohit:
This table shows (hopefully clearly :) ) how the rate of memory usage increases faster when using `Cmd` than with `ThinCmd`. It also covers a `ThinCmd` implementation that uses `Vec` instead of `Bytes` internally, which makes it slightly smaller for single usage, but makes cloning `ThinCmd` more expensive.
[memory improvement.xlsx](https://github.com/redis-rs/redis-rs/files/11240613/memory.improvement.xlsx)

shachlanAmazon:
relevant to https://github.com/redis-rs/redis-rs/issues/615
@philippgl, is this still relevant to you?

nihohit:
relevant to https://github.com/redis-rs/redis-rs/issues/615
@philippgl, is this still relevant to you

nihohit:
Another potentially significant perf improvement would be if commands weren't encoded by default as arrays of byte-safe strings. Commands whose arguments don't contain any byte-unsafe characters can be encoded as simple strings, and instead only used this more verbose format when some of the args are byte unsafe.

philippgl:
Just to clarify: I currently use `Cmd::query_async` and especially `Script::invoke_async` (this automatically uploads the script on script failure), i.e. I use the `aio::ConnectionLike` trait, so I will likely not benefit from this change.

Using `Pipeline` provided a really huge performance improvement for me (~tripled the number of requests, the redis can handle). I also use a custom `aio::ConnectionLike` connection, that bundles requests from different tokio tasks (which mainly simplifies usage of `Pipeline`) and it is unlikely, that I will benefit much from this. (unless I could combine Cmd and Bytes into a single Pipeline somehow, which sounds complicated).

I do not clone `Cmd`s at all, so optimizing that is not really useful for me (I do not know, if others will benefit more from this).

Just to provide some feedback.

jaymell:
@nihohit What are your thoughts on making breaking changes to the interface, i.e., by requiring the existing methods to accept `Bytes` or `Vec<u8>` instead of `Cmd` directly?

jaymell:
As discussed elsewhere, I think breaking changes here are preferred, and I think it would be particularly good to resolve some of the poorly named methods on the async `ConnectionLike` trait, 

- `req_packed_command` -- renamed to `req_command`
- New implementation of `req_packed_command` that mirrors the sync trait as much as possible, *possibly* with new names to avoid confusing breakage for existing users, accepting `Bytes` or `Vec<u8>` (or a separate method for each). 
- `req_packed_commands` -- rename to `req_commands`??

As far as the `ThinCmd` changes, can we defer those to a separate PR? It seems reasonable but also secondary to the above.

jaymell:
Can we update the PR title?

jaymell:
And do you mind updating the top-level comment to reflect the current state of things? :pray:

nihohit:
@jaymell updated.

redis/src/cluster_async/mod.rs
```diff
@@ -110,14 +110,14 @@ struct ClusterConnInner<C> {
 #[derive(Clone)]
 enum CmdArg<C> {
     Cmd {
-        cmd: Arc<Cmd>,
-        func: fn(C, Arc<Cmd>) -> RedisFuture<'static, Response>,
+        cmd: Bytes, // TODO - use Bytes instead, after rebasing on the routing change.
```
jaymell, line 113: Remove TODO?

redis/src/cluster_async/mod.rs
```diff
@@ -146,16 +146,28 @@ impl<C> CmdArg<C> {
         }
 
         match self {
-            Self::Cmd { ref cmd, .. } => route_for_command(cmd),
+            Self::Cmd { ref cmd, .. } => {
+                let value = parse_redis_value(cmd).expect("Command wasn't a valid redis value");
```
jaymell, line 150: These expects are a bit scary. Is it possible to handle them earlier in the request / without panicking?

redis/src/cluster_async/mod.rs
nihohit: done

redis/src/cluster_async/mod.rs
nihohit: Mapped them to None for now. I believe this should be an error, but since I'm refactoring this area in https://github.com/redis-rs/redis-rs/pull/843, I'll implement the correct behavior once it is rebased.

jaymell: Ok, I think let's wait for #843 to go in first then.

nihohit:
@jaymell rebased

nihohit:
Notice that the async cluster's `send_packed_command`/`send_packed_commands` don't have to be public, but I was planning on making the routing choice public in any case, and this seemed like a good time for it.
There are some issues with cluster_routing which will take a while to fix, so this seems to me like a good workaround, for now. It allows the user to choose their own routing, if they disagree with redis-rs' choices.

nihohit:
converted to draft - I'm reconsidering whether to use Bytes, Vec<u8>, or Box<[u8]>.

PR #830, opened 2023-04-17
Track the caller of execute

davidspies:
Since the unwrap can fail due to eg. a syntax error, I think we want it to fail with the caller's line number


nihohit:
can you please post sample stack trace before/after the change?

nihohit:
execute has been deprecated. so this change is no longer needed.

Issue #802, opened 2023-03-15
Are keydb specific commands in scope?


ibotty:
I am using redis-rs with keydb and use keydb specific commands, mostly `EXPIREMEMBER`.  Would you consider a PR implementing this and e.g. `KEYDB.CRON`?  Should it be with a specific feature flag (`keydb`)?

nihohit:
As mentioned in the PR, Keydb support won't be added.

PR #831, opened 2023-04-19
add KeyDB specific expiremember command

ibotty:


ibotty:
fixes #802.

jaymell:
With apologies for not replying sooner, I don't think this is appropriate for redis-rs. I would recommend just constructing the command manually.

PR #832, opened 2023-04-21
fix: use `SMISMEMBER` if multiple keys are given

insertish:
Ran into an issue trying to use this when I found out it isn't implemented.
I basically copied how `GET`/`MGET` works, hopefully that's fine.

If there's anything wrong here I'll be happy to ammend the PR.

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -507,9 +507,14 @@ implement_commands! {
         cmd("SINTERSTORE").arg(dstkey).arg(keys)
     }
 
-    /// Determine if a given value is a member of a set.  If key is a vec this becomes an `SMISMEMBER`.
+    /// Determine if a given value is a member of a set.
     fn sismember<K: ToRedisArgs, M: ToRedisArgs>(key: K, member: M) {
-        cmd(if member.is_single_arg() { "SISMEMBER" } else { "SMISMEMBER" }).arg(key).arg(member)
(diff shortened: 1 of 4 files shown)


jaymell:
Hi, thanks for the PR. What you did is totally reasonable, however I'd actually prefer it if we just added a separate command. Do you mind adding a separate command for `SMISMEMBER` instead?

insertish:
Done

jaymell:
Can we add a simple test for it as well?

redis/src/commands/mod.rs
```diff
@@ -512,6 +512,11 @@ implement_commands! {
         cmd("SISMEMBER").arg(key).arg(member)
     }
 
+    /// Determine if multiple values are members of a set.
+    fn smismember<K: ToRedisArgs, M: ToRedisArgs>(key: K, member: M) {
```
shachlanAmazon, line 516: minor: please rename member to members, to clarify that the function can take multiple values there.

nihohit:
@Zomatree why did you add changes to the pub/sub to this PR?

Zomatree:
My mistake I was doing other changes, I didn't release this pr was set for this branch.

nihohit:
@Zomatree @insertish 
sorry, somehow missed your comment.
Please fix the comment about the argument name and add a test, and then let's merge this :)

nihohit:
implemented in https://github.com/redis-rs/redis-rs/pull/1002

PR #818, opened 2023-03-31
Add request timeout to MultiplexedConnection.

nihohit:
https://github.com/redis-rs/redis-rs/issues/683

PR #833, opened 2023-05-02
Expose connection timeout in ConnectionInfo.

nihohit:
This will allow users to implement `cluster_async::Connect` with a timeout passed from the cluster params, and thus passing the connection timeout parameter to the connection defined in `ClusterClient::get_async_generic_connection`. This adds the minimal logic change to the crate, but is a breaking change because of the change in `ConnectionInfo`.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -838,6 +838,7 @@ pub(crate) fn get_connection_info(
             username: cluster_params.username,
             ..Default::default()
         },
+        connection_timeout: cluster_params.connection_timeout,
     })
 }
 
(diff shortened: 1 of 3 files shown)


nihohit:
Note: This is a proposed alternative to https://github.com/redis-rs/redis-rs/pull/834

nihohit:
This might be expanded by adding a type-erased field to `ConnectionInfo`, that will allow the user to pass any information to the `C::connect` function - for example, request timeout (https://github.com/redis-rs/redis-rs/pull/818), or exponential backoff (https://github.com/redis-rs/redis-rs/pull/804) configurations.

nihohit:
3rd option: https://github.com/redis-rs/redis-rs/pull/835

PR #834, opened 2023-05-02
Add connection timeout to the cluster client

nihohit:
This allows the users to define a connection timeout to the async cluster, which will cause the `refresh_slots` action to timeout if it takes too long to connect to a node.

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -339,7 +339,7 @@ impl MultiplexedConnection {
     }
 
     /// Constructs a new `MultiplexedConnection` out of a `AsyncRead + AsyncWrite` object
-    /// and a `ConnectionInfo`. The new object will wait on opeartions for the given `response_timeout`.
+    /// and a `ConnectionInfo`. The new object will wait on operations for the given `response_timeout`.
     pub async fn new_with_response_timeout<C>(
         connection_info: &RedisConnectionInfo,
         stream: C,
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -26,18 +26,18 @@ use std::task::{self, Poll};
 use tokio_util::codec::Decoder;
 
 // Senders which the result of a single request are sent through
-type PipelineOutput<O, E> = oneshot::Sender<Result<Vec<O>, E>>;
+type PipelineOutput<O> = oneshot::Sender<Result<Vec<O>, RedisError>>;
 
-struct InFlight<O, E> {
-    output: PipelineOutput<O, E>,
+struct InFlight<O> {
+    output: PipelineOutput<O>,
     expected_response_count: usize,
     current_response_count: usize,
     buffer: Vec<O>,
-    first_err: Option<E>,
+    first_err: Option<RedisError>,
 }
 
-impl<O, E> InFlight<O, E> {
-    fn new(output: PipelineOutput<O, E>, expected_response_count: usize) -> Self {
+impl<O> InFlight<O> {
+    fn new(output: PipelineOutput<O>, expected_response_count: usize) -> Self {
         Self {
             output,
             expected_response_count,
@@ -49,9 +49,9 @@ impl<O, E> InFlight<O, E> {
 }
 
 // A single message sent through the pipeline
-struct PipelineMessage<S, I, E> {
+struct PipelineMessage<S, I> {
     input: S,
-    output: PipelineOutput<I, E>,
+    output: PipelineOutput<I>,
     response_count: usize,
 }
 
@@ -59,19 +59,18 @@ struct PipelineMessage<S, I, E> {
 /// items being output by the `Stream` (the number is specified at time of sending). With the
 /// interface provided by `Pipeline` an easy interface of request to response, hiding the `Stream`
 /// and `Sink`.
-struct Pipeline<SinkItem, I, E>(mpsc::Sender<PipelineMessage<SinkItem, I, E>>);
+struct Pipeline<SinkItem, I>(mpsc::Sender<PipelineMessage<SinkItem, I>>);
 
-impl<SinkItem, I, E> Clone for Pipeline<SinkItem, I, E> {
+impl<SinkItem, I> Clone for Pipeline<SinkItem, I> {
     fn clone(&self) -> Self {
         Pipeline(self.0.clone())
     }
 }
 
-impl<SinkItem, I, E> Debug for Pipeline<SinkItem, I, E>
+impl<SinkItem, I> Debug for Pipeline<SinkItem, I>
 where
     SinkItem: Debug,
     I: Debug,
-    E: Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_tuple("Pipeline").field(&self.0).finish()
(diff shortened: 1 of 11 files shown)


jaymell:
I think this looks good -- simple and non-breaking!

nihohit:
3rd option: https://github.com/redis-rs/redis-rs/pull/835

nolik:
> @jaymell right, moved to ready for review. Also added support for sync cluster, for consistency - I didn't want to have a configuration in `ClusterParams` that applies only to want flavor of the client.

hey, mates. I'm using ConnectionManager and really waiting for the connection t/0 configuration for ConnectionManager / MultiplexedConnection.

@jaymell  @nihohit if I able to speed up this t/o feature delivery - just let me know.

nolik:
@jaymell Hey, mate. Would you be able to review it? 
it seems it's passing all tests

nolik:
@nihohit seems you have a conflicts after the latest merges.

nolik:
@jaymell any chance to review this PR ?

jaymell:
> @jaymell any chance to review this PR ?

Yep, will get to it. Thanks for your patience.

nihohit:
@nolik you're welcome to join in on the discussion in https://github.com/redis-rs/redis-rs/pull/866 on what would be the best awy to implement this.

nihohit:
rebased over https://github.com/redis-rs/redis-rs/pull/875

jaymell:
I've been trying to benchmark the response-timeout changes locally and haven't found a significant difference, though I don't particularly trust my results. Any lingering doubts about the performance impacts of those changes?

nihohit:
I've received roughly the same results in repeated benchmarks. I tried moving things around, but can't find either the reason, or a solution.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -329,6 +334,19 @@ impl MultiplexedConnection {
...
         stream: C,
     ) -> RedisResult<(Self, impl Future<Output = ()>)>
+    where
+        C: Unpin + AsyncRead + AsyncWrite + Send + 'static,
+    {
+        Self::new_with_response_timeout(connection_info, stream, std::time::Duration::MAX).await
+    }
+
+    /// Constructs a new `MultiplexedConnection` out of a `AsyncRead + AsyncWrite` object
+    /// and a `ConnectionInfo`. The new object will wait on opeartions for the given `response_timeout`.
```
jaymell, line 344:
```suggestion
    /// and a `ConnectionInfo`. The new object will wait on operations for the given `response_timeout`.
```

redis/src/client.rs
```diff
@@ -164,11 +200,12 @@ impl Client {
     #[cfg_attr(docsrs, doc(cfg(feature = "tokio-comp")))]
     pub async fn create_multiplexed_tokio_connection(
         &self,
+        response_timeout: std::time::Duration,
```
jaymell, line 203: Should we avoid a breaking change here?

redis/src/client.rs
```diff
@@ -181,12 +218,15 @@ impl Client {
     #[cfg_attr(docsrs, doc(cfg(feature = "async-std-comp")))]
     pub async fn create_multiplexed_async_std_connection(
         &self,
+        response_timeout: std::time::Duration,
```
jaymell, line 221: Should we avoid a breaking change here?

redis/src/cluster_async/mod.rs
```diff
@@ -1066,25 +1067,33 @@ where
 /// and obtaining a connection handle.
 pub trait Connect: Sized {
     /// Connect to a node, returning handle for command execution.
-    fn connect<'a, T>(info: T) -> RedisFuture<'a, Self>
+    fn connect<'a, T>(
```
jaymell, line 1070: Anyway, I guess we can live with the breaking changes, obviously we'll want to wait until after the next minor release though.

nihohit:
fixed breaking changes.

redis/src/cluster_client.rs
```diff
@@ -79,6 +81,8 @@ impl From<BuilderParams> for ClusterParams {
             read_from_replicas: value.read_from_replicas,
             tls: value.tls,
             retry_params: value.retries_configuration,
+            connection_timeout: value.connection_timeout.unwrap_or(Duration::MAX),
```
jaymell, line 84: I think we should set a sane default of 1 second or less for `connection_timeout`.

redis/src/cluster_client.rs
```diff
@@ -238,6 +242,22 @@ impl ClusterClientBuilder {
...
+    /// If enabled, the cluster will only wait the given time on each connection attempt to each node.
+    pub fn connection_timeout(mut self, connection_timeout: Duration) -> ClusterClientBuilder {
+        self.builder_params.connection_timeout = Some(connection_timeout);
+        self
+    }
+
+    /// Enables timing out on slow responses.
+    ///
+    /// If enabled, the cluster will only wait the given time to each response from each node.
+    pub fn response_timeout(mut self, connection_timeout: Duration) -> ClusterClientBuilder {
```
kamulos, line 256: The variable name here is also `connection_timeout`

kamulos:
I tried to use this branch in my service, and it is broken for me. If I trigger a timeout I get a panic in `get_random_connection` (https://github.com/redis-rs/redis-rs/blob/main/redis/src/cluster_async/mod.rs#L1292) and the connection stays broken after that. I have to reestablish a completely new connection to continue making Redis requests.

## How to Reproduce

I am using a 1 node Redis cluster for testing. There I use the command `DEBUG SLEEP 5` to simulate a situation that would ideally result in a timeout on the `redis-rs` side.

<details>

<summary>The code I use</summary>

```rust
use redis::{cluster::ClusterClientBuilder, cluster_async::ClusterConnection, Value};
use std::{
    sync::atomic::{AtomicU64, Ordering},
    time::Duration,
};

static REQUEST_COUNTER: AtomicU64 = AtomicU64::new(0);

#[tokio::main]
async fn main() {
    let connection = ClusterClientBuilder::new(vec!["redis://:password@redis-0"])
        .connection_timeout(Duration::from_millis(400))
        .response_timeout(Duration::from_millis(400))
        .build()
        .unwrap()
        .get_async_connection()
        .await
        .unwrap();

    loop {
        let connection = connection.clone();
        tokio::spawn(do_request(connection));

        tokio::time::sleep(Duration::from_millis(250)).await;
    }
}

async fn do_request(mut connection: ClusterConnection) {
    redis::cmd("role")
        .query_async::<_, Value>(&mut connection)
        .await
        .unwrap();
    let counter = REQUEST_COUNTER.fetch_add(1, Ordering::SeqCst);
    println!("request {counter}");
}
```

</details>

nihohit:
@kamulos can you please write a self-contained test to demonstrate the issue? It's unclear how/when `DEBUG SLEEP 5` comes into action here.

redis/src/cluster_client.rs
jaymell: I agree that it should be renamed (or just call the var `timeout` in both this and the above method)

jaymell (approved):

jaymell:
Let's get it in!

nihohit:
@kamulos I'd like your input before merging this. Can you still reproduce the issue?
@jaymell I think I'll wait for reply until wed the 13th, and if we won't hear anything more, I'll merge this.

kamulos:
@nihohit I finally got to to testing it again. I am not entirely sure what is happening, but my educated guess would be something along these lines:

In my Test I have a Redis Node, where I execute the command `DEBUG SLEEP 5`. This is difficult to handle for `redis-rs` because not only the request will time out, but also any attempt to reconnect to the Cluster. First a timeout in `refresh_connections` will occur, and after that a panic in `get_random_connection` (`redis-rs/redis/src/cluster_async/mod.rs:1301:10`): "Connections is empty".

I don't believe this merge request is at fault, because the whole panic in `get_random_connection` obviously is a preexisting condition. There is however a very bad interaction here:

When the panic occurs, it ultimately is caught by tokio and other tasks are able to continue running. **However the ClusterConnection is broken at that point and does not recover until the whole program is restarted.** This is quite critical in my use-case.

If I would speculate about why the Connection is broken, I think it is just not panic-safe and we somehow end up in an inconsistent state. What I observe is that after the panic the first few requests end with the Error `Unable to receive command`. Maybe those requests were already in-flight during the panic? Later requests end with the Error `Unable to send command`.

For some reason the mpsc is broken, but the `stream` future (in `CusterConnection::new`) does not seem to have terminated.

In my application I solved the timeouts, by just wrapping the `ClusterConnection` and putting a timeout around `req_packed_command`. This seems to work fine and is able to recover in this case.

So in summary I think there is a critical flaw in the `cluster_async` module, that needs to be solved, but this pull request probably can't do anything about it.

nihohit:
Yes, I assume that the timeouts caused connections to be removed from the connections map, which eventually causes the panic. I believe https://github.com/redis-rs/redis-rs/pull/968 will help there, but as you mentioned, it's not caused by this change.
Thanks for the detective work!

PR #835, opened 2023-05-03
Cluster: Allow passing a type-erased configuration

nihohit:
This allows users to pass any additional information needed by the generic connection, using a type-erased field. Since all additions are optional this change should be backwards compatible.

For example, if a user wanted to implement a timeout on the connection attempt, they could do it with this class:
```rs
use redis::{
    aio::{ConnectionLike, MultiplexedConnection},
    cluster_async::Connect,
};
use std::io;
use std::time::Duration;

pub(crate) struct ConnectionConfiguration {
    pub(crate) connection_timeout: Duration,
}

#[derive(Clone)]
pub struct Connection(MultiplexedConnection);

impl ConnectionLike for Connection {
    fn req_packed_command<'a>(
        &'a mut self,
        cmd: &'a redis::Cmd,
    ) -> redis::RedisFuture<'a, redis::Value> {
        self.0.req_packed_command(cmd)
    }

    fn req_packed_commands<'a>(
        &'a mut self,
        cmd: &'a redis::Pipeline,
        offset: usize,
        count: usize,
    ) -> redis::RedisFuture<'a, Vec<redis::Value>> {
        self.0.req_packed_commands(cmd, offset, count)
    }

    fn get_db(&self) -> i64 {
        self.0.get_db()
    }
}

impl Connect for Connection {
    fn connect<'a, T>(_info: T) -> redis::RedisFuture<'a, Self>
    where
        T: redis::IntoConnectionInfo + Send + 'a,
    {
        panic!("Unused, should use `connect_extended`")
    }

    fn connect_extended<'a, T>(
        info: T,
        additional_data: Option<std::sync::Arc<dyn std::any::Any + Send + Sync>>,
    ) -> redis::RedisFuture<'a, Self>
    where
        T: redis::IntoConnectionInfo + Send + 'a,
    {
        futures::FutureExt::boxed(async move {
            let connection_info = info.into_connection_info()?;
            let connection_timeout = additional_data
                .map(|val| {
                    val.downcast::<ConnectionConfiguration>()
                        .unwrap()
                        .connection_timeout
                })
                .unwrap_or(Duration::MAX);
            let client = redis::Client::open(connection_info)?;
            tokio::time::timeout(
                connection_timeout,
                client.get_multiplexed_tokio_connection(),
            )
            .await
            .map_err(|_| io::Error::from(io::ErrorKind::TimedOut).into())
            .and_then(|res| res)
            .map(|connection| Connection(connection))
        })
    }
}

impl Unpin for Connection {}
```

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -43,7 +43,6 @@ use std::time::Duration;
 
 use rand::{seq::IteratorRandom, thread_rng, Rng};
 
-use crate::cluster_client::RetryParams;
 use crate::cluster_pipeline::UNROUTABLE_ERROR;
 use crate::cluster_routing::{
     MultipleNodeRoutingInfo, ResponsePolicy, SingleNodeRoutingInfo, SlotAddr,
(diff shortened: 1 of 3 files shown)


jaymell:
I think this looks like a reasonable potential extension to the trait, but as far as connection timeouts are concerned, the simpler option to me is still the most appealing. The `Connect` trait (at least as I understand it, I didn't write it) is primarily there for the mock tests, so I would prefer to avoid leaving the connection timeouts up to the implementation of the trait.

Issue #703, opened 2022-11-03
redis-sentinel


omer-shtivi:
Hello,
Is there a plan to support redis with Sentinel in order to discover which replica is the master?
https://redis.io/docs/management/sentinel/#:~:text=Configuration%20provider.%20Sentinel%20acts%20as%20a%20source%20of%20authority%20for%20clients%20service%20discovery%3A%20clients%20connect%20to%20Sentinels%20in%20order%20to%20ask%20for%20the%20address%20of%20the%20current%20Redis%20master%20responsible%20for%20a%20given%20service.%20If%20a%20failover%20occurs%2C%20Sentinels%20will%20report%20the%20new%20address.

djc:
There is no plan that I am aware of, though @jaymell may have better context on this.

omer-shtivi:
Thanks @djc  for the response.
If I will contribute the code myself, will you accept it?
Looks like a very common deployment module, and I would like to get rid of our haproxy which seams is causing us troubles.

nihohit:
Sentinel support was added.

Issue #539, opened 2021-09-12
Sentinel support?


guaychou:
How to create client with sentinel support that can auto listen and watch the master node changes?

doyshinda:
I took a stab at implementing a SentinelClient in my [fork](https://github.com/doyshinda/redis-rs/tree/sentinel). It supports querying for the master directly or subscribing to updates (you'll receive a message through a crossbeam channel when master node changes).

**NOTE:** This code has not been fully tested and should not be used in production.

aljazerzen:
For anyone in the future: I've needed this and made [my own library](https://github.com/aljazerzen/redis-sentinel-rs) that correctly implements sentinel connection and can be used as extension to this crate.

It's a fork of some other library, but with a major overhaul.

PR #836, opened 2023-05-05
Add basic Sentinel functionality

felipou:
Hi! I've created this implementation of some Redis Sentinel logic, which I think would be a nice addition to this crate. I'm going to need something like this anyway for my projects, so I've decided to try and contribute it here first.

This is my first Rust open-source contribution, and I have very little Rust experience, so if anything seems weird please enlighten me!

Before doing this, I've seen that there are two issues about adding sentinel support:
https://github.com/redis-rs/redis-rs/issues/539
https://github.com/redis-rs/redis-rs/issues/703

And two attempts to implement this, one in a fork, and one in a separate crate:
https://github.com/redis-rs/redis-rs/compare/main...doyshinda:redis-rs:sentinel
https://github.com/aljazerzen/redis-sentinel-rs/blob/main/src/client.rs

Both are quite old and not quite the way I'd like then to be. Also, I've been wanting to practice writing some Rust code, so I decided to implement something myself, which got me to this PR.

I'm not very confident I got the API structured in a good way, so any suggestions on how to improve it would be greatly appreciated. I tried to come up with something along the lines of the Python client (https://github.com/redis/redis-py) but also trying the follow the recommendations for sentinel clients (https://redis.io/docs/reference/sentinel-clients).

### DIFF
--- a/redis/src/sentinel.rs
+++ b/redis/src/sentinel.rs
@@ -27,7 +27,7 @@
 //! use redis::sentinel::{ SentinelServerType, SentinelClient };
 //!
 //! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/", "redis://127.0.0.1:6377/"];
-//! let mut master_client = SentinelClient::build(nodes, String::from("master_name"), None, SentinelServerType::Master).unwrap();
+//! let mut master_client = SentinelClient::build(nodes.clone(), String::from("master_name"), None, SentinelServerType::Master).unwrap();
 //! let mut replica_client = SentinelClient::build(nodes, String::from("master_name"), None, SentinelServerType::Replica).unwrap();
 //! let mut master_conn = master_client.get_connection().unwrap();
 //! let mut replica_conn = replica_client.get_connection().unwrap();
@@ -69,7 +69,7 @@
 //!     .master_for(
 //!         "master_name",
 //!         Some(&SentinelNodeConnectionInfo {
-//!             tls_mode: Some(redis::cluster::TlsMode::Secure),
+//!             tls_mode: Some(redis::sentinel::TlsMode::Secure),
 //!             redis_connection_info: None,
 //!         }),
 //!     )
@@ -83,11 +83,12 @@
 //! use redis::{ Commands, RedisConnectionInfo };
 //! use redis::sentinel::{ SentinelServerType, SentinelClient, SentinelNodeConnectionInfo };
 //!
+//! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/", "redis://127.0.0.1:6377/"];
 //! let mut master_client = SentinelClient::build(
 //!     nodes,
 //!     String::from("master1"),
 //!     Some(SentinelNodeConnectionInfo {
-//!         tls_mode: Some(redis::cluster::TlsMode::Insecure),
+//!         tls_mode: Some(redis::sentinel::TlsMode::Insecure),
 //!         redis_connection_info: Some(RedisConnectionInfo {
 //!             db: 0,
 //!             username: Some(String::from("user")),
--- a/redis/src/sentinel.rs
+++ b/redis/src/sentinel.rs
@@ -403,127 +403,6 @@ async fn async_try_single_sentinel<T: FromRedisValue>(
     }
 }
 
-// async methods
-#[cfg(feature = "aio")]
-impl Sentinel {
-    /// Async version of [Sentinel::try_all_sentinels].
-    async fn async_try_all_sentinels<T: FromRedisValue>(&mut self, cmd: Cmd) -> RedisResult<T> {
-        let mut last_err = None;
-        for (connection_info, cached_connection) in self
-            .sentinels_connection_info
-            .iter()
-            .zip(self.async_connections_cache.iter_mut())
-        {
-            match async_try_single_sentinel(cmd.clone(), connection_info, cached_connection).await {
-                Ok(result) => {
-                    return Ok(result);
-                }
-                Err(err) => {
-                    last_err = Some(err);
-                }
-            }
-        }
-
-        // We can unwrap here because we know there is at least one connection info.
-        // Maybe we should have a more specific error for this situation?
-        Err(last_err.expect("There should be at least one connection info"))
-    }
-
-    /// Async version of [Sentinel::get_sentinel_masters].
-    async fn async_get_sentinel_masters(&mut self) -> RedisResult<Vec<HashMap<String, String>>> {
-        self.async_try_all_sentinels(sentinel_masters_cmd()).await
-    }
-
-    async fn async_get_sentinel_replicas<'a>(
-        &mut self,
-        service_name: &'a str,
-    ) -> RedisResult<Vec<HashMap<String, String>>> {
-        self.async_try_all_sentinels(sentinel_replicas_cmd(service_name))
-            .await
-    }
-
-    async fn async_find_master_address<'a>(
-        &mut self,
-        service_name: &str,
-        node_connection_info: &SentinelNodeConnectionInfo,
-    ) -> RedisResult<ConnectionInfo> {
-        let masters = self.async_get_sentinel_masters().await?;
-        async_find_valid_master(masters, service_name, node_connection_info).await
-    }
-
-    async fn async_find_valid_replica_addresses<'a>(
-        &mut self,
-        service_name: &str,
-        node_connection_info: &SentinelNodeConnectionInfo,
-    ) -> RedisResult<Vec<ConnectionInfo>> {
-        let replicas = self.async_get_sentinel_replicas(service_name).await?;
-        Ok(async_get_valid_replicas_addresses(replicas, node_connection_info).await)
-    }
-
-    /// Determines the masters address for the given name, and returns a client for that
-    /// master.
-    ///
-    /// To determine the master we ask the sentinels one by one until one of
-    /// then correctly responds to the SENTINEL MASTERS command, and then we check to
-    /// see if there is a master with the given name; if there is, we create a
-    /// client with that address (ip and port) and the extra information from the
-    /// `node_connection_info` parameter.
-    pub async fn async_master_for(
-        &mut self,
-        service_name: &str,
-        node_connection_info: Option<&SentinelNodeConnectionInfo>,
-    ) -> RedisResult<Client> {
-        let address = self
-            .async_find_master_address(service_name, node_connection_info.unwrap_or_default())
-            .await?;
-        Client::open(address)
-    }
-
-    /// Connects to the first available replica of the given master name.
-    ///
-    /// After listing the valid replicas (those without a sdown or odown flag) for the
-    /// given master name, we start trying to connect to then one by one, starting at a
-    /// random index (simple random derived from the nanoseconds part of the current
-    /// unix timestamp).
-    pub async fn async_replica_for(
-        &mut self,
-        service_name: &str,
-        node_connection_info: Option<&SentinelNodeConnectionInfo>,
-    ) -> RedisResult<Client> {
-        let addresses = self
-            .async_find_valid_replica_addresses(
-                service_name,
-                node_connection_info.unwrap_or_default(),
-            )
-            .await?;
-        let start_index = random_replica_index(addresses.len());
-        try_connect_to_first_replica(&addresses, start_index)
-    }
-
-    /// Connects to the first available replica of the given master name, starting in a
-    /// different (incremented) index each time the function is called.
-    ///
-    /// After listing the valid replicas (those without a sdown or odown flag) for the
-    /// given master name, we start trying to connect to then one by one, starting at a
-    /// rotating index which gets incremented on each call to this function.
-    pub async fn async_replica_rotate_for<'a>(
-        &mut self,
-        service_name: &str,
-        node_connection_info: Option<&SentinelNodeConnectionInfo>,
-    ) -> RedisResult<Client> {
-        let addresses = self
-            .async_find_valid_replica_addresses(
-                service_name,
-                node_connection_info.unwrap_or_default(),
-            )
-            .await?;
-        if !addresses.is_empty() {
-            self.replica_start_index = (self.replica_start_index + 1) % addresses.len();
-        }
-        try_connect_to_first_replica(&addresses, self.replica_start_index)
-    }
-}
-
 fn reconnect(
     connection: &mut Option<Connection>,
     connection_info: &ConnectionInfo,
@@ -593,7 +472,7 @@ impl Sentinel {
     /// error of the last attempt.
     ///
     /// For each sentinel, we first check if there is a cached connection, and if not
-    /// we attemp to connect to it (skipping that sentinel if there is an error during
+    /// we attempt to connect to it (skipping that sentinel if there is an error during
     /// the connection). Then, we attempt to execute the given command with the cached
     /// connection. If there is an error indicating that the connection is invalid, we
     /// reconnect and try one more time in the new connection.
@@ -616,13 +495,11 @@ impl Sentinel {
         }
 
         // We can unwrap here because we know there is at least one connection info.
-        // Maybe we should have a more specific error for this situation?
         Err(last_err.expect("There should be at least one connection info"))
     }
 
     /// Get a list of all masters (using the command SENTINEL MASTERS) from the
-    /// sentinels (we'll return the result of the first sentinel who responds to that
-    /// command without errors).
+    /// sentinels.
     fn get_sentinel_masters(&mut self) -> RedisResult<Vec<HashMap<String, String>>> {
         self.try_all_sentinels(sentinel_masters_cmd())
     }
@@ -654,12 +531,6 @@ impl Sentinel {
 
     /// Determines the masters address for the given name, and returns a client for that
     /// master.
-    ///
-    /// To determine the master we ask the sentinels one by one until one of
-    /// then correctly responds to the SENTINEL MASTERS command, and then we check to
-    /// see if there is a master with the given name; if there is, we create a
-    /// client with that address (ip and port) and the extra information from the
-    /// `node_connection_info` parameter.
     pub fn master_for(
         &mut self,
         service_name: &str,
@@ -670,12 +541,7 @@ impl Sentinel {
         Client::open(connection_info)
     }
 
-    /// Connects to the first available replica of the given master name.
-    ///
-    /// After listing the valid replicas (those without a sdown or odown flag) for the
-    /// given master name, we start trying to connect to then one by one, starting at a
-    /// random index (simple random derived from the nanoseconds part of the current
-    /// unix timestamp).
+    /// Connects to a randomly chosen replica of the given master name.
     pub fn replica_for(
         &mut self,
         service_name: &str,
@@ -687,12 +553,10 @@ impl Sentinel {
         try_connect_to_first_replica(&addresses, start_index)
     }
 
-    /// Connects to the first available replica of the given master name, starting in a
-    /// different (incremented) index each time the function is called.
-    ///
-    /// After listing the valid replicas (those without a sdown or odown flag) for the
-    /// given master name, we start trying to connect to then one by one, starting at a
-    /// rotating index which gets incremented on each call to this function.
+    /// Attempts to connect to a different replica of the given master name each time.
+    /// There is no guarantee that we'll actually be connecting to a different replica
+    /// in the next call, but in a static set of replicas (no replicas added or
+    /// removed), on average we'll choose each replica the same number of times.
     pub fn replica_rotate_for(
         &mut self,
         service_name: &str,
@@ -700,6 +564,112 @@ impl Sentinel {
     ) -> RedisResult<Client> {
         let addresses = self
             .find_valid_replica_addresses(service_name, node_connection_info.unwrap_or_default())?;
+        if !addresses.is_empty() {
+            self.replica_start_index = (self.replica_start_index + 1) % addresses.len();
+        }
+        try_connect_to_first_replica(&addresses, self.replica_start_index)
+    }
+}
+
+// Async versions of the public methods above, along with async versions of private
+// methods required for the public methods.
+#[cfg(feature = "aio")]
+impl Sentinel {
+    async fn async_try_all_sentinels<T: FromRedisValue>(&mut self, cmd: Cmd) -> RedisResult<T> {
+        let mut last_err = None;
+        for (connection_info, cached_connection) in self
+            .sentinels_connection_info
+            .iter()
+            .zip(self.async_connections_cache.iter_mut())
+        {
+            match async_try_single_sentinel(cmd.clone(), connection_info, cached_connection).await {
+                Ok(result) => {
+                    return Ok(result);
+                }
+                Err(err) => {
+                    last_err = Some(err);
+                }
+            }
+        }
+
+        // We can unwrap here because we know there is at least one connection info.
+        Err(last_err.expect("There should be at least one connection info"))
+    }
+
+    async fn async_get_sentinel_masters(&mut self) -> RedisResult<Vec<HashMap<String, String>>> {
+        self.async_try_all_sentinels(sentinel_masters_cmd()).await
+    }
+
+    async fn async_get_sentinel_replicas<'a>(
+        &mut self,
+        service_name: &'a str,
+    ) -> RedisResult<Vec<HashMap<String, String>>> {
+        self.async_try_all_sentinels(sentinel_replicas_cmd(service_name))
+            .await
+    }
+
+    async fn async_find_master_address<'a>(
+        &mut self,
+        service_name: &str,
+        node_connection_info: &SentinelNodeConnectionInfo,
+    ) -> RedisResult<ConnectionInfo> {
+        let masters = self.async_get_sentinel_masters().await?;
+        async_find_valid_master(masters, service_name, node_connection_info).await
+    }
+
+    async fn async_find_valid_replica_addresses<'a>(
+        &mut self,
+        service_name: &str,
+        node_connection_info: &SentinelNodeConnectionInfo,
+    ) -> RedisResult<Vec<ConnectionInfo>> {
+        let replicas = self.async_get_sentinel_replicas(service_name).await?;
+        Ok(async_get_valid_replicas_addresses(replicas, node_connection_info).await)
+    }
+
+    /// Determines the masters address for the given name, and returns a client for that
+    /// master.
+    pub async fn async_master_for(
+        &mut self,
+        service_name: &str,
+        node_connection_info: Option<&SentinelNodeConnectionInfo>,
+    ) -> RedisResult<Client> {
+        let address = self
+            .async_find_master_address(service_name, node_connection_info.unwrap_or_default())
+            .await?;
+        Client::open(address)
+    }
+
+    /// Connects to a randomly chosen replica of the given master name.
+    pub async fn async_replica_for(
+        &mut self,
+        service_name: &str,
+        node_connection_info: Option<&SentinelNodeConnectionInfo>,
+    ) -> RedisResult<Client> {
+        let addresses = self
+            .async_find_valid_replica_addresses(
+                service_name,
+                node_connection_info.unwrap_or_default(),
+            )
+            .await?;
+        let start_index = random_replica_index(addresses.len());
+        try_connect_to_first_replica(&addresses, start_index)
+    }
+
+    /// Attempts to connect to a different replica of the given master name each time.
+    /// There is no guarantee that we'll actually be connecting to a different replica
+    /// in the next call, but in a static set of replicas (no replicas added or
+    /// removed), on average we'll choose each replica the same number of times.
+    pub async fn async_replica_rotate_for<'a>(
+        &mut self,
+        service_name: &str,
+        node_connection_info: Option<&SentinelNodeConnectionInfo>,
+    ) -> RedisResult<Client> {
+        let addresses = self
+            .async_find_valid_replica_addresses(
+                service_name,
+                node_connection_info.unwrap_or_default(),
+            )
+            .await?;
         if !addresses.is_empty() {
             self.replica_start_index = (self.replica_start_index + 1) % addresses.len();
         }
@@ -720,7 +690,10 @@ pub enum SentinelServerType {
 /// on-demand based on information fetched from the sentinels. Uses the Sentinel type
 /// internally. This is basic an utility to help make it easier to use sentinels but
 /// with an interface similar to the client (`get_connection` and
-/// `get_async_connection`).
+/// `get_async_connection`). The type of server (master or replica) and name of the
+/// desired master are specified when constructing an instance, so it will always
+/// return connections to the same target (for example, always to the master with name
+/// "mymaster123", or always to replicas of the master "another-master-abc").
 pub struct SentinelClient {
     sentinel: Sentinel,
     service_name: String,
--- a/redis/src/sentinel.rs
+++ b/redis/src/sentinel.rs
@@ -101,7 +101,7 @@
 //! ```
 //!
 
-use std::collections::HashMap;
+use std::{collections::HashMap, num::NonZeroUsize};
 
 #[cfg(feature = "aio")]
 use futures_util::StreamExt;
@@ -209,21 +209,14 @@ fn is_replica_valid(replica_info: &HashMap<String, String>) -> bool {
         && replica_info["port"].parse::<u16>().is_ok()
 }
 
-/// Generates a random value in the 0..max range. If max is zero, the returned value is
-/// also zero.
-fn random_replica_index(max: usize) -> usize {
-    if max == 0 {
-        // This is just to avoid panicking in the gen_range function, this value does
-        // not have any special meaning, and is not expected to be used.
-        0
-    } else {
-        rand::thread_rng().gen_range(0..max)
-    }
+/// Generates a random value in the 0..max range.
+fn random_replica_index(max: NonZeroUsize) -> usize {
+    rand::thread_rng().gen_range(0..max.into())
 }
 
 fn try_connect_to_first_replica(
     addresses: &Vec<ConnectionInfo>,
-    start_index: usize,
+    start_index: Option<usize>,
 ) -> Result<Client, crate::RedisError> {
     if addresses.is_empty() {
         fail!((
@@ -231,6 +224,8 @@ fn try_connect_to_first_replica(
             "No valid replica found in sentinel for given name",
         ));
     }
+
+    let start_index = start_index.unwrap_or(0);
 
     let mut last_err = None;
     for i in 0..addresses.len() {
@@ -463,7 +458,7 @@ impl Sentinel {
             sentinels_connection_info,
             connections_cache,
             async_connections_cache,
-            replica_start_index: random_replica_index(1000000),
+            replica_start_index: random_replica_index(NonZeroUsize::new(1000000).unwrap()),
         })
     }
 
@@ -549,7 +544,7 @@ impl Sentinel {
     ) -> RedisResult<Client> {
         let addresses = self
             .find_valid_replica_addresses(service_name, node_connection_info.unwrap_or_default())?;
-        let start_index = random_replica_index(addresses.len());
+        let start_index = NonZeroUsize::new(addresses.len()).map(random_replica_index);
         try_connect_to_first_replica(&addresses, start_index)
     }
 
@@ -567,7 +562,7 @@ impl Sentinel {
         if !addresses.is_empty() {
             self.replica_start_index = (self.replica_start_index + 1) % addresses.len();
         }
-        try_connect_to_first_replica(&addresses, self.replica_start_index)
+        try_connect_to_first_replica(&addresses, Some(self.replica_start_index))
     }
 }
 
@@ -651,7 +646,7 @@ impl Sentinel {
                 node_connection_info.unwrap_or_default(),
             )
             .await?;
-        let start_index = random_replica_index(addresses.len());
+        let start_index = NonZeroUsize::new(addresses.len()).map(random_replica_index);
         try_connect_to_first_replica(&addresses, start_index)
     }
 
@@ -673,7 +668,7 @@ impl Sentinel {
         if !addresses.is_empty() {
             self.replica_start_index = (self.replica_start_index + 1) % addresses.len();
         }
-        try_connect_to_first_replica(&addresses, self.replica_start_index)
+        try_connect_to_first_replica(&addresses, Some(self.replica_start_index))
     }
 }
 
--- a/redis/src/sentinel.rs
+++ b/redis/src/sentinel.rs
@@ -1,0 +1,558 @@
+//! Defines a Sentinel type that connects to Redis sentinels and creates clients to
+//! master or replica nodes.
+//!
+//! # Example
+//! ```rust,no_run
+//! use redis::Commands;
+//! use redis::sentinel::Sentinel;
+//!
+//! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/", "redis://127.0.0.1:6377/"];
+//! let sentinel = Sentinel::build(nodes).unwrap();
+//! let mut master = sentinel.master_for("master_name").unwrap().get_connection().unwrap();
+//! let mut replica = sentinel.replica_for("master_name").unwrap().get_connection().unwrap();
+//!
+//! let _: () = master.set("test", "test_data").unwrap();
+//! let rv: String = replica.get("test").unwrap();
+//!
+//! assert_eq!(rv, "test_data");
+//! ```
+//!
+//! There is also a SentinelClient which acts like a regular Client, providing the
+//! `get_connection` and `get_async_connection` methods, internally using the Sentinel
+//! type to create clients on demand for the desired node type (Master or Replica).
+//!
+//! # Example
+//! ```rust,no_run
+//! use redis::Commands;
+//! use redis::sentinel::{ SentinelServerType, SentinelClient };
+//!
+//! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/", "redis://127.0.0.1:6377/"];
+//! let master_client = SentinelClient::build(nodes, "master_name", SentinelServerType::Master).unwrap();
+//! let replica_client = SentinelClient::build(nodes, "master_name", SentinelServerType::Replica).unwrap();
+//! let mut master_conn = master_client.get_connection().unwrap();
+//! let mut replica_conn = replica_client.get_connection().unwrap();
+//!
+//! let _: () = master_conn.set("test", "test_data").unwrap();
+//! let rv: String = replica_conn.get("test").unwrap();
+//!
+//! assert_eq!(rv, "test_data");
+//! ```
+//!
+
+use std::{
+    collections::HashMap,
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+#[cfg(feature = "aio")]
+use std::future::Future;
+
+#[cfg(feature = "aio")]
+use futures_util::StreamExt;
+
+use crate::{
+    connection::ConnectionInfo, types::RedisResult, Client, Connection, ErrorKind, FromRedisValue,
+    IntoConnectionInfo, Value,
+};
+
+/// The Sentinel type, serves as a special purpose client which builds other clients on
+/// demand.
+#[derive(Debug, Clone)]
+pub struct Sentinel {
+    sentinels_connection_info: Vec<ConnectionInfo>,
+    replica_start_index: usize,
+}
+
+fn sentinel_masters_cmd() -> crate::Cmd {
+    let mut cmd = crate::cmd("SENTINEL");
+    cmd.arg("MASTERS");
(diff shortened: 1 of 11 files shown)


redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+    /// Determines the masters address for the given name, and returns a client for that
+    /// master.
+    ///
+    /// To determine the master we ask the sentinels one by one until one of
+    /// then correctly responds to the SENTINEL MASTERS command, and then we check to
+    /// see if there is a master with the given name; if there is, we create a
+    /// client with that address (ip and port).
+    pub fn master_for(&self, service_name: &str) -> RedisResult<Client> {
+        let address = self.find_master_address(service_name)?;
+        Client::open(address)
```
nihohit, line 449: minor: why a client, and not connection info?

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+
+    fail!((
+        ErrorKind::MasterNameNotFound,
+        "Master with given name not found in sentinel",
+    ))
+}
+
+#[cfg(feature = "aio")]
+async fn async_check_role(ip: &str, port: u16, target_role: &str) -> bool {
+    if let Ok(client) = Client::open((ip, port)) {
```
nihohit, line 236: I believe this will fail if the server uses TLS. Either ask the user whether to use TLS or not, or assume that if the sentinel uses TLS, then the server does, too.

redis/tests/support/sentinel.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub fn with_modules(
+        masters: u16,
+        replicas_per_master: u16,
+        sentinels: u16,
+        modules: &[Module],
+    ) -> RedisSentinelCluster {
+        let mut servers = vec![];
+        let mut folders = vec![];
+        let mut master_ports = vec![];
+        let start_port = 7000;
```
nihohit, line 121: don't use a hard-coded port. Find an available port, like [here](https://github.com/redis-rs/redis-rs/blob/6be07049f4982af3b34c1be5cbd7cc0b7a06c6c0/redis/tests/support/mod.rs#L100).

Better that you find an available port for each node, too, instead of assuming that you have enough available ports in a row.

redis/tests/support/sentinel.rs
```diff
@@ -0,0 +1,271 @@
...
+                servers.push(spawn_replica_server(replica_port, port, &tempdir, modules));
+                folders.push(tempdir);
+            }
+        }
+
+        // Wait for replicas to sync so that the sentinels discover them on the first try
+        sleep(Duration::from_millis(50));
+
+        let mut sentinel_servers = vec![];
+        let start_port = 27000;
```
nihohit, line 148: same

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,404 @@
...
+    redis::cmd("SET")
+        .arg("{x}key1")
+        .arg(b"foo")
+        .execute(&mut master_con);
+    redis::cmd("SET")
+        .arg(&["{x}key2", "bar"])
+        .execute(&mut master_con);
+
+    // Read commands would go to the replica nodes
+    assert_eq!(
```
nihohit, line 61: this test will be flakey, since it assumes that the replica will sync in time. Instead of testing set & get (here and in the previous test), just send an `INFO REPLICATION` command and check that the node is indeed a master or replica, and if it is a replica, that it is a replica of the correct master.

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,404 @@
...
+    redis::cmd("SET")
+        .arg("{x}key1")
+        .arg(b"foo")
+        .execute(&mut master_con);
+    redis::cmd("SET")
+        .arg(&["{x}key2", "bar"])
+        .execute(&mut master_con);
+
+    for _ in 0..20 {
+        let mut replica_con = sentinel
```
nihohit, line 90: same, just test that you get a different name on INFO

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,404 @@
...
+    let sentinel = context.sentinel_mut();
+
+    for _ in 0..20 {
+        let mut replica_con = sentinel
+            .replica_rotate_for("master1")
+            .unwrap()
+            .get_connection()
+            .unwrap();
+
+        assert_eq!(
```
nihohit, line 137: same throughout

nihohit (commented):
Very impressive work, well done!
I think that the general design is good, most of the comments are about the implementation and the tests.

nihohit:
@jaymell a couple of questions of code order - 
1. Should this be behind a feature flag?
2. For future extensibility, maybe the module should be under its own folder - sentinel/client.rs, sentinal/connection.rs ?

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+            replica_start_index: random_replica_index(1000000),
+        })
+    }
+
+    fn try_all_sentinels<F, T>(&self, f: F) -> RedisResult<T>
+    where
+        F: Fn(&ConnectionInfo) -> RedisResult<T>,
+    {
+        let mut last_err = None;
+        for connection_info in self.sentinels_connection_info.iter() {
```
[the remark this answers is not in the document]
felipou: To do that, I'd have to require a mutable reference, or use interior mutability, right? What's the most common/recommended approach in this case?

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+            !flags.contains("s_down") && !flags.contains("o_down")
+        })
+        && replica_info["port"].parse::<u16>().is_ok()
+}
+
+fn random_replica_index(max: usize) -> usize {
+    if max == 0 {
+        0
+    } else {
+        let nanos = SystemTime::now()
```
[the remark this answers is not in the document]
felipou: Sorry, I should have commented (both in the PR and the code) why I did this. I mentioned this in my reply to you last comment below, but basically I wanted to avoid an extra dependency just for this, since it seems rather simple and harmless enough that any seemingly random number would suffice. Specially if this isn't going to be behind a feature flag. Does that make sense? Or am I being overzealous about adding new dependencies?

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+    let mut cmd = crate::cmd("SENTINEL");
+    cmd.arg("SLAVES"); // For compatibility with older redis versions
+    cmd.arg(master_name);
+    cmd
+}
+
+#[cfg(feature = "aio")]
+async fn async_try_sentinel_masters(
+    connection_info: &ConnectionInfo,
+) -> RedisResult<Vec<HashMap<String, String>>> {
```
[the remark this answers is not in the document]
felipou: This is because I'm using the `SENTINEL MASTERS` command, which may return multiple masters, instead of `SENTINEL get-master-addr-by-name master-name`. The latter is actually the suggested command in the sentinel client doc, but I the former is what redis-py uses, and I believe it's because only this command returns the master's flags (which are used to identify whether a master is valid).

The return of this method is then used by `async_get_first_valid_master` to find the master with the desired name (and now I'm thinking that this method name doesn´t make much sense, since there can be only one master with each name).

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+        .await?;
+    Ok(result)
+}
+
+fn try_sentinel_masters(
+    connection_info: &ConnectionInfo,
+) -> RedisResult<Vec<HashMap<String, String>>> {
+    let sentinel_client = Client::open(connection_info.clone())?;
+    let mut conn = sentinel_client.get_connection()?;
+    let result: Vec<HashMap<String, String>> = sentinel_masters_cmd().query(&mut conn)?;
```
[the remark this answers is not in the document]
felipou: Yes, good point! Fixing it now. I'm doing the same in the async version.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+}
+
+fn try_sentinel_replicas(
+    connection_info: &ConnectionInfo,
+    master_name: &str,
+) -> RedisResult<Vec<HashMap<String, String>>> {
+    let sentinel_client = Client::open(connection_info.clone())?;
+    let mut conn = sentinel_client.get_connection()?;
+    let result: Vec<HashMap<String, String>> =
+        sentinel_replicas_cmd(master_name).query(&mut conn)?;
```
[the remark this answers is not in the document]
felipou: Also fixing now, along with async version.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+    let mut last_err = None;
+    for i in 0..addresses.len() {
+        let index = (i + start_index) % addresses.len();
+        match Client::open(addresses[index].clone()) {
+            Ok(client) => return Ok(client),
+            Err(err) => last_err = Some(err),
+        }
+    }
+
+    // We can unwrap here because we know there is at least one address.
```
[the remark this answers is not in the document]
felipou: I'll try to improve this comment further. My point was that if we have at least one address, we can guarantee there will be an error there (since any address would either store an error or just return a client).

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+    Err(last_err.expect("There should be at least one address"))
```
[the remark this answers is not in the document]
felipou: Also trying to improve the message here. This actually reminded me that I wasn't making sure there is at least one address. I'll add this check to the `Sentinel::build` function!

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+            let port = info["port"].parse::<u16>().unwrap();
+            (ip, port)
+        })
+}
+
+fn check_role_result(result: &RedisResult<Vec<Value>>, target_role: &str) -> bool {
+    if let Ok(values) = result {
+        if !values.is_empty() {
+            if let Ok(role) = String::from_redis_value(&values[0]) {
+                if role.to_ascii_lowercase() == target_role {
```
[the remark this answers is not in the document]
felipou: Changing it now.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+        }
+    }
+    false
+}
+
+fn check_role(ip: &str, port: u16, target_role: &str) -> bool {
+    if let Ok(client) = Client::open((ip, port)) {
+        if let Ok(mut conn) = client.get_connection() {
+            let result: RedisResult<Vec<Value>> = crate::cmd("ROLE").query(&mut conn);
+            if check_role_result(&result, target_role) {
```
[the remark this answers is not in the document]
felipou: Changing it now.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,558 @@
...
+        "Master with given name not found in sentinel",
+    ))
+}
+
+#[cfg(feature = "aio")]
+async fn async_check_role(ip: &str, port: u16, target_role: &str) -> bool {
+    if let Ok(client) = Client::open((ip, port)) {
+        if let Ok(mut conn) = client.get_async_connection().await {
+            let result: RedisResult<Vec<Value>> = crate::cmd("ROLE").query_async(&mut conn).await;
+            if check_role_result(&result, target_role) {
```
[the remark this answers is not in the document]
felipou: Changing it now.

redis/src/sentinel.rs
felipou: I just hadn't thought about this possibility, you think it would be better? Is a connection info useful for anything else than creating a client? :thinking:

redis/tests/support/sentinel.rs
felipou: I agree, I just did it like this because the cluster tests are like this, and I copied some code from it: https://github.com/redis-rs/redis-rs/blob/6be07049f4982af3b34c1be5cbd7cc0b7a06c6c0/redis/tests/support/cluster.rs#L78

I'll try and change to use a random port now.

redis/tests/support/sentinel.rs
felipou: Ok, I'll try to improve here too!

redis/tests/test_sentinel.rs
felipou: Ok, makes sense, I'll change it now!

redis/tests/test_sentinel.rs
felipou: Changing it now.

redis/tests/test_sentinel.rs
felipou: Changing it now.

felipou:
Just pushed 3 commits fixing some of the things pointed out.

Two other things I'll try to do later today: support for TLS connections (to masters and replicas), and random ports (instead of fixed ones) in the tests.

For the other points, I've replied the comments and will wait for the responses before modifying anything.

redis/src/sentinel.rs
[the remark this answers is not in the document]
nihohit: interesting question.
In general I would've gone with interior mutability if you believe this method might be called in parallel from multiple sources, and mutable reference if you believe that this method usually would be called once every so often.
If you're going with interior mutability, make sure that you're using a mutex that is safe to hold over an await point, since the mutex might be used in both sync and async methods.

redis/src/sentinel.rs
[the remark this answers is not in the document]
nihohit: both `oorandom`, `rand` and `fastrand` are used by in redis-rs dependency tree.
IMO, there's no problem with adding a direct dependency on one of them, but I'll give @jaymell the final word on this.

redis/src/sentinel.rs
nihohit: well, for starters, connection info can be enriched to add TLS :)
not a big deal, it's my preference to return the basic data.

redis/tests/test_sentinel.rs
```diff
@@ -1,9 +1,54 @@
...
+fn parse_replication_info(value: &str) -> HashMap<&str, &str> {
+    let info_map: std::collections::HashMap<&str, &str> = value
+        .split("\r\n")
+        .filter(|line| !line.trim_start().starts_with('#'))
+        .filter_map(|line| line.split_once(':'))
+        .collect();
+    info_map
+}
+
+fn check_master_role(conn: &mut Connection) {
```
nihohit, line 18: assert_is_connection_to_master

redis/tests/test_sentinel.rs
```diff
@@ -1,9 +1,54 @@
...
+    let info: String = redis::cmd("INFO").arg("REPLICATION").query(conn).unwrap();
+
+    let info_map = parse_replication_info(&info);
+    assert_eq!(info_map.get("role"), Some(&"master"));
+}
+
+fn check_replica_info_result(info: String, master_client: &Client) {
```
nihohit, line 25: consider:
`assert_connection_is_replica_of_correct_master(connection_info: String, expected_master_info: &ConnectionInfo)`

redis/tests/test_sentinel.rs
```diff
@@ -93,12 +98,7 @@ fn test_sentinel_read_from_multiple_replicas() {
             .get_connection()
             .unwrap();
 
-        assert_eq!(
-            redis::cmd("MGET")
-                .arg(&["{x}key1", "{x}key2"])
-                .query(&mut replica_con),
-            Ok(("foo".to_string(), b"bar".to_vec()))
-        );
+        check_replica_role_and_master(&mut replica_con, &master_client);
```
nihohit, line 101: you're checking that all of the connections are replicas of the correct master, but you're not testing the "multiple" part of the test name - you're not verifying that these are connections to separate replicas.

redis/tests/test_sentinel.rs
```diff
@@ -162,32 +147,44 @@ fn test_sentinel_client() {
...
+    use crate::{check_replica_info_result, parse_replication_info, support::*};
+
+    async fn async_check_master_role(conn: &mut Connection) {
+        let info: String = redis::cmd("INFO")
+            .arg("REPLICATION")
+            .query_async(conn)
+            .await
+            .unwrap();
+
+        let info_map = parse_replication_info(&info);
```
nihohit, line 175: why not use `check_master_role`?

redis/tests/test_sentinel.rs
```diff
@@ -294,11 +247,7 @@ pub mod async_tests {
                     .get_async_connection()
                     .await?;
 
-                let mget_result: RedisResult<(String, Vec<u8>)> = redis::cmd("MGET")
-                    .arg(&["{x}key1", "{x}key2"])
-                    .query_async(&mut replica_con)
-                    .await;
-                assert_eq!(mget_result, Ok(("foo".to_string(), b"bar".to_vec())));
+                async_check_replica_role_and_master(&mut replica_con, &master_client).await;
```
nihohit, line 250: same comment as for sync test

nihohit (commented):
good progress.

redis/src/sentinel.rs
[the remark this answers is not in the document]
nihohit: OK, so this isn't intuitive. Please document the functions, to clarify the intended flow. alternatively, call `async_get_first_valid_master ` from this function, and return the single master.

redis/tests/support/sentinel.rs
felipou: It was far easier than I initially thought, done!

redis/tests/test_sentinel.rs
felipou: Good suggestion, changed it!

redis/tests/test_sentinel.rs
felipou: Done!

redis/tests/test_sentinel.rs
felipou: Good point, so I just changed this to check that we actually connect to all replicas. I've also changed the number of replicas because there was only one. In the SentinelClient test I only get the connection (no client, so no address), so I couldn't do the same. But I'm thinking of replacing ConnectionInfo instead of Client as you suggested, this will probably allow me to improve the SentinelClient test too.

redis/tests/test_sentinel.rs
felipou: I hadn't extracted this part as a separate function yet, the `check_master_role` didn't use async. But while renaming with your suggestions, I've also extracted this part to a separate function.

redis/tests/test_sentinel.rs
felipou: Ok, done!

redis/tests/support/sentinel.rs
```diff
@@ -118,10 +119,10 @@ impl RedisSentinelCluster {
         let mut servers = vec![];
         let mut folders = vec![];
         let mut master_ports = vec![];
-        let start_port = 7000;
 
-        for node in 0..masters {
-            let port = start_port + (node * (replicas_per_master + 1));
+        for _ in 0..masters {
+            let port = get_random_available_port();
+            sleep(Duration::from_millis(25));
```
nihohit, line 125: why do you sleep after getting the port number? and why don't you sleep after lines 135?

redis/tests/test_sentinel.rs
```diff
@@ -106,20 +116,31 @@ fn test_sentinel_server_down() {
...
 
-    for _ in 0..20 {
-        let mut replica_con = sentinel
-            .replica_rotate_for("master1")
-            .unwrap()
-            .get_connection()
-            .unwrap();
+    let mut replica_conn_infos = vec![];
+
+    for _ in 0..3 {
```
nihohit, line 121: is this a verbatim copy of lines 85 and on? if so, please extract to a function

redis/tests/support/sentinel.rs
felipou: Just forgot it there really, I was having an issue with the random ports so I put it there to see if it helped, but it turns out it was something else entirely. I'll do some clean up in those tests, I've also noticed I've left some commented lines in the "spawn" functions which can be removed.

redis/tests/test_sentinel.rs
felipou: Yes, and there is something similar in the async tests, I'll extract both to a function.

redis/src/sentinel.rs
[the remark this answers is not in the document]
felipou: I tried adding some comments and renaming the `async_get_first_valid_master` function. I was not sure if I should repeat the comments for the async functions, so for now I just reference the non-async function.

nihohit:
when referencing other functions, please use the [rust linking format](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html)
Looking fine. I'm waiting for TLS support and connection caching, and then I'll go through the code again.

redis/src/sentinel.rs
nihohit: I think that's a good solution, but if you're looking to the simplest solution, it's just returning host and port.

felipou:
> when referencing other functions, please use the [rust linking format](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html)

Nice, didn't know about that, thanks! Guess I have to read that book now :smile: 

> Looking fine. I'm waiting for TLS support and connection caching, and then I'll go through the code again.

I just pushed the TLS support!

About the connection caching, I was about to start but I noticed/remembered that async connections are a different struct. Should I just have two vectors of connections, one for sync and other for async connections? I think I'll try implementing it that way and see how it goes.

felipou:
I've noticed some tests failed, so I pushed some commits to try and fix that.

But one of the problems was that I used the TlsMode enum that is in the cluster module, so it didn't exist in tests without this feature enabled. To fix it in the short term I just duplicated that enum inside the sentinel module, but maybe it's an opportunity to move this up? I think this could be useful in more places (for example, replacing the insecure attribute of the ConnectionAddr::TcpTls type). If not, any suggestions as to how we should specify a TLS connection?

nihohit:
Hi, great work. You found a reasonable way around the way the API is structured today in a way which is consistent with the current API language of the library.
But, I think that the last 2 commits significantly expand the size of this PR, and they contains near-duplication of existing structures, in order to maintain this consistency. This makes me think that maybe I was wrong, and the sentinel should only return a host-port tuple, and let the user handle connections, TLS, and caching. 
So, essentially, I think that this PR in it's current state is fine in the broad strokes - I still didn't go line-by-line on the last 2 commits, but the quick read-through looked mostly fine. The question is whether the library wants to provide the minimal API, which is just returning host-port tuples, or to support in-house all the features and behaviors that a user might want.
This kind of strategic choice should be done by @jaymell, not me.

@felipou I do think you've done a good job, and I hope that if we'll decide to minimize the blast zone of this change you won't take it as a critique of your contribution.

felipou:
> @felipou I do think you've done a good job, and I hope that if we'll decide to minimize the blast zone of this change you won't take it as a critique of your contribution.

Not at all, you make a fine point, and the API strategy surely comes first. But thanks for saying that anyway! In any case, this has been a great Rust programming and design exercise for me!

There's just one thing I'd like to point out: in the current state of the code, we do connect to the sentinel's monitored servers anyway, as part of the recommended strategy for sentinel clients, to query for the role of the server (using the ROLE command). If the goal is to minimize the API by removing the extra connection attributes, we'd have to either remove that check, or maybe just decide that we won't support TLS in this case?

Also, one thing I've been thinking since before creating the PR but forgot to mention: maybe this code should be in a new crate? After all, it is fairly independent. I even considered just creating the crate first, since I'm going to need this code anyway, and then just deprecating it if this got merged. But I guess for the most part I just believe this sits well within this crate, and as a user I would prefer it to be more feature complete than having to rely on another crate that maybe won't have the same level of quality and maintainability (this seems to be a common discussion as of late, minimal libs vs large libs).

redis/src/sentinel.rs
```diff
@@ -123,6 +122,16 @@ pub struct Sentinel {
     replica_start_index: usize,
 }
 
+/// TlsMode indicates use or do not use verification of certification.
+/// Check [ConnectionAddr](ConnectionAddr::TcpTls::insecure) for more.
+#[derive(Clone, Copy)]
+pub enum TlsMode {
```
nihohit, line 128: I agree that TlsMode should be moved out of cluster instead of copied.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
+    Master,
+    /// Replica connections only
+    Replica,
+}
+
+/// An alternative to the Client type which creates connections from clients created
+/// on-demand based on information fetched from the sentinels. Uses the Sentinel type
+/// internally. This is basic an utility to help make it easier to use sentinels but
+/// with an interface similar to the client (`get_connection` and
+/// `get_async_connection`).
```
nihohit, line 723: Just to clarify, please add to the doc that the client is limited to only returning one of replicas or masters, not both.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
+        service_name: &str,
+        node_connection_info: Option<&SentinelNodeConnectionInfo>,
+    ) -> RedisResult<Client> {
+        let addresses = self
+            .find_valid_replica_addresses(service_name, node_connection_info.unwrap_or_default())?;
+        let start_index = random_replica_index(addresses.len());
+        try_connect_to_first_replica(&addresses, start_index)
+    }
+
+    /// Connects to the first available replica of the given master name, starting in a
```
nihohit, line 690: This doc is a bit vague - what is the "first available replica", and what does the index index?

Specifically, there seems to be a hidden assumption that the list of replicas is the same between calls to the function, which is wrong if any replica was replaced.
I don't have a clearer phrasing off the top of my head, and I'm also not 100% sure that this is a meaningful problem, so my request here is just to see if you have better phrasing. If not, I believe this is fine.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
+        &mut self,
+        service_name: &str,
+        node_connection_info: Option<&SentinelNodeConnectionInfo>,
+    ) -> RedisResult<Client> {
+        let connection_info =
+            self.find_master_address(service_name, node_connection_info.unwrap_or_default())?;
+        Client::open(connection_info)
+    }
+
+    /// Connects to the first available replica of the given master name.
```
nihohit, line 673: This doc is a bit vague - what is the "first available replica"? I would just say "a replica" or "a randomly chosen replica", and drop the second paragraph.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
+            }
+        }
+
+        // We can unwrap here because we know there is at least one connection info.
+        // Maybe we should have a more specific error for this situation?
+        Err(last_err.expect("There should be at least one connection info"))
+    }
+
+    /// Get a list of all masters (using the command SENTINEL MASTERS) from the
+    /// sentinels (we'll return the result of the first sentinel who responds to that
```
nihohit, line 624: remove the part in parenthesis, it's an implementation detail.

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
+            replica_start_index: random_replica_index(1000000),
+        })
+    }
+
+    /// Try to execute the given command in each sentinel, returning the result of the
+    /// first one that executes without errors. If all return errors, we return the
+    /// error of the last attempt.
+    ///
+    /// For each sentinel, we first check if there is a cached connection, and if not
+    /// we attemp to connect to it (skipping that sentinel if there is an error during
```
nihohit, line 596: attemp -> attempt

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
+                    return Ok(result);
+                }
+                Err(err) => {
+                    last_err = Some(err);
+                }
+            }
+        }
+
+        // We can unwrap here because we know there is at least one connection info.
+        // Maybe we should have a more specific error for this situation?
```
nihohit, line 619: IMO this line can be removed

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
```
nihohit, line 428: IMO this line can be removed

redis/src/sentinel.rs
```diff
@@ -0,0 +1,795 @@
...
+        } else {
+            Err(err)
+        }
+    } else {
+        result
+    }
+}
+
+// async methods
+#[cfg(feature = "aio")]
```
nihohit, line 407: minor:
Maybe move this to under the sync implementation, since some of the comments reference functions from there?

Also, please be consistent, and make sure that any function comment that can reference the sync implementation does so. you don't have to add a comment just for the reference, but if you already have one, make sure that the repetition of content is minimal.

nihohit (approved):
Great work!

nihohit:
> There's just one thing I'd like to point out: in the current state of the code, we do connect to the sentinel's monitored servers anyway, as part of the recommended strategy for sentinel clients, to query for the role of the server (using the ROLE command). If the goal is to minimize the API by removing the extra connection attributes, we'd have to either remove that check, or maybe just decide that we won't support TLS in this case?

So we could return `(host, port)` pairs, but we won't know which are primaries and which are replicas? Then I think you've made the right decision.

> Also, one thing I've been thinking since before creating the PR but forgot to mention: maybe this code should be in a new crate?

A valid question, but I think it's a question to you. IMO this fits in this crate, as a basic Redis capability, as this is the most popular Rust Redis client. If you want to maintain your own crate, that's up to you.

felipou:
> > There's just one thing I'd like to point out: in the current state of the code, we do connect to the sentinel's monitored servers anyway, as part of the recommended strategy for sentinel clients, to query for the role of the server (using the ROLE command). If the goal is to minimize the API by removing the extra connection attributes, we'd have to either remove that check, or maybe just decide that we won't support TLS in this case?
> 
> So we could return `(host, port)` pairs, but we won't know which are primaries and which are replicas? Then I think you've made the right decision.
>

We know from the point of view of the sentinels, because when we want a primary we ask specifically for that, and the same for the replicas (there's a separate command to ask Redis for the replicas). I think the docs recommend that we check the role (with the ROLE command) because there can be some delays when syncing the server's roles with the sentinels. So it is possible that we could ask for a master, and the `(host, port)` returned is already a replica because the sentinel isn't in sync with that server at that point. But if we're returning just the address, there is always the possibility that server will have its role changed even if we ask for it, because there will always be some delay after getting the address, and before we actually connect to that address.

Thinking about it as I write this, I believe there could be some changes to best guarantee the role of a server when using the SentinelClient. Perhaps we should make the Sentinel just return the address as you suggested, and the SentinelClient can ask for the role after connecting to a server (and before returning the connection).

> > Also, one thing I've been thinking since before creating the PR but forgot to mention: maybe this code should be in a new crate?
> 
> A valid question, but I think it's a question to you. IMO this fits in this crate, as a basic Redis capability, as this is the most popular Rust Redis client. If you want to maintain your own crate, that's up to you.

I certainly do not want to maintain my own crate for this, I much prefer for this code to be in this crate!

nihohit:
> Thinking about it as I write this, I believe there could be some changes to best guarantee the role of a server when using the SentinelClient. Perhaps we should make the Sentinel just return the address as you suggested, and the SentinelClient can ask for the role after connecting to a server (and before returning the connection).

IMO the Sentinel struct should expose a behavior that matches the Redis best practices. We shouldn't split the correct behavior between multiple structs. So, I think the code is fine as-is.

redis/src/types.rs
```diff
@@ -99,6 +99,12 @@ pub enum ErrorKind {
     ExtensionError,
     /// Attempt to write to a read-only server
     ReadOnly,
+    /// Requested name not found among masters
+    MasterNameNotFound,
```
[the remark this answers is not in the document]
felipou: Ok, just updated it!

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,456 @@
...
+    sentinel: &mut Sentinel,
+    master_client: &Client,
+    node_conn_info: &SentinelNodeConnectionInfo,
+    number_of_replicas: u16,
+) -> Vec<ConnectionAddr> {
+    let mut replica_conn_infos = vec![];
+
+    for _ in 0..number_of_replicas {
+        let replica_client = sentinel
+            .replica_rotate_for("master1", Some(node_conn_info))
```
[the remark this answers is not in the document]
felipou: Good point. I guess I ignored this part because I was using just "master1" everywhere. But I've added it as a parameter to this function and all similar ones. I've also extracted the string "master1" to a single variable per test (I've thought about a single global one for the whole file, but thought it was best to just replicate it for each test).

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,456 @@
...
+    node_conn_info: &SentinelNodeConnectionInfo,
+    number_of_connections: u32,
+) {
+    for _ in 0..number_of_connections {
+        let replica_client = sentinel
+            .replica_rotate_for("master1", Some(node_conn_info))
+            .unwrap();
+        let mut replica_con = replica_client.get_connection().unwrap();
+
+        assert!(replica_conn_infos.contains(&replica_client.get_connection_info().addr));
```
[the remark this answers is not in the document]
felipou: Yes, this is related to the next comment, I'll reply below.

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,456 @@
...
+        assert!(!replica_conn_infos.contains(&replica_client.get_connection_info().addr));
+        replica_conn_infos.push(replica_client.get_connection_info().addr.clone());
+
+        assert_connection_is_replica_of_correct_master(&mut replica_con, master_client);
+    }
+
+    replica_conn_infos
+}
+
+fn assert_connect_to_known_replicas(
```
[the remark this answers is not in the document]
felipou: Now that you've asked, I'm not really sure. I guess I just wanted to make sure that a connecting a bunch of times would always end up in the same replicas found by the function `connect_to_all_replicas`.

The idea is that `connect_to_all_replicas` connects N times (the number of replicas), checking that each client is to a different address (a different replica), while also asserting that it is a connection to a replica of the target master, and storing the addresses to be returned.

These addresses are then passed to the `assert_connect_to_known_replicas` which will just connect a bunch of times to replicas, checking that every connection ends up in one of the known replicas (which were found by the `connect_to_all_replicas`), while also checking that the replica is of the target master.

So yes, both are quite similar. I'm not sure if it makes sense to keep the `assert_connect_to_known_replicas` function, if you don't see a point in keeping it, I can just remove it.

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,456 @@
...
+        sentinel,
+        replica_conn_infos,
+        &master_client,
+        &node_conn_info,
+        10,
+    );
+}
+
+#[test]
+fn test_sentinel_server_down() {
```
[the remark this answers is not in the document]
felipou: Just tests that everything works the same, we'll just skip the non-working sentinel and use the next one in the list. It's exactly the same code as `test_sentinel_connect_to_multiple_replicas`, but stopping the first sentinel before connecting to the replicas. Not sure why I didn't stop the sentinel before connecting to the master though, now I'm thinking it makes more sense to stop it earlier perhaps?

redis/tests/test_sentinel.rs
```diff
@@ -0,0 +1,456 @@
...
+            )
+            .await;
+
+            Ok::<(), RedisError>(())
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn test_sentinel_server_down_async() {
```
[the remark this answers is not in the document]
felipou: Same idea as above, almost the same test as `test_sentinel_connect_to_multiple_replicas_async`.

redis/src/sentinel.rs
felipou: Ok, done!

redis/src/sentinel.rs
felipou: I think this "first available replica" comes from a previous state of the code, and is related to using the function `try_connect_to_first_replica`. I tried to rewrite this whole description to try and make it clearer!

redis/src/sentinel.rs
felipou: Like I said above, I agree that it doesn't make sense, I've replaced it as you've suggested!

redis/src/sentinel.rs
felipou: Ok, done!

redis/src/sentinel.rs
felipou: Ok, done!

redis/src/sentinel.rs
felipou: Ok, done!

redis/src/sentinel.rs
felipou: I moved it below, makes sense. But I've also removed the references, and just expanded the comment above, is that ok? I've also kept the doc comments in the public methods, but they're exactly the same as the sync versions. You think in this case we should just reference the sync versions to avoid repetition? I'm not sure what would be the best practice in this case.

felipou:
> > Thinking about it as I write this, I believe there could be some changes to best guarantee the role of a server when using the SentinelClient. Perhaps we should make the Sentinel just return the address as you suggested, and the SentinelClient can ask for the role after connecting to a server (and before returning the connection).
> 
> IMO the Sentinel struct should expose a behavior that matches the Redis best practices. We shouldn't split the correct behavior between multiple structs. So, I think the code is fine as-is.

It's just that currently there is this problem where the role of the server can change after we check it, and before the connection is created. So I'm thinking we should actually return the connection that we used to check for the role (I've thought about that before, but not because of this, it just seemed that we were "wasting" that connection). If I recall correctly, that's more in line with the recommended practices, because in this case if the role changes, the connections are dropped.

But changing the Sentinel to return connections makes the SentinelClient almost redundant? Maybe not, but I'm not sure yet.

nihohit (approved):
LGTM, @jaymell

nihohit (approved):
LGTM, @jaymell

jaymell:
Thanks for reviewing @nihohit. I'll give this a review soon. Please rebase @felipou when able :pray:

wojciechbator:
Hey, great job. When can we expect it to enter?

felipou:
Sorry for taking so long, I expect to be able to work on this until the end of the week!

wojciechbator:
@felipou good news, fingers crossed!

redis/src/sentinel.rs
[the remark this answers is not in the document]
jaymell: Yeah, we already use `rand` elsewhere, we can just make it non-optional. 

For what it's worth, I think it's also considering making the sentinel functionality itself a feature. It would be easy to do, since the code is so self-contained.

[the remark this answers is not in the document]
wojciechbator: @jaymell any workaround to connect just to master sentinel node using your library? If I just specify protocol as `redis-sentinel` in a connection string, the thing panics. Would be glad for some solution ad-hoc until this is merged.

felipou:
Sorry for delay, rebase done!

jaymell:
@felipou Thanks for the PR! I'm largely going to defer to @nihohit's excellent review here, but a couple concerns:
1. I think we should go ahead and put the sentinel functionality behind a feature flag, and implement the proper random selection logic as discussed in prior comments. It's consistent w/ how we've handled other functionality like clustered support, so I think it's a good approach here as well.
2. Can we use `aio::MultiplexedConnection` and avoid the use of `aio::Connection`? Unfortunately the latter is a bit of a mess that will hopefully be deprecated/obsolete soon, but as it stands has serious risks (e.g., #325) and should be avoided when possible.

felipou:
>     1. I think we should go ahead and put the sentinel functionality behind a feature flag, and implement the proper random selection logic as discussed in prior comments. It's consistent w/ how we've handled other functionality like clustered support, so I think it's a good approach here as well.

Ok, will do!

>     2. Can we use `aio::MultiplexedConnection` and avoid the use of `aio::Connection`? Unfortunately the latter is a bit of a mess that will hopefully be deprecated/obsolete soon, but as it stands has serious risks (e.g., [Corrupt connections #325](https://github.com/redis-rs/redis-rs/issues/325)) and should be avoided when possible.

Ok, good to know, I'll replace it!

felipou:
All done @jaymell! Replace the async connection with multiplexed one, and added the feature flag (along with the proper rand method).

felipou:
I noticed there was a conflict with Cargo.toml, so I did another rebase.

felipou:
The tests failed, and the error made me notice that I had to handle a special value in the random replica index function, I've just pushed a commit changing that.

But I still don't understand why the tests failed, since that specific situation wasn't expected to happen in the tests, there should always be some valid replicas in the tests.

redis/src/sentinel.rs
nihohit: @felipou please expose `TlsMode` instead of copying it

redis/src/sentinel.rs
```diff
@@ -0,0 +1,768 @@
...
+        && replica_info["port"].parse::<u16>().is_ok()
+}
+
+/// Generates a random value in the 0..max range. If max is zero, the returned value is
+/// also zero.
+fn random_replica_index(max: usize) -> usize {
+    if max == 0 {
+        // This is just to avoid panicking in the gen_range function, this value does
+        // not have any special meaning, and is not expected to be used.
+        0
```
nihohit, line 218: use Option in order to better convey these semantics.

nihohit (approved):

nihohit:
> But I still don't understand why the tests failed, since that specific situation wasn't expected to happen in the tests, there should always be some valid replicas in the tests.

My guess is that it's a timing issue. You're waiting a 100ms assuming that's sufficient for the servers to start and connect, but that's an assumption which might not hold for a weak, slow CI machine. You probably need to actively check that the primary has found the replicas and finished syncing, something like the [check here](https://github.com/redis-rs/redis-rs/blob/29b470247c1c14ca8d9be37e7d2997902b7688d0/redis/tests/support/cluster.rs#L285). You can read INFO replication on the primary and see how many replicas it has.

felipou:
> > But I still don't understand why the tests failed, since that specific situation wasn't expected to happen in the tests, there should always be some valid replicas in the tests.
> 
> My guess is that it's a timing issue. You're waiting a 100ms assuming that's sufficient for the servers to start and connect, but that's an assumption which might not hold for a weak, slow CI machine. You probably need to actively check that the primary has found the replicas and finished syncing, something like the [check here](https://github.com/redis-rs/redis-rs/blob/29b470247c1c14ca8d9be37e7d2997902b7688d0/redis/tests/support/cluster.rs#L285). You can read INFO replication on the primary and see how many replicas it has.

I actually did something similar here: https://github.com/felipou/redis-rs/blob/add-sentinel/redis/tests/support/sentinel.rs#L280 (the `TestSentinelContext` type is actually a copied and adapted version of the `TestClusterContext` type). That 100ms wait in the creation of the sentinel cluster is so that this check doesn't fail. You're right that I should probably use something similar there instead of the 100ms wait (maybe even use the same method, I'll try to do that now), but in any case it should always fail in the `wait_for_cluster_up` method before anything else (and now that I think about, I'm not totally sure this wasn't the case, I'll try to check the failed tests again).

redis/src/sentinel.rs
felipou: What about NonZeroUsize? After your comment I thought about it, just pushed a commit with it. Not sure if there is any downside to using it though.

felipou:
@nihohit I removed the duplicated TlsMode, so it isn't letting me reply to your comment above. For now I just used the `redis::cluster::TlsMode` directly, since it is already exposed. Do you think we should move it? I actually think we should, but couldn't think of where it would fit best, any suggestions?

jaymell:
> Do you think we should move it? I actually think we should, but couldn't think of where it would fit best, any suggestions?

I would just put it under the `connection` module and add it to the `pub use` in lib.rs.

felipou:
> > Do you think we should move it? I actually think we should, but couldn't think of where it would fit best, any suggestions?
> 
> I would just put it under the `connection` module and add it to the `pub use` in lib.rs.

Ok, done!

felipou:
The tests were failing probably because of the part where we wait for the cluster to be ready, so I've re-organized that part a bit, now we shouldn't have any more problems.

felipou:
I've reorganized the "wait" functions of sentinel tests setup to give more information while waiting for the cluster to be up, while also reusing some bits of code in both functions.

felipou:
My changes helped me see the problem: it seems there was a bug in version 7.0.0, the "SENTINEL SLAVES" command was missing: https://github.com/redis/redis/blob/7.0/00-RELEASENOTES#L412

Is it possible to update the Redis 7 version to at least 7.0.1?

I didn't understand the error with redis 6 and Rust 1.60.0, is this something that is happening only on my branch?

jaymell:
>  I didn't understand the error with redis 6 and Rust 1.60.0, is this something that is happening only on my branch?

The issue with 1.60.0 just cropped up in last day or two, some transitive dependency on Rustix that we can't even explain is causing builds to fail, so unrelated to your changes. So you ignore those for now.

I'll try to push an update to the redis version used in CI shortly.

jaymell:
@felipou Want to rebase and we'll kick off tests again?

felipou:
Ok, done!

jaymell (approved):

jaymell:
Merging. Thanks again for the great contribution!

felipou:
It was my pleasure! Thanks a lot for the reviews and suggestions @nihohit and @jaymell, and also for maintaining this project! This sure feels like high quality open-source!

PR #838, opened 2023-05-09
Improve error logging.

nihohit:
This change adds the error's kind to the string representation of the error. So the error that was
`An error was signalled by the server: 10279 127.0.0.1:39192` 
will now be
`An error was signalled by the server - Moved: 10279 127.0.0.1:39192` 
which makes the error much clearer.


jaymell:
Good call -- thanks!

jaymell (approved):

PR #839, opened 2023-05-09
Add support for fallible high-level transactions

threema-danilo:
Add a new `transaction_opt` function. This function allows running a high-level transaction _without_ automatic retry. Instead, `None` is returned if the transaction fails due to a conflict of a WATCHed key.

Additionally, transaction tests were improved and extended (previously there was no test that ensured that `WATCH` and automatic retry of `transaction` works as intended).

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -1287,6 +1287,7 @@ impl Msg {
 ///
 /// ```rust,no_run
 /// use redis::Commands;
+///
 /// # fn do_something() -> redis::RedisResult<()> {
 /// # let client = redis::Client::open("redis://127.0.0.1/").unwrap();
 /// # let mut con = client.get_connection().unwrap();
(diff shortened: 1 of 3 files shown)


threema-danilo:
Note: Another implementation approach would be to modify `transaction` and to add a `retry` parameter. This parameter could then be `Retry::Never`, `Retry::Times(usize)` or `Retry::Infinitely` (or something like that).

However, this is a breaking change, and the parameter would always be required, so we thought it might be better to add a second function instead.

jaymell:
Hi, thanks for the PR, and apologies for the delay reviewing. I wonder, why not use your second option on the new method? So, allow `transaction_opt` to take the retry parameter you suggest?

threema-danilo:
> Hi, thanks for the PR, and apologies for the delay reviewing. I wonder, why not use your second option on the new method? So, allow `transaction_opt` to take the retry parameter you suggest?

Hmm, because it would be a bit weird that `transaction_opt` would behave exactly like `transaction` if `Retry::Infinitely` were passed in (just with a different return type).

I think modifying `transaction` with such a parameter would be the cleanest option, but it's a breaking change. What do you think about it?

jaymell:
Definitely no breaking change needed here. I don't think it's strange to have the method allow an argument that would _potentially_ behave the same as the other method.

PR #842, opened 2023-05-18
Fix AllMasters routing in sync cluster

nihohit:
This fix ensures that commands that should only be routed to master won't be routed to replicas.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -446,17 +446,10 @@ where
         let mut results = HashMap::new();
 
         // TODO: reconnect and shit
-        for slot in slots.values() {
-            let addr = slot.slot_addr(&SlotAddr::Master).to_string();
+        for addr in slots.all_unique_addresses(only_primaries) {
+            let addr = addr.to_string();
             if let Some(connection) = connections.get_mut(&addr) {
                 results.insert(addr, func(connection)?);
-            }
-
-            if !only_primaries {
-                let addr = slot.slot_addr(&SlotAddr::Replica).to_string();
-                if let Some(connection) = connections.get_mut(&addr) {
-                    results.insert(addr, func(connection)?);
-                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


redis/src/cluster.rs
```diff
@@ -448,17 +448,28 @@ where
         Ok(result)
     }
 
-    fn execute_on_all_nodes<T, F>(&self, mut func: F) -> RedisResult<T>
+    fn execute_on_all_nodes<T, F>(&self, mut func: F, only_primaries: bool) -> RedisResult<T>
```
barshaul, line 451: Not related to this PR, but I really don't think we should keep this function:
If we execute a command to all nodes/primaries in this function, we're not handling CME errors (e.g. MOVED, ASK..) nor refreshing topology.
I think we should have a wrapper function that decide which nodes to send the command to, and an inner function with the logic of executing the request.
e.g. 
```
fn request () {
  nodes = get_nodes_to_be_executed()
  for node in nodes:
     node.execute_request()
}

fn execute_request(...) {
        loop {
            // Get target address and response.
            let (addr, rv) = {
   ...
                (addr, func(conn))
            };
            match rv {
                Ok(rv) => return Ok(rv),
                Err(err) => {
       ...
                    }
}
}

barshaul (approved):

redis/src/cluster.rs
nihohit: You're talking about [this logic](https://github.com/redis-rs/redis-rs/blob/5c81077c63aa30afc523e7fcb1842fa900ae364a/redis/src/cluster.rs#L488)?
If so, I think you're right. We should open an issue (here and on redis-rs?) to fix this once we get to the sync implementation.

nihohit: AFAIS we don't have this issue on the async implementation

nihohit:
was contained in https://github.com/redis-rs/redis-rs/pull/843

PR #843, opened 2023-05-18
Support fan-out commands in cluster-async

nihohit:
Commands that should be routed to all nodes or all masters will now be routed correctly.

This Pr includes:
1. https://github.com/redis-rs/redis-rs/pull/842, in order to maintain consistent behavior between cluster clients.
2. Moved the `Route` struct into `RoutingInfo` as the `SpecificNode` variant, instead of the `ReplicaSlot`/`MasterSlot` variants. This was done because the current `RoutingInfo`->`Route` mapping functions lost information from `RoutingInfo` in variants that don't map to a specific route.
3. Instead of returning a `(String, RedisResult<Response>)`, cluster_async operations now return `(OperationTarget, RedisResult<Response>)`, where operation target disambiguates single node operations from fan-out operations. This is required, since fan-out operations can't return a single address, and so error handling has to know how to handle this.
4. Moved the routing information from `RequestInfo` into the `CmdArg` variants. This is done because the two variants require different routing information - pipelines can only be routed to single nodes, while Cmd can also be fanned-out.
5. Moved the structs required for getting connections into a `ConnectionsAndMetadata` struct, which is wrapped in an `Arc`. This allows us to moved the objects between functions without using `self`, which allows us to create futures that can run in parallel - in order to fan out concurrently, instead of serially like in the sync cluster. The `Arc` also ensures that clones are cheap, and we only clone/create expensive structs at the last possible moment.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -138,16 +138,14 @@ fn route_pipeline(pipeline: &crate::Pipeline) -> Option<Route> {
         }
     }
 
-    let mut iter = pipeline.cmd_iter();
-    let slot = iter.next().map(route_for_command)?;
-    for cmd in iter {
-        if slot != route_for_command(cmd) {
-            return None;
-        }
-    }
-    slot
+    // Find first specific slot and send to it. There's no need to check If later commands
+    // should be routed to a different slot, since the server will return an error indicating this.
+    pipeline
+        .cmd_iter()
+        .map(route_for_command)
+        .find(|route| route.is_some())
+        .flatten()
 }
-
 enum Response {
     Single(Value),
     Multiple(Vec<Value>),
@@ -1133,4 +1131,43 @@ where
         .expect("Connections is empty")
         .clone();
     (addr, conn)
+}
+
+#[cfg(test)]
+mod pipeline_routing_tests {
+    use super::route_pipeline;
+    use crate::{
+        cluster_routing::{Route, SlotAddr},
+        cmd,
+    };
+
+    #[test]
+    fn test_first_route_is_found() {
+        let mut pipeline = crate::Pipeline::new();
+
+        pipeline
+            .add_command(cmd("FLUSHALL")) // route to all masters
+            .get("foo") // route to slot 12182
+            .add_command(cmd("EVAL")); // route randomly
+
+        assert_eq!(
+            route_pipeline(&pipeline),
+            Some(Route::new(12182, SlotAddr::Replica))
+        );
+    }
+
+    #[test]
+    fn test_ignore_conflicting_slots() {
+        let mut pipeline = crate::Pipeline::new();
+
+        pipeline
+            .add_command(cmd("FLUSHALL")) // route to all masters
+            .set("baz", "bar") // route to slot 4813
+            .get("foo"); // route to slot 12182
+
+        assert_eq!(
+            route_pipeline(&pipeline),
+            Some(Route::new(4813, SlotAddr::Master))
+        );
+    }
 }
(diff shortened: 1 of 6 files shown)


jaymell:
Hey, this is great! I've got a pending (unrelated) PR that touches a lot of the same code, which I'd like to get in first, purely for my own benefit :stuck_out_tongue:, if you don't mind. I will try to have it raised in the next few days for your review.

nihohit:
Such be life 🤷 . I'd appreciate it if you could browse through the code and point out design choices that will clash with your change, or if you can integrate parts that are relevant to your PR, in order to reduce merge complexity further down the line

jaymell:
I think we'll be ok. Our concerns are mostly orthogonal, though there is some overlap in code touched. For instance, I've ripped out the cluster async `get_connection` method entirely, but overall I think we'll be able to resolve things without much trouble. Anyway, give me a couple more days, and I'll get a proper PR raised. 

Overall I think this looks pretty good. I would appreciate it if you could add a bit more description of changes made/motivations for them in your PR description.

nihohit:
@jaymell added more details to the description.

jaymell:
Still not quite ready, but raised draft of #844 for reference.

nihohit:
@jaymell rebased

redis/src/cluster_async/mod.rs
```diff
@@ -94,70 +94,96 @@ where
 type ConnectionFuture<C> = future::Shared<BoxFuture<'static, C>>;
 type ConnectionMap<C> = HashMap<String, ConnectionFuture<C>>;
 
-struct ClusterConnInner<C> {
-    connections: ConnectionMap<C>,
+struct ConnectionsAndMetadata<C> {
```
jaymell, line 97: With apologies for the inconvenience here, I feel like adding this struct and related connection-handling refactorings are in themselves a large change-set. Is it possible to bring those in first in a separate PR? :pray:

nihohit: https://github.com/redis-rs/redis-rs/pull/864

nihohit: also https://github.com/redis-rs/redis-rs/pull/842

redis/src/cluster_async/mod.rs
```diff
@@ -94,70 +94,96 @@ where
...
-                let route = iter.next().map(route_for_command)?;
-                for cmd in iter {
-                    if route != route_for_command(cmd) {
-                        return None;
-                    }
-                }
-                route
-            }
-        }
+#[cfg(test)]
```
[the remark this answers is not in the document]
nihohit: done

redis/src/cluster_async/mod.rs
```diff
@@ -242,17 +234,49 @@ enum Next<I, C> {
...
+    slots: &SlotMap,
+) -> Option<(String, Option<ConnectionFuture<C>>)>
+where
+    C: Connect + Clone,
+{
+    let addr = initial_route.and_then(|route| slots.slot_addr_for_route(route));
+    get_connection(addr, redirect, connections)
+}
+
+fn get_connection<C>(
```
jaymell, line 258: Minor, but can we either rename this or consolidate with the above function, which I believe is the only place it's used? We now have a bunch of functions with the some variation of the name `get_connection` -- this one is particularly irksome :laughing:

nihohit: removed entirely - it was made unnecessary by https://github.com/redis-rs/redis-rs/pull/877

jaymell (approved):

jaymell:
Thanks!!

PR #844, opened 2023-05-19
Refactor cluster error handling

jaymell:
This make some changes to how errors are handled in the redis cluster clients, most notably:
* Remove exclusions -- in practice, exclusions create a lot of unnecessary connection rebuilding/slot refreshes when a failed command results in a `MOVED` error.
* Attempt to refresh failed connections when IO errors occur instead of adding exclusions and ultimately refreshing slots due to the above-described scenario
* Only retry given a subset of error types, otherwise return error immediately.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -279,9 +279,6 @@ where
 
                 if let Ok(mut conn) = self.connect(addr) {
                     if conn.check_connection() {
-                        conn.set_read_timeout(*self.read_timeout.borrow()).unwrap();
-                        conn.set_write_timeout(*self.write_timeout.borrow())
-                            .unwrap();
                         return Some((addr.to_string(), conn));
                     }
                 }
@@ -357,6 +354,8 @@ where
             // If READONLY is sent to primary nodes, it will have no effect
             cmd("READONLY").query(&mut conn)?;
         }
+        conn.set_read_timeout(*self.read_timeout.borrow())?;
+        conn.set_write_timeout(*self.write_timeout.borrow())?;
         Ok(conn)
     }
 
@@ -523,11 +522,7 @@ where
                         ErrorKind::IoError => {
                             if *self.auto_reconnect.borrow() {
                                 if let Ok(mut conn) = self.connect(&addr) {
-                                    // FIXME: Duplicated from `refresh_slots`
                                     if conn.check_connection() {
-                                        conn.set_read_timeout(*self.read_timeout.borrow()).unwrap();
-                                        conn.set_write_timeout(*self.write_timeout.borrow())
-                                            .unwrap();
                                         self.connections.borrow_mut().insert(addr, conn);
                                     }
                                 }
@@ -716,7 +711,7 @@ pub enum TlsMode {
     Insecure,
 }
 
-// FIXME -- This function can panic and should probably
+// TODO: This function can panic and should probably
 // return an Option instead:
 fn get_random_connection<C: ConnectionLike + Connect + Sized>(
     connections: &mut HashMap<String, C>,
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -50,7 +50,7 @@ use crate::connection::{
     connect, Connection, ConnectionAddr, ConnectionInfo, ConnectionLike, RedisConnectionInfo,
 };
 use crate::parser::parse_redis_value;
-use crate::types::{ErrorKind, HashMap, HashSet, RedisError, RedisResult, Value};
+use crate::types::{ErrorKind, HashMap, RedisError, RedisResult, Value};
 use crate::IntoConnectionInfo;
 use crate::{
     cluster_client::ClusterParams,
@@ -329,7 +329,6 @@ where
                 }
 
                 new_slots = Some(SlotMap::from_slots(&slots_data, self.read_from_replicas));
-
                 break;
             }
         }
@@ -375,7 +374,7 @@ where
         } else {
             // try a random node next.  This is safe if slots are involved
             // as a wrong node would reject the request.
-            Ok(get_random_connection(connections, None))
+            Ok(get_random_connection(connections))
         }
     }
 
@@ -472,7 +471,6 @@ where
         };
 
         let mut retries = self.retries;
-        let mut excludes = HashSet::new();
         let mut redirected = None::<String>;
         let mut is_asking = false;
         loop {
@@ -489,8 +487,8 @@ where
                         is_asking = false;
                     }
                     (addr.to_string(), conn)
-                } else if !excludes.is_empty() || route.is_none() {
-                    get_random_connection(&mut connections, Some(&excludes))
+                } else if route.is_none() {
+                    get_random_connection(&mut connections)
                 } else {
                     self.get_connection(&mut connections, route.as_ref().unwrap())?
                 };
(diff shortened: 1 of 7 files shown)


redis/tests/support/mock_cluster.rs
```diff
@@ -121,6 +121,34 @@ pub fn respond_startup(name: &str, cmd: &[u8]) -> Result<(), RedisResult<Value>>
     }
 }
 
+pub fn respond_startup_two_nodes(name: &str, cmd: &[u8]) -> Result<(), RedisResult<Value>> {
```
nihohit, line 124: instead of this, maybe just pull https://github.com/redis-rs/redis-rs/pull/843/files#diff-ae2ede8b12dd85b78d0a380121b8b51caedf3f46452a564b18454dc8b5fe354c ? That way we won't have logic conflicts later.

redis/src/cluster_routing.rs
```diff
@@ -210,7 +210,39 @@ impl<'a> IntoIterator for &'a SlotAddrs {
     }
 }
 
-pub(crate) type SlotMap = BTreeMap<u16, SlotAddrs>;
+#[derive(Debug, Default)]
+pub(crate) struct SlotMap(BTreeMap<u16, SlotAddrs>);
```
nihohit, line 214: IMO this is a self contained change, that can be moved to a separate PR. Merging this early will make the review process for the larger PR faster, since it will reduce the amount of code to review.

redis/src/types.rs
```diff
@@ -581,6 +581,32 @@ impl RedisError {
         };
         Self { repr }
     }
+
+    #[cfg(feature = "cluster")] // Used to avoid "unused method" warning
+    pub(crate) fn is_retryable(&self) -> bool {
```
nihohit, line 586: IMO this should be pub, not pub(crate) - this is usable to anyone who wants to implement a client over redis-rs. 
This is also a self contained change, that can be moved to a separate PR. Merging this early will make the review process for the larger PR faster, since it will reduce the amount of code to review.

nihohit (commented):
this was a high-level review, I didn't go into each line.
I think it's a great change, I'll be happy to see it land soon.

nihohit (commented):
this was a high-level review, I didn't go into each line.
I think it's a great change, I'll be happy to see it land soon.

redis/src/types.rs
jaymell: I think I'd like to leave this in the PR since integrating it into both clients will require a lot of the larger refactoring that's in here. I'll admit this is PR is a bit more unwieldy than I'd hoped and didn't quite end up where I intended it to :sweat_smile:. 

I'm not opposed to making `is_retryable` public, but for the moment I'm not sure whether all of the variants are handled properly so thought I'd avoid exposing it for now.

redis/tests/test_cluster.rs
```diff
@@ -433,3 +439,71 @@ fn test_cluster_replica_read() {
...
+                completed.fetch_add(1, Ordering::SeqCst);
+                Err(parse_redis_value(b"-ERR mock\r\n"))
+            }
+        },
+    );
+
+    let value = cmd("GET").arg("test").query::<Option<i32>>(&mut connection);
+
+    match value {
+        Ok(_) => panic!("This is not ok, man"),
```
nihohit, line 502: same

redis/src/cluster_async/mod.rs
```diff
@@ -282,38 +287,42 @@ where
...
-                        // Refresh slots and request again.
-                        request.info.excludes.clear();
-                        return Next::Err {
-                            request: this.request.take().unwrap(),
-                            error: err,
-                        }
-                        .into();
-                    } else if error_code == "TRYAGAIN" || error_code == "CLUSTERDOWN" {
+                match err.kind() {
+                    // TODO: Proper Ask implementation
```
nihohit, line 291: please add what's missing from the current implementation

redis/src/cluster_async/mod.rs
```diff
@@ -496,47 +513,26 @@ where
...
+            let route = info.route.as_ref().unwrap();
+            if let Some(addr) = self.slots.slot_addr_for_route(route) {
+                let addr = addr.to_string();
+                if let Some(conn) = self.connections.get(&addr) {
+                    (addr, conn.clone())
+                } else {
+                    get_random_connection(&self.connections)
+                }
+            } else {
+                get_random_connection(&self.connections)
```
nihohit, line 534: minor:
this 3-fold do {} else {get_random_connection}

makes me think that the structure of lines 522-534 can be terser. Maybe something like this:

```rs
 let (addr, conn) = info.route
   .and_then(|route| self.slots.slot_addr_for_route(route)
   .and_then(|addr| {
      let addr = addr.to_string();
      self.connections.get(&addr).map(|conn| (addr, conn))
   }).unwrap_or_else(||get_random_connection(&self.connections));
```

redis/src/types.rs
```diff
@@ -587,6 +587,32 @@ impl RedisError {
...
+        match self.kind() {
+            ErrorKind::ExecAbortError => true,
```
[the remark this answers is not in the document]
jaymell: `ExecAbortError` definitely not.

`I'm not clear on `ExtensionError` -- it's a bit of a catch-all.

`Readonly` and `ClusterDown` -- I'm also not sure but thought they might potentially arise due to intermittent cluster flakiness?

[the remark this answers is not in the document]
nihohit: > I'm not clear on `ExtensionError` -- it's a bit of a catch-all.

IMO without additional information, we can't know whether this is a transitory error or not. In such a situation, we should pass if to the user, and let them decide, given their knowledge of what modules they use and how the modules behave, whether to retry or not.

> Readonly and ClusterDown -- I'm also not sure but thought they might potentially arise due to intermittent cluster flakiness?

fair enough.

redis/src/cluster_async/mod.rs
jaymell: Yep, looks great!

redis/src/cluster.rs
```diff
@@ -505,41 +503,41 @@ where
...
-
-                    excludes.insert(addr);
-
-                    let connections = self.connections.borrow();
-                    if excludes.len() >= connections.len() {
-                        return Err(err);
+                        ErrorKind::IoError => {
+                            if *self.auto_reconnect.borrow() {
+                                if let Ok(mut conn) = self.connect(&addr) {
+                                    // FIXME: Duplicated from `refresh_slots`
```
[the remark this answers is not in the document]
jaymell: Moved into `ClusterConnection::connect`. I don't see a reason not to consolidate there.

redis/src/cluster_async/mod.rs
jaymell: I'm working on a followup PR that should add the implementation shortly, so I think this is ok for now.

redis/src/cluster_async/mod.rs
```diff
@@ -398,14 +407,32 @@ where
...
+    fn refresh_connections(
+        &mut self,
+        addrs: Vec<String>,
+    ) -> impl Future<Output = ConnectionMap<C>> {
+        let mut connections = mem::take(&mut self.connections);
+        let params = self.cluster_params.clone();
+        async move {
+            for addr in addrs {
+                let conn =
+                    Self::get_or_create_conn(&addr, connections.remove(&addr), params.clone())
```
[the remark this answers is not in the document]
jaymell: I went ahead and parallelized it.

redis/src/types.rs
```diff
@@ -587,6 +587,32 @@ impl RedisError {
         };
         Self { repr }
     }
+
+    #[cfg(feature = "cluster")] // Used to avoid "unused method" warning
+    pub(crate) fn is_retryable(&self) -> bool {
```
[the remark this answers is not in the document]
jaymell: I think it's a good idea but would need to consider further. I think there's probably still larger refactoring needed around retries generally (i.e., in non-clustered use cases as well), so would probably defer this for now.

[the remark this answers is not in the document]
nihohit: fair enough. Would you mind adding a TODO, so we won't forget?

redis/src/cluster_routing.rs
```diff
@@ -237,10 +237,6 @@ impl SlotMap {
     pub fn values(&self) -> std::collections::btree_map::Values<u16, SlotAddrs> {
         self.0.values()
     }
-
-    pub fn len(&self) -> usize {
```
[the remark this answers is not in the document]
jaymell: Added back in -- was temporarily unused during refactoring :smirk:

nihohit (approved):

nihohit:
LGTM

PR #845, opened 2023-05-22
Newtype SlotMap

jaymell:
Make `SlotMap` a struct for better abstraction
and code de-duplication

### DIFF
--- a/redis/src/cluster_routing.rs
+++ b/redis/src/cluster_routing.rs
@@ -210,7 +210,38 @@ impl<'a> IntoIterator for &'a SlotAddrs {
     }
 }
 
-pub(crate) type SlotMap = BTreeMap<u16, SlotAddrs>;
+#[derive(Debug, Default)]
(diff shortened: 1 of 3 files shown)


redis/src/cluster_routing.rs
```diff
@@ -210,7 +210,39 @@ impl<'a> IntoIterator for &'a SlotAddrs {
...
+        Self(
+            slots
+                .iter()
+                .map(|slot| (slot.end(), SlotAddrs::from_slot(slot, read_from_replicas)))
+                .collect(),
+        )
+    }
+
+    pub fn slot_addr_for_route(&self, route: &Route) -> Option<&str> {
+        if let Some((_, node_addrs)) = self.0.range(route.slot()..).next() {
```
nihohit, line 231: minor - can remove the if and just map
```rs
        self.0
            .range(route.slot()..)
            .next()
            .map(|(_, node_addrs)| node_addrs.slot_addr(route.slot_addr()))
```

nihohit (approved):
LGTM

PR #849, opened 2023-05-23
fix: unnecessarily mutable variable

kamulos:
newer Rust versions warn of a variable, that is unnecessarily mut

Just removed the `mut`


jaymell (approved):

jaymell:
Thanks!

PR #857, opened 2023-06-05
Improved redirection error handling

jaymell:
This PR adds proper `ASK` redirection for the async cluster implementation, so rather than simply rebuilding slots, which may not be an appropriate response to `ASK` anyway, it issues the relevant `ASKING` command to the redirected node and issues the query.

A `Redirect` enum has been added for better handling of both redirect scenarios.

The async cluster `MOVED` implementation has also been tweaked. Instead of immediately refreshing slots, the redirected node will be queried, and slots will be refreshed after the redirected request has been made. The same approach has not been taken on the sync implementation for now because doing so would require greater refactoring than is appropriate here.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -44,7 +44,7 @@ use std::time::Duration;
 use rand::{seq::IteratorRandom, thread_rng, Rng};
 
 use crate::cluster_pipeline::UNROUTABLE_ERROR;
-use crate::cluster_routing::{Routable, RoutingInfo, Slot, SLOT_SIZE};
+use crate::cluster_routing::{Redirect, Routable, RoutingInfo, Slot, SLOT_SIZE};
 use crate::cmd::{cmd, Cmd};
 use crate::connection::{
     connect, Connection, ConnectionAddr, ConnectionInfo, ConnectionLike, RedisConnectionInfo,
@@ -470,20 +470,23 @@ where
         };
 
         let mut retries = self.retries;
-        let mut redirected = None::<String>;
-        let mut is_asking = false;
+        let mut redirected = None::<Redirect>;
+
         loop {
             // Get target address and response.
             let (addr, rv) = {
                 let mut connections = self.connections.borrow_mut();
-                let (addr, conn) = if let Some(addr) = redirected.take() {
+                let (addr, conn) = if let Some(redirected) = redirected.take() {
+                    let (addr, is_asking) = match redirected {
+                        Redirect::Moved(addr) => (addr, false),
+                        Redirect::Ask(addr) => (addr, true),
+                    };
                     let conn = self.get_connection_by_addr(&mut connections, &addr)?;
                     if is_asking {
                         // if we are in asking mode we want to feed a single
                         // ASKING command into the connection before what we
                         // actually want to execute.
                         conn.req_packed_command(&b"*1\r\n$6\r\nASKING\r\n"[..])?;
-                        is_asking = false;
                     }
                     (addr.to_string(), conn)
                 } else if route.is_none() {
@@ -504,15 +507,17 @@ where
 
                     match err.kind() {
                         ErrorKind::Ask => {
-                            redirected = err.redirect_node().map(|(node, _slot)| node.to_string());
-                            is_asking = true;
+                            redirected = err
+                                .redirect_node()
+                                .map(|(node, _slot)| Redirect::Ask(node.to_string()));
                         }
                         ErrorKind::Moved => {
                             // Refresh slots.
                             self.refresh_slots()?;
                             // Request again.
-                            redirected = err.redirect_node().map(|(node, _slot)| node.to_string());
-                            is_asking = false;
+                            redirected = err
+                                .redirect_node()
+                                .map(|(node, _slot)| Redirect::Moved(node.to_string()));
                         }
                         ErrorKind::TryAgain | ErrorKind::ClusterDown => {
                             // Sleep and retry.
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -37,7 +37,7 @@ use crate::{
     aio::{ConnectionLike, MultiplexedConnection},
     cluster::{get_connection_info, parse_slots, slot_cmd},
     cluster_client::ClusterParams,
-    cluster_routing::{Route, RoutingInfo, Slot, SlotAddr, SlotMap},
+    cluster_routing::{Redirect, Route, RoutingInfo, Slot, SlotAddr, SlotMap},
     Cmd, ConnectionInfo, ErrorKind, IntoConnectionInfo, RedisError, RedisFuture, RedisResult,
     Value,
 };
@@ -200,6 +200,7 @@ impl<C> fmt::Debug for ConnectionState<C> {
 struct RequestInfo<C> {
     cmd: CmdArg<C>,
     route: Option<Route>,
+    redirect: Option<Redirect>,
 }
 
 pin_project! {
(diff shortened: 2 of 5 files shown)


redis/tests/test_cluster.rs
```diff
@@ -507,3 +507,41 @@ fn test_cluster_non_retryable_error_should_not_retry() {
...
+                // Error twice with io-error, ensure connection is reestablished w/out calling
+                // other node (i.e., not doing a full slot rebuild)
+                let count = completed.fetch_add(1, Ordering::SeqCst);
+                match port {
+                    6379 => match count {
+                        0 => Err(parse_redis_value(b"-ASK 14000 node:6380\r\n")),
+                        _ => panic!("Node should not be called now"),
+                    },
+                    6380 => match count {
+                        1 => Err(Ok(Value::Okay)),
```
nihohit, line 534: minor: please check that this is an ASK command.

redis/tests/test_cluster_async.rs
```diff
@@ -380,52 +380,60 @@ fn test_async_cluster_tryagain_exhaust_retries() {
...
-                    Value::Int(1),
+            let i = i.fetch_add(1, atomic::Ordering::SeqCst);
+
+            match i {
+                // Respond that the key exists elswehere (the slot, 123, is unused in the
+                // implementation)
+                0 => Err(parse_redis_value(
+                    format!("-MOVED 14000 {name}:6380\r\n").as_bytes(),
+                )),
+                1 => {
```
nihohit, line 410: this might happen concurrently with the node refresh process, no? We're not promised that this will happen before the `CLUSTER SLOTS` command is sent.
I worry that the test will become flaky.

redis/tests/test_cluster_async.rs
```diff
@@ -380,52 +380,60 @@ fn test_async_cluster_tryagain_exhaust_retries() {
...
+                    // Check that the correct node receives the request after rebuilding
```
nihohit, line 411: this happens before rebuilding, according to the atomic, no?

redis/tests/test_cluster_async.rs
```diff
@@ -380,52 +380,60 @@ fn test_async_cluster_tryagain_exhaust_retries() {
...
+                    assert_eq!(port, 6380);
+                    Err(Ok(Value::Data(b"123".to_vec())))
+                }
+                // Respond with the new masters
+                2 => Err(Ok(Value::Bulk(vec![
```
nihohit, line 416: please verify that this is `CLUSTER SLOTS`.

redis/tests/test_cluster_async.rs
```diff
@@ -604,3 +613,46 @@ fn test_async_cluster_non_retryable_error_should_not_retry() {
...
+                respond_startup_two_nodes(name, cmd)?;
+                // Error twice with io-error, ensure connection is reestablished w/out calling
+                // other node (i.e., not doing a full slot rebuild)
+                let count = completed.fetch_add(1, Ordering::SeqCst);
+                match port {
+                    6379 => match count {
+                        0 => Err(parse_redis_value(b"-ASK 14000 node:6380\r\n")),
+                        _ => panic!("Node should not be called now"),
+                    },
+                    6380 => match count {
```
nihohit, line 640: same comment as in sync

nihohit (approved):
The implementation is fine, a couple of comments on the tests

redis/tests/test_cluster_async.rs
jaymell: Of course you have me doubting myself now :sweat_smile: , but I think this is a safe assumption. The request is added to `in_flight_requests` (rather than `pending_requests` as before) [after redirection](https://github.com/redis-rs/redis-rs/pull/857/files#diff-5ddfd8b2718c21504d43c50b0490bc22de4189208d97f5604a2297124e529577R664), and the refresh should not happen until after `in_flight_requests` have been drained.

jaymell: Ok, clearly I'm wrong, and furthermore I realize I've altered the original intent of this test. Anyway, a useful problem to run into, because it's made clear to me that following the redirect for `MOVED` rather than immediately rebuilding is probably not a good idea given our current impl.... if a new node has been added to the cluster, the current implementation will retry x number of times on a random connection rather than immediately rebuild the slots, which is suboptimal, and probably the most likely scenario if a `MOVED` error is actually received. I think I need to strip everything out but the ask implementation here :grimacing:

nihohit: 😱 
If you want, I have a fix just for ASK rebased over https://github.com/redis-rs/redis-rs/pull/843, so if you prefer we can merge that and I'll open a PR.

redis/tests/test_cluster_async.rs
jaymell: I believe this comment is no longer relevant, but please confirm.

redis/tests/test_cluster_async.rs
jaymell: I've made another pass at this w/ the `MOVED` changes in place and tried to better validate expected behavior in new/existing tests.

redis/src/cluster_async/mod.rs
```diff
@@ -288,11 +289,20 @@ where
...
+                            .redirect_node()
+                            .map(|(node, _slot)| Redirect::Ask(node.to_string()));
+                        Next::Retry { request }.into()
+                    }
+                    ErrorKind::Moved => {
+                        let mut request = this.request.take().unwrap();
+                        request.info.redirect = err
+                            .redirect_node()
+                            .map(|(node, _slot)| Redirect::Moved(node.to_string()));
+                        Next::RefreshSlots { request }.into()
```
nihohit, line 304: not to fix in this PR: if I understand the code correctly, this result is queued into `in_flight_requests`. This means that if we decide to refresh slots, but there are other requests on there, the slot refresh will be delayed. Should we add some kind of mechanism to give slot refresh preference, and run that first?

redis/tests/test_cluster_async.rs
nihohit: you're right.

redis/src/cluster_async/mod.rs
```diff
@@ -520,23 +530,61 @@ where
...
-            .as_ref()
-            .and_then(|route| self.slots.slot_addr_for_route(route))
-            .and_then(|addr| {
-                self.connections
-                    .get(addr)
-                    .map(|conn| (addr.to_string(), conn.clone()))
-            })
-            .unwrap_or_else(|| get_random_connection(&self.connections));
+
+        let random_conn = get_random_connection(&self.connections);
```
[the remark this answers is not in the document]
jaymell: Agreed. The issue as it stands at the moment is the need to avoid the `self` reference in the async blocks.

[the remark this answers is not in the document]
nihohit: well, https://github.com/redis-rs/redis-rs/pull/857 will help us get rid for `self` in async blocks :)

redis/tests/test_cluster_async.rs
```diff
@@ -399,34 +401,95 @@ fn test_async_cluster_rebuild_with_extra_nodes() {
...
+        cmd("GET")
+            .arg("test")
+            .query_async::<_, Option<i32>>(&mut connection),
+    );
+
+    assert_eq!(value, Ok(Some(123)));
+}
+
+#[test]
+fn test_async_cluster_ask_error_with_new_node() {
```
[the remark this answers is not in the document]
nihohit: and maybe rename to `test_async_cluster_ask_error_when_new_node_is_added`?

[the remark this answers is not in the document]
jaymell: I changed the names but decided to keep both test variants. I think it's worthwhile to have an ASK test when a new node is/isn't involved, even at the risk of some redundancy.

redis/src/cluster_async/mod.rs
```diff
@@ -200,6 +200,7 @@ impl<C> fmt::Debug for ConnectionState<C> {
 struct RequestInfo<C> {
     cmd: CmdArg<C>,
     route: Option<Route>,
+    redirect: Option<Redirect>,
```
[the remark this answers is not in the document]
jaymell: I can't simply overwrite b/c the redirect requires an additional async call to potentially get a new connection. I have tried to simplify it a bit, however.

redis/src/cluster.rs
```diff
@@ -470,20 +470,23 @@ where
...
+                    let (addr, is_asking) = match redirected {
+                        Redirect::Moved(addr) => (addr, false),
+                        Redirect::Ask(addr) => (addr, true),
+                    };
                     let conn = self.get_connection_by_addr(&mut connections, &addr)?;
                     if is_asking {
                         // if we are in asking mode we want to feed a single
                         // ASKING command into the connection before what we
                         // actually want to execute.
                         conn.req_packed_command(&b"*1\r\n$6\r\nASKING\r\n"[..])?;
```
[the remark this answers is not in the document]
jaymell: Yes it seems reasonable but in a future PR.

redis/src/cluster_async/mod.rs
jaymell: Agreed. I don't think it's a huge concern at the moment but would be ideal if we can do so w/out further complicating the code.

PR #858, opened 2023-06-06
Add flexibility to configure TlsConnector with client key and cert fo…

sp-angel:
This PR adds a `TLSClientBuilder` that allows the creation of a TLS client with extra parameters _CA root certificate_, _client key_ and _client certificate_.

For this change I'm focusing *only* on *Rustls*. Consequently it requires enabling the features `tls-rustls` and `tokio-rustls-comp`, the latter to enable async.

Example:
```rust
    let mut tls_builder = TLSClientBuilder::new(server.to_string(), port);

    let cert_file = File::open(cert_file).expect("cannot open private cert file");
    let mut reader = BufReader::new(cert_file);
    tls_builder
        .client_certs(&mut reader)
        .expect("Unable to handle client certs");

    let key_file = File::open(key_file).expect("Cannot open private key file");
    let mut reader = BufReader::new(key_file);
    tls_builder
        .client_key(&mut reader)
        .expect("Unable to handle client key");

    let ca_file = File::open(ca_file).expect("Cannot open CA cert file");
    let mut reader = BufReader::new(ca_file);
    tls_builder
        .root_cert(&mut reader)
        .expect("Unable to load CA cert");
   
    let client = tls_builder.build().expect("Unable to build client");
    
    let mut con = client.get_connection()?;   // sync

    // let mut con = client.get_async_connection().await?;  // from an 'async' function
```


esemeniuc:
would like to see this merged!

redis/src/connection.rs
```diff
@@ -73,11 +80,42 @@ pub enum ConnectionAddr {
...
+        match (self, other) {
+            (ConnectionAddr::Tcp(host1, port1), ConnectionAddr::Tcp(host2, port2)) => {
+                host1 == host2 && port1 == port2
+            }
+            (
+                ConnectionAddr::TcpTls {
+                    host: host1,
+                    port: port1,
+                    insecure: insecure1,
+                    ..
```
robjtede, line 102: i'd advise to add all fields explicitly for future additions to be caught

```diff
                      insecure: insecure1,
+                     tls_params: _,
```

redis/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "redis"
-version = "0.23.0"
+version = "0.24.0-alpha"
```
robjtede, line 3: they'll probably ask you to keep the version as it is for now

redis/src/tls.rs
```diff
@@ -0,0 +1,174 @@
...
+///     .expect("Unable to handle client key");
+/// let ca_file = File::open(ca_file).expect("Cannot open CA cert file");
+/// let mut reader = BufReader::new(ca_file);
+/// tls_builder
+///     .root_cert(&mut reader)
+///     .expect("Unable to load CA cert");
+///
+/// let client = tls_builder.build().expect("Unable to build client");
+/// ```
+///
```
robjtede, line 49:
```suggestion
```

redis/src/tls.rs
```diff
@@ -0,0 +1,174 @@
...
+                    insecure: false,
+                    tls_params,
+                },
+                redis: RedisConnectionInfo::default(),
+            },
+        })
+    }
+
+    /// Loads a client certificate in PEM format as a byte stream
+    /// In case of errors loading the certificate it returns an error
```
robjtede, line 105: doc nit:

```suggestion
    /// Loads a client certificate in PEM format as a byte stream.
    ///
    /// In case of errors loading the certificate it returns an error.
```

redis/src/tls.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    /// Loads a client key in PEM format as a byte stream
+    /// In case of errors loading the key it returns an error
+    pub fn client_key(&mut self, reader: &mut dyn BufRead) -> RedisResult<()> {
+        self.client_key = Some(load_key(reader)?);
+        Ok(())
+    }
+
+    /// Loads a CA root certificate in PEM format as a byte stream
+    /// In case of errors loading the certificate it returns an error
```
robjtede, line 121: doc nit:

```suggestion
    /// Loads a CA root certificate in PEM format as a byte stream.
    ///
    /// In case of errors loading the certificate it returns an error.
```

redis/src/tls.rs
```diff
@@ -0,0 +1,174 @@
...
+        match read_one(&mut reader)? {
+            Some(rustls_pemfile::Item::ECKey(key)) => {
+                return Ok(PrivateKey(key));
+            }
+            Some(rustls_pemfile::Item::RSAKey(key)) => {
+                return Ok(PrivateKey(key));
+            }
+            Some(rustls_pemfile::Item::PKCS8Key(key)) => {
+                return Ok(PrivateKey(key));
+            }
```
robjtede, line 163:
```suggestion
            Some(rustls_pemfile::Item::ECKey(key))
            | Some(rustls_pemfile::Item::RSAKey(key))
            | Some(rustls_pemfile::Item::PKCS8Key(key)) => {
                return Ok(PrivateKey(key));
            }
```

redis/src/tls.rs
```diff
@@ -0,0 +1,174 @@
...
+            None => {
+                break;
+            }
+            _ => {}
+        }
+    }
+    Err(RedisError::from(Error::new(
+        IOErrorKind::Other,
+        "invalid proto buf message type",
```
robjtede, line 172: error seems to references protobuf

robjtede (changes requested):

redis/src/connection.rs
```diff
@@ -689,9 +734,23 @@ pub(crate) fn create_rustls_config(insecure: bool) -> RedisResult<rustls::Client
...
         .with_safe_default_kx_groups()
-        .with_protocol_versions(rustls::ALL_VERSIONS)?
-        .with_root_certificates(root_store)
-        .with_no_client_auth();
+        .with_protocol_versions(rustls::ALL_VERSIONS)?;
+
+    let config = if let Some(tls_params) = tls_params {
+        config
+            .with_root_certificates(tls_params.root_cert)
+            .with_single_cert(tls_params.client_certs, tls_params.client_key)
```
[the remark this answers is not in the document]
sp-angel: going for `client_cert` - at the moment use cases contemplate a single cert only

redis/Cargo.toml
sp-angel: reverted

redis/src/aio/mod.rs
```diff
@@ -28,12 +34,12 @@ pub(crate) trait RedisRuntime: AsyncStream + Send + Sync + Sized + 'static {
     /// Performs a TCP connection
     async fn connect_tcp(socket_addr: SocketAddr) -> RedisResult<Self>;
 
-    // Performs a TCP TLS connection
```
nihohit, line 31: why removed?

redis/src/aio/tokio.rs
```diff
@@ -20,12 +20,18 @@ use native_tls::TlsConnector;
 use crate::connection::create_rustls_config;
 #[cfg(feature = "tls-rustls")]
 use std::{convert::TryInto, sync::Arc};
-#[cfg(feature = "tls-rustls")]
+#[cfg(all(feature = "tls-rustls"))]
```
nihohit, line 23: why `all`, if there's only a single entry?

redis/src/connection.rs
```diff
@@ -689,9 +738,23 @@ pub(crate) fn create_rustls_config(insecure: bool) -> RedisResult<rustls::Client
...
+        .with_protocol_versions(rustls::ALL_VERSIONS)?;
+
+    let config = if let Some(tls_params) = tls_params {
+        config
+            .with_root_certificates(tls_params.root_cert)
+            .with_client_auth_cert(tls_params.client_cert, tls_params.client_key)
+            .map_err(|_| {
+                RedisError::from((
+                    ErrorKind::InvalidClientConfig,
+                    "Unable to build client with TLS parameters provided",
```
nihohit, line 750: why not log the error, or problematic parameters?

redis/src/tls.rs
```diff
@@ -0,0 +1,207 @@
...
+    host: String,
+    port: u16,
+    client_key: Option<PrivateKey>,
+    client_cert: Vec<Certificate>,
+    root_cert: RootCertStore,
+}
+
+/// The builder takes parameters all parameters necessary to create a TLS connection.
+///
+/// It must be created with a _server_ and a _port_ and should be provided with
```
nihohit, line 21: if the builder is incomplete without the binary stream, its creation structure shouldn't allow `build` to be called without the stream, so the stream should be passed in constructor, or in `build`. Maybe pass it in an enum denoting the different values, instead of using the builder pattern?

nihohit (commented):
I don't understand TLS well, so I can't comment on whether what you're doing is correct. These are more structural comments.

redis/src/tls.rs
sp-angel: yeah, makes sense. 

Since this is a case for tls clients, maybe something like a `TlsParamBuilder` struct that can be "loaded" with keys and certificates, and used to generate a `TlsParam` value if they are consistent.
Then this `TlsParam` structure could  be used to eventually build the client.

Something like 
```
let mut tls_param_builder = TlsParamBuilder::new();

// set CA, client certificates and key
// and handle errors in case of inconsistent data

let tls_param: TlsParam = tls_param_builder.build().expect("parameters have been properly set"); 

let tls_client = redis::Client::open_tls(server, port, tls_param);
```

wdyt?

nihohit: I'm not sure what're the dependencies between the builder functions here. is the builder valid if only one of `client_cert`, `client_key`, or `root_cert` are called? Is it valid if all 3 are called? only some?
Essentially, in the 6 combinatorial options we have for the builder to be prepared before being built - which are valid and which aren't?

I assume that there are dependencies which aren't reflected in the API between the files' contents, and I'm not sure that this can be representable in the API, but if we can be certain that if function A is called, then function B has to / should not be called, we should endeavor to create an API in which it is impossible to not call B / to call B after A.

sp-angel: No combinations, mutual TLS needs all 3 parameters: `client_cert`, `client_key`, and `root_cert`.

These 3 parameters are contents of PEM files passed as bytestreams to the builder, and we must ensure they are all valid (convertible into `Certificate`, `PrivateKey` and `RootCertStore` respec) before we can use them to establish a mTLS connection. 

So the idea is to split the API in 2 parts: 

### 1 - Tls parameter creation

Parameter builder provides a way to the user to have these parameters properly validated before proceeding to create the tls_client.

Extending the client code above:
```rust
 let mut tls_param_builder = TlsParamBuilder::new();

// client certificate
 let mut reader = BufReader::new(cert_file);
 tls_param_builder
     .client_cert(&mut reader)
     .expect("Unable to handle client certs");

// client key
 let key_file = File::open(key_file).expect("Cannot open private key file");
 let mut reader = BufReader::new(key_file);
 tls_param_builder
     .client_key(&mut reader)
     .expect("Unable to handle client key");

// root cert
 let ca_file = File::open(ca_file).expect("Cannot open CA cert file");
 let mut reader = BufReader::new(ca_file);
 tls_param_builder
     .root_cert(&mut reader)
     .expect("Unable to load CA cert");

// A
let tls_param: TlsParam = tls_param_builder.build().expect("parameters have not been properly set"); 
```

I feel like enforcing the order of parameters supplied (`cert_file`, then `key_file`, then `root_cert`) would make some scenarios confusing:
_e.g._ if `cert_file` and `key_file` provided are fine, then an inconsistent `root_cert` would return an error and abort the whole builder?

Therefore `TlsParamBuilder`'s `build()` function would return `Ok(TlsParam)` only if the parameters were properly set, and issues during setting the parameters could be handled w/o destroying the builder.

### 2 - TLS client creation

Provided we have valid TLS parameters, user can create the client

```rust
// B
let tls_client = redis::Client::open_tls(server, port, tls_param);
```

nihohit: but this still uses the builder pattern, that implies that the client can be built without any or all of the 3 - you essentially move the check that all 3 were provided to runtime / document reading time. why not dump the builder pattern and enforce that these are required arguments in the function signature?

`Client::build_with_tls(host, port, client_cert, client_key, root_cert)`? This way the user can't try to build without the files, and there's less room for confusion.

redis/src/tls.rs
```diff
@@ -0,0 +1,212 @@
...
+
+#[derive(Debug, Clone)]
+pub struct TlsConnParams {
+    pub(crate) client_key: PrivateKey,
+    pub(crate) client_cert: Vec<Certificate>,
+    pub(crate) root_cert: RootCertStore,
+}
+
+/// The TLS client builder type
+struct TlsClientBuilder {
```
nihohit, line 88: why keep this struct? it adds almost 100 lines to the file. AFAIS TlsClientBuilder ::build can be inlined into  build_with_tls, or kept as a free function, and the builder can be removed.

sp-angel: dang, you are right

nihohit (approved):
LGTM @jaymell 
as mentioned before - this was only a review of the structure of the code, since I don't understand TLS.

redis/src/tls.rs
nihohit: thank you :)

sp-angel:
Fyi, noticed issues to connect to a TLS-enabled cluster using mTLS (_i.e._ set with `tls-auth-clients yes`) with my previous change. 
I am investigating that.

robjtede (changes requested):

redis/CHANGELOG.md
```diff
@@ -33,7 +33,7 @@
 * Bump MSRV to 1.60 ([#846](https://github.com/redis-rs/redis-rs/pull/846))
 * Improve error logging. ([#838](https://github.com/redis-rs/redis-rs/pull/838) @nihohit)
 * Improve documentation, add references to `redis-macros` ([#769](https://github.com/redis-rs/redis-rs/pull/769) @daniel7grant)
-* Allow creating Cmd with capacity. ([#817](https://github.com/redis-rs/redis-rs/pull/817) @nihohit)
+* Allow creating Cmd with capacity. ([#817](https://github.com/redis-rs/redis-rs/pull/817) @
```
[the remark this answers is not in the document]
sp-angel: fixed. weird how it happened :man_shrugging:

jaymell:
Hi, can you remove all the files w/ unrelated changes from the PR and also rebase? I see a lot of whitespace changes and/or changes that have already been made in `main`.

.github/workflows/rust.yml
jaymell, line 1: Please remove file from PR.

redis/CHANGELOG.md
jaymell, line 1: Please remove from PR

redis/src/cluster_client.rs
```diff
@@ -229,6 +276,28 @@ impl ClusterClientBuilder {
...
 
+    /// Sets raw TLS certificates for the new ClusterClient.
+    ///
+    /// `tls` parameter is required since certificates are only used in TLS context
+    /// All certificates are byte streams loaded from PEM files
+    /// their consistency is checked during `build()` call
+    #[cfg(feature = "tls-rustls")]
+    pub fn certs(
+        mut self,
+        tls: TlsMode,
```
jaymell, line 287: Does `tls` need to be supplied with this method? Does it make sense to supply configuration for mutual TLS but to have `TlsMode` be insecure?

jaymell:
Alright, apologies for the delay in reviewing here.... I've been procrastinating on reading into mutual TLS. If we can get conflicts resolved here, I'll try to get this merged soon. :pray:

redis/src/tls.rs
```diff
@@ -0,0 +1,160 @@
...
+        client_cert,
+        root_cert,
+    })
+}
+
+#[derive(Debug, Clone)]
+pub struct TlsConnParams {
+    pub(crate) client_key: PrivateKey,
+    pub(crate) client_cert: Vec<Certificate>,
+    pub(crate) root_cert: RootCertStore,
```
jaymell, line 139: Can `root_cert` be optional? I.e., is it possible that the CA cert is installed at the OS level and that we therefore don't need to pass it here?

redis/src/tls.rs
```diff
@@ -0,0 +1,160 @@
...
+///         .arg(&["key1", "key2"])
+///         .query_async(&mut con)
+///         .await;
+///     assert_eq!(result, Ok(("foo".to_string(), b"bar".to_vec())));
+///     println!("Result from MGET: {result:?}");
+///
+///     Ok(())
+/// }
+/// ```
+pub fn build_with_tls<C: IntoConnectionInfo>(
```
jaymell, line 66: I think this might make more sense as a method of `Client`, since it's basically a constructor, right?

redis/src/tls.rs
sp-angel: I guess it's possible, but not the use case I've seen around. 
My idea is to tackle the case in which we create a client only if all certificates/key are consistent in a single method, instead of relying on a builder with multiple steps to take and check consistence of certificates and keys.

redis/src/tls.rs
sp-angel: it's a constructor aye, but kept it here to avoid setting lot of attributes to condition the compilation to only when `rustls` is enabled. 
I could use an internal build method in `tls.rs` and call from a constructor wrapper in the client that would only be defined if the feature is available. Wdyt?

redis/src/tls.rs
jaymell: This is part of the public API, though, so if it's ever feasible that a client may retrieve the root CA certs from the OS, then I think we should leave open that possibility. The only changes I see required for that would be modifying `create_rustls_config` to use `root_store` instead in `with_with_certificates` if `tls_params.root_cert` is `None`, e.g.,

```
    let config = if let Some(tls_params) = tls_params {
        config
            .with_root_certificates(tls_params.root_cert.unwrap_or(root_store))
           ...
    } else { 
    ...
```

Practically speaking this might not work at the moment *anway*, b/c I don't think `rustls` will check OS-level certs unless we make some additional changes. But we can test/fix that later if needed. Right now I just want to make sure we're defining the public API the best way possible.

jaymell: Actually I'll turn my point on its head as well. Is it not possible we might want to enable explicitly specifying a specific root certificate, but don't actually have client-side auth enabled (i.e., `tls-auth-clients` is `no`) so don't care about supplying client cert and key?

sp-angel: this way I see the possible scenarios are:
- TLS using CA in trust store (already implemented)
- TLS using CA provided 
- mTLS using CA in trust store and client cert/key
- mTLS using provided CA cert + client cert/key (what we have at [this](https://github.com/redis-rs/redis-rs/pull/858/commits/8aa5943d21545a0946930bfbf21f1b86f8194204#diff-8ce03c631ce32ec6617737c96699023373f6d8d35700bf9820a8b32951cb8e54R60) point)

if that's the case, maybe it would make sense to use a Client ConfigBuilder approach similar to (rustls client builder here)[https://github.com/redis-rs/redis-rs/blob/8aa5943d21545a0946930bfbf21f1b86f8194204/redis/src/connection.rs#L790-L792]

redis/src/tls.rs
sp-angel: yeah, the idea is add a constructor, but use int method to avoid lots of directives to dependencies conditioned to `rustls`: `retrieve_tls_certificates`, `ConnectionAddr`, `RedisError` and `ErrorKind`...

redis/src/tls.rs
jaymell: Well, why not just make the client artifacts their own type, which would enforce the conditions for each potential scenario? E.g.,
```
struct ClientTlsParams {
    client_key: PrivateKey,
    client_cert: Vec<Certificate>,
}

pub struct TlsConnParams {
    client_params: Option<ClientTlsParams>,
    root_cert: Option<RootCertStore>,
}
```

sp-angel: sorry, took few days off and only got some time to work on this couple days ago.

I've changed both `Client::build_with_tls` and `ClusterClientBuilder` to make `root_cert` and `client_tls` pair optional at the API, which allow setting TLS parameters as you suggested. 

Pls let me know what you think :+1:

redis/src/tls.rs
sp-angel: done

redis/src/client.rs
```diff
@@ -287,6 +293,78 @@ impl Client {
...
+    ///     assert_eq!(result, Ok(("foo".to_string(), b"bar".to_vec())));
+    ///     println!("Result from MGET: {result:?}");
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    #[cfg(feature = "tls-rustls")]
+    pub fn build_with_tls<C: IntoConnectionInfo>(
+        conn_info: C,
+        client_tls_params: Option<(&mut dyn BufRead, &mut dyn BufRead)>,
```
jaymell, line 363: Minor, but I wonder if we should add a type here so it's a bit more obvious which is the key and which is the cert. (same comment for cluster_client)

jaymell:
I think overall this looks great! It does make me a bit nervous that there are no tests here. Any thoughts about the feasibility of adding at least one basic test to confirm that mutual TLS works? I feel like it's possible with our current test-support code though am not sure off the top of my head how much refactoring would required.

redis/src/client.rs
sp-angel: Was thinking about this and I'm feeling tempted to make these structs public and using `RawCertificates` as the set TLS parameters for both clustered and single client constructor. 
```
pub struct ClientTlsBinary {
    pub client_cert: Vec<u8>,
    pub client_key: Vec<u8>,
}

pub struct CertificatesBinary {
    pub client_tls: Option<ClientTlsBinary>,
    pub root_cert: Option<Vec<u8>>,
}
```
Wdyt?

sp-angel:
Regarding tests, totally agree. Seems there are mechanisms to set a cluster of servers w/TLS enabled enforcing mTLS, but I'm a bit struggling to run these directly on my box to change / experiment.

I managed to run them locally only with the help of  [act](https://github.com/nektos/act) but it takes ages to complete, the dev cycle is a bit painful :sweat:

jaymell:
> I managed to run them locally only with the help of [act](https://github.com/nektos/act) but it takes ages to complete, the dev cycle is a bit painful 😓

Are you not able to to run a simple `make test` on your system to get tests running? Sorry, am only familiar with Linux these days so not sure what local issues you face. 

The test code we have for generating certs looks like it can easily be extended to support creating a client-side key/cert, so we can probably add some methods on the test-context to get the paths for those, which can then be used to build the client. If you want to take a stab at getting something working, I can try to help with testing if it's not something you're able to do locally.

redis/src/client.rs
jaymell: Looks good to me :+1:

sp-angel: cool. done (hopefully final form now :crossed_fingers: )

sp-angel:
I'm on Linux too, but seems the issue I was having was due to stale crates and was solved with a `cargo update` :sweat_smile: 

On the bright side I've managed to get authentication tests working w/certs generated by `build_keys_and_certs_for_tls()` for all `test_cluster`, but still need some refactoring to get the code less hacky and pass `TlsFilePaths` around, and also handle it only when `tls-rustls` is enabled. 

I'm working on this now.

sp-angel:
In the process enabling TLS tests, I came across `sentinel` functionality that wasn't working because it wasn't implemented. Just pushed a change for that before adding the tests

Note this somehow breaks the existing API: it now requires a `certificates: Option<CertificatesBinary>` field within the `SentinelNodeConnectionInfo`.

Simple change (`certificates: None` if TLS isn't used) but a change anyway

jaymell:
> In the process enabling TLS tests, I came across sentinel functionality that wasn't working because it wasn't implemented. Just pushed a change for that before adding the tests

Perhaps a bit late, but I think it's fine if you want to leave sentinel support out of this PR. We can always merge that in a subsequent PR.

redis/tests/support/mod.rs
```diff
@@ -286,15 +307,71 @@ pub struct TestContext {
     pub client: redis::Client,
 }
 
+pub(crate) fn is_tls_enabled() -> bool {
```
jaymell, line 310: I think it might be a simpler approach to just have a few very basic tests (i.e., for sync, async, clustered, etc.) that verify the new functionality works. I would prefer we avoid enabling it for all tls-enabled tests.

redis/src/cluster_client.rs
```diff
@@ -108,13 +119,34 @@ impl ClusterClientBuilder {
...
 
+        #[cfg(not(feature = "tls-rustls"))]
+        let tls_params = None;
+
+        #[cfg(feature = "tls-rustls")]
+        let tls_params = {
+            let retrieved_tls_params = self
+                .builder_params
+                .certs
+                .clone()
```
thalesfragoso, line 140: Is this clone necessary ?

redis/src/cluster_client.rs
```diff
@@ -108,13 +119,34 @@ impl ClusterClientBuilder {
...
+                .map(retrieve_tls_certificates);
+
+            if let Some(tls_params) = retrieved_tls_params {
+                Some(tls_params?)
+            } else {
+                None
+            }
```
thalesfragoso, line 147: I think this can be replaced with `transpose()?` and be more legible.

redis/src/cluster_client.rs
```diff
@@ -229,6 +263,23 @@ impl ClusterClientBuilder {
...
+    /// Sets raw TLS certificates for the new ClusterClient.
+    ///
+    /// When set, enforces the connection must be TLS secured.
+    ///
+    /// All certificates must be provided as byte streams loaded from PEM files their consistency is
+    /// checked during `build()` call.
+    ///
+    /// - `client_tls` - Pair containing client certificate and key respectively used for
+    ///   client-side authentication.
+    /// - `root_cert` - CA root certificate. If not provided, it uses certificates in trust store.
```
thalesfragoso, line 275: These docs seem outdated.

redis/src/connection.rs
```diff
@@ -31,9 +31,20 @@ use rustls::OwnedTrustAnchor;
...
+    not(feature = "tls-rustls-webpki-roots")
+))]
 use rustls_native_certs::load_native_certs;
 
+#[cfg(feature = "tls-rustls")]
+use crate::tls::TlsConnParams;
+
+#[cfg(not(feature = "tls-rustls"))]
+#[derive(Clone, Debug)]
+pub struct TlsConnParams;
```
thalesfragoso, line 46: Maybe add a private field for future proofing ? Together with a "constructor" function, of course.

redis/src/tls.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    let client_tls_params = if let Some(ClientTlsBinary {
+        client_cert,
+        client_key,
+    }) = client_tls
+    {
+        let mut certs = rustls_pemfile::certs(&mut Cursor::new(client_cert))?;
+        let client_cert_chain = certs.drain(..).map(Certificate).collect();
+
+        let client_key = load_key(&mut Cursor::new(client_key))?;
```
thalesfragoso, line 75: Is this `Cursor` necessary ? `&[u8]` implements `BufRead`.

redis/src/tls.rs
```diff
@@ -0,0 +1,146 @@
...
+        Some(ClientTlsParams {
+            client_cert_chain,
+            client_key,
+        })
+    } else {
+        None
+    };
+
+    let root_cert_store = if let Some(root_cert) = root_cert {
+        let certs = rustls_pemfile::certs(&mut Cursor::new(root_cert))?;
```
thalesfragoso, line 86: Same as above.

redis/src/tls.rs
```diff
@@ -0,0 +1,146 @@
...
+/// - `client_tls`: binaries of clientkey and certificate within a `ClientTlsBinary` structure if mTLS is used
+/// - `root_cert`: binary CA certificate in PEM format if CA is not in local truststore
+///
+#[derive(Clone)]
+pub struct CertificatesBinary {
+    /// 'ClientTlsBinary' containing client certificate and key if mTLS is to be used
+    pub client_tls: Option<ClientTlsBinary>,
+    /// root certificate byte stream in PEM format if the local truststore is *not* to be used
+    pub root_cert: Option<Vec<u8>>,
+}
```
thalesfragoso, line 27: Not very important, but why the `Binary` in the name ? It kinda alludes to DER encoding in this situation, which isn't the case here, since we're using PEM.

redis/src/client.rs
```diff
@@ -287,6 +290,91 @@ impl Client {
...
         &self.connection_info
     }
+
+    /// Constructs a new `Client` with parameters necessary to create a TLS connection.
+    ///
+    /// - `conn_info` - URL using the `rediss://` scheme.
+    /// - `client_tls_params` - Optional pair of byte readers yielding PEM formatted files for:
+    ///   0. client's certificate chain
+    ///   1. client's private key
+    /// - `root_cert` - Optional byte readers yielding PEM formatted file for root certificates.
```
thalesfragoso, line 300: Is this documentation outdated ?

redis/src/client.rs
```diff
@@ -287,6 +290,91 @@ impl Client {
...
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    #[cfg(feature = "tls-rustls")]
+    pub fn build_with_tls<C: IntoConnectionInfo>(
+        conn_info: C,
+        tls_certs: CertificatesBinary,
+    ) -> RedisResult<Client> {
+        inner_build_with_tls(conn_info, tls_certs)
```
thalesfragoso, line 376: Another nitpick, but there isn't really a need for `inner_build_with_tls` to be generic. `build_with_tls` should be the one to call `into_connection_info`. This will help with compilation times and binary sizes in some situations.

You probably won't notice the difference in most situations (again, nitpick), but it's a good practice and it's easy enough in this case.

thalesfragoso (commented):
Thanks for this PR, I left a few suggestions/questions.

I'm interested in moving this forward, so let me know if you need any help.

redis/src/cluster_client.rs
sp-angel: Not strictly necessary, but it's a simpler way to handle `tls_params` extraction only if `tls-rustls` feature is defined. 

I don't seem a lot of perf penalty here since it should happen only 1x per client build.

redis/src/connection.rs
sp-angel: not really sure what you meant here... could you pls jot down some code to exemplify?

redis/src/tls.rs
sp-angel: I thought of `Binary` cos it contains just the binary streams for keys and certificates, and to make clear that's not a PEM validated structure yet.
... but if it's confusing maybe `Raw` would work better?

sp-angel:
> > In the process enabling TLS tests, I came across sentinel functionality that wasn't working because it wasn't implemented. Just pushed a change for that before adding the tests
> 
> Perhaps a bit late, but I think it's fine if you want to leave sentinel support out of this PR. We can always merge that in a subsequent PR.

Seems a bit pity to revert that since seems it's working for sentinel now :smiling_face_with_tear: 
...but I understand that would break compatibility. 

If the choice is for keeping compatibility I can leave that `sentinel` change in another branch and revert for this PR

jaymell:
> > > In the process enabling TLS tests, I came across sentinel functionality that wasn't working because it wasn't implemented. Just pushed a change for that before adding the tests
> > 
> > 
> > Perhaps a bit late, but I think it's fine if you want to leave sentinel support out of this PR. We can always merge that in a subsequent PR.
> 
> Seems a bit pity to revert that since seems it's working for sentinel now 🥲 ...but I understand that would break compatibility.
> 
> If the choice is for keeping compatibility I can leave that `sentinel` change in another branch and revert for this PR

Hey, sorry for delayed reply. I think it would be good to leave the changes you've made for sentinel in a different PR and let's look at those changes separately. I don't want to hold up this PR any further than necessary.

jaymell:
Once you've removed the sentinel changes, resolved merge conflicts, and feel it's ready for approval, please drop a line. I haven't looked at the recent changes and am not sure where things stand at the moment. Thanks!

sp-angel:
> Once you've removed the sentinel changes, resolved merge conflicts, and feel it's ready for approval, please drop a line. I haven't looked at the recent changes and am not sure where things stand at the moment. Thanks!

Hey, I 've reverted al `sentinel` changes and supporting code. Also rebased this branch on `main` and fixed the existing conflicts, From my side I believe it's in a good shape for another check :rocket: 

Btw a heads up: I'll be taking a couple of weeks off, will be able to respond by end of Oct.

.github/workflows/rust.yml
jaymell, line 1: Whitespace-only changes -- can we remove this file from PR?

redis/CHANGELOG.md
jaymell, line 1: Whitespace-only changes -- can we remove this file from PR?

redis/src/connection.rs
```diff
@@ -1138,7 +1214,13 @@ impl ConnectionLike for Connection {
     }
 
     fn check_connection(&mut self) -> bool {
-        cmd("PING").query::<String>(self).is_ok()
+        match cmd("PING").query::<String>(self) {
```
jaymell, line 1217: I'm assuming you added this for debugging purposes. Can we remove?

redis/src/tls.rs
jaymell: I think @thalesfragoso makes a fair point. What about just `ClientTlsConfig` and `TlsCertificates` instead of  ClientTlsBinary and CertificatesBinary, respectively?

redis/src/connection.rs
robjtede: Adding `#[non_exchaustive]` will also make this type non-construct-able publicly.

jaymell (commented):
Ok last comment regarding the PR: 

1) The test code is just duplicating other tests, right? At a minimum, can we refactor the tests so the code isn't duplicated?
2) Can we add a test that ensures a failure if we try to connect w/out the TLS params or w/ incorrect ones?

.github/workflows/rust.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.63.0
+          - 1.65.0
```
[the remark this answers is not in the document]
sp-angel: sure :+1:

sp-angel:
On 2 above, good call.

On 1 the duplication was mostly to deal with symbols that are only defined when `tls-rustls` feature is on. Most of the tentative deduplication I tried resulted in a bunch of "dead code" warnings, since any `enable_mtls` become useless when feature is off.

I could group all mtls tests under a test module only enabled by that feature, that would still keep some duplication but least would avoid littering tests with ` #[cfg(feature = "tls-rustls")]` blocks. 

wdyt?

sp-angel:
Also reviewing the tests for (2) above I realized most of the `_with_mtls` tests add little value since it makes sense to validate if  the connection can be made. 

So I pruned few test cases and added some to ensure connection is only established if tls is available on server-side and certificates are provided.

redis/tests/test_cluster.rs
```diff
@@ -747,3 +760,132 @@ fn test_cluster_split_multi_shard_command_and_combine_arrays_of_values() {
...
+        assert_eq!(
+            redis::cmd("MGET")
+                .arg(&["{x}key1", "{x}key2"])
+                .query(&mut con),
+            Ok(("foo".to_string(), b"bar".to_vec()))
+        );
+    }
+
+    #[test]
+    fn test_cluster_should_not_connect_without_mtls() {
```
jaymell, line 792: What happens if you try to run a SET/GET using a new client that doesn't have TLS enabled? Can we make some assertions based on that?

redis/tests/test_cluster.rs
```diff
@@ -747,3 +760,132 @@ fn test_cluster_split_multi_shard_command_and_combine_arrays_of_values() {
...
+            }
+            _ => {
+                if let Err(e) = connection {
+                    panic!("Must be able to connect without client credentials if server does NOT accept TLS: {e:?}");
+                }
+            }
+        }
+    }
+
+    #[test]
```
jaymell, line 815: I don't know that we need the remaining tests here. As you mentioned, they don't add much value. I think just a basic test and then one asserting that requests fail if the client doesn't enable mTLS is sufficient.

redis/tests/test_cluster_async.rs
```diff
@@ -1433,3 +1451,98 @@ fn test_async_cluster_non_retryable_error_should_not_retry() {
...
+                .clone()
+                .query_async(&mut connection)
+                .await?;
+            assert_eq!(res, "test_data");
+            Ok::<_, RedisError>(())
+        })
+        .unwrap();
+    }
+
+    #[test]
```
jaymell, line 1484: Same points as above w/ tests in this file

redis/tests/test_cluster.rs
sp-angel: In this case it shouldn't even be able to create a connection, unless the server is only accepting TCP
Iiuc I believe that's already handled by this test case when credentials are not provided (`mtls_enabled:false`)

jaymell: Ok, yes, that makes sense -- thanks for clarifying!

redis/tests/test_async.rs
```diff
@@ -680,10 +686,101 @@ fn test_connection_manager_reconnect_after_delay() {
...
         let result: redis::Value = manager.set("foo", "bar").await.unwrap();
         assert_eq!(result, redis::Value::Okay);
     });
 }
+
+#[cfg(feature = "tls-rustls")]
+mod mtls_test {
+    use super::*;
+
+    #[test]
```
jaymell, line 702: We can get rid of one of these two tests as well, right?

jaymell: Otherwise, I think this is ready to go in!

robjtede:
Ignoring the MSRV jobs, having a hard time tricking the rest of the jobs to pass successfully. Retrying yields different results each time. Is it expected atm?

jaymell:
> Ignoring the MSRV jobs, having a hard time tricking the rest of the jobs to pass successfully. Retrying yields different results each time. Is it expected atm?

Yes the tests are a bit of a dumpster fire at the moment and haven't looked into it recently so not sure offhand has caused the recent instability. I will try to re-run the specific failures (besides 1.63) a few times and as long as things ultimately pass will likely merge thereafter.

jaymell (approved):

jaymell:
Ok, I'm going to go ahead and merge this. Thanks for all the hard work here @robjtede -- this is a great feature!

PR #859, opened 2023-06-06
Allow configuration of cluster retry wait duration

nihohit:
This change allows the user to configure the wait duration between retry
attempts in the cluster clients. The change doesn't the current wait time,
but it does contain a fix to the tests, which will allow us to reduce the defaults
without breaking the tests.

This change also introduces jitter, so that if multiple requests
are waiting, they won't be all sent at exactly the same time.


jaymell (approved):

nihohit:
@jaymell rebased

PR #862, opened 2023-06-12
Fix minor async client bug.

nihohit:
`connections.remove(addr),` in the old code always returned `None`, because `connections` in the fold operation was an empty map. This change ensures that connections from the external `connections` variable will be used in the fold operation.

`len` was removed from SlotMap because it's no longer in use.


jaymell (approved):

jaymell:
Thanks for fixing this!

PR #863, opened 2023-06-12
refresh_slots - run CLUSTER SLOTS on all nodes.

nihohit:
This change runs CLUSTER SLOTS concurrently on all slots, instead of serially. Using `select_ok` means that the first success will stop all other attempts. This allows us to complete the process faster, if some connections have higher latency than others, or simply fail.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -436,26 +436,18 @@ where
         let mut connections = mem::take(&mut self.connections);
         let cluster_params = self.cluster_params.clone();
         async move {
-            let mut result = Ok(SlotMap::new());
-            for (_, conn) in connections.iter_mut() {
-                let mut conn = conn.clone().await;
-                let value = match conn.req_packed_command(&slot_cmd()).await {
-                    Ok(value) => value,
-                    Err(err) => {
-                        result = Err(err);
-                        continue;
-                    }
-                };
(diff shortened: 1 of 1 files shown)


jaymell:
This seems reasonable.  For the sake of argument, though, is generating mostly unnecessary load on all nodes in the cluster potentially undesirable?

nihohit:
Fair question. AFAIK a Redis server is supposed to handle x0,000< TPS, so even assuming a user has thousands of active clients refreshing slots at the same time, its still ~10% of the lowest load the server should be able to handle. If the clients that are refreshing slots are also not sending other requests, this isn't an increased load at all.
On the other hand, there's also the increased CPU/network load on the client's machine. Assuming a server has 500> shards (the maximum in ElastiCache) and ~3-4 nodes per shard, this could be expensive, but again - thousands of concurrent network calls seem ok to me. 
We can try to throttle this in some way - chunk the number of concurrent attempts, if that's what you prefer.

barshaul:
I think we can randomly take Min(~3, cluster_size) nodes at a time and query them in parallel. It can be a lot of overhead to query all nodes in large clusters (from the client perspective, not from the server side). In a future approach we might want to compare the output of CLUSTER SLOTS of some nodes to get the most common topology view, and also then we wouldn't want to query all nodes for that matter.

nihohit:
closing for now.

PR #864, opened 2023-06-13
cluster_async: Wrap internal state with Arc.

nihohit:
This change allows us to send the internal state of the client to async functions without the `self` reference imposing a lifetime.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -548,10 +548,12 @@ where
 
         let read_guard = core.conn_lock.read().await;
         // Ideally we would get a random conn only after other means failed,
-        // but referencing self in the async block is problematic:
+        // but we have to release the lock before any `await`, otherwise the
+        // operation might be suspended until ending a refresh_slots call,
+        // which will be blocked on the lock:
         let random_conn = get_random_connection(&read_guard.0);
 
-        let addr = match info.redirect.take() {
+        let conn = match info.redirect.take() {
             Some(Redirect::Moved(moved_addr)) => Some(moved_addr),
             Some(Redirect::Ask(ask_addr)) => {
                 asking = true;
@@ -562,17 +564,21 @@ where
                 .as_ref()
                 .and_then(|route| read_guard.1.slot_addr_for_route(route))
                 .map(|addr| addr.to_string()),
-        };
+        }
+        .map(|addr| {
+            let conn = read_guard.0.get(&addr).cloned();
+            (addr, conn)
+        });
+        drop(read_guard);
 
-        let conn_future = match addr {
-            Some(addr) => {
-                let conn = read_guard.0.get(&addr).cloned();
+        let conn_future = match conn {
+            Some((addr, conn)) => {
                 let params = core.cluster_params.clone();
 
                 async move { (Self::get_or_create_conn(&addr, conn, params).await, addr) }
                     .map(|(result, addr)| {
                         result
-                            .map(|conn| (addr, async { conn }.boxed().shared()))
+                            .map(|conn| (addr, async move { conn }.boxed().shared()))
                             .unwrap_or(random_conn)
                     })
                     .boxed()
@@ -580,7 +586,6 @@ where
             }
             None => async move { random_conn }.boxed().shared(),
         };
-        drop(read_guard);
 
         let (addr, conn) = conn_future.await;
         let mut conn = conn.await;
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -49,11 +49,10 @@ use futures::{
     prelude::*,
     ready, stream,
 };
-use futures_util::lock::Mutex;
 use log::trace;
 use pin_project_lite::pin_project;
 use rand::{seq::IteratorRandom, thread_rng};
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{mpsc, oneshot, RwLock};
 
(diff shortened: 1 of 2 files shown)


redis/src/cluster_async/mod.rs
```diff
@@ -94,17 +94,21 @@ where
 type ConnectionFuture<C> = future::Shared<BoxFuture<'static, C>>;
 type ConnectionMap<C> = HashMap<String, ConnectionFuture<C>>;
 
-struct ClusterConnInner<C> {
-    connections: ConnectionMap<C>,
+struct ConnectionsAndMetadata<C> {
```
jaymell, line 97: Minor, but a more terse name here would be great. `ConnsMeta` is perhaps not appropriate but saves us 13 columns.

redis/src/cluster_async/mod.rs
```diff
@@ -355,18 +357,24 @@ where
     ) -> RedisResult<Self> {
         let connections =
             Self::create_initial_connections(initial_nodes, cluster_params.clone()).await?;
-        let mut connection = ClusterConnInner {
-            connections,
+        let connections_and_metadata = Arc::new(ConnectionsAndMetadata {
+            connections: Mutex::new(connections),
```
jaymell, line 361: I'm honestly not sure -- are we adding overhead by introducing the Mutex?

redis/src/cluster_async/mod.rs
```diff
@@ -94,17 +94,21 @@ where
...
-struct ClusterConnInner<C> {
-    connections: ConnectionMap<C>,
+struct ConnectionsAndMetadata<C> {
+    connections: Mutex<ConnectionMap<C>>,
     slots: SlotMap,
+    cluster_params: ClusterParams,
+}
+
+struct ClusterConnInner<C> {
+    connections_and_metadata: Arc<ConnectionsAndMetadata<C>>,
```
jaymell, line 104: Or maybe just calling this field `inner` would be sufficient and not too confusing? Anything to cut down on the width here would be spectacular.

nihohit: inner is fine with me. once we'll decide what to call the type, I'll push a commit.

redis/src/cluster_async/mod.rs
nihohit: I'm not 100% sure, but:
1. The mutex surrounds a Map, and most of the time it is only held for get/remove calls, which should be very fast, so even if the mutex is locked, the contention should be low.
2. The places where the mutex is held for long are functions that take &mut self, so they should be exclusive, and taking an unlocked mutex should have a negligible cost.

So as long as we're sure to either use it in exclusive context or for quick shared calls, this should be ok.

nihohit: also, it means that we don't leave the client in a weird state, like we do when we empty the connection map - ` let mut connections = mem::take(&mut self.connections);`

redis/src/cluster_async/mod.rs
jaymell: "Core" is 4 characters. I'll take it!

nihohit: renamed.

redis/src/cluster_async/mod.rs
nihohit: done.

nihohit:
@jaymell looks like we have some more flakyness - 2 test suites run for hours.
https://github.com/redis-rs/redis-rs/actions/runs/5279416076/jobs/9550056369?pr=864
https://github.com/redis-rs/redis-rs/actions/runs/5279416076/jobs/9550056536?pr=864

jaymell:
> @jaymell looks like we have some more flakyness - 2 test suites run for hours. https://github.com/redis-rs/redis-rs/actions/runs/5279416076/jobs/9550056369?pr=864 https://github.com/redis-rs/redis-rs/actions/runs/5279416076/jobs/9550056536?pr=864

Kicked off the jobs again. Have we seen this issue outside of this PR? I don't recall.

nihohit:
@jaymell I'm not sure, I think I might've noticed this in the past. 
1. one action seems to be stuck again.
2. The actions view doesn't show any cancelled/timed out action, including the two I linked to, so there might be more which are simply not available in the view.

nihohit:
But both runs seem to have been stuck on `test_async_cluster_basic_script`, so this branch might've added a deadlock there. I'll investigate.

nihohit:
I believe I found &solved the deadlock.

jaymell:
Those deadlocks in the prior test failures have me a bit nervous. I think these latest changes look good, but I'm no longer confident I understand the risks of having an `RwLock` or `Mutex` in our implementation here.

Also, what was your reasoning for switching to `RwLock`?

nihohit:
> I'm no longer confident I understand the risks of having an RwLock or Mutex in our implementation here.

The risks of using a Mutex / RwLock are always that there might be a deadlock possible. On the other hand, the risks of not doing this are that the async cluster client won't behave correctly, or will implement fan-out commands serially instead of concurrently. IMO, since the test suite managed to find several issues, I'm feeling pretty confident about this.

> what was your reasoning for switching to RwLock?

The ability to hold multiple read locks during try_request. Although the locks are held very briefly each time, I would still prefer not to block concurrent requests at all.

jaymell (approved):

jaymell:
Sorry, rebase is again required, but let's get this in!

nihohit:
@jaymell rebased

PR #869, opened 2023-06-19
Amend README for macOS users

sarisssa:
The current guidelines in the README for testing and development are incorrect for macOS users since we do not have access to `librejson.so`. Instead, macOS users need to link the env variable to `librejson.dylib` which has added to the README along with a link to the Redis JSON repo.


jaymell (approved):

jaymell:
Thanks!

PR #870, opened 2023-06-20
Fix redirect routing on request with no route.

nihohit:
If a request's `route` field is set to None and the request was redirected, it should go to the redirected address regardless of the original route. Currently it goes to a random node.


jaymell:
Thanks -- looks good to me!

jaymell (approved):

PR #872, opened 2023-06-21
Fix async connect when ns resolved to multi ip

### LINKED ISSUE #871: Different behavior of sync and async get connection methods
When I create a connection using `redis://localhost:6379`, the synchronous method `Client::get_connection` succeeds, but the asynchronous method `Client::get_async_connection` throws an error.

```rust
let rc = redis::Client::open("redis://localhost:6379").unwrap();
let mut conn = rc.get_connection().unwrap();

redis::cmd("PING").query::<String>(&mut conn).unwrap()

// output: "PONG"


let rc = redis::Client::open("redis://localhost:6379").unwrap();
let conn = rc.get_async_connection().await;

let mut conn = match conn {
    Ok(conn) => conn,
    Err(e) => {
        println!("Failed to connect: {}", e);
        return;
    }
};

redis::cmd("PING").query_async::<_, String>(&mut conn).await.unwrap()

// output: Failed to connect: Connection refused (os error 111)
```

The issue seems to be that localhost corresponds to two IP addresses (`::1` and `127.0.0.1`), and the asynchronous method only uses the first one (`::1`) returned by `get_socket_addrs`. This causes a connection error because my Redis instance is running on the IPv4 address.

On the other hand, the synchronous method uses the standard library's `TcpStream::connect` method which tries each IP address in sequence until it succeeds.

hugefiver:
fix #871

### DIFF
--- a/redis/src/aio/connection.rs
+++ b/redis/src/aio/connection.rs
@@ -368,7 +368,7 @@ async fn get_socket_addrs(
     #[cfg(feature = "tokio-comp")]
     let socket_addrs = lookup_host((host, port)).await?;
     #[cfg(all(not(feature = "tokio-comp"), feature = "async-std-comp"))]
-    let mut socket_addrs = (host, port).to_socket_addrs().await?;
+    let socket_addrs = (host, port).to_socket_addrs().await?;
 
     let mut socket_addrs = socket_addrs.peekable();
     match socket_addrs.peek() {
@@ -383,21 +383,24 @@ async fn get_socket_addrs(
 pub(crate) async fn connect_simple<T: RedisRuntime>(
     connection_info: &ConnectionInfo,
 ) -> RedisResult<T> {
-    
     Ok(match connection_info.addr {
         ConnectionAddr::Tcp(ref host, port) => {
             let socket_addrs = get_socket_addrs(host, port).await?;
-            let mut ret = Err(RedisError::from((
-                ErrorKind::InvalidClientConfig,
-                "No address found for host",
-            )));
-            for socket_addr in socket_addrs {
-                ret = <T>::connect_tcp(socket_addr).await;
-                if ret.is_ok() {
-                    break;
-                }
-            }
-            ret?
+            futures::stream::iter(socket_addrs)
+                .map(|socket_addr| async move { <T>::connect_tcp(socket_addr).await })
+                .fold(
+                    Err(RedisError::from((
+                        ErrorKind::InvalidClientConfig,
+                        "No address found for host",
+                    ))),
+                    |acc, res| async {
+                        match acc {
+                            ok @ Result::Ok(_) => ok,
+                            Err(_) => res.await,
+                        }
+                    },
+                )
+                .await?
         }
 
         #[cfg(any(feature = "tls-native-tls", feature = "tls-rustls"))]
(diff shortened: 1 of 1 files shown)


redis/src/aio/connection.rs
```diff
@@ -361,13 +361,18 @@ where
...
-async fn get_socket_addrs(host: &str, port: u16) -> RedisResult<SocketAddr> {
+async fn get_socket_addrs(
+    host: &str,
+    port: u16,
+) -> RedisResult<impl Iterator<Item = SocketAddr> + Send + '_> {
     #[cfg(feature = "tokio-comp")]
-    let mut socket_addrs = lookup_host((host, port)).await?;
+    let socket_addrs = lookup_host((host, port)).await?;
     #[cfg(all(not(feature = "tokio-comp"), feature = "async-std-comp"))]
     let mut socket_addrs = (host, port).to_socket_addrs().await?;
```
nihohit, line 371: if you didn't need `mut` for tokio, you probably don't need it for std-async either.

redis/src/aio/connection.rs
```diff
@@ -378,10 +383,21 @@ async fn get_socket_addrs(host: &str, port: u16) -> RedisResult<SocketAddr> {
 pub(crate) async fn connect_simple<T: RedisRuntime>(
     connection_info: &ConnectionInfo,
 ) -> RedisResult<T> {
+    
     Ok(match connection_info.addr {
         ConnectionAddr::Tcp(ref host, port) => {
-            let socket_addr = get_socket_addrs(host, port).await?;
-            <T>::connect_tcp(socket_addr).await?
+            let socket_addrs = get_socket_addrs(host, port).await?;
```
nihohit, line 389: consider:
```rs
  let socket_addrs = get_socket_addrs(host, port).await?;
  for socket_addr in socket_addrs {
      let ret = <T>::connect_tcp(socket_addr).await;
      if ret.is_ok() {
          return ret;
      }
  }
 Err(RedisError::from((
      "No address found for host",
  )))
```
or even simpler

```rs
match futures::future::select_ok(get_socket_addrs(host, port).await?.map(|socket_addr|<T>::connect_tcp(socket_addr)).await {
  Ok((res, _) => Ok(res),
  Err(_) =>  Err(RedisError::from((
                "No address found for host",
            )))
}
```
(I hope these compile. Just wanted to provide suggestions)

redis/src/aio/connection.rs
```diff
@@ -390,8 +406,18 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
             port,
             insecure,
         } => {
-            let socket_addr = get_socket_addrs(host, port).await?;
-            <T>::connect_tcp_tls(host, socket_addr, insecure).await?
+            let socket_addrs = get_socket_addrs(host, port).await?;
+            let mut ret = Err(RedisError::from((
```
nihohit, line 410: same as above

nihohit (approved):
minor comments. Looking good!

redis/src/aio/connection.rs
hugefiver: didn't notice that the dependencies include the `futures` package earlier :)

redis/src/aio/connection.rs
```diff
@@ -380,8 +386,22 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
 ) -> RedisResult<T> {
     Ok(match connection_info.addr {
         ConnectionAddr::Tcp(ref host, port) => {
-            let socket_addr = get_socket_addrs(host, port).await?;
-            <T>::connect_tcp(socket_addr).await?
+            let socket_addrs = get_socket_addrs(host, port).await?;
+            stream::iter(socket_addrs)
+                .map(|socket_addr| async move { <T>::connect_tcp(socket_addr).await })
+                .fold(
```
nihohit, line 392: Truly minor, feel free to ignore - 
can you either remove the `map` and call `connect_tcp` in the `fold` operation?
Alternatively, can the fold operation be moved to function, to reduce repetition?

nihohit (approved):
LGTM @jaymell

nihohit:
@hugefiver do you think this is possible to test in CI, or does it require a specific machine config?

hugefiver:
Maybe it's possible to simulate it by modifying `hosts` file, such as binding the `l1` domain to `127.0.0.{1,2,3}`, etc. The example in the issue is in this way. maybe running a DNS server alse works, or directly mock the function of querying DNS in tokio library (not sure if it's possible)

I have no idea which one to choose.

nihohit:
yeah, all of the above sound over the top. nevermind.

jaymell (approved):

jaymell:
Thanks for this!

PR #873, opened 2023-06-22
Remove connection checking on every request.

nihohit:
Currently every request is prepended with a PING, and is only sent when the ping completes. This slows down performance significantly - removing this increased TPS on my test machine by 50%.


jaymell (approved):

jaymell:
Thanks for catching this!

PR #874, opened 2023-06-23
Reduce the number of unnecessary clones.

nihohit:
Started looking into the performance degradation in cluster_async. This isn't a big deal, but every little bit helps :)


jaymell (approved):

jaymell:
:sunglasses:

PR #875, opened 2023-06-26
Aio: add connection / response timeouts.

nihohit:
Cluster timeouts will be added in a separate PR.

### DIFF
--- a/redis-test/Cargo.toml
+++ b/redis-test/Cargo.toml
@@ -23,5 +23,4 @@ aio = ["futures", "redis/aio"]
 
 [dev-dependencies]
 redis = { version = "0.23.0", path = "../redis", features = ["aio", "tokio-comp"] }
-tokio = { version = "1", features = ["rt", "macros", "rt-multi-thread"] }
-
+tokio = { version = "1", features = ["rt", "macros", "rt-multi-thread", "time"] }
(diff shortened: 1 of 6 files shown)


nihohit:
I see a 5-10% reduction in TPS, when benchmarking using https://github.com/nihohit/redis_benchmark.

nihohit:
> Any chance that may be mitigated by https://github.com/redis-rs/redis-rs/pull/877?

for this I benchmarked direct connections, not cluster connections, so the PRs are orthogonal.

> I wonder if it's worth considering making this an optional feature

It would introduce differences in implementation between the implementation with and without the feature. I'm not very keen on that. 


I'll try to put some more time for this task, and see if the perf hit can be reduced / if the benchmark's results were wrong.

jaymell:
Also applicable to the other timeout-related PRs: Do we need the `futures-time` crate? I'm not familiar with it, but is it necessary just to handle timeouts?

nihohit:
It gives us a runtime-agnostic timeout function. Without it we'll need to use feature flags to disambiguate different functions for each runtime.

nihohit:
I pinged @yoshuawuyts on twitter, maybe he'll be able to suggest a way to mitigate this performance hit.

jaymell:
Can we close this one and handle through #834 instead?

PR #877, opened 2023-06-27
Improve async cluster client performance.

nihohit:
This change reverts most of the performance degradation introduced recently. Using [this benchmark](https://github.com/nihohit/redis_benchmark) I measured ~320K TPS on version 0.23, ~160K TPS on main, and ~295K TPS after this change.

The cause of the degradation seems to be that existing connections, which were received on the happy path, were wrapped 4 times - 2 boxes, 2 shares. Maybe more, if the compiler had to move the connection to heap for the boxed/shared generator. This change removes all of these unnecessary allocations.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -430,7 +430,6 @@ where
                             &addr,
                             connections.remove(&addr),
                             &inner.cluster_params,
-                            true,
                         )
                         .await;
                         if let Ok(conn) = conn {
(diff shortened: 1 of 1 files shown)


nihohit:
I also checked, and `ConnectionMap` can be changed to hold connections instead of connection futures as values. I don't see a meaningful perf change there, so didn't add it to this PR.

jaymell:
This looks good -- thanks for cleaning this up!

PR #879, opened 2023-06-29
Extended set command to support SetOptions

### LINKED ISSUE #876: Extending SET command to support more options
I'm using your package in my learning project and found that `SET` command lacks support for options that redis supports, like NX/PX etc.

These extra options can be found here: https://redis.io/commands/set/

Of course I was able to make it work with just building the command using `cmd` and `arg`, but it's quite cumbersome:
```rust
let redis_result: Result<Value, redis::RedisError> = redis::cmd("SET")
     .arg(ket)
     .arg(value)
     .arg("NX")
     .arg("GET")
     .arg("EX")
     .arg(20)
     .query(&mut connection);
```

Maybe instead of that it would be better to extend `set` command to support some sort of options argument, similar to what node.js [redis](https://github.com/redis/node-redis/tree/master#redis-commands) packages does:

```js
await client.set('key', 'value', {
  EX: 10,
  NX: true
});
```

What do you think?

RokasVaitkevicius:
Added new `set` command parameter: `options` which supports `SET` options described in Redis documentation: https://redis.io/commands/set/ 

Closes #876

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -2082,7 +2082,7 @@ impl ToRedisArgs for Direction {
 /// ) -> RedisResult<Vec<usize>> {
 ///     let opts = SetOptions::default()
 ///         .conditional_set(ExistenceCheck::NX)
-///         .fetch(true)
+///         .get(true)
 ///         .with_expiration(SetExpiry::EX(60));
 ///     con.set_options(key, value, opts)
 /// }
@@ -2102,8 +2102,8 @@ impl SetOptions {
     }
 
     /// Set the GET option for the SET command
-    pub fn fetch(mut self, fetch: bool) -> Self {
-        self.get = fetch;
+    pub fn get(mut self, get: bool) -> Self {
+        self.get = get;
         self
     }
 
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -3,7 +3,7 @@
 use crate::cmd::{cmd, Cmd, Iter};
 use crate::connection::{Connection, ConnectionLike, Msg};
 use crate::pipeline::Pipeline;
-use crate::types::{FromRedisValue, NumericBehavior, RedisResult, ToRedisArgs, RedisWrite, Expiry};
+use crate::types::{FromRedisValue, NumericBehavior, RedisResult, ToRedisArgs, RedisWrite, Expiry, SetExpiry, ExistenceCheck};
 
 #[macro_use]
 mod macros;
@@ -83,8 +83,8 @@ implement_commands! {
     }
 
     /// Set the string value of a key.
-    fn set<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V) {
-        cmd("SET").arg(key).arg(value)
+    fn set<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V, options: SetOptions) {
+        cmd("SET").arg(key).arg(value).arg(options)
     }
 
     /// Sets multiple keys to their values.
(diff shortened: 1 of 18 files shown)


RokasVaitkevicius:
I opened this to seek opinions and guidance on a couple of matters.

1. I'm unsure how to maintain backward compatibility for the set function because my implementation requires users to add a new function parameter to their set calls. I see two possible approaches here: either leave it as incompatible or add a new set function with options. Alternatively, there might be a way to ensure compatibility that I've overlooked, because I'm relatively new to Rust. If so, let me know.
2. I've introduced a new enum called SetExpiry, and it shares many values with the existing Expiry enum. Is it acceptable to keep them both, or should I extract the shared values into a separate enum? If extraction is recommended, how would you suggest accomplishing it?

Once these questions are answered, I'll make the necessary changes, add tests, and update the documentation. I didn't want to dive too deep into the implementation until these uncertainties were addressed. Thank you!

nihohit:
1. IMO we should add a new function (`set_ext`/`set_with_options`?) instead of breaking backwards compatibility.
2. IMO the enum should be separate, seeing as they have different values.

jaymell:
Yep, please no breakage here. `set_options` would probably be the closest we have to a standard name, given that the stream commands also use this naming approach.

RokasVaitkevicius:
Okay, I'll do that then.

RokasVaitkevicius:
@jaymell Should be good to go. Let me know what you think.

nihohit:
not sure I like the `upsert` term - AFAIK it's from RDBMS, not the NoSQL world. maybe conditional_update, conditional_set?

RokasVaitkevicius:
Good call. I've ranamed it to `conditional_set`.

I've also renamed `fetch` to `get`, it's a bit more straightforward that way I think and it says what it does exactly gets :D

redis/src/commands/mod.rs
```diff
@@ -83,8 +83,8 @@ implement_commands! {
     }
 
     /// Set the string value of a key.
-    fn set<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V) {
-        cmd("SET").arg(key).arg(value)
+    fn set<K: ToRedisArgs, V: ToRedisArgs>(key: K, value: V, options: SetOptions) {
```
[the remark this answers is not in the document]
RokasVaitkevicius: Yeah, I've reverted my changes and based on @jaymell recommendations added a new method called `set_options`

RokasVaitkevicius:
Sorry, forgot to run fmt, now it should pass the checks

redis/tests/test_basic.rs
```diff
@@ -1184,3 +1184,30 @@ fn test_multi_generics() {
...
+fn test_options() {
+    let empty = SetOptions::default();
+    assert_eq!(ToRedisArgs::to_redis_args(&empty).len(), 0);
+
+    let opts = SetOptions::default()
+        .conditional_set(ExistenceCheck::NX)
+        .get(true)
+        .with_expiration(SetExpiry::PX(1000));
+
+    assert_args!(&opts, "NX", "GET", "PX", "1000");
```
jaymell, line 1212: Any reason not to be a bit redundant here and include all variants in the test?

jaymell:
Thanks, this looks great! I'll leave it up to you, but might not be a bad idea to add a simple integration test of the command as well, just to cover all our bases.

redis/tests/test_basic.rs
RokasVaitkevicius: Okay, will do

RokasVaitkevicius:
@jaymell I've added more tests for set options struct. 

Regarding integrations tests I have this one: https://github.com/RokasVaitkevicius/redis-rs/blob/50669d679e74656d6111f54588f548ed79f1849c/redis/tests/test_basic.rs#L1189

Is it sufficient or you want me to add more tests like this?

jaymell (approved):

jaymell:
> @jaymell I've added more tests for set options struct.
> 
> Regarding integrations tests I have this one: https://github.com/RokasVaitkevicius/redis-rs/blob/50669d679e74656d6111f54588f548ed79f1849c/redis/tests/test_basic.rs#L1189
> 
> Is it sufficient or you want me to add more tests like this?

This looks great. Please fix the outstanding lint issue and we'll get it merged :pray:

RokasVaitkevicius:
Linting issue fixed, need to add it as pre-commit hook or something :D

jaymell:
Merged -- thanks again!

PR #880, opened 2023-06-29
Benchmark CI

jaymell:


jaymell:
Still testing... stay tuned

jaymell:
Well, I am sufficiently demoralized by benchmarking thus far that I'm almost reluctant to merge this, but it's *possibly* better than not running benchmarks at all, so it may be worth letting it run on PRs for awhile to see how things look over time. 

I think it's probably a futile/unwise endeavor to try and assess pass/failure running in CI, and none of the tooling here is really designed for that anyway, so likely better for now to just have the results available to look at.

Issue #885, opened 2023-07-06
ConnectionManager: IO error behavior consideration


DoumanAsh:
I noticed that if connection gets stale for long, `ConnectionManager` tends to fail with broken pipe.
Unfortunately the way `ConnectionManager` handles IO error is rather inconvenient: instead of retrying on new connection established, it immediately return error which is of no use.

I think it would be a better approach to wait for re-connect and attempt again

The way it is right now, user has to basically wrap `ConnectionManager` and implement proper interface to avoid encountering useless errors with broken pipe

DoumanAsh:
ideally you need to distinguish between send and  recv
As you can be sure command is enqueued after send, any error during `send` should be re-tried.
But if `recv` fails, current handling is more than enough

DoumanAsh:
So currently manager uses this function to send and receive reply
https://github.com/redis-rs/redis-rs/blob/main/redis/src/aio/connection_manager.rs#L196-L215

While it would be a bit of code repetition, but it could be instead done by using individual methods and then handling would be refined by re-trying on `send` errors and immediately failing on `recv`

nihohit:
where would you reasonably get a broken pipe error on `send`?

Let's roll out the operations - 

1. ConnectionManager's req_packed_command is called
2. **a `MultiplexedConnection` is cloned from the guard**
3. if there's an io error, start reconnection and return error. 
4. `send_packed_command` is called on the multiplexed connection.
5. the command is turned into a vec, and sent to the pipeline with a oneshot response channel
6. **the command is sent through the TCP stream**
7. **once the TCP stream returns an answer (simplified, let's not go into details) the oneshot channel is called with the response**
8. the multiplexer calls the passes the response to the connection manager
9. if the response is a connection error, start reconnection and return error

The bolded lines are those in which an error might happen. of the 3 lines, only line 3 is safely retryable, because no data was sent to the server. On the other hand, it only errors if the previous reconnection attempt failed, which means that there was a disconnect attempt, and then the reconnect failed. This implies that the server is no longer available, and this should probably be communicated to the user ASAP, since the user defined the number of retries and the backoff strategy for reconnects, so you don't double the number of attempts behind the user's back.

DoumanAsh:
> where would you reasonably get a broken pipe error on send?

Attempting to write on TCP stream while receiver end is closed by remote peer

> the command is sent through the TCP stream

The command sending process is not atomic
[...]

PR #886, opened 2023-07-11
Enable keep alive on tcp connections via feature

DoumanAsh:
This enables keep alive on tcp sockets via `keep-alive` feature

~Do you want to have it configurable so that only `ConnectionManager` would enable it?
There is no harm in enabling it for short living connections as default intervals are quite long~

Note that this doesn't really fix anything, but for long living connections it will speed up failure detection

Related to #885

> package `socket2 v0.5.3` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.60.0

Unfortunate sad state of rust networking

Just for reference redis(server) default value for keep-alive is 300 seconds ([Ref](https://raw.githubusercontent.com/redis/redis/7.0/redis.conf))

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -42,6 +42,7 @@ futures-util = { version = "0.3.15", default-features = false, optional = true }
 pin-project-lite = { version = "0.2", optional = true }
 tokio-util = { version = "0.7", optional = true }
 tokio = { version = "1", features = ["rt", "net", "time"], optional = true }
+socket2 = { version = "0.4", default-features = false, optional = true }
 
 # Only needed for the connection manager
 arc-swap = { version = "1.1.0", optional = true }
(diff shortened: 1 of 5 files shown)


jaymell:
Thanks for this!

> Do you want to have it configurable so that only ConnectionManager would enable it?

Is there any reason to limit it to `ConnectionManager`? Also, please correct me if I'm wrong, but currently the sync connection would need a similar setup in order to use keep-alives, correct? If so, I think it might be reasonable to feature-gate keep-alives/the `socket2` dependency behind a new feature we could enable by default and let the whole lib use it. What do you think?

DoumanAsh:
> Is there any reason to limit it to ConnectionManager

It is unclear how user will use regular connection and whether he wants to use keep-alive.
but in general keep-alive itself has impact only on long living connections, so that's why I asked whether you'd prefer to enable it on all connections or only long living one.

> Also, please correct me if I'm wrong, but currently the sync connection would need a similar setup in order to use keep-alives, correct

Yes, in similar manner it can be enabled for sync clients. 

If you prefer to feature gate the whole keep-alive on all connections, then I'll do it for all connections

DoumanAsh:
I refactored PR to hide `socket2` behind `keep-alive` feature

jaymell:
> If you prefer to feature gate the whole keep-alive on all connections, then I'll do it for all connections

I think it sounds like a good approach!

jaymell:
Sorry for the delay, but one more request: Can we enable the feature by default?

nihohit:
not for this PR, in order not to block it, but we should probably also use this in sync connections, not only in async.

jaymell (approved):

jaymell:
@nihohit Sorry I'm not sure I follow your last comment.

PR #887, opened 2023-07-13
Update routing data for commands.

nihohit:
The choices were based on the request_policy tip from the COMMAND output.

### DIFF
--- a/redis/src/cluster_routing.rs
+++ b/redis/src/cluster_routing.rs
@@ -35,12 +35,37 @@ impl RoutingInfo {
     {
         let cmd = &r.command()?[..];
         match cmd {
-            b"FLUSHALL" | b"FLUSHDB" | b"SCRIPT" => Some(RoutingInfo::AllMasters),
-            b"ECHO" | b"CONFIG" | b"CLIENT" | b"SLOWLOG" | b"DBSIZE" | b"LASTSAVE" | b"PING"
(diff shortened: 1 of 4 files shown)


nihohit:
tasks for after this PR:
implement multi-shard commands
implement response policies

jaymell:
I think this looks good! Can you list out which commands you've removed from the routing rules? I think they are all appropriate for random routing?

nihohit:
readonly commands:
added:
	LOLWUT
	MEMORY USAGE
	OBJECT ENCODING
	OBJECT FREQ
	OBJECT IDLETIME
	OBJECT REFCOUNT
	XINFO CONSUMERS
	XINFO GROUPS
	XINFO STREAM

(the OBEJCT & XINFO commands were expanded from only OBJECT & XINFO in the old version)
removed:
CLIENT
ECHO
LASTSAVE
STRALGO (which is not a real command)

AllNodes routing - 
all removed, except for SLOWLOG & CONFIG SET
DBSIZE, PING, INFO, KEYS were moved to AllMasters list.
BGREWRITEAOF, CLIENT LIST	CLIENT LIST, ECHO, other CONFIG commands, other CLIENT commands, LASTSAVE, don't have explicit routing now.

AllMasters:
SCRIPT FLUSH, SCRIPT LOAD, have been moved to AllNodes. SCRIPT DEBUG, SCRIPT HELP now don't have explicit routing.

jaymell (approved):

PR #888, opened 2023-07-14
async cluster: Group responses by response_policy. 

nihohit:
This change implements the response_policy tip in the async cluster.
https://redis.io/docs/reference/command-tips/#response_policy

This means that the results from fan-out commands will be aggregated
differently, based on the sent command.


jaymell (approved):

jaymell:
Looks great!

PR #899, opened 2023-07-23
Allow passing routing information to cluster.

nihohit:
This PR is based over https://github.com/redis-rs/redis-rs/pull/900 .

This change allows the user to specify routing information for commands. This is especially helpful for commands that don't have an explicit request policy, such as BGSAVE, and so are sent to a random node.

### DIFF
--- a/redis/tests/test_cluster_async.rs
+++ b/redis/tests/test_cluster_async.rs
@@ -809,7 +809,7 @@ fn test_async_cluster_route_according_to_passed_argument() {
     cmd.arg("test");
     let _ = runtime.block_on(connection.send_packed_command(
         &cmd,
-        Some(RoutingInfo::MultiNode(MultipleNodeRoutingInfo::AllMasters)),
+        RoutingInfo::MultiNode(MultipleNodeRoutingInfo::AllMasters),
         None,
     ));
     {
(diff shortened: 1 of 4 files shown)


nihohit:
@jaymell rebased

redis/tests/test_cluster_async.rs
```diff
@@ -779,6 +780,58 @@ fn test_cluster_fan_out_once_even_if_primary_has_multiple_slot_ranges() {
     );
 }
 
+#[test]
```
jaymell, line 783: Can we NOT mock this test and possibly test a few different combinations of custom routes/response policies? I know it slows things down further, but I think it's better to actually assert the results of a real server request when feasible.

jaymell:
Thanks for the PR! Looking beautiful so far.

redis/tests/test_cluster_async.rs
```diff
@@ -80,6 +80,109 @@ fn test_async_cluster_basic_script() {
...
+        assert_eq!(
+            connection
+                .route_command(&redis::cmd("FLUSHALL"), routing, None)
+                .await
+                .unwrap(),
+            Value::Okay
+        );
+        let res: String = connection.get("foo").await.unwrap();
+        assert_eq!(res, "bar".to_string());
+        let res2: Option<String> = connection.get("bar").await.unwrap();
```
jaymell, line 104: For the glory of pedantry, might we assert that "bar" is `Some` prior to the FLUSHALL command?

jaymell (approved):

PR #900, opened 2023-07-23
Add support for multi-shard commands.

nihohit:
This PR is based over https://github.com/redis-rs/redis-rs/pull/888 .

This allows the async cluster client to split keyed commands so that
keys will be grouped by the slot they belong in, and sent only to the
relevant shard.

### DIFF
--- a/redis/src/cluster_routing.rs
+++ b/redis/src/cluster_routing.rs
@@ -155,12 +155,22 @@ pub(crate) fn combine_array_results(values: Vec<Value>) -> RedisResult<Value> {
     Ok(Value::Bulk(results))
 }
 
+/// Combines multiple call results in the `values` field, each assume to be an array of results,
+/// into a single array. `sorting_order` defines the order of the results in the returned array -
+/// for each array of results, `sorting_order` should contain a matching array with the indices of
+/// the results in the final array.
 pub(crate) fn combine_and_sort_array_results<'a>(
     values: Vec<Value>,
-    sorting_order: impl IntoIterator<Item = &'a Vec<usize>> + ExactSizeIterator,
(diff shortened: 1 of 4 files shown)


nihohit:
@jaymell rebased

redis/src/cluster_routing.rs
```diff
@@ -145,6 +157,83 @@ pub(crate) fn combine_array_results(values: Vec<Value>) -> RedisResult<Value> {
...
+fn get_route(is_readonly: bool, key: &[u8]) -> Route {
+    let slot = get_slot(key);
+    if is_readonly {
+        Route::new(slot, SlotAddr::Replica)
+    } else {
+        Route::new(slot, SlotAddr::Master)
+    }
+}
+
+fn multi_shard<R>(
```
jaymell, line 203: Small comment on this function please

jaymell (approved):
I think this looks good, but not gonna lie, trying to read through a few parts of this was pretty cumbersome. I'm not sure if it's possible to make things any more readable, so just asking for a couple of comments on the denser parts for the benefit of anyone who has to look through it again later.

nihohit:
@jaymell I added docs, a sample test, and slightly changed the implementation of `multi_slot` to return a 0-based indices. Hopefully that'll make the result more readable and understandable.

jaymell:
Thanks for adding the comments! As for the actual code changes, I'll leave it up to you. If you think the code was more robust prior to the latest changes, for feel free to revert them. Otherwise let's go ahead and merge!

tsmt-iris:
Hey, does this mean that pipelines using key on different shards are working with async cluster?

nihohit:
> Hey, does this mean that pipelines using key on different shards are working with async cluster?

no, it doesn't.

PR #901, opened 2023-07-23
Changed cluster creation to wait for all servers to change their state to OK

barshaul:
It is now required that the cluster creation waits for all servers to change their state to OK before it returns, rather than waiting only for the first one to change its state.

This change prevents flaky tests due to CLUSTER DOWN errors by not starting testing before all servers are ready.


nihohit (approved):
@jaymell LGTM

jaymell:
Thanks!

jaymell (approved):

PR #904, opened 2023-07-26
async transaction support

bgrieder:
Hello,

This is a proposal to support `async` transactions (because I could not figure out how to do it otherwise).

```rust
let res: Vec<String> = transaction_async!(
        mgr.clone(),
        &["key"],
        |mut mgr: ConnectionManager, mut pipeline: Pipeline| async move {
            pipeline
                .set("key", "blah")
                .ignore()
                .get("key")
                .query_async(&mut mgr)
                .await
        }
    )?;
    assert_eq!(res, vec!["blah".to_string()]);
```

I am unsure what features must be enabled to support this; the test passes with the current list, but that needs review.

Also, the way the test is written in `redis/tests/test_async_transaction.rs` is probably not what you want (It expects a local running Redis server with storage enabled)

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -56,7 +56,7 @@ r2d2 = { version = "0.8.8", optional = true }
 crc16 = { version = "0.4", optional = true }
 rand = { version = "0.8", optional = true }
 # Only needed for async_std support
-async-std = { version = "1.8.0", optional = true}
+async-std = { version = "1.8.0", optional = true }
 async-trait = { version = "0.1.24", optional = true }
 
 # Only needed for native tls
@@ -83,7 +83,19 @@ log = { version = "0.4", optional = true }
 [features]
 default = ["acl", "streams", "geospatial", "script", "keep-alive"]
 acl = []
-aio = ["bytes", "pin-project-lite", "futures-util", "futures-util/alloc", "futures-util/sink", "tokio/io-util", "tokio-util", "tokio-util/codec", "tokio/sync", "combine/tokio", "async-trait"]
+aio = [
+    "bytes",
+    "pin-project-lite",
+    "futures-util",
+    "futures-util/alloc",
+    "futures-util/sink",
+    "tokio/io-util",
+    "tokio-util",
+    "tokio-util/codec",
+    "tokio/sync",
+    "combine/tokio",
+    "async-trait",
+]
 geospatial = []
 json = ["serde", "serde/derive", "serde_json"]
 cluster = ["crc16", "rand"]
@@ -93,7 +105,11 @@ tls-rustls = ["rustls", "rustls-native-certs"]
 tls-rustls-insecure = ["tls-rustls", "rustls/dangerous_configuration"]
 tls-rustls-webpki-roots = ["tls-rustls", "webpki-roots"]
 async-std-comp = ["aio", "async-std"]
-async-std-native-tls-comp = ["async-std-comp", "async-native-tls", "tls-native-tls"]
+async-std-native-tls-comp = [
+    "async-std-comp",
+    "async-native-tls",
+    "tls-native-tls",
+]
 async-std-rustls-comp = ["async-std-comp", "futures-rustls", "tls-rustls"]
 tokio-comp = ["aio", "tokio", "tokio/net"]
 tokio-native-tls-comp = ["tokio-comp", "tls-native-tls", "tokio-native-tls"]
@@ -106,7 +122,9 @@ sentinel = ["rand"]
 
 # Deprecated features
 tls = ["tls-native-tls"] # use "tls-native-tls" instead
-async-std-tls-comp = ["async-std-native-tls-comp"] # use "async-std-native-tls-comp" instead
+async-std-tls-comp = [
+    "async-std-native-tls-comp",
+] # use "async-std-native-tls-comp" instead
 
 [dev-dependencies]
 rand = "0.8"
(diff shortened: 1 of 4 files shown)


nihohit:
Hi @bgrieder, thanks for the suggestion!
transactions can be used with the `pipeline` object, after calling the `atomic` function. You can find [an example here](https://github.com/redis-rs/redis-rs/blob/a6b4c74d7bfe99a86f31a44fcf8d29d3bd9064fd/redis/tests/test_async.rs#L79). Please tell me if that answers your requirements.

If you feel that the documentation didn't guide you well, please say where you checked, so that I could fix the documentation (or better yet, please open a PR with documentation improvement 😄  ).

bgrieder:
Hello. Thanks but correct me if I am wrong, the example you gave does not leverage the WATCH keys capability with automatic retry (in cas optimistic locking fails)

What I am trying to do here is to have a facility similar to that of the transaction() function that does all this for you, in an async context

redis/src/aio/async_transactions.rs
```diff
@@ -0,0 +1,127 @@
...
+    ) -> Pin<Box<dyn Future<Output = Result<Option<T>, E>> + Send>>,
+>(
+    mut mgr: ConnectionManager,
+    keys: &[K],
+    mut func: F,
+) -> Result<T, E> {
+    loop {
+        cmd("WATCH").arg(keys).query_async(&mut mgr).await?;
+        let mut p = pipe();
+        let response: Option<T> = func(mgr.clone(), p.atomic()).await?;
```
nihohit, line 25: the `?` means that UNWATCH isn't called, correct?

The tests need to ensure that this leaves the connection in a clean state even when encountering an error.

redis/src/aio/async_transactions.rs
```diff
@@ -0,0 +1,127 @@
...
+pub async fn transaction_async<
+    T,
+    E: From<RedisError>,
+    K: ToRedisArgs,
+    F: FnMut(
+        ConnectionManager,
+        &mut Pipeline,
+    ) -> Pin<Box<dyn Future<Output = Result<Option<T>, E>> + Send>>,
+>(
+    mut mgr: ConnectionManager,
```
nihohit, line 18: why only `ConnectionManager` and not `impl aio::ConnectionLike`?

redis/Cargo.toml
```diff
@@ -56,7 +56,7 @@ r2d2 = { version = "0.8.8", optional = true }
 crc16 = { version = "0.4", optional = true }
 rand = { version = "0.8", optional = true }
 # Only needed for async_std support
-async-std = { version = "1.8.0", optional = true}
+async-std = { version = "1.8.0", optional = true }
```
nihohit, line 59: please remove all of these auto-format changes

redis/src/aio/async_transactions.rs
```diff
@@ -0,0 +1,127 @@
...
+    keys: &[K],
+    mut func: F,
+) -> Result<T, E> {
+    loop {
+        cmd("WATCH").arg(keys).query_async(&mut mgr).await?;
+        let mut p = pipe();
+        let response: Option<T> = func(mgr.clone(), p.atomic()).await?;
+        match response {
+            None => {
+                println!("retrying transaction");
```
nihohit, line 28: use trace instead

redis/src/aio/async_transactions.rs
```diff
@@ -0,0 +1,127 @@
...
+                return Ok(response);
+            }
+        }
+    }
+}
+
+/// Asynchronous transaction macro for Redis operations.
+///
+/// This macro encapsulates the boilerplate required to establish a Redis transaction
+/// and apply a function to it. What it
```
nihohit, line 44: minor: weird linebreak

redis/tests/test_async_transaction.rs
```diff
@@ -0,0 +1,63 @@
+use redis::{
+    aio::ConnectionManager, transaction_async, AsyncCommands, Client, Pipeline, RedisError,
+    RedisResult,
+};
+
+const REDIS_URL: &str = "redis://localhost:6379";
+
+#[tokio::test]
+pub async fn test_async_transaction() -> RedisResult<()> {
```
nihohit, line 9: please separate into several tests with informative names.

please also test the error paths - specifically, both EXECABORTERROR and errors when running the transaction's commands.

please verify at the end of every test that UNWATCH was called.

redis/src/aio/async_transactions.rs
```diff
@@ -0,0 +1,127 @@
...
+/// from querying the pipeline with the connection.
+///
+/// The end result of the transaction is then available as the return
+/// value from the function call.
+///
+/// # Parameters
+///
+/// - `$mgr`: A cloned connection manager for Redis.
+/// - `$key`: A key (or array of keys) for the Redis operation.
+/// - `$func`: Either a path to a function or a lambda function. This function
```
nihohit, line 63: clarify that this function might be executed more than once, and that should be taken into account if it contains side effects.

redis/src/aio/async_transactions.rs
```diff
@@ -0,0 +1,127 @@
+use std::pin::Pin;
```
nihohit, line 1: @jaymell do you want this in a separate file, or in `pipeline.rs`?

nihohit (commented):
first round.

paramako:
any update on this?

HatemMn:
@bgrieder @paramako and all interested parties, please refer to https://github.com/redis-rs/redis-rs/pull/1485

Issue #890, opened 2023-07-17
Serializing large values via RedisWrite (aka making RedisWrite compatible with std::io::Write)


swwu:
We have a somewhat unusual use case in that we need to serialize large-ish structs—in the kb-Mb range—to redis. To generate these serializations we use many different libraries, including serde, most of which support writing to a `std::io::Write`. Many of these implementations leverage the fact that `std::io::Write`'s `write` method can be called multiple times to successively append chunks of the final object.

Unfortunately, it seems like the `RedisWrite` trait (and consequently, the `ToRedisArgs` trait) doesn't really support doing this without first performing a copy, as each call to `RedisWrite::write_arg` writes a _new_ argument rather than appending to an existing one. The only way to get around this is to first write the serialization to a buffer, then call `write_arg` on that buffer—but since our serializations are rather large, and we do a lot of them, this ends up having significant performance consequences.

I have some proposals for changing the interface to allow doing this without needing this extra copy. I'm happy to work on either of these, if people think they're worth adding to the API.

### Proposal 1
Add methods `start_buffered_arg`, `write_buffered_arg`, `finish_buffered_arg` to `RedisWrite`. Their implementation would be simple (method names are, obviously, not final):
```rust
impl RedisWrite for Cmd {
    // ...
    fn start_buffered_arg(&mut self) { }
    fn write_buffered_arg(&mut self, arg: &[u8]) { self.data.extend_from_slice(arg); }
    fn finish_buffered_arg(&mut self) { self.args.push(Arg::Simple(self.data.len())); }
}
impl RedisWrite for Vec<Vec<u8>> {
    // ..
    fn start_buffered_arg(&mut self) { self.push(vec![]); }
    fn write_buffered_arg(&mut self, arg: &[u8]) { self.last().unwrap().extend_from_slice(arg); }
    fn finish_buffered_arg(&mut self) { }
}
```
(This could also be a slightly more complex interface where `start_` returned a guard, and the other methods were only callable on that guard.)
This would allow implementing `ToRedisArgs` for values with complex "chunked" serializations via interacting with the existing passed-in `RedisWrite` instance (and would also directly allow implementing a fairly-trivial `std::io::Write` adapter for `RedisWrite` instances).

### Proposal 2
[...]

PR #905, opened 2023-07-26
Add buffered write methods to RedisWrite

swwu:
As discussed in #890.

The goal of this is to allow compatibility with `std::io::Write`-based serializers in `ToRedisArgs` implementations, and more generally to allow serializers that write to a buffer in chunks to be able to write directly to `redis::Cmd`'s internal buffer, instead of having to write it to an external buffer, then copy that into `redis::Cmd`'s internal buffer.

This does not affect `redis::Cmd`'s own serialization behavior, which will still copy that internal buffer once. However, one copy is better than two for this use case.

I also have a guard implementation for `std::io::Write` here https://gist.github.com/swwu/f3e21a5c939e06c20f445735ee75d358 that I'd be happy to throw in, but it's a bit more opinionated and I'm not completely clear on where it would go.

I'm not certain about the method names, and am happy to change them if people have better suggestions.

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -279,12 +279,25 @@ impl RedisWrite for Cmd {
         self.args.push(Arg::Simple(self.data.len()));
     }
 
-    fn start_buffered_arg(&mut self) {}
-    fn write_buffered_arg(&mut self, arg: &[u8]) {
-        self.data.extend_from_slice(arg);
-    }
-    fn finish_buffered_arg(&mut self) {
-        self.args.push(Arg::Simple(self.data.len()));
+    fn writer_for_next_arg(&mut self) -> impl std::io::Write + '_ {
+        struct CmdBufferedArgGuard<'a>(&'a mut Cmd);
+        impl Drop for CmdBufferedArgGuard<'_> {
+            fn drop(&mut self) {
+                self.0.args.push(Arg::Simple(self.0.data.len()));
+            }
+        }
+        impl std::io::Write for CmdBufferedArgGuard<'_> {
+            fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+                self.0.data.extend_from_slice(buf);
+                Ok(buf.len())
+            }
+
+            fn flush(&mut self) -> std::io::Result<()> {
+                Ok(())
+            }
+        }
+
+        CmdBufferedArgGuard(self)
     }
 }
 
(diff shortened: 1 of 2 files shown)


swwu:
woops, drafting this for now; forgot tests

swwu:
Actually, I can't seem to find where tests for the `RedisWrite` interface live; there don't seem to be any calls at all to `RedisWrite::write_arg` in the tests directory.

@nihohit could you point me in the right direction here?

nihohit:
I don't think that there are tests specifically for `RedisWrite`. Most of the explicit `Cmd` tests are in cmd.rs, and, well, almost every other test in the repo implicitly tests this. So, I think you can add the tests for this in cmd.rs. Please add tests that verify that Cmd is still in a serializable state even mid-buffering. 

Seeing as I hope that #829 will be merged in the coming months, I would like to hear how you think these changes should interface.

swwu:
Hi @nihohit ! Finally got some spare cycles to come around to this.

I actually changed the implementation slightly as per your comments about serializing mid-stream; there's now only one new method on `RedisWrite`: 
```rust
fn writer_for_next_arg(&mut self) -> impl std::io::Write + '_;
```
This appends a new, empty arg, then returns an opaque writer to it. This writer holds a mutable reference to the underlying `Cmd`, so it should no longer be possible at all to serialize the command until the writer is closed.

I added a test that checks that writing to this writer produces the same serialized result as writing that data in one shot using `RedisWrite::write_arg`; happy to add any other tests you think should be necessary here.

Also, left the second commit as separate for now, but happy to squash them down to one if that's preferred.

Going to mark this as reviewable since I finally got that test in...

redis/src/cmd.rs
```diff
@@ -278,6 +278,27 @@ impl RedisWrite for Cmd {
         write!(self.data, "{arg}").unwrap();
         self.args.push(Arg::Simple(self.data.len()));
     }
+
+    fn writer_for_next_arg(&mut self) -> impl std::io::Write + '_ {
+        struct CmdBufferedArgGuard<'a>(&'a mut Cmd);
+        impl Drop for CmdBufferedArgGuard<'_> {
+            fn drop(&mut self) {
+                self.0.args.push(Arg::Simple(self.0.data.len()));
```
nihohit, line 286: I think that this design is fine, but I think it will break if someone calls `writer_for_next_arg` and then drops the writer without writing to it.
I'd prefer a design that only adds an arg if data was written to the writer - it can be as simple as saving the original length of the `args` field and checking that the current length is larger.

redis/src/cmd.rs
```diff
@@ -646,11 +667,78 @@ pub fn pipe() -> Pipeline {
...
     use super::Cmd;
 
     #[test]
+    fn test_cmd_writer_for_next_arg() {
+        use crate::RedisWrite;
+        use std::io::Write;
+
+        // Test that a write split across multiple calls to `write` produces the
+        // same result as a single call to `write_arg`
+        {
```
nihohit, line 680: please split into multiple tests. these checks are orthogonal (as evidenced by the separate scopes), and the benefit of splitting tests is that if something breaks the first condition, the second condition will run in a separate test and will still be checked.

nihohit (commented):
Thanks for picking this up. Very good job, just a couple of small comments.

redis/src/cmd.rs
swwu: @nihohit i think it won't actually break anything, or at least not differently than is already possible. This will just write an empty byte buffer, which makes it both logically and in outcome equivalent to calling `cmd.write_arg(b"");`.

swwu: Thinking about it more, I would actually find it surprising if it _didn't_ work this way, because it's possible I'm e.g. serializing a ZST marker type into redis whose binary representation is a 0-byte sequence, and in that case only appending the arg if the buffer is non-empty would cause weird inconsistencies like:

```rust

// empty ZST
#[derive(Serialize)]
pub struct Empty;

#[derive(Serialize, Default)
pub struct NonEmpty(u64);

fn build_set_cmd<T: Serialize>(thing: &T) {
  let mut c = cmd("SET").arg("some_key");
  bincode::serialize_with(thing, c.writer_for_next_arg());
  c
}


build_set_cmd(&NonEmpty(0)).query(con); // this generates a well-formed command
build_set_cmd(&Empty).query(con); // this generates a malformed command `SET "key"` with no value

```

Overall I feel that getting an empty arg by a buffer creating a buffer you don't need is a lot easier to debug/less unexpected than this.

redis/src/cmd.rs
swwu: Fixed!

redis/src/cmd.rs
nihohit: Fair enough. If it wasn't a valid Redis key I would've argued against it, but (yet again) I'm beaten by the logic of the Redis database :)

nihohit (approved):

nihohit:
Thanks for the contribution!

PR #910, opened 2023-07-30
Sync Pub/Sub - cache received pub/sub messages.

nihohit:
When sending subscibre/unsubscribe messages to the server, the pub/sub client might receive messages that it can mistake to be the response, and not return to the next `get_message` call. This fix ensures that pub/sub messages will be queued until the next
`get_message` call, so that no message will be missed.


jaymell (approved):
Looks good!

Issue #737, opened 2022-12-07
Remove “tokio” mention from name of runtime agnostic `Client::get_tokio_connection_manager()`


LeoniePhiline:
Connection manager appears to be runtime agnostic.

LeoniePhiline:
I would prepare a PR, but I am not sure what you would prefer:

- Add a method with an agnostic name (`Client::get_connection_manager()`), which does the same? (Deprecating `Client::get_tokio_connection_manager()`)
- Or straight-out renaming `Client::get_tokio_connection_manager()` to `Client::get_connection_manager()`? This would of course be a breaking change.

djc:
Probably makes sense to add the agnostic name while deprecating the tokio name?

nihohit:
https://github.com/redis-rs/redis-rs/pull/913

PR #913, opened 2023-07-31
Deprecate functions that erroneously use `tokio` in their name.

nihohit:
as per the request here:
https://github.com/redis-rs/redis-rs/issues/737

### DIFF
--- a/redis/src/client.rs
+++ b/redis/src/client.rs
@@ -208,8 +208,31 @@ impl Client {
     /// [multiplexed-connection]: aio/struct.MultiplexedConnection.html
     #[cfg(feature = "connection-manager")]
     #[cfg_attr(docsrs, doc(cfg(feature = "connection-manager")))]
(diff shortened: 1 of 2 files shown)


redis/src/client.rs
```diff
@@ -208,6 +208,7 @@ impl Client {
     /// [multiplexed-connection]: aio/struct.MultiplexedConnection.html
     #[cfg(feature = "connection-manager")]
     #[cfg_attr(docsrs, doc(cfg(feature = "connection-manager")))]
+    #[deprecated(note = "use get_connection_manager_with_backoff instead")]
```
jaymell, line 211: Can we continue to have a non-`with_backoff` method that supplies the defaults (i.e., that just removes `tokio` from the name of the current method)?

nihohit: done

jaymell (approved):

redis/src/client.rs
jaymell: Awesome, thanks!

PR #914, opened 2023-07-31
Split async pubsub connection into SplitSink

dwfait:
This is a rebase and some tweaks to: https://github.com/redis-rs/redis-rs/pull/633

To be able to subscribe or unsubscribe to channels in redis pub sub, we need to be able to split the connection into stream / sink, so that we can own each half of the connection to be able to send a subscription message on the sink while the stream is waiting for messages.

This means we must remove the on_message() functionality and return a tuple for sink / stream from the into_pubsub() function on the async connection.

This will fix issues such as: https://github.com/redis-rs/redis-rs/issues/852

Changes to the original pull request: 

- rebased with latest main branch
- some changes to some of the code in reference to using tokio's decoder to split the stream
- Changed the MONITOR command stream to return a stream of Values instead of pubsub Msgs
- Added an example for usage of the async monitor command
- Left out the parser changes as they did not seem to be relevant to changes needed

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -173,6 +173,10 @@ name = "async-pub-sub"
 required-features = ["aio"]
 
 [[example]]
+name = "async-monitor"
+required-features = ["aio"]
+
+[[example]]
 name = "async-scan"
 required-features = ["aio"]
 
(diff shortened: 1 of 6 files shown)


dwfait:
Please see https://github.com/redis-rs/redis-rs/pull/633

Issue #922, opened 2023-08-08
Redis sentinel 'could not find `aio` in the crate root' error


brocaar:
If enabling the `sentinel` feature without `aoi` enabled, building the code will fail with the following error:

```
$ cargo check
    Checking redis v0.23.1
error[E0432]: unresolved import `crate::aio`
   --> /home/brocaar/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-0.23.1/src/sentinel.rs:111:5
    |
111 |     aio::MultiplexedConnection as AsyncConnection, connection::ConnectionInfo, types::RedisResult,
    |     ^^^ could not find `aio` in the crate root

For more information about this error, try `rustc --explain E0432`.
error: could not compile `redis` due to previous error
```

I believe this import should be behind a `#[cfg(feature = "aio")]`:
https://github.com/redis-rs/redis-rs/blob/main/redis/src/sentinel.rs#L111

/cc @felipou

jaymell:
Thanks for the report. Want to raise a PR for this?

PR #923, opened 2023-08-08
Fix AsyncConnection usage in sentinel if aio feature is not used.

brocaar:
Hello @jaymell I believe this should fix the issue with the `aio` import within `sentinel` (#922).


nihohit:
@brocaar how can the issue be replicated? I tried running `cargo build --features sentinel` and it completed succesfully.

felipou:
I replicated it by using the redis crate in another crate, using only the sentinel feature. I don't know why the build command is not failing though.

nihohit:
Well, it's a real shame if we can't add to the CI some way of checking whether dependencies are broken in this way.
@jaymell, any ideas?

jaymell:
@nihohit I've seen errors like this get masked previously by the `redis-test` crate. If you explicitly build only the `redis` crate, you will see the issue, i.e. with `cargo build -p redis --features sentinel`.

It looks like there is a tool tokio is using called [cargo-hack](https://github.com/taiki-e/cargo-hack) that can test features individually, which we can probably wire up easily in CI to do a `check` per dependency. Let me see if I can throw something together.

jaymell (approved):

jaymell:
Thanks for the quick turnaround on the PR. @brocaar if you rebase your PR, the checks should all pass.

brocaar:
Thanks for the review. @jaymell I have just pushed my rebased changes.

PR #929, opened 2023-08-16
merge main into resp3 and update redis 6.2.4 to 6.2.13

altanozlu:


svix-james:
Alright, my git-fu is not the strongest, but do we want this to be squashed? It would be nice to see the full commit history here, as well, as subsequent changes made to resolve conflicts, etc.

Also, if you don't mind, I think the CI changes would be better put in a separate PR to main?

svix-james:
Also, I'm not sure. Do we ultimately want to rebase this when we merge it to main? @nihohit Any thoughts there?

altanozlu:
> Also, I'm not sure. Do we ultimately want to rebase this when we merge it to main? @nihohit Any thoughts there?

I tried to preserve to full commit history but couldn't be able to show changes to solve conflicts they became one commit . Do you mean by "Do we ultimately want to rebase this when we merge it to main?" we should rebase resp3 branch before merging into main ?

nihohit:
On the one hand, I don't think this is critical. This branch won't contain a lot of changes - 
https://github.com/redis-rs/redis-rs/pull/898
https://github.com/redis-rs/redis-rs/pull/813
https://github.com/redis-rs/redis-rs/pull/779
and maybe one more?

So the chance for significant drift isn't very large.

On the other hand, in the general case of long living branches, I think that periodically aligning long-living branches with main is wise - it reduces the chance for drift.
IMO the correct thing to do in the general case isn't to merge the branches, but to rebase the long living branch over main. This way once the branch is merged into main, there's no merge commit to add to the process, and if someone wants to examine the branch as a whole, it's possible.
Sometimes some commits in the branch are contain changes that later commits remove, in which case a rebase is more problematic. I don't think this is the situation right now. 

Regardless of whether we use rebase or merge, I don't trust git to record the changes needed in order align two distant branches, so it requires manual recording of all fixes needed in order to align the branches. Were any such fixes needed here?

altanozlu:
> 

There were some conflicts but they weren't too hard to fix and I think rebase is more okay.

altanozlu:
@svix-james @nihohit I would like to merge everything in one go, can I do it ?

nihohit:
@altanozlu @jaymell 
I think we should merge this to main after the 0.25 release, and after https://github.com/redis-rs/redis-rs/pull/779 is merged to the resp3 branch. IMO there's no need to merge main into resp3 beforehand - it would just do double work.
What do you think?

altanozlu:
Sure it can wait, I was going to work on resp3 branch and my concern is it's probably diverged a lot with main.

nihohit:
opened a draft PR with the merge
https://github.com/redis-rs/redis-rs/pull/1058

Issue #931, opened 2023-08-19
 Determine key type


bruaba:
/// Determine key type.
    fn type<K: ToRedisArgs>(key: K) {
        cmd("TYPE").arg(key)
    }

add get type of a key to redis async connection

jaymell:
Wanna open a PR for this? :pray:

bruaba:
OK, no worries.

jaymell:
Fixed by #933

PR #933, opened 2023-08-21
Determine the type of a key

bruaba:


bruaba:
add get type of a key to redis async connection
follow https://github.com/redis-rs/redis-rs/issues/931

jaymell:
Thanks for the PR! 

I personally think I prefer `key_type` instead of `type_key`. What do you think? Could we also add a small test -- in `test_basic.rs` should be fine.

bruaba:
@jaymell 
"I personally think I prefer key_type instead of type_key. What do you think?"
yes you are right it sounds better in English, I was thinking in French hence the type key.

"Could we also add a small test -- in test_basic.rs should be fine."
I added some tests in the test_basic.rs file, hope they will be fine for you.

jaymell (approved):

robjtede (approved):

jaymell:
@nihohit LGTM

PR #934, opened 2023-08-25
Address clippy lints

robjtede:


jaymell:
Can we instead just fix the errors as suggested by Clippy?

robjtede:
Two of them were fixable, the other 3 are false positives from Clippy; they cannot be fixed in the naïve way since their initial values are `Err(_)`.

redis/src/aio/connection.rs
```diff
@@ -386,6 +386,7 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
     Ok(match connection_info.addr {
         ConnectionAddr::Tcp(ref host, port) => {
             let socket_addrs = get_socket_addrs(host, port).await?;
+            #[allow(clippy::manual_try_fold)]
```
jaymell, line 389: I think the two instances in this file should probably be refactored to not use a fold, honestly. I don't think there's any reason to evaluate all the potential addresses, so we can probably just do something like this instead:

```
future::select_ok(
    socket_addrs.into_iter().map(|socket_addr| {
        Box::pin(async move { <T>::connect_tcp(socket_addr).await })
    })).await?.0
```

Wanna try it?

redis/tests/test_cluster_async.rs
```diff
@@ -175,13 +174,10 @@ async fn test_failover(env: &TestClusterContext, requests: i32, value: i32) {
                         .iter_mut()
                         .map(do_failover)
                         .collect::<stream::FuturesUnordered<_>>()
-                        .fold(
-                            Err(anyhow::anyhow!("None")),
-                            |acc: Result<(), _>, result: Result<(), _>| async move {
-                                acc.or(result)
-                            },
-                        )
+                        .filter_map(|map| future::ready(map.ok()))
```
jaymell, line 177: I'm not sure this change is appropriate. If I'm understanding this correctly, it changes the behavior a bit, doesn't it? If I'm not mistaken, with your change, the first success will cause early termination, whereas the original goal of the test was to issue the command on all the nodes. Personally, I'm fine w/ just leaving the override in place here as you had previously.

jaymell:
Issue fixed by #937

PR #939, opened 2023-09-01
remove unnecessary heap allocation

thechampagne:


redis/src/types.rs
```diff
@@ -1310,7 +1310,7 @@ impl FromRedisValue for bool {
 impl FromRedisValue for CString {
     fn from_redis_value(v: &Value) -> RedisResult<CString> {
         match *v {
-            Value::Data(ref bytes) => Ok(CString::new(bytes.clone())?),
+            Value::Data(ref bytes) => Ok(CString::new(bytes.as_slice())?),
```
jaymell, line 1313: This is also ultimately a clone, though, isn't it?

nihohit: `Vec::as_slice` doesn't allocate, but `CString::new` does allocate.
IMO this looks legit.

jaymell (approved):

jaymell:
Thanks for the PR!

PR #940, opened 2023-09-01
Release redis 0.23.3 / redis-test 0.2.3

jaymell:


redis/CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
...
+* Async cluster: Group responses by response_policy. ([#888](https://github.com/redis-rs/redis-rs/pull/888))
+
+#### Fixes
+* Remove unnecessary heap allocation ([#939](https://github.com/redis-rs/redis-rs/pull/939) @thechampagne)
+* Sentinel tests: Ensure no ports are used twice ([#915](https://github.com/redis-rs/redis-rs/pull/915))
+* Fix lint issues ([#937](https://github.com/redis-rs/redis-rs/pull/937))
+* Fix JSON serialization error test ([#928](https://github.com/redis-rs/redis-rs/pull/928))
+* Remove unused dependencies ([#916](https://github.com/redis-rs/redis-rs/pull/916))
+
+<a name="0.23.2"></a>
```
nihohit, line 18: minor: 
what is this line? I didn't add it for 0.23.2 (which is why the removed line is 0.23.1), and I see that some releases have this tag, and some don't.

redis-test/Cargo.toml
```diff
@@ -13,7 +13,7 @@ rust-version = "1.63"
 bench = false
 
 [dependencies]
-redis = { version = "0.23.2", path = "../redis" }
+redis = { version = "0.23.3", path = "../redis" }
```
nihohit, line 16: minor: we probably don't need the version tag here and below, because we're fetching the dependency from a path, so cargo has only a single version to choose from.

nihohit (approved):
LGTM.

The comments are not meant as things to fix for this PR, just a couple of things for us to think about.

PureWhiteWu:
Hi, is it possible to bring this PR https://github.com/redis-rs/redis-rs/pull/941 together in the release?
`TCP_NODELAY` is quite important for us.
Thanks!

redis-test/Cargo.toml
jaymell: Yep that is correct. Let's get it next time.

redis/CHANGELOG.md
jaymell: Honestly, no idea. I just mindlessly copy-pasted this in previous releases b/c it has been there off and on for a long time, but I don't think it does anything. We can simply remove in the future.

PR #941, opened 2023-09-01
feat(redis): set TCP_NODELAY to reduce latency

### LINKED ISSUE #354: Can this module set TCP_NODELAY option to TCP socket?
I think this seems it cannot.
This is important for some processes that needs low latency.

PureWhiteWu:
`TCP_NODELAY` is quite important for redis client because this greatly affects latency.
This PR adds this option by default.
This PR also implements writev for tokio tcp connection (maybe for future optimization).
Fixes #354 .

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -103,6 +103,7 @@ streams = []
 cluster-async = ["cluster", "futures", "futures-util", "log"]
 keep-alive = ["socket2"]
 sentinel = ["rand"]
+tcp_nodelay = []
 
 # Deprecated features
 tls = ["tls-native-tls"] # use "tls-native-tls" instead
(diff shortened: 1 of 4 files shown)


nihohit:
Hi @PureWhiteWu, thanks for the contribution!
I agree that there should be some way of disabling Nagle's algorithm for low-latency transmissions, but doing so by default without giving users a choice isn't the right solution for the library. After all, there's a good reason Nagle is set on by default :)

Please change the PR so that users will opt-in to this change, either by using a feature flag, or by using a runtime configuration (IMO the second is the better option). see https://github.com/redis-rs/redis-rs/pull/886 for an example for the first option.

> This PR also implements writev for tokio tcp connection (maybe for future optimization).

I'm sorry, but I don't understand this change. Is the function used somewhere? tested anywhere? If not, then IMO we shouldn't add code that we can't know that works correctly. Please consider adding that as a separate, more fully-fleshed PR.

PureWhiteWu:
@nihohit Hi, thanks very much for your suggestions!
I've added a feature flag to opt-in the tcp_nodelay behaviour. Also, I've removed the support for `writev`, maybe I will add it later in another pr.

nihohit (approved):
LGTM @jaymell

PureWhiteWu:
Hi, kindly ping if you missed this PR.
This is really important for us, thanks!

jaymell:
Thanks for the PR!

PR #942, opened 2023-09-01
Fix parser infinite loop

enzious:
- Issue: #487
- Problem
  - If the ValueCodec parses an invalid line, it will not consume the line and continuously parse the same invalid line in an infinite loop, in some cases consuming the resources of the thread
- Fix
  - Consume the invalid line and return an error
  - Issue referenced avoided by: #633

### DIFF
--- a/redis/src/parser.rs
+++ b/redis/src/parser.rs
@@ -154,13 +154,23 @@ where
                     })
                 };
 
+                let unknown = || {
+                    line().map(|line| {
+                        RedisError::from((
+                            ErrorKind::ResponseError,
+                            "Unknown response",
+                            format!("Unknown line: {}", line),
+                        ))
+                    })
+                };
+
                 combine::dispatch!(b;
                     b'+' => status().map(Ok),
                     b':' => int().map(|i| Ok(Value::Int(i))),
                     b'$' => data().map(Ok),
                     b'*' => bulk(),
                     b'-' => error().map(Err),
-                    b => combine::unexpected_any(combine::error::Token(b))
+                    _ => unknown().map(Err)
                 )
             })
     ))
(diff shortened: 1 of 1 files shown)


nihohit:
hi, 
I ran the test without the change to the parser, and it doesn't go into an infinite loop, just returns `Err(..)` instead of `Ok(Some(Err(..)))`. Can you please add an explanation to the PR, and maybe clarify the test?

enzious:
The issue is more how the Err(_) is handled. Looking at it now, this could be better handled by abandoning or recovering:
https://github.com/enzious/redis-rs/blob/feature/wip/split-stream-pubsub/redis/src/aio/connection.rs#L320

On that line, the error case just uses `break`, and since it just doesn't consume it will enter the loop there.

nihohit:
In which situations in #633 did you get an invalid line? That's not something that should happen - the server is supposed to send parsable messages.

> this could be better handled by abandoning here

You mean, closing the stream if we receive a bad input from the server would be better handling?
And just to properly understand - this change would cause the user to receive the parse error as part of the stream, and the user could continue to receive more message on the stream?

enzious:
> In which situations in #633 did you get an invalid line? That's not something that should happen - the server is supposed to send parsable messages.
> 
> > this could be better handled by abandoning here
> 
> You mean, closing the stream if we receive a bad input from the server would be better handling? And just to properly understand - this change would cause the user to receive the parse error as part of the stream, and the user could continue to receive more message on the stream?

@nihohit I'm closing the issue as thinking about it... this is not the way to fix the issue. The issue in #487 is caused by waiting for a response when a pubsub message is received:
image
Here you can see that the channel is read as the response, leaving the line starting with "277...". That causes an error to be thrown, but the buffer is not consumed and the current code filters out errors, meaning it will throw error over and over again waiting for a valid value:
image
I can fix this with my other PR this weekend.

nihohit:
Ah, I believe that now I understand the issue. Thanks :)
Seems like the problem is that we assume that we know the type of messages that we'll receive - either "OK" or a pubsub message, and so we handle messages without knowing for sure what they are. My guess is that the solution is instead to demux them - have a way to pass non-pubsub messages back to the sink, and pubsub messages to the stream.

PR #945, opened 2023-09-04
Call CLIENT SETINFO on new connections.

nihohit:
CLIENT SETINFO is a new command to Redis 7.2.0.

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -806,6 +806,23 @@ pub fn connect(
     setup_connection(con, &connection_info.redis)
 }
 
+pub(crate) fn client_set_info_pipeline() -> Pipeline {
+    let mut pipeline = crate::pipe();
+    pipeline
+        .cmd("CLIENT")
+        .arg("SETINFO")
+        .arg("LIB-NAME")
+        .arg("redis-rs")
+        .ignore();
+    pipeline
+        .cmd("CLIENT")
+        .arg("SETINFO")
+        .arg("LIB-VER")
+        .arg(env!("CARGO_PKG_VERSION"))
+        .ignore();
+    pipeline
+}
+
 fn setup_connection(
     con: ActualConnection,
     connection_info: &RedisConnectionInfo,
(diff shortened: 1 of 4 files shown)


redis/src/connection.rs
```diff
@@ -806,6 +806,23 @@ pub fn connect(
     setup_connection(con, &connection_info.redis)
 }
 
+pub(crate) fn client_set_info_pipeline() -> Pipeline {
```
chayim, line 809: In other clients, we're:

1. Making this optional (but default on)
2. Allowing the variables used to be overrideable. 

The reason behind #2 is to allow folks building on top of a library, to embed a name within the string, for example: redis-rs(myfooclient)

nihohit: 1. ATM we have pretty thin configurations. I'm not against doing this, but it will require a small paradigm change in how we configure our connections.
2. This I don't understand. The intention behind the CLIENT SETINFO command is explicitly to save *client* data, not application data - CLIENT SETNAME is usually used for that*. 

Since anyone can all CLIENT SETINFO and overwrite the previous value, I'm not eager to complicate our existing API in order to enable this, unless there's a concrete user request for that.

*the story I heard is that historically CLIENT SETINFO was intended to be used to save client data, but was co-opted by applications.

jaymell:
What is the actual utility of adding this at this point?

nihohit:
> What is the actual utility of adding this at this point?

@chayim might be able to expand on this, but AFAIK this doesn't help directly the users of the client. This serve a more backend-centric service - it allows anyone who looks at the connected clients (by using CLIENT LIST, or some sort of backend metrics system) to know which clients are used in each connection. This might be used in analytics, understanding erroneous behavior, etc..

jaymell:
Thanks for the explanation. 

@chayim As nihohit alluded, we don't have particularly rich support for configuring the connections directly. Until that changes, what's here at the moment seems reasonable to me.

jaymell (approved):

PR #954, opened 2023-09-14
Update and use better maintained github actions.

nihohit:
actions-rs is no longer maintained.


nihohit:
You can see that now there are no warnings in the Github actions view.

jaymell (approved):

PR #958, opened 2023-09-18
feat: optimize write_command by using Vec instead of io::Write

PureWhiteWu:
In fact, we don't need to use `io::Write` here, because in all callers we use Vec<u8>. Thus, we can optimize this by using a Vec directly, which can greatly reduce some unnecessary checks in `io::Write::write_all`.

However, this still cannot guaranteed to remove the unnecessary bound checks, which may forbid the compiler to optimize further and generate optimized vectored(SIMD) code. To achieve this, we need to opt-in for unsafe code, which I think is not needed at least for now.

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -242,16 +242,18 @@ where
     write_command(cmd, args, cursor).unwrap()
 }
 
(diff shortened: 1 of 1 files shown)


nihohit:
have you measured the perf difference here?

PureWhiteWu:
You are right, there's no performance difference, the compiler can optimize the unnecessary checks out.

nihohit:
Thanks for checking!

Issue #959, opened 2023-09-21
Possibly incorrect command ordering for XREADGROUP


prabhpreet:
From the source code, the XREADGROUP command is sent in the following order:

`XREADGROUP BLOCK [ms] COUNT [count-number] [NOACK] GROUP [group] [consumer] STREAMS [stream-name] [id]`

This works out great with Redis (works with redis_version:7.2.1).

However, Redis docs state this order: 
`XREADGROUP GROUP [group] [consumer] [COUNT count] [BLOCK milliseconds] [NOACK] STREAMS [key ...] [id ...]`

https://redis.io/commands/xreadgroup/

Upstash Redis ( @upstash ) uses the order on the Redis site, and does not accept any other order.

Not sure who is at "fault" here, but I think we should follow the order in the Redis docs.

PR #960, opened 2023-09-21
Fix XREADGROUP command ordering as per Redis Docs, and compatibility with Upstash Redis

prabhpreet:
Fixes https://github.com/redis-rs/redis-rs/issues/959

Ordering also reported with Upstash Redis: https://github.com/upstash/issues/issues/74

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1760,7 +1760,7 @@ implement_commands! {
     ///     STREAMS key_1 key_2 ... key_N
     ///     ID_1 ID_2 ... ID_N
     ///
-    /// XREADGROUP [BLOCK <milliseconds>] [COUNT <count>] [NOACK] [GROUP group-name consumer-name]
+    /// XREADGROUP [GROUP group-name consumer-name] [BLOCK <milliseconds>] [COUNT <count>] [NOACK] 
     ///     STREAMS key_1 key_2 ... key_N
     ///     ID_1 ID_2 ... ID_N
     /// ```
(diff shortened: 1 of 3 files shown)


nihohit:
@prabhpreet thanks!
can you please add a test?

prabhpreet:
@nihohit I've updated the command order in the tests which had specified the other ordering before, and passes.

There should be additional tests with Upstash Redis if @upstash could provide an Redis instance for test CI, but I'd suggest it be a separate PR due to adding support to test against a non localhost remote server, and the likeliness other commands will also come up with similar ordering issues.

nihohit:
https://github.com/redis-rs/redis-rs/blob/2718fdd8f466bbc70e928a7026dbad8fbd03824d/redis/src/commands/mod.rs#L1763
this doc also needs to be fixed

I see that there are tests running XREADGROUP, so there's no need to anything else AFAIS.

nihohit (approved):

prabhpreet:
Not sure why Rust / build (7.2.0, nightly) fails for `test_async_cluster_basic_cmd` in https://github.com/prabhpreet/redis-rs/blob/43b39087ea196febcfc0d007af14e3d3ac92c163/redis/tests/test_cluster_async.rs#L24

nihohit:
@prabhpreet lint still broken.

nihohit:
Thanks for the fix @prabhpreet !

prabhpreet:
@nihohit Thanks for the review and approval. Can this be published into a release? When are new releases cut for cargo?

nihohit:
I'm not sure when will we make the new release. If you need the feature, you can use the crate from a git commit instead of from crates.io.

Issue #944, opened 2023-09-03
Missing blocking commands for Redis Sorted set


gheorghitamutu:
Hello,

I was looking through Sorted Set functions (wrappers over commands) in this library and I noticed that there are no blocking commands exposed for Redis Sorted Set. More specifically:

https://redis.io/commands/bzpopmin
https://redis.io/commands/bzpopmax
https://redis.io/commands/bzmpop

Are there any plans to implement these in the near future?
I also found an issue that's related to this - https://github.com/redis-rs/redis-rs/issues/476 - but there's no answer to it.

It is indeed true that one could just use:
```
let (key, member, score): (String, String, u64) = redis::cmd("BZPOPMIN")
            .arg(key)
            .arg(1)
            .arg(0)
            .query_async(&mut connection)
            .await?;
```

But I was expecting to have a wrapper similar to List blocking commands:

```
    fn blpop<K: ToRedisArgs>(key: K, timeout: usize) {
        cmd("BLPOP").arg(key).arg(timeout)
    }
```

nihohit:
> Are there any plans to implement these in the near future?

I don't think any of the maintainers are planning on doing this, but we're always happy to accept contributions. Usually adding commands is very easy.

> I also found an issue that's related to this - https://github.com/redis-rs/redis-rs/issues/476 - but there's no answer to it.

Answered there. Thanks for pointing that out.

jaymell:
I'm going to close this as a duplicate of #476, but please feel free to submit a PR if you'd like @gheorghitamutu.

Issue #476, opened 2021-04-12
How can I use bzpopmax or bzpopmin?


uheee:
I want my program to **block** when using ```bzpopmax``` until the first available element appears in the zset. I did not find any related functions in the documentation, and ```zpopmax``` will return regardless of whether there is an element in zset.

nihohit:
Hi, can you write a sample test that will demonstrate the issue?

jaymell:
I think the issue here is just that we're missing the wrappers for the blocking commands, correct? If so, it should be an easy PR to add these, as @nihohit mentioned on #944.

somechris:
It seems support for `bzpopmax` meanwhile got added to `implement_commands!` and is documented.

PR #962, opened 2023-09-24
Sorted set blocking commands

gheorghitamutu:
Hello,

The feature has been requested in #944 & #476.

The commands implemented are:
https://redis.io/commands/bzpopmin
https://redis.io/commands/bzpopmax
https://redis.io/commands/bzmpop

Tests were added for all 3 of them.

### DIFF
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -1280,3 +1280,32 @@ fn test_set_options_options() {
 
     assert_args!(&opts, "EX", "1000");
 }
+
+#[test]
+fn test_blocking_sorted_set_api() {
+    let ctx = TestContext::new();
+    let mut con = ctx.connection();
+
+    assert_eq!(con.zadd("a", "1a", 1), Ok(()));
+    assert_eq!(con.zadd("b", "2b", 2), Ok(()));
+    assert_eq!(con.zadd("c", "3c", 3), Ok(()));
+    assert_eq!(con.zadd("d", "4d", 4), Ok(()));
+    assert_eq!(con.zadd("a", "5a", 1), Ok(()));
+    assert_eq!(con.zadd("b", "6b", 2), Ok(()));
+    assert_eq!(con.zadd("c", "7c", 3), Ok(()));
+    assert_eq!(con.zadd("d", "8d", 4), Ok(()));
+
+    let result_multiple_min = con
+        .bzmpop_min::<&str, (String, Vec<Vec<(String, String)>>)>(0, vec!["a", "b", "c", "d"].as_slice(), 1);
+    let result_multiple_max  = con
+        .bzmpop_min::<&str, (String, Vec<Vec<(String, String)>>)>(0, vec!["a", "b", "c", "d"].as_slice(), 1);
+
+    let result_min = con.bzpopmin::<&str, (String, String, String)>("b", 0);
+    let result_max = con.bzpopmin::<&str, (String, String, String)>("b", 0);
+
+    assert_eq!(result_multiple_min.unwrap().1[0][0], (String::from("1a"), String::from("1")));
+    assert_eq!(result_multiple_max.unwrap().1[0][0], (String::from("5a"), String::from("1")));
+
+    assert_eq!(result_min.unwrap(), (String::from("b"), String::from("2b"), String::from("2")));
+    assert_eq!(result_max.unwrap(), (String::from("b"), String::from("6b"), String::from("2")));
+}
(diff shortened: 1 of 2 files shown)


gheorghitamutu:
As a side note:
The `BZMPOP` command is available from version 7.0.0 and it will 
probably fail testing with Redis version 6.2.4. If there's a workaround/check that I can do in order to specify Redis' version, please let me know. If not, I'll remove this specific command test.

redis/tests/test_basic.rs
```diff
@@ -1280,3 +1280,32 @@ fn test_set_options_options() {
...
+    assert_eq!(con.zadd("a", "1a", 1), Ok(()));
+    assert_eq!(con.zadd("b", "2b", 2), Ok(()));
+    assert_eq!(con.zadd("c", "3c", 3), Ok(()));
+    assert_eq!(con.zadd("d", "4d", 4), Ok(()));
+    assert_eq!(con.zadd("a", "5a", 1), Ok(()));
+    assert_eq!(con.zadd("b", "6b", 2), Ok(()));
+    assert_eq!(con.zadd("c", "7c", 3), Ok(()));
+    assert_eq!(con.zadd("d", "8d", 4), Ok(()));
+
+    let result_multiple_min = con
```
jaymell, line 1298: Did you mean to only test the `min` command variants?

jaymell:
RE the version checking, I'm surprised, but not aware of any place where we've previously excluded tests based on version. It should be possible to parse the server version out of `INFO` command and exclude the the specific assertions as needed, so I'd prefer to do that if you think you can tackle it.

nihohit:
> It should be possible to parse the server version out of INFO command and exclude the the specific assertions as needed

maybe add a helper function that takes a version and returns bool to tests/utilities/mod.rs to do that, and return early from tests when the version is too early?
alternatively, maybe we can get the version from env variable?
 If you need help, please say.

redis/tests/test_basic.rs
gheorghitamutu: You're right, the max variables should have been the result of max functions. My mistake. 
I fixed it in [a08aba1](https://github.com/redis-rs/redis-rs/pull/962/commits/a08aba18f1b2ac8ebf50e7328a271761f0048c6a).

gheorghitamutu:
@jaymell I did not see any tests for other (non) blocking functions added as commands that were introduced in Redis version 7.0.0 so that might be why there was no need for a version checking helper function.

@jaymell @nihohit I can probably do both of them as helpers for version checking but if you want a specific one, please let me know. Unfortunately, I'm not aware of a standard env variable that specifies the Redis version. If there's no such thing, I'd go with @jaymell suggestion and parse the INFO command for version information. If there's a custom env var set up, anyone who tries to create a build & test workflow needs to make sure that they also set it up and it sounds like another small dependency added.

nihohit:
you're right, best to test the server version.

gheorghitamutu:
@jaymell @nihohit please let me know what you think about statements below.

I've looked a bit into how I'd be doing this for just my case.
The quick and dirty code block I came with is something along these lines:
```rs
    let major_version : u32 = {
        let raw_info: String = redis::cmd("INFO").query(&mut con).unwrap();
        let pattern = "redis_version:";
        let version_position = raw_info.find(pattern).unwrap();
        let major_version = raw_info.chars().nth(version_position + pattern.len()).unwrap();
        major_version.to_digit(10).unwrap()
    };
```

What this solves is solely my current issue (partially, as non multi commands - `BZPOPMIN` & `BZPOPMAX` are only available from Redis version 5 as well).
What it doesn't is along these lines:
- there will probably be more tests in the future depending on the Redis version and thus this is not the right place nor the ergonomic implementation for this function/code block
- the function should probably be a helper for TestContext (and maybe keep it as metadata/ structure member) and this is out of the scope of the current PR (it should also take in account the actual format of the version `<major>.<minor>.<path>` in case there are cases we also want to look into that)
- the future design should also probably somehow remove the platform attribute (windows) for several tests and rely on the actual Redis version
- I'd rather open an entirely new issue based on the above

If you agree I can either merge the above code block in my test only or I can remove asserting/calling `BZMPOP` command that requires Redis version 7.

nihohit:
@gheorghitamutu please try rebasing over https://github.com/redis-rs/redis-rs/pull/963 and using the functions I added there.

gheorghitamutu:
Hello, @nihohit, thank you!

nihohit:
@jaymell LGTM

gheorghitamutu:
There was a small fix to be done in [8f26d05](https://github.com/redis-rs/redis-rs/pull/962/commits/8f26d056309dcdacad2ff95bedd9fa0efede4a39).

nihohit (approved):

jaymell:
Thanks for the PR!

PR #965, opened 2023-10-02
Cluster: fix read from replica & missing slots.

nihohit:
This PR refactors the cluster `SlotMap` to account for 4 existing issues:
1. ATM clusters can't be created with partial slot coverage. This is erroneous, because given the right configuration the Redis cluster can work in such a situation, so we shouldn't fail it.
This was fixed by removing the asserts enforcing this from the cluster connections, and adding to the SlotMap the start slot of each range, so that it could return `None` when trying to get the address for a missing slot.

2. Since SlotAddrs saves at most 1 replica (or none, if
`read_from_replica` is false), it means that when there are 2 or more
replicas in the cluster, `AllNodes` commands aren't actually sent to
all nodes.
This was fixed by saving a Vec of addresses in SlotAddrs instead of a single replica.

3. `SlotAddr::Replica` carries 2 separate semantic meanings - it could
mean that the command is a readonly command, or that the user asked to
route the command to a replica.
The first case should be affected by the `read_from_replica` client
flag, but the second case shouldn't - if the user requested that this
specific command be routed to the replica, the specific request should
override the prior configuration.
This was fixed by separating `SlotAddr::Replica` to `SlotAddr::ReplicaOptional` and `SlotAddr::ReplicaRequired`.

4. MultiSlot commands were some routes are to missing slots. In this case we silently ignore the missing slots, in order to allow the rest of the commands to be sent.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -224,7 +224,7 @@ where
     ) -> RedisResult<Self> {
         let connection = Self {
             connections: RefCell::new(HashMap::new()),
-            slots: RefCell::new(SlotMap::new()),
+            slots: RefCell::new(SlotMap::new(cluster_params.read_from_replicas)),
             auto_reconnect: RefCell::new(true),
             read_timeout: RefCell::new(None),
             write_timeout: RefCell::new(None),
@@ -375,10 +375,7 @@ where
         for conn in samples.iter_mut() {
             let value = conn.req_command(&slot_cmd())?;
             if let Ok(slots_data) = parse_slots(value, self.cluster_params.tls) {
-                new_slots = Some(SlotMap::from_slots(
-                    &slots_data,
-                    self.cluster_params.read_from_replicas,
-                ));
+                new_slots = Some(SlotMap::from_slots(slots_data, self.cluster_params.read_from_replicas));
                 break;
             }
         }
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -36,6 +36,7 @@
 //!     .query(&mut connection).unwrap();
 //! ```
 use std::cell::RefCell;
+use std::collections::HashSet;
 use std::iter::Iterator;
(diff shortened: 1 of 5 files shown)


redis/tests/test_cluster_async.rs
```diff
@@ -1457,6 +1457,42 @@ fn test_async_cluster_non_retryable_error_should_not_retry() {
     assert_eq!(completed.load(Ordering::SeqCst), 1);
 }
 
+#[test]
+fn test_async_cluster_can_be_created_with_partial_slot_coverage() {
+    let name = "test_async_cluster_can_be_created_with_partial_slot_coverage";
```
jaymell, line 1462: Is it ever possible that slots would _overlap_? I've never encountered this but wondering if possible.

redis/src/cluster.rs
```diff
@@ -509,52 +485,101 @@ where
         Ok(result)
     }
 
+    fn exectue_on_all<'a>(
```
jaymell, line 488:
```suggestion
    fn execute_on_all<'a>(
```
Same issue w/ following methods.

redis/tests/test_cluster_async.rs
nihohit: It's not possible, according to the redis spec. Each slot can be managed only by one shard.

jaymell: Yeah, makes sense. I was wondering more in the scenario of a cluster that has briefly lost its mind, could it possibly serve out a list of nodes w/ overlapping slots? But I'm guessing not, and in any case I think the code here would probably handle it without issue.

nihohit: Putting aside the possibility of bugs in Redis code (see gif below), a single node won't return overlapping slots. Different nodes might have different views of the server, but that's a tomorrow-problem.

image

PR #966, opened 2023-10-03
Bump aHash to v0.8.6

aumetra:
This is a breaking change, bumping the `ahash` dependency in a semver-incompatible way

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -76,7 +76,7 @@ serde = { version = "1.0.82", optional = true }
 serde_json = { version = "1.0.82", optional = true }
 
 # Optional aHash support
-ahash = { version = "0.8.3", optional = true }
+ahash = { version = "0.8.6", optional = true }
 
 log = { version = "0.4", optional = true }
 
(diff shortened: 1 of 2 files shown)


jaymell:
Hi, can you explain the reason for the breaking changes?

aumetra:
Sure! The first breaking change here is the bump from v0.7 to v0.8 itself. That's done to simply stay up-to-date with the newest release and to include a fix for their use of AES instructions on `aarch64` architectures (which will be needed in the future)

The second breaking change is the removal of the `RandomState` parameter from `AHashMap`. The reason behind this is two-fold:

1. The `AHashMap` only implements `FromIterator` with the `ahash::RandomState` as the hasher
2. The only way to instantiate an `ahash::AHasher` hasher via the current API (apart from the `Default` impl) is via `ahash::RandomState`, so having the parameter freely choosable is somewhat unnecessary anyways (and could be brought back with a future release without breaking semver)

jaymell:
Thanks for the explanation, which certainly sounds reasonable to me. We still need to figure out our next release and whether it's breaking, so I'm going to leave this un-merged for now. @aumetra do you mind updating the version specified to a non-yanked version?

jaymell (approved):

PR #964, opened 2023-09-25
Improve cluster retries

jaymell:
Currently we don't sleep for certain classes of errors, which in my testing leads to certain potentially suboptimal results. Most notably, `MOVED` errors are likely to be thrown when a cluster is in an unhealthy state, such as during a node loss. With our current strategy of immediately retrying `MOVED` errors without pause, we eliminate the client's ability to recover in many scenarios.

In any case, I think it's reasonable that we sleep for all retryable errors.

Relates to #906.

PR #968, opened 2023-10-04
Fix proposal - sleep when connection not found.

nihohit:
This change should ensure that if a connection wasn't found, after redirecting to node, or when no random connection is available, then the cluster connection will refresh slots and sleep before retrying.


jaymell:
I'll find the time soon to test this locally against the failure scenarios I was testing previously.

redis/src/cluster_async/mod.rs
```diff
@@ -356,23 +370,34 @@ where
...
                     OperationTarget::FanOut => {
                         // Fanout operation are retried per internal request, and don't need additional retries.
                         self.respond(Err(err));
                         return Next::Done.into();
                     }
+                    OperationTarget::NotFound => {
+                        println!("target not found. retry: {}", request.retry);
+                        // TODO - this is essentially a repeat of the retriable error. probably can remove duplication.
+                        let request = this.request.take().unwrap();
+                        return Next::RefreshSlots {
```
jaymell, line 392: I've tested this branch against a common failure scenario, complete loss of a node. Based on that testing, we may want to do something like this here:
```
                        let mut request = this.request.take().unwrap();
                        request.info.redirect = None; 
```                        
Else, the request is likely to be repeatedly retried against the failed node, the issue being that when the node initially fails, the client will get an erroneous `MOVED` error pointing to the failed node. If we don't unset the redirect somehow, the request will never connect to the correct node after a slot-refresh has occurred.

nihohit: did you test this change, or is this only an assumption?

where exactly in the code do you think that the redirect should be reset?

jaymell: I tested your branch locally, using same scenarios I'd previously tested [here](https://github.com/redis-rs/redis-rs/pull/964#issuecomment-1736744056), so it's not just an assumption.

I'm not immediately sure the best way to accomplish this, but the code snippet I posted above at line 391 at least causes the request to eventually recover. Without it, it will go through a cycle of 16 retries before eventually failing, long after the cluster has recovered.

jaymell: The more general scenario I suppose, is that a request with a redirect may, in the process of retrying, need to try the original request again without the redirect. Right now the code doesn't allow for this flexibility.

nihohit: Thanks for the data! I'll give it a thought and see how to handle this. Hopefully another patch next week :)

jaymell:
@nihohit Are we good to take this out of "draft" mode?

nihohit:
> @nihohit Are we good to take this out of "draft" mode?

I'm waiting for your input, whether this solves the failover issue. If it works, sure!

nihohit:
you might want to review https://github.com/redis-rs/redis-rs/pull/965 first

jaymell:
Thanks, I didn't realize you had rebased this. Anyway, I've tested locally and the problem I reported earlier is definitely resolved :sunglasses:

nihohit:
@jaymell ready for review.

There's still the situation where we get a complete disconnect from the server - all nodes at the same time are disconnected - which needs handling, but I'll leave that to the next PR. This is plenty large enough as-is.

jaymell:
I think these changes look great, but I also feel like they should be broken up into more clearly structured commits, and the PR needs better title. 

The redirection / handling connection-not-found changes are significant. Can we put those changes and the sleep functionality in separate commits and merge w/ the individual commits intact?

nihohit:
also - I forgot to add tests 🤦

Updated PR name & description.

re: commits - The first 2 commits are the proposed changes - sleep & retry, remove redirect when node isn't found.
the 3rd commit is just improving the code readability around the changes.
the 4-5th commits are fixes to the first 2 commits.
So, besides squashing the last 2 commits to earlier commits, how would you like the commits to be rearranged?

jaymell:
> re: commits - The first 2 commits are the proposed changes - sleep & retry, remove redirect when node isn't found.
the 3rd commit is just improving the code readability around the changes.
the 4-5th commits are fixes to the first 2 commits.
So, besides squashing the last 2 commits to earlier commits, how would you like the commits to be rearranged?

IMO there's a lot going on in that first commit. But at a minimum let's remove the last two and make sure the others have appropriate titles/descriptions.

nihohit:
@jaymell rearranged commits, added tests, and covered another scenario - a complete disconnect from server.

redis/tests/test_cluster_async.rs
```diff
@@ -1492,6 +1532,75 @@ fn test_async_cluster_can_be_created_with_partial_slot_coverage() {
...
+        3,
+        0,
+        |builder| builder.retries(2),
+        false,
+    );
+
+    block_on_all(async move {
+        let mut connection = cluster.async_connection().await;
+        drop(cluster);
+        for _ in 0..5 {
```
jaymell, line 1575: Not totally following -- what's the benefit of doing this block repeatedly?

jaymell (approved):
Beautiful -- thanks for restructuring! 

And for adding even more code! :stuck_out_tongue:

redis/tests/test_cluster_async.rs
nihohit: each attempt removes some connections. We can't be sure that they'd all be removed in the first try.

Issue #977, opened 2023-10-13
StreamReadReply silently discards data


mxbrt:
I am trying to write and read a redis stream using `xadd` and `xread`. While I can see that the data exists in redis, I get an empty `StreamReadReply` from `xread_options`:

```rust
let _: () = redis::pipe()
    .xadd("stream", "*", &[("key", "val")])
    .query_async(&mut conn)
    .await?;
let options = StreamReadOptions::default().block(0).count(1);
let reply: StreamReadReply = redis::pipe()
    .xread_options(&["stream"], &["0"], &options)
    .query_async(&mut conn)
    .await?;
eprintln!("reply {:?}", reply); // Prints: reply StreamReadReply { keys: [] }
```
If I do not convert the query into a `StreamReadReply`, but print it as a `redis::Value`, i can see that some data is returned, but it is wrapped in one more `Value::Bulk` then `redis-rs` expects. The data is silently discarded during conversion to `StreamReadReply`.

Example:

```rust
use anyhow::Result;
use redis::{streams::StreamReadReply, FromRedisValue, Value};

fn main() -> Result<()> {
    // This is the data that I get from redis
    // bulk(bulk(bulk(string-data('"stream"'), bulk(bulk(string-data('"0-1"'), bulk(string-data('"key"'), string-data('val'))))))))
    let v1 = Value::Bulk(vec![Value::Bulk(vec![Value::Bulk(vec![
        Value::Data("stream".into()),
        Value::Bulk(vec![Value::Bulk(vec![
            Value::Data("0-1".into()),
            Value::Bulk(vec![Value::Data("key".into()), Value::Data("val".into())]),
        ])]),
    ])])]);
    // returns an empty StreamReadReply
    let v1 = StreamReadReply::from_redis_value(&v1)?;
    // This is the data how redis-rs expects it
    // bulk(bulk(string-data('"stream"'), bulk(bulk(string-data('"0-1"'), bulk(string-data('"key"'), string-data('val')))))))
    let v2 = Value::Bulk(vec![Value::Bulk(vec![
        Value::Data("stream".into()),
        Value::Bulk(vec![Value::Bulk(vec![
            Value::Data("0-1".into()),
            Value::Bulk(vec![Value::Data("key".into()), Value::Data("val".into())]),
        ])]),
    ])]);
    let v2 = StreamReadReply::from_redis_value(&v2)?;
    eprintln!("\nv1 {:?}\nv2 {:?}\n", v1, v2);

    // output:
[...]

PR #978, opened 2023-10-21
Fix HashMap conversion (#977)

mxbrt:
When the list in a bulk value contains an odd number of elements, the last element would be silently discarded by MapIter. The list is now checked to be of even lenth during initialization.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -219,7 +219,13 @@ impl Value {
     /// Returns an iterator of `(&Value, &Value)` if `self` is compatible with a map type
     pub fn as_map_iter(&self) -> Option<MapIter<'_>> {
         match self {
-            Value::Bulk(items) => Some(MapIter(items.iter())),
+            Value::Bulk(items) => {
+                if items.len() % 2 == 0 {
+                    Some(MapIter(items.iter()))
+                } else {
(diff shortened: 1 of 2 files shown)


nihohit:
can you please add a test?

nihohit:
I'm not sure that silently returning None is the right solution here. Shouldn't this be an error?
@jaymell what do you think?

mxbrt:
While as_map_iter returns None, this is later converted to an error in [from_redis_value](https://github.com/redis-rs/redis-rs/blob/304eb93ca0215bd754865e4e6a7bbb1aeaed0228/redis/src/types.rs#L1374). That's also checked by the added assert in test_hashmap in this commit.

nihohit:
yes, but the error doesn't explain why the value wasn't hashmap compatible - instead of returning None, you could return Err(SomeErrorKind, "Map contained an odd number of elements");

jaymell:
I'm not opposed to merging this. IMO the use and/or design of `MapIter` could probably be improved (do we even need it at all?), but I don't know that we need to do that here.

nihohit (approved):

nihohit:
@mxbrt Thanks for your contribution!

PR #979, opened 2023-10-23
Bump ahash dependency version

joshlf:
The previous version had a bug, and was yanked.


jaymell:
Thanks for the PR. We already have another PR (#966) that addresses the issue here, so I'm going to close this for now.

PR #983, opened 2023-11-05
Fix sync cluster behavior with transactions.

nihohit:
The leading MULTI prevents the transaction from being properly routed, because there's no routing for the multi command, and the routing check doesn't iterate through the contained commands. In order to route the request correctly, we drop the leading MULTI from the routable.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -46,7 +46,7 @@ use rand::{seq::IteratorRandom, thread_rng, Rng};
 use crate::cluster_client::RetryParams;
 use crate::cluster_pipeline::UNROUTABLE_ERROR;
 use crate::cluster_routing::{
-    MultipleNodeRoutingInfo, ResponsePolicy, SingleNodeRoutingInfo, SlotAddr,
+    MultipleNodeRoutingInfo, ResponsePolicy, Routable, SingleNodeRoutingInfo, SlotAddr,
 };
 use crate::cmd::{cmd, Cmd};
 use crate::connection::{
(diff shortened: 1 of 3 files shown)


nihohit:
@jaymell this feels to me like a bit of a hack around inherent correctness issues in this API, which we'll need to consider if/when we implement a similar API for the async connections.
1. This API allows the user to send the wrong number of requests to the server - for example, sending multiple commands packed in a single Vec on `req_packed_command` or giving the wrong command count to `req_packed_commands`. Both will cause a similar issue to that of the `aio::Connection` dropped requests - the connection returns responses to the request in received order, so miscounts will mean that requests receive the wrong response. 
2. `req_packed_commands` might receive more than one `MULTI`-`EXEC` pair packed in a single Vec, so the `offset`/`count` arguments won't help to return the correct array.
3. `req_packed_commands` might receive multiple commands not in a transaction, which need to be routed separately. ATM we don't have a way of iterating through the commands in a packed Vec, so we can't split the commands and route them separately.
4. Because we can't iterate through the commands, even after this fix, routing is done according to the first command in a transaction. This is unlike the [async pipeliene behavior](https://github.com/redis-rs/redis-rs/blob/c56c72a924ce95ca6b07f875beec16732a0aef9d/redis/src/cluster_async/mod.rs#L214), where we find the first command with a concrete route in the pipeline, and route it according to that command.

All of these issues stem from "wrong" user behavior, but the API (or the documentation) doesn't warn or protect against these issues, and ATM I don't see a way to mitigate this.

jaymell (approved):
This looks good. The comments about the limitations you make here are important and definitely valid. It's probably worth adding your summary to a separate issue for better tracking in the future.

PR #986, opened 2023-11-12
Prefer routing to primary in a transaction.

nihohit:
This change will cause transactions with commands that are routed to replicas and primaries to prefer the primary routing over the replica.


jaymell (approved):
Looks good!

PR #988, opened 2023-11-15
Changing timeouts from usize and isize to f64

eythorhel19:
Changing timeout values of functions:
- blmove
- blmpop
- blpop
- brpop
- brpoplpush
- bzpopmax
- bzpopmin
- bzmpop_max
- bzmpop_min

The reason behind it is that the timeout is in seconds and being limited to a minimum of one second is challenging.

The timeout value of all of these functions is a double value in redis versions greater than 6.0.0.


nihohit:
Thanks for the contribution!

PR #993, opened 2023-11-26
Revert redis-test versioning changes

jaymell:
Publishing to crates.io without the version specification 
is currently not possible. Reverting these changes for now 
until we have a better solution/workaround.

### DIFF
--- a/redis-test/Cargo.toml
+++ b/redis-test/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.65"
 bench = false
 
 [dependencies]
-redis = { path = "../redis" }
+redis = { version = "0.23.4", path = "../redis" }
 
 bytes = { version = "1", optional = true }
 futures = { version = "0.3", optional = true }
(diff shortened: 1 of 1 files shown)


jaymell:
@nihohit I'll let you review this one before merging. I think it probably makes sense to leave the version specification in (as, for example, tokio does for its test crate), though perhaps we don't need to update it for minor releases of the `redis` crate. 

For reference https://github.com/rust-lang/cargo/issues/11133

nihohit:
LGTM. Will keeping both the version and the path arguments be an issue? Might this fall if we don't update the version correctly in both places (which might be good verification)?

nihohit (approved):

jaymell:
> LGTM. Will keeping both the version and the path arguments be an issue? Might this fall if we don't update the version correctly in both places (which might be good verification)?

We should be ok as long as the version specified in the test crate is part of the same major release as the redis crate. For example, specifying a redis version of 0.23.0 would also work in the current PR, but builds will fail if it's 0.22.0 or 0.24.0. So, we should generally have warning if we forget to update the versions.

PR #994, opened 2023-11-26
Tests: Add retries to test cluster creation

jaymell:
Add retries to both the server and cluster creation process. I believe the core
fix here is just adding a sleep after the server child process is spawned, however
I'm leaving the retries in as I don't think they hurt anything.

### DIFF
--- a/redis/tests/support/cluster.rs
+++ b/redis/tests/support/cluster.rs
@@ -108,6 +108,8 @@ impl RedisCluster {
             is_tls = true;
         }
 
+        let max_attempts = 5;
+
         for node in 0..nodes {
             let port = start_port + node;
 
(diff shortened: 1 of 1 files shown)


redis/tests/support/cluster.rs
```diff
@@ -118,77 +120,111 @@ impl RedisCluster {
...
+                            cmd.arg("--tls-cluster").arg("yes");
+                            if replicas > 0 {
+                                cmd.arg("--tls-replication").arg("yes");
+                            }
+                        }
+                        cmd.current_dir(tempdir.path());
+                        folders.push(tempdir);
+                        addrs.push(format!("127.0.0.1:{port}"));
+                        let mut process = cmd.spawn().unwrap();
+                        sleep(Duration::from_millis(100));
```
jaymell, line 156: After some trial and error, it seems like this `sleep` prevents most of the issues we've been seeing recently.

redis/tests/support/cluster.rs
```diff
@@ -118,77 +120,111 @@ impl RedisCluster {
...
+                            .tempdir()
+                            .expect("failed to create tempdir");
+                        let acl_path = tempdir.path().join("users.acl");
+                        let acl_content = format!(
+                            "user {} on allcommands allkeys >{}",
+                            Self::username(),
+                            Self::password()
+                        );
+                        std::fs::write(&acl_path, acl_content).expect("failed to write acl file");
+                        cmd.arg("--cluster-enabled")
```
nihohit, line 136: you add args to `cmd` in every loop, which sounds incorrect. shouldn't the loop only encompass `cmd.spawn`? or maybe create a base `cmd` before the loop, and in the loop only create the tempdir and add it to a clone of `cmd`?

PR #997, opened 2023-12-01
Order in usage of ClusterParams.

nihohit:
Moved `tls_params` into `ClusterParams`, and made the sync cluster use `ClusterParams` directly instead of copying its fields.


jaymell (approved):
Looks good!

PR #1002, opened 2023-12-11
feat: add SMISMEMBER command

Zacaria:
Hello

I have just found that the command SMISMEMBERS is missing from the lib.

I think it can be great for checking presence of multiple members with one network connection.

As I didn't found a Contributing section, I'm naively proposing this change.

Please do not hesitate to tell me what changes are needed along with my initial commit.

Please note that SMISMEMBERS appears at redis version 6.2.0

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -497,6 +497,11 @@ implement_commands! {
         cmd("SISMEMBER").arg(key).arg(member)
     }
 
+    /// Determine if given values are members of a set.
+    fn smismember<K: ToRedisArgs, M: ToRedisArgs>(key: K, members: M) {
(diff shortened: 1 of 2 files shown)


nihohit:
thanks for the addition!
can you please add a test?

nihohit:
Well, if the tests before the JSON module tests passed, that's good enough. If you want to be able to run all tests, you'll need to build the JSON module library - see https://github.com/redis-rs/redis-rs/blob/main/README.md#development.

Please fix the lint issues.

nihohit (approved):

FallingSnow:
Do we close #797 and #832?

nihohit:
> Do we close #797 and #832?

yes, thanks for pointing them out.

PR #1007, opened 2023-12-16
Fix the problem that AsyncIter may be blocked

tryor:
When executing: 
```
  async_conn.hscan_match::<_, _, Vec<u8>>(name, prefix.as_slice()).await?
```
or
```
  cmd("HSCAN")
        .arg(name).cursor_arg(0).arg("match").arg(prefix).clone()
        .iter_async::<Vec<u8>>(&mut async_conn)
        .await?;
```

When no data with prefix is found, it will be blocked.

Viewing the source code is due to the blockage caused by the cursor not being 0 but the return result set being empty, which will always attempt to obtain more data.

The following is the result of directly executing the Redis command in Redis cli:
```
127.0.0.1:6379> hscan "map001" 0 match "10206*" count 10
1) "1572864"
2) (empty array)
```

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -149,6 +149,10 @@ required-features = ["json", "serde/derive"]
 name = "test_cluster_async"
 required-features = ["cluster-async"]
 
+[[test]]
+name = "test_iter_async"
+required-features = ["tokio-comp"]
+
 [[bench]]
 name = "bench_basic"
 harness = false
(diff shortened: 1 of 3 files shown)


nihohit:
hi, can you please add tests for both scenarios?

tryor:
Executing 'test_iter_async.rs' will fail in 0.24 and earlier versions. Increasing the value of let max = 100_000; will increase the blocking time.

nihohit:
Thanks :)

1. Can you please add a `mod` around the test file? I know we haven't done that in previous files, but it really helps with running just these tests.

2. I ran your tests locally, and it seems like either one of the added checks is sufficient in order to make them pass. Can you please explain why both of the checks are needed?

tryor:
Sorry, this change will cause other new problems. I will close this PR!

PR #1012, opened 2023-12-21
Use enum instead of boolean for RESP version.

nihohit:
Since there's a discussion starting about what might become RESP4 (https://github.com/redis/redis/issues/12873), this
PR will make it easier to add more RESP versions in the future.


jaymell (approved):
Looks great!

PR #1014, opened 2023-12-27
Add support for some big number types

AkiraMiyakoda:
This PR enables `redis-rs` to read/write some third-party "big number" types directly.

## Motivation

We can store a number with any number of digits in Redis as a string. Although we can read such a value as a string and convert it into a desired numeric type, it's convenient when we can read it directly as a number like this:
```rust
use num_bigint::BigInt;

let v: BigInt = connection.get("foo").await?;
```

This library already works with Rust's intrinsic numeric types. This PR adds a feature called `bignum` to it so that it can handle some 'big number' types from some third-party crates which are mature and widely used.

## What is this?

This PR contains `FromRedisValue` and `ToRedisArgs` implementation for following types:
 * `rust_decimal::Decimal`
 * `bigdecimal::BigDecimal`
 * `num_bigint::BigInt`
 * `num_bigint::BigUint`

It includes tests and a feature flag as well.


Makefile
```diff
@@ -26,7 +26,7 @@ test:
 	@echo "===================================================================="
 	@echo "Testing Connection Type UNIX"
 	@echo "===================================================================="
-	@REDISRS_SERVER_TYPE=unix RUST_BACKTRACE=1 cargo test -p redis --test parser --test test_basic --test test_types --all-features -- --test-threads=1 --skip test_module
+	@REDISRS_SERVER_TYPE=unix RUST_BACKTRACE=1 cargo test -p redis --test parser --test test_basic --test test_types --test test_bignum --all-features -- --test-threads=1 --skip test_module
```
nihohit, line 29: revert

redis/tests/test_bignum.rs
```diff
@@ -0,0 +1,120 @@
+use redis::{ErrorKind, FromRedisValue, RedisResult, ToRedisArgs, Value};
+use std::str::FromStr;
+
+#[test]
+fn test_rust_decimal() {
+    use rust_decimal::Decimal;
+
+    let content = "-79228162514264.337593543950335";
```
nihohit, line 8: the tests mostly repeat themselves - can the logic be moved to a function?

redis/src/bignum.rs
```diff
@@ -0,0 +1,74 @@
...
+            format!("{:?} (response was {:?})", $det, $v),
+        ))
+    };
+}
+
+macro_rules! from_redis_value_for_bignum_internal {
+    ($t:ty, $v:expr) => {{
+        let v = $v;
+        match *v {
+            Value::Int(val) => match <$t>::try_from(val) {
```
nihohit, line 26: you can use `map_err` instead of the matches.

redis/src/bignum.rs
```diff
@@ -0,0 +1,74 @@
+use std::str::from_utf8;
+
+use super::{ErrorKind, RedisError, RedisResult};
+use crate::types::{FromRedisValue, RedisWrite, ToRedisArgs, Value};
+
+macro_rules! invalid_type_error {
```
nihohit, line 6: these are copies from types.rs. why not just add the logic there?

nihohit (commented):
This looks mostly OK to me. 
@jaymell since these are separate crates, do you think that these should be added as a single feature, or split between multiple features?

Makefile
AkiraMiyakoda: @nihohit Thanks. Reverted.

redis/src/bignum.rs
AkiraMiyakoda: Changed to map_err

redis/src/bignum.rs
AkiraMiyakoda: Moved to types.rs

redis/tests/test_bignum.rs
AkiraMiyakoda: Done

nihohit (approved):

nihohit (approved):
@jaymell LGMT, just not sure whether we want this as a single feature, or split the feature between crates, so that each bignum crate will have its own feature. WDYT?

jaymell:
This is great! But I definitely think it makes sense to not force everything into one feature. @AkiraMiyakoda Can we split this into multiple features, one for each external crate?

AkiraMiyakoda:
Hi @jaymell 
Thanks for the review. I split the `bignum` feature into three which correspond to target crates.
New features' names are `rust_decimal`, `bigdecimal` and `num-bigint` just the same as the crates.

nihohit:
Thanks for the contribution, and the excellent work!

Issue #1017, opened 2024-01-03
Response's value does not match the expected command's response value 


x1qqq:
I am having an issue with this library when trying to fetch a json object via the function `json_get`.
It outputs the following error logs.

![image](https://github.com/redis-rs/redis-rs/assets/87322933/84221a84-72c8-40bc-8289-2ae58e4cf4d0)

Eventually after a few tries, it manages to fetch the correct value, but it always has this unexpected behavior for around 2 or 3 times with the same following values: int(1),  ok

I have no idea where the problem is coming from, I am supposing there might be a problem with the library, I just don't know what it specifically can be.

Note:
I am using the ConnectionManager.
I am using redis-macros for implementing FromRedisValue via its derive macro for my struct object `InteractionStep`.

jaymell:
Can you post a reproducible example?

x1qqq:
I have tried reproducing on a test function, everything seems to be working fine there somehow.

The only difference from my code and the test function, is that I'm passing the ConnectionManager to a [TypeMap](https://docs.rs/typemap_rev/latest/typemap_rev/struct.TypeMap.html#), everything else functionality wise is the same.

To give a little bit more context on what I am doing, I am building a discord bot with [serenity-rs](https://docs.rs/serenity/latest/serenity/) and using this library (redis-rs) to fetch json objects.

Here is my test function

```rust
#[cfg(test)]
mod tests {
    use std::time::Duration;

    use redis::aio::ConnectionLike;
    use redis::{aio::ConnectionManager, JsonAsyncCommands};
    use serenity::all::InteractionType;

    use crate::database::{InteractionStepKind, InteractionStep};
    use crate::slash_command::SlashCommandFunctionality;
    use crate::payload::captcha::UpdateSettings;
        

    #[tokio::test]
    async fn redis_json_get() {
        // Init redis
        let host = "127.0.0.1";
        let port = 6379;
        let user = "default";
        let password = "development";
        let database = 0;

        let client = redis::Client::open(format!("redis://{}:{}@{}:{}/{}", user, password, host, port, database)).unwrap();
        let mut conn_manager = ConnectionManager::new_with_backoff(client, 2, 2, 5).await.unwrap();

        let object = InteractionStep::new(
            &ulid::Ulid::new().to_string(),
            InteractionStepKind::ConfirmAction { confirm_button_id: "aaa".to_string(), cancel_button_id: "bbb".to_string() },
[...]

PR #1020, opened 2024-01-09
Hide the `req_packed_commands` from docs.

nihohit:
`req_packed_commands` is meant for internal usage by the `Pipeline` object, and is an easy to misuse footgun.
See https://github.com/redis-rs/redis-rs/issues/1017#issuecomment-1878158571

### DIFF
--- a/redis/src/aio/mod.rs
+++ b/redis/src/aio/mod.rs
@@ -65,6 +65,11 @@ pub trait ConnectionLike {
     /// Sends multiple already encoded (packed) command into the TCP socket
     /// and reads `count` responses from it.  This is used to implement
     /// pipelining.
+    /// Important - this function is meant for internal usage, since it's
+    /// easy to pass incorrect `offset` & `count` parameters, which might
+    /// cause the connection to enter an erroneous state. Users shouldn't
+    /// call it, instead using the Pipeline::query_async function.
+    #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


jaymell:
What if we just update the docs to better explain the parameters and otherwise warn against using the method, as you've done here -- I wonder if that will accomplish the goal without hiding it completely? I'm not sure how likely it is, but it seems like hiding it might frustrate downstream users/libraries unnecessarily.

nihohit:
As far as I understand the #[doc(hidden)] command, it just hides this from generated documentation. It doesn't prevent usage of this function - hence, no semver-breaking tag on this PR. So, I'm not sure what might cause frustration here - users who already use this function will be able to keep using it, and those that don't (and probably shouldn't) will have a harder time being aware of the function. Anyone who wants to implement `ConnectionLike` will still be prompted by the compiler to implement these functions.
I don't have a strong opinion on this, though. If you strongly believe that we should limit this change to just a documentation update, I'm fine with that.

jaymell:
>  I don't have a strong opinion on this, though. If you strongly believe that we should limit this change to just a documentation update, I'm fine with that.

I don't have a real strong opinion either, but it feels a bit wrong to me to hide methods of public traits from the documentation. I understand the reasoning though, so not opposed to it.

jaymell (approved):

PR #1022, opened 2024-01-10
Fix documentaion error

rcl-viveksharma:
docs have 
assert_eq!(redis::cmd("GET").arg("my_key").arg("$").query_async(&mut con).await, Ok(String::from(r#"[{"item":42}]"#)));

should be 

assert_eq!(redis::cmd("JSON.GET").arg("my_key").arg("$").query_async(&mut con).await, Ok(String::from(r#"[{"item":42}]"#)));


nihohit (approved):

nihohit:
Thanks for the contribution!

PR #1029, opened 2024-01-17
Add Support for UUIDs

Rabbitminers:
This allows for UUID's to be used directly as an optional feature flag. Aside from the added ease of use this also skips the added cast to a string if we where to read the UUID as string then parse it.

In short `FromRedisValue` and `ToRedisArgs` implementations for the Uuid type from the [uuid](https://docs.rs/uuid/latest/uuid/) crate

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -355,6 +355,20 @@ impl From<rustls_pki_types::InvalidDnsNameError> for RedisError {
     }
 }
 
+#[cfg(feature = "uuid")]
+impl From<uuid::Error> for RedisError {
+    fn from(err: uuid::Error) -> RedisError {
+        RedisError {
+            repr: ErrorRepr::WithDescriptionAndDetail(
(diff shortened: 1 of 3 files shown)


nihohit (approved):

redis/src/types.rs
```diff
@@ -1616,8 +1630,28 @@ impl FromRedisValue for bytes::Bytes {
...
+        out.write_arg(self.as_bytes());
+    }
+}
+
 /// A shortcut function to invoke `FromRedisValue::from_redis_value`
 /// to make the API slightly nicer.
 pub fn from_redis_value<T: FromRedisValue>(v: &Value) -> RedisResult<T> {
     FromRedisValue::from_redis_value(v)
-}
+}
```
nihohit, line 1657: please revert this

nihohit (approved):
@Rabbitminers Thanks for the contribution! please fix the lint, and the comment.

Issue #1025, opened 2024-01-15
Async Cluster: Connection is created on each request


kamulos:
I am playing around with the async cluster timeouts, and ended up in a strange situation: for each redis request I am triggering, `redis-rs` will call `connect_and_check` and create a new connection.

Initially I got in this situation by triggering a timeout and the following reconnection also lead to a timeout. My request is a simple `GET`.

What I am observing is that in `get_connection` the branch with `connect_and_check` is taken and succeeds. But during the next request the same branch is taken, because the connection is still `None`.

Is it possible, that the connection, that is created in `connect_and_check` in `get_connection` needs to be stored back into the `conn_lock`?

My text is just wild guessing, but the issue is quite reproducible for me, so I can dig deeper if needed.

nihohit:
Sounds like you reach this line, where the connection is created but not saved - 
https://github.com/redis-rs/redis-rs/blob/a980a5fc588e7fd822da43a74208bdafb01107ec/redis/src/cluster_async/mod.rs#L873
I believe that the solution is something like 
```rs
match connect_and_check::<C>(&addr, core.cluster_params.clone()).await {
    Ok(conn) => {
        let conn_clone = conn.clone();
        core.conn_lock
            .write()
            .await
            .0
            .insert(addr.clone(), async { conn_clone }.boxed().shared());
        Some((addr, conn))
    }
    Err(_) => None,
}
```

but I can't recreate this situation in a test, so can't verify it.
Can you please manually test this fix? and if possible, create a test case that is locally reproducible?

nihohit:
Ok, different take - does this solve your issue?
https://github.com/redis-rs/redis-rs/pull/1032

IMO both fixes are correct and relevant, but I want to be sure that what I think will work actually will work :)

nihohit:
Managed to write the world's most artificial tests for the first fix, hurrah!
https://github.com/redis-rs/redis-rs/pull/1033

kamulos:
> Ok, different take - does this solve your issue? #1032
> 
> IMO both fixes are correct and relevant, but I want to be sure that what I think will work actually will work :)

Yes it does :blush: I am especially enthusiastic about this one, because it also resolves the issue of panics when I use the `DEBUG SLEEP` command on the server side.

Those two fixes are great, thank you!

kamulos:
Fixed in release 0.25.0

PR #1033, opened 2024-01-18
Save reconnected connections during retries.

nihohit:
This change ensures that reconnect attempts that happen during retries, or new connections that happen after MOVED/ASKING errors, will be saved instead of constantly reconnecting until slots are refreshed.

https://github.com/redis-rs/redis-rs/issues/1025


jaymell (approved):
Nice!

PR #1035, opened 2024-01-24
Avoid panic on connection failure

jaymell:
This code is not unreachable, so let's handle the error instead of panic.

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -384,7 +384,10 @@ impl MultiplexedConnection {
                     driver
                 }
                 futures_util::future::Either::Right(((), _)) => {
-                    unreachable!("Multiplexed connection driver unexpectedly terminated")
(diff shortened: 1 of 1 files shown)


nihohit (approved):

nihohit:
How did you reach that code?

NathanFlurry:
If the `MultiplexedConnection` panics, then all future commands also panic. This means that one bad `AUTH` and connection (and therefore the entire server) is unusable until it's restarted.

Our Redis hosting provider triggers this panic any time they enter a maintenance window. I recommend cutting a release with this fix soon. :)

---

I can consistently reproduce this issue (before this is merged) by:

1. Adding a timeout in the authenticate function to give us time to inject a fault

image 21

2. Add a timeout _after_ the authentication so the `driver` future finishes first

image 22

3. Open a connection and kill the Redis server during the `authenticate` function.

I tested this same code with this PR, can confirm that the connection recovers correctly now.

PR #1037, opened 2024-01-24
Update dependency versions and limit `half` package version.

nihohit:
Limit the `half` indirect dependency to a version compatible with our CI.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -42,7 +42,7 @@ futures-util = { version = "0.3.15", default-features = false, optional = true }
 pin-project-lite = { version = "0.2", optional = true }
 tokio-util = { version = "0.7", optional = true }
 tokio = { version = "1", features = ["rt", "net", "time"], optional = true }
-socket2 = { version = "0.4", default-features = false, optional = true }
+socket2 = { version = "0.5", default-features = false, optional = true }
 
 # Only needed for the connection manager
 arc-swap = { version = "1.1.0", optional = true }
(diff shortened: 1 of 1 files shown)


jaymell (approved):

jaymell:
Looks good, though now `anstyle` is causing similar problem. I wish I knew where this dependency is even coming from -- `cargo tree` is not very illuminating here.

nihohit:
> Looks good, though now anstyle is causing similar problem. I wish I knew where this dependency is even coming from -- cargo tree is not very illuminating here.

It's also coming from the criterion dependency.
If we take the route of this PR, we'll need to start chasing dependencies. I wonder if testing on lower rust versions is really required. Maybe we can lint (using cargo-msrv) that we're working with the declared msrv, but not run our tests on the lower rust version? that will remove the need to limit the MSRV of our dev/test dependencies - we'll only need to limit the MSRV on our core dependencies.
WDYT?

jaymell:
> Maybe we can lint (using cargo-msrv) that we're working with the declared msrv, but not run our tests on the lower rust version? that will remove the need to limit the MSRV of our dev/test dependencies - we'll only need to limit the MSRV on our core dependencies.
WDYT?

It definitely sounds like a possible option. I was also wondering whether we should consider adding a lock file, given that those are no longer ignored for libraries.

nihohit:
A lock file is also a good solution, but we will need to massage the initial lock file to get versions that are compatible with our MSRV.
Either solution is fine by me.

jaymell:
Offhand I think a proper lock file w/ test runs against different versions of Rust sounds preferable. It seems more failproof than this other mechanism. But tbh I'm not clear offhand about all the implications of a lock-file for users of the lib, so not sure if there are any problems that approach would cause.

nihohit:
trying https://github.com/redis-rs/redis-rs/pull/1039

nihohit:
using a lockfile works.

PR #1039, opened 2024-01-25
Add lock file to keep MSRV constant.

nihohit:


jaymell:
Before we merge this -- what are the potential downsides here? If I understand correctly (and I'm not sure that I do), the lockfile won't impact users of the crate, just a `cargo build` of the crate itself? If that's so, are we potentially missing problems that downstream users will encounter? I guess so, but in theory that isn't our problem. We guarantee a MSRV, but only with whatever particular versions of deps we have in our otherwise-ignored lockfile. 

Is that ok? Is there a chance we get bombarded with a lot of issues/pull requests from people who can't build the crate? 

Anyway, I think it's probably fine, I'm just trying to understand the pros/cons.

nihohit:
I asked on the Rust discord, and the suggestions I received were to either do this, increase MSRV, or fix the indirect dependencies versions.
There was no mention of other considerations, but that might just be the people on the discord at the time. Do you have a Rust expert to ask?

jaymell (approved):
Let's just give this a shot. Since the lockfile isn't actually used downstream, I think the biggest implication is that those who for whatever reason use an older version of Rust may have to jump through a few more unexpected hoops to keep their builds working. 

We should probably remember to occasionally `cargo update` things on our end, though :laughing:

PR #1042, opened 2024-01-29
Fail CI if lock file isn't updated.

nihohit:


Makefile
```diff
@@ -6,7 +6,7 @@ test:
 	@echo "===================================================================="
 	@echo "Testing Connection Type TCP without features"
 	@echo "===================================================================="
-	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 cargo test -p redis --no-default-features -- --nocapture --test-threads=1
+	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 cargo test --locked -p redis --no-default-features -- --nocapture --test-threads=1
```
jaymell, line 9: I'm not sure I fully understand what `--locked` does in the `cargo test` context. Should we consider doing this in the other runs as well? If we don't, are we still going to potentially hit the same compatibility issues with older Rust versions?

nihohit: https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options

```
--frozen
--locked
Either of these flags requires that the Cargo.lock file is up-to-date. If the lock file is missing, or it needs to be updated, Cargo will exit with an error. The --frozen flag also prevents Cargo from attempting to access the network to determine if it is out-of-date.
These may be used in environments where you want to assert that the Cargo.lock file is up-to-date (such as a CI build) or want to avoid network access.
```

It does the same thing in `cargo test` as in `cargo build` - fail the build if the lock file is out of date. 
I thought it makes sense to put it in the tests, since that will run on the oldest Rust version. We can move it to a separate step, if you prefer it out of the makefile.

> Should we consider doing this in the other runs as well?

Not sure which runs are you referring to. In the makefile we hit 8 out of 9 CI workflows. Would you like this also in the `lint` workflow? That workflow is only tested on the most recent Rust, AFAIK.

> are we still going to potentially hit the same compatibility issues with older Rust versions?

I don't think so, since this makefile is called from the workflows with the older rust version matrix.

jaymell: Sorry, I mean the specific test invocations within the Makefile. The first one -- "Testing Connection Type TCP without features" -- doesn't test any of the features. I was thinking that we're more likely to have issues w/ the subsequent invocations that explicitly use `--all-features` and bring in all the additional dependencies. But maybe it doesn't matter?

nihohit: no, that's a good point. Maybe we'll just start with `cargo build --locked --all-features`?

jaymell: Sounds good to me 👍

nihohit:
@jaymell round

jaymell:
Again, this could just be flawed understanding, but if we don't run each iteration of the tests in the Makefile with `--locked`, are we not still likely to encounter an issue related to unlocked dependency versions bumping MSRV?

nihohit:
I'm not sure that it's possible, but adding the flag doesn't cost us money, so why not?

jaymell (approved):

PR #1044, opened 2024-02-01
[opt] preallocate buffer for evalsha in `Script`

framlog:
a.t.

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -330,6 +330,13 @@ impl Cmd {
         }
     }
 
+    /// Get the capacities for the internal buffers.
+    #[cfg(test)]
+    #[allow(dead_code)]
+    pub(crate) fn capacity(&self) -> (usize, usize) {
+        (self.args.capacity(), self.data.capacity())
+    }
+
     /// Appends an argument to the command.  The argument passed must
     /// be a type that implements `ToRedisArgs`.  Most primitive types as
     /// well as vectors of primitive types implement it.
--- a/redis/src/script.rs
+++ b/redis/src/script.rs
@@ -212,12 +212,44 @@ impl<'a> ScriptInvocation<'a> {
         cmd
     }
 
+    fn estimate_buflen(&self) -> usize {
+        self
+            .keys
+            .iter()
+            .chain(self.args.iter())
+            .fold(0, |acc, e| acc + e.len())
+            + 7 /* "EVALSHA".len() */
+            + self.script.hash.len()
+            + 4 /* Slots reserved for the length of keys. */
+    }
+
     fn eval_cmd(&self) -> Cmd {
-        let mut cmd = cmd("EVALSHA");
-        cmd.arg(self.script.hash.as_bytes())
+        let args_len = 3 + self.keys.len() + self.args.len();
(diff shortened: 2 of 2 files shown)


redis/src/script.rs
```diff
@@ -213,6 +213,15 @@ impl<'a> ScriptInvocation<'a> {
...
+        let args_len = 3 + self.keys.len() + self.args.len();
+        let buf_len = self
+            .keys
+            .iter()
+            .chain(self.args.iter())
+            .fold(0, |acc, e| acc + e.len())
+            + 7 /* "EVALSHA".len() */
+            + self.script.hash.len();
+        let mut cmd = Cmd::with_capacity(args_len, buf_len);
         let mut cmd = cmd("EVALSHA");
```
nihohit, line 225: this overwrites the previous `cmd`, and you've essentially done nothing. you'll need to replace this with `arg`.

framlog: 0.0..

Sorry, it's a mistake...

redis/src/script.rs
```diff
@@ -213,7 +213,15 @@ impl<'a> ScriptInvocation<'a> {
...
+        let args_len = 3 + self.keys.len() + self.args.len();
+        let buf_len = self
+            .keys
+            .iter()
+            .chain(self.args.iter())
+            .fold(0, |acc, e| acc + e.len())
+            + 7 /* "EVALSHA".len() */
+            + self.script.hash.len();
+        let mut cmd = Cmd::with_capacity(args_len, buf_len);
         cmd.arg(self.script.hash.as_bytes())
```
nihohit, line 225: now you didn't set the "EVALSHA" argument, and this will fail. 😸

framlog: 🧎

I'll fix that.

redis/src/script.rs
```diff
@@ -212,12 +212,44 @@ impl<'a> ScriptInvocation<'a> {
...
 
+    fn estimate_buflen(&self) -> usize {
+        self
+            .keys
+            .iter()
+            .chain(self.args.iter())
+            .fold(0, |acc, e| acc + e.len())
+            + 7 /* "EVALSHA".len() */
+            + self.script.hash.len()
+            + 1 /* Require a little more since rustc seems to reallocate vecs when length matches exactly. */
```
nihohit, line 223: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=26ff6d3d5ddfaf6ad99aeb0c11c20679

I don't see this happening. Can you please explain?

framlog: weird...

framlog: I found the real reason is the length of keys.

redis/src/cmd.rs
```diff
@@ -330,6 +330,11 @@ impl Cmd {
         }
     }
 
+    /// Get the capacities for the internal buffers.
+    pub fn capacity(&self) -> (usize, usize) {
```
nihohit, line 334: please change to pub(crate) and add #[cfg(test)]

nihohit (approved):
Thanks for the contribution! It was a bit of a process, but I think this will be a good perf boost :)

Issue #984, opened 2023-11-08
Connection not usable after parsing error. "broken pipe" when connection manager is used.


tsmt-iris:
Hey

We have an application making large use of an async connection pool, timeseries and pipelines. Under the following conditions, a connection from the connection pool can "break" and result in a "broken pipe" error on next connection use:

- pipeline of `TS.RANGE` commands, which fail because the key is not available. `ERR TSDB: the key does not exist`
- pipeline **must be atomic**
- it must be **multiple commands** (single command in atomic seems to be ok)

I think timeseries is not really the issue here, but actually the fact that the connection breaks if the parsing fails and the parsing error is happening with the timeseries responses.

I was able to reproduce it in a minimal redis-rs repo. Notice that this won't produce the **broken pipe** error, but the check_connection() indicates that after the call is connection is somehow broken. Also, the `value: parse error` in the unwrap is something that we can see in our larger async system. It's just that the connection is not really usable anymore after the `pipe-multi-atomic` call (last one before the unwrap()).

It seems that in our larger system with connection manager this leads to a `broken pipe` error when the connection is checked out the next time. After we removed the `atomic()` in our larger systems, we did not see these errors anymore.

```rust
fn main() -> Result<(), Box<dyn std::error::Error>> {
    let client = Client::open("redis://127.0.0.1:6379")?;
    let mut connection = client.get_connection().unwrap();
    let flush = redis::cmd("FLUSHALL");

    println!("connection status pre: {}", connection.check_connection());
    let _: Result<Value, RedisError> = redis::cmd("TS.RANGE").arg("NOT_AVAILABLE").arg("-").arg("+").query(&mut connection);
    println!("connection status post single: {}", connection.check_connection());

    let _: Result<Value, RedisError> = redis::pipe().cmd("TS.RANGE").arg("NOT_AVAILABLE").arg("-").arg("+").query(&mut connection);
    println!("connection status post pipe-single: {}", connection.check_connection());

    let _: Result<Value, RedisError> = redis::pipe().cmd("TS.RANGE").arg("NOT_AVAILABLE").arg("-").arg("+").atomic().query(&mut connection);
    println!("connection status post pipe-single-atomic: {}", connection.check_connection());

    let _: Result<Value, RedisError> = redis::pipe().cmd("TS.RANGE").arg("NOT_AVAILABLE").arg("-").arg("+").cmd("TS.RANGE").arg("NOT_AVAILABLE").arg("-").arg("+").query(&mut connection);
[...]

PR #1051, opened 2024-02-08
Reconnect on parsing errors.

nihohit:
Parsing errors mean that the connection received a response it doesn't know how to handle. This means that it cannot make sense of the next values sent over the connection, and the connection must be replaced.

related:
https://github.com/redis-rs/redis-rs/issues/984#issuecomment-1930956524


jaymell (approved):
Looks good!

PR #1053, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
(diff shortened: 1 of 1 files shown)


jaymell:
Hi -- I think this is probably ok. The main reason I've been avoiding it is just b/c is 1) ignorance, 2) I don't want a lot of noise PRs that would cause us to inadvertently bump MSRV, but overall this is probably a good thing.

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
```
jaymell, line 8: I'm not clear -- what does package-ecosystem `docker` do here?

oriontvv:
> that would cause us to inadvertently bump MSRV

We can ignore specific dependency if you wish so. Just tell the list of them.

.github/dependabot.yml
jaymell: Should we remove it then?

oriontvv: done

jaymell (approved):
Alright, thanks! Let's see what happens :crossed_fingers:

nihohit:
Thanks for the contribution!

PR #1056, opened 2024-02-12
Handle server errors in array response.

nihohit:
Currently server errors stop the parser and return a RedisError. This caused errors that are returned inside an array, such as transaction errors, to cause the rest of the array to not be parsed. 
This is solved by adding an internal value type that includes the server errors, so when parsing to the internal value type, the array will finish parsing, and then extracting the error.

Pros:
- backwards compatible
- allows us to replace the parsing backend (for [redis-protocol](https://crates.io/crates/redis-protocol)?) without affecting the users.

cons:
- repetition - some changes to Value might need to be copied to InternalValue.
- adds some conversion overhead

### DIFF
--- a/redis/src/parser.rs
+++ b/redis/src/parser.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use crate::types::{
-    make_extension_error, ErrorKind, InternalValue, RedisError, RedisResult, Value,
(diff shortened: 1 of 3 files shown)


nihohit:
possible solution to https://github.com/redis-rs/redis-rs/issues/984

nihohit:
@jaymell ready for review

jaymell (approved):

PR #1057, opened 2024-02-12
Handle server errors in array response 2.

nihohit:
Currently server errors stop the parser and return a RedisError. This caused errors that are returned inside an array, such as transaction errors, to cause the rest of the array to not be parsed. This is solved by adding a `ServerError` variant to `Value` that contains the server errors, so when parsing to the server message, the array will finish parsing, and then extracting the error.

Pros:
- Gives the user/contributor a separation between errors returned from the server and connection / logic errors.
- Gives us the ability to return the actual array of responses on pipe operations, which allows further improvements down the line. For example, the `setup_connection` functions could be made faster by using a pipeline.

Cons:
- not backwards compatible - users who `match` on Value will need to handle the extra variant. This will be less painful if we ship this in the same version with the RESP3 update.

### DIFF
--- a/redis/src/parser.rs
+++ b/redis/src/parser.rs
@@ -55,6 +55,7 @@ where
 
 const MAX_RECURSE_DEPTH: usize = 100;
 
+//TODO - change Output to ValueInternal instead of result
 fn value<'a, I>(
     count: Option<usize>,
 ) -> impl combine::Parser<I, Output = RedisResult<Value>, PartialState = AnySendSyncPartialState>
@@ -105,7 +106,7 @@ where
                 let data = || {
                     int().then_partial(move |size| {
                         if *size < 0 {
-                            combine::value(Value::Nil).left()
+                            combine::produce(|| Value::Nil).left()
                         } else {
                             take(*size as usize)
                                 .map(|bs: &[u8]| Value::Data(bs.to_vec()))
@@ -118,7 +119,7 @@ where
                 let bulk = || {
                     int().then_partial(move |&mut length| {
                         if length < 0 {
-                            combine::value(Value::Nil).map(Ok).left()
+                            combine::produce(|| Value::Nil).map(Ok).left()
                         } else {
                             let length = length as usize;
                             combine::count_min_max(length, length, value(Some(count + 1)))
@@ -159,7 +160,7 @@ where
                     b':' => int().map(|i| Ok(Value::Int(i))),
                     b'$' => data().map(Ok),
                     b'*' => bulk(),
-                    b'-' => error().map(Err),
+                    b'-' => error().map(|err|Ok(Value::ServerError(err))),
                     b => combine::unexpected_any(combine::error::Token(b))
                 )
             })
@@ -207,7 +208,7 @@ mod aio_support {
 
             bytes.advance(removed_len);
             match opt {
-                Some(result) => Ok(Some(result)),
+                Some(result) => Ok(Some(result.and_then(Value::remove_error))),
                 None => Ok(None),
             }
         }
@@ -260,7 +261,7 @@ mod aio_support {
                     }
                 }
             }),
-            Ok(result) => result,
+            Ok(result) => result.and_then(Value::remove_error),
         }
     }
 }
@@ -318,7 +319,7 @@ impl Parser {
                     }
                 }
             }),
-            Ok(result) => result,
+            Ok(result) => result.and_then(Value::remove_error),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


nihohit:
possible solution to https://github.com/redis-rs/redis-rs/issues/984

jaymell:
I think both of the drafts look nice, but I'm not clear what's gained by exposing the `ServerError` variant to the user if we ultimately surface it as a `RedisError` anyway.

nihohit:
> if we ultimately surface it as a RedisError anyway.

that's the question - **if** we want to break backwards compatibility / expose a new interface, then the user can receive a `Vec<Value>` from a pipeline operation, and know exactly which operations failed and which succeeded.

jaymell:
> if we want to break backwards compatibility / expose a new interface, then the user can receive a Vec<Value> from a pipeline operation, and know exactly which operations failed and which succeeded.

But this particular PR does not get us that, right? If not, I don't see the advantage to changing `Value`, though I'm probably missing something.

nihohit:
> But this particular PR does not get us that, right? If not, I don't see the advantage to changing Value, though I'm probably missing something.

no, you're not missing anything :). This PR only adds the variant to the enum, and populates that variant in the parser - it enables such future change, although it isn't a lot of work -  in order to allow users to access the errors we'd need remove the various `.and_then(Value::remove_error)` calls from the parser, and that's it... and change a lot of tests, because it's backwards incompatible with the existing behavior.

jaymell:
I'm pretty sure we've discussed this elsewhere, but what does the user experience look like if we have `Value::ServerError`? It seems like a potentially large change to the library, so maybe the non-breaking PR is the better approach if we want this relatively quickly. What do you think?

nihohit:
> what does the user experience look like if we have Value::ServerError?

I hope that this is what you mean - it allows the user to tailor their error handling according to which command failed.
```rs
let pipe = pipe();
pipe.set("foo", "bar");
pipe.set("baz", "foobar");

let results: Vec<Value> = pipe.query(&mut conn)?;

let backup_pipe = pipe();
// here we check which operations failed
if results[0].is_error() {
  // set the first value
}
if results[1].is_error() {
  // set the second value
}
```

or, for a more concrete example - https://github.com/redis-rs/redis-rs/blob/26c977c517c2c8ea4668a8e99eb71b67ae75d13f/redis/src/connection.rs#L952 the `setup_connection` function sends multiple back-and-forth requests to the server, which could've been sent as a single pipeline if we had errors per command in the pipeline. We don't, and since each failed request has its own error handling, we can't use a pipeline there, which I estimate makes the setup process take 3x longer than it should. That 3x will grow with the amount of things we automatically do on connection - for example, if we wanted to automatically reconnect to pubsub channels, or set client side caching, then each additional step is another back-and-forth to the server, because we can't pipeline the results.

> It seems like a potentially large change to the library, so maybe the non-breaking PR is the better approach if we want this relatively quickly. What do you think?

We could definitely do the second change first, and then implement this change later if we choose to. They're not mutually exclusive.
I do believe that eventually we'll want this - this change has been requested - https://github.com/redis-rs/redis-rs/issues/805 , https://github.com/redis-rs/redis-rs/pull/813, so there is user demand to it, and IMO if we introduce it with the resp3 merge it will be less jarring to users.

nihohit:
1. True. This PR solves the parser misbehaving on pipeline error. Of the 2 proposed solution, the upside of this PR is that it brings us closer to solving other issues, but doesn't actually solve them.

2. That's a good question. Both behaviors are possible. I tend towards `Ok(Value::ServerError)`, which is the greater breaking behavior, but:
* it's consistent with the behavior of the pipeline
* you can always do `con.incr("test-key", 1).and_then(Value::remove_error)` and get the old behavior.

kamulos:
I think in general it is the correct solution for the future to expose the `ServerError` in some way to the user of the library. I am not entirely sure if `ServerError` should contain the whole `RedisError`, because states like `ServerError(IoError)` do not make any sense, but would be representable with this solution, I think.

Just as one more datapoint, why it is nice to have this modeled very closely to the behavior of Redis: #813 was opened by a formed colleague of mine. At that point we were not using cluster, but needed more performance. So we started collecting requests for a defined time on the client side and sent them as one pipeline. The answers (and errors) were redistributed to all the tasks. This traded a bit of increased latency for each request against a much higher throughput.

nihohit:
will implement this after merging RESP3.

Abioy:
> I think in general it is the correct solution for the future to expose the `ServerError` in some way to the user of the library. I am not entirely sure if `ServerError` should contain the whole `RedisError`, because states like `ServerError(IoError)` do not make any sense, but would be representable with this solution, I think.

IMHO, exposing `IoError`s which take place during a pipeline that mainly contains read cmds (get/hget/...) would be quite useful. 'Cause it allows user to re-send failing cmds only, and thus reduces unnecessary requests.

nihohit:
@Abioy I see what you mean, but alas, the pipeline API is oriented towards complete results - solving the issue of IoErrors in the middle of pipelines means also handling receiving only some of the responses and then disconnecting, timing out in the middle of a pipeline, and other similar cases that require handling of partial results. I think that if we want to handle partial responses to a pipeline we need to properly design around this scenario - just adding IoErrors won't solve the issue.

Abioy:
@nihohit Is it acceptable to add a new `query` API that returns `RedisResult<Vec<RedisResult<Value>>>` for pipeline? 

If we treat pipeline just as an underlying method to send multi cmds, maybe keeping all responses(including error ones) in a vector would be more natural?

nihohit:
> @nihohit Is it acceptable to add a new query API that returns RedisResult<Vec<RedisResult<Value>>> for pipeline?

What does the external RedisResult mean, and what do the internal ones mean?

Abioy:
> What does the external RedisResult mean, and what do the internal ones mean?

The external RedisResult stands for a completely fail/error. e.g. The connection is stuck in pubsub state,

nihohit:
Off the top of my head this sounds reasonable. If you want to whip up a prototype, let's see if it's easy to implement.

Abioy:
@nihohit I just make a pull request, would you mind reviewing this https://github.com/redis-rs/redis-rs/pull/1385

PR #1058, opened 2024-02-13
Merge the `resp3` branch.

nihohit:


nihohit:
@altanozlu @jaymell 
since the resp3 branch was long-living, this wasn't a simple merge. I'd appreciate your review.

redis/tests/support/mod.rs
```diff
@@ -540,20 +581,48 @@ where
...
+        Value::Attribute {
+            ref data,
+            ref attributes,
+        } => {
+            encode_map(attributes, writer, "|")?;
+            encode_value(data, writer)?;
+            Ok(())
+        }
+        Value::Set(ref values) => encode_iter(values, writer, "~"),
+        // Value::Nil => write!(writer, "_\r\n"), //TODO is it okey to use $-1 in resp3 ?
```
altanozlu, line 602: I think this can be removed :D

altanozlu (commented):
Really nice PR, it seems like I need to improve my git skills :D 
I'm currently working on client side caching and I have a working prototype. It seems like client side caching won't brake much things, so if we feel this RESP3 model is fine I don't see any reason to not merge.

redis/src/types.rs
```diff
@@ -147,44 +152,181 @@ pub enum Value {
     /// is why this library generally treats integers and strings
     /// the same for all numeric responses.
     Int(i64),
-    /// An arbitary binary data.
-    Data(Vec<u8>),
-    /// A bulk response of more data.  This is generally used by redis
+    /// An arbitrary binary data, usually represents a binary-safe string.
+    BulkString(Vec<u8>),
```
jaymell, line 156: I'm just looking at this in the context of the resp3 spec again. Should we call this `BlobString` to be more consistent w/ resp3 terminology?

nihohit: So, [here](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md) it's called a blob string, while [here](https://redis.io/docs/reference/protocol-spec/) it's called a bulk string. I'm inclined to prefer the documentation over the .md file, but mostly because I'm lazy 🤷

jaymell:
Let's do it!

jaymell (approved):

jaymell:
:heavy_check_mark:

PR #1065, opened 2024-03-05
Simplify Sink polling logic.

nihohit:
This removes unnecessary `match`es and `map` from the code, and moves the usage of `poll_recover` into `poll_flush`, so as not to block new requests while trying to recover a connection.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -1039,42 +1039,30 @@ where
         Ok((addr, conn))
     }
 
-    fn poll_recover(
-        &mut self,
-        cx: &mut task::Context<'_>,
-        future: RecoverFuture,
-    ) -> Poll<Result<(), RedisError>> {
-        match future {
-            RecoverFuture::RecoverSlots(mut future) => match future.as_mut().poll(cx) {
-                Poll::Ready(Ok(_)) => {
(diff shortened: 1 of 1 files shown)


redis/src/cluster_async/mod.rs
```diff
@@ -1039,42 +1039,30 @@ where
...
+        };
+        match recover_future {
+            RecoverFuture::RecoverSlots(ref mut future) => match ready!(future.as_mut().poll(cx)) {
+                Ok(_) => {
                     trace!("Recovered!");
                     self.state = ConnectionState::PollComplete;
                     Poll::Ready(Ok(()))
                 }
-                Poll::Pending => {
-                    self.state = ConnectionState::Recover(RecoverFuture::RecoverSlots(future));
```
jaymell, line 1055: What are the consequences for the recovery action if we no longer reset the state here when the future is still pending?

nihohit: notice that I removed the `replace` call in line 1292 - `self.state` will remain `ConnectionState::Recover(RecoverFuture::RecoverSlots(future))`, instead of moving the value out and then back in.

jaymell:
Overall the code simplification here looks good. I'm not clear what the practical impact of the changes here are, but it seems reasonable to me to not have to potentially poll repeatedly just to enqueue commands.

jaymell (approved):

nihohit:
We've tested this on our internal failover framework, and saw no difference in availability.

PR #1067, opened 2024-03-05
fix #1066: impl `Script::from_hash`

### LINKED ISSUE #1066: Script object from hash
When instansiating `Script` object, one must suppy the script's code (`pub fn new(code: &str) -> Script`).
However, my use case is that a script is already loaded, and it's hash is given, and I need to call it. Surprisingly, theer is no API to create `Script` object given that hash.

tomers:
Allow for instantiation of a Script object given a hash.

### DIFF
--- a/redis/src/script.rs
+++ b/redis/src/script.rs
@@ -9,7 +9,7 @@ use crate::Cmd;
 /// Represents a lua script.
 #[derive(Debug, Clone)]
 pub struct Script {
-    code: String,
+    code: Option<String>,
     hash: String,
 }
 
@@ -34,8 +34,16 @@ impl Script {
         let mut hash = Sha1::new();
         hash.update(code.as_bytes());
         Script {
-            code: code.to_string(),
+            code: Some(code.to_string()),
             hash: hash.digest().to_string(),
+        }
+    }
+
+    /// Creates a new script object given hash.
+    pub fn from_hash(hash: &str) -> Script {
+        Script {
+            code: None,
+            hash: hash.to_string(),
         }
     }
 
@@ -103,6 +111,12 @@ impl Script {
         .invoke_async(con)
         .await
     }
+
+    /// Checks whether script is loadable.
+    #[inline]
+    pub fn is_loadable(&self) -> bool {
+        self.code.is_some()
+    }
 }
 
 /// Represents a prepared script call.
@@ -146,8 +160,8 @@ impl<'a> ScriptInvocation<'a> {
         match eval_cmd.query(con) {
             Ok(val) => Ok(val),
             Err(err) => {
-                if err.kind() == ErrorKind::NoScriptError {
-                    self.load_cmd().query(con)?;
+                if err.kind() == ErrorKind::NoScriptError && self.script.is_loadable() {
+                    self.load_cmd()?.query(con)?;
                     eval_cmd.query(con)
                 } else {
                     Err(err)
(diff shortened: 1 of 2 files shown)


nihohit:
This looks like a bit of a footgun, so I'd like to understand the reasons for this.
AFAIS the main benefit of  `Script` & `ScriptInvocation` usage is automatic reloading of the script. Without that it's just a wrapper to `Cmd`. In your usecase, what're you gaining from using `Script` and not `Cmd`?

tomers:
@nihohit,
Thanks for the prompt response!
My use case is that I have a rather large script which is already loaded to the database.
In a separate process, I would like to execute that script from Rust code. I do not want to load it with all the script's code, as I know for sure that it is already loaded in Redis. I just want to get a Script object so I can invoke, and I want to get this object given the hash, which I got from elsewhere.

nihohit:
Right, but why do you need `Script` for that? Can't you just use `Cmd` with the EVALSHA command?

tomers:
What is the purpose of this library, then, if not to provide users with
better API.
I think the following code is rather low level, and we should try to
prevent our users from having to write such code:
let mut cmd = Cmd::with_capacity(args_len, self.estimate_buflen());
cmd.arg("EVALSHA")
.arg(self.script.hash.as_bytes())
.arg(self.keys.len())
.arg(&*self.keys)
.arg(&*self.args);

nihohit:
> What is the purpose of this library, then, if not to provide users with better API.

This is exactly what I'm doing here - making sure that our users get better API. In this case, preventing the creation of `Script` objects that don't serve their purpose - abstracting the full lifetime of a script, not only the EVALSHA calls. As I said, having a script object that might reload or might not, is a footgun. I prefer to minimize the potential for mistakes in our APIs, when there's no significant gain to offset the risk.

If you feel that the code you presented is too verbose, why not wrap it in a function? Or if you think that an object will serve the purpose better, let's add another object, with a its own distinct purpose.

tomers:
Introducing a new class, `CachedScript` will still inflict changes on `ScriptInvocation`, which will need to support both `Script` and `CachedScript`. Some functionality (load) exist only in `Script` and not in `CachedScript`, and some exist in both, so `ScriptInvocation::load` and `ScriptInvocation::load_async` will only work with `Script` and not with `CachedScript`.

As far as I see, this solution has the same issues my original solution has. To keep backward compatibility by keeping current API would be hard and result in a convoluted code. I guess we can use traits to make some implementation possible, but I am not sure at all that the code will remain simple and readable.

I find the workarouing to just use `Cmd` with `EVALSHA` command just as well missing the purpose of making it easier on users.

I am puzzled here, and would like to get some suggested, @nihohit. wdyt?

redis/src/script.rs
```diff
@@ -172,8 +186,8 @@ impl<'a> ScriptInvocation<'a> {
             }
             Err(err) => {
                 // Load the script into Redis if the script hash wasn't there already
-                if err.kind() == ErrorKind::NoScriptError {
-                    self.load_cmd().query_async(con).await?;
+                if err.kind() == ErrorKind::NoScriptError && self.script.is_loadable() {
```
jaymell, line 189: I'm not sure the extra condition here adds any value.

redis/src/script.rs
```diff
@@ -146,8 +160,8 @@ impl<'a> ScriptInvocation<'a> {
         match eval_cmd.query(con) {
             Ok(val) => Ok(val),
             Err(err) => {
-                if err.kind() == ErrorKind::NoScriptError {
-                    self.load_cmd().query(con)?;
+                if err.kind() == ErrorKind::NoScriptError && self.script.is_loadable() {
```
jaymell, line 163: I'm not sure the extra condition here adds any value.

jaymell:
To me this change looks okay. I question the utility of a lot of the higher-level APIs in this library :laughing:, but I don't see the danger in adding this. It should continue to work as it did before for users of the existing API and otherwise provides a bit of convenience for those that would like to BYO-SHA. 

Some possibilities:
* `from_hash` could require a call to `SCRIPT EXISTS` before returning? 
* A much-beloved mutable method to explicitly load the script on the `Script` struct after instantiation?

nihohit:
@tomers why not just add an `evalsha(hash, keys, args)` command? what purpose does adding/changing a struct serve?

nihohit:
closing, since the discussion hasn't progressed in 2 years.

PR #1074, opened 2024-03-11
Avoid library dependency on futures-time

djc:
The futures-time crate brings in a bunch of extra dependencies from the async-std ecosystem even when this crate is configured to use `tokio-comp`. Additionally, it hasn't had a release in 2 years, so it depends on older versions of its transitive dependencies (for example, rustix). It is relatively straightforward to replace it with some direct calls to the runtime.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -86,7 +86,6 @@ num-bigint = { version = "0.4.4", optional = true }
 ahash = { version = "0.8.6", optional = true }
 
 log = { version = "0.4", optional = true }
-futures-time = { version = "3.0.0", optional = true }
 
 # Optional uuid support
 uuid = { version = "1.6.1", optional = true }
@@ -94,7 +93,7 @@ uuid = { version = "1.6.1", optional = true }
 [features]
 default = ["acl", "streams", "geospatial", "script", "keep-alive"]
 acl = []
-aio = ["bytes", "pin-project-lite", "futures-util", "futures-util/alloc", "futures-util/sink", "tokio/io-util", "tokio-util", "tokio-util/codec", "tokio/sync", "combine/tokio", "async-trait", "futures-time"]
+aio = ["bytes", "pin-project-lite", "futures-util", "futures-util/alloc", "futures-util/sink", "tokio/io-util", "tokio-util", "tokio-util/codec", "tokio/sync", "combine/tokio", "async-trait"]
 geospatial = []
 json = ["serde", "serde/derive", "serde_json"]
 cluster = ["crc16", "rand"]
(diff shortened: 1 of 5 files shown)


redis/src/types.rs
```diff
@@ -827,6 +830,7 @@ impl RedisError {
             ErrorKind::ParseError => RetryMethod::Reconnect,
             ErrorKind::AuthenticationFailed => RetryMethod::Reconnect,
             ErrorKind::ClusterConnectionNotFound => RetryMethod::Reconnect,
+            ErrorKind::TimeoutElapsed => RetryMethod::Reconnect,
```
nihohit, line 833: why reconnect?

redis/src/types.rs
```diff
@@ -133,6 +133,8 @@ pub enum ErrorKind {
     NotBusy,
     /// Used when a cluster connection cannot find a connection to a valid node.
     ClusterConnectionNotFound,
+    /// A timeout elapsed before the operation could complete.
+    TimeoutElapsed,
```
nihohit, line 137: don't forget the `is_timeout` function.

redis/Cargo.toml
```diff
@@ -131,6 +130,7 @@ socket2 = "0.4"
 assert_approx_eq = "1.0"
 fnv = "1.0.5"
 futures = "0.3"
+futures-time = "3"
```
nihohit, line 133: let's remove it entirely

nihohit (commented):
I'm not very keen on having 2 representations of timeouts in RedisError.
Except for this and the failing CI, looks ok.

redis/src/types.rs
djc: Not sure, what should it be instead?

redis/src/types.rs
djc: Fixed.

redis/Cargo.toml
djc: How do you want to do this? `test_block_on_all_panics_from_spawns()` seems to already rely on tokio unconditionally (via the use of `tokio::task::spawn()`), is it okay to do that for the `sleep()` calls, too? `Runtime` isn't a part of the public API (and I'm honestly not motivated to spend a lot of time on cleaning up the tests, since the dev-dependencies don't bother me).

djc:
> I'm not very keen on having 2 representations of timeouts in RedisError.

How would you like to represent it, instead?

redis/Cargo.toml
nihohit: fair enough, not critical.

nihohit:
> How would you like to represent it, instead?

Using the existing timeout representation - as an IoError. and If we do duplicate this error, then both should behave the same on all RedisError functions.

nihohit (approved):

nihohit:
Thanks for the improvement!

djc:
Would it be possible to get a quick follow-up release with this and #1073?

nihohit:
@djc release 0.25.1

djc:
Thanks!

Issue #1078, opened 2024-03-14
Bug with multiplexed async connections in combination with pipeline (redis::pipe)?


brocaar:
I'm storing metrics data in Redis. Depending the time-range, this will retrieve one or multiple keys for which using `redis::pipe()` to `HGETALL` these keys in one go. E.g. these keys could be `metrics:20240310`, `metrics:20240311`, `metrics:20240312`. This was all working very well until I migrated the async connections to async multiplexed connections after upgrading to v0.25 and seeing the deprecation notices.

Suddenly I got errors like:

> Response was of incompatible type - TypeError: "Response type not hashmap compatible" (response was string-data('"field1"'))

When using  `get_async_connection()`, all was working fine again. This is very simple to reproduce, see the below code example. 

### Code example to reproduce the issue

```rust
use std::collections::HashMap;

/* In Redis:

redis> HSET myhash field1 "Hello"
(integer) 1
redis> HSET myhash field2 "World"
(integer) 1
redis> HGETALL myhash
1) "field1"
2) "Hello"
3) "field2"
4) "World"

redis>
*/

#[tokio::main]
async fn main() {
    let client = redis::Client::open("redis://127.0.0.1/").unwrap();

    // This works
    let mut con = client.get_async_connection().await.unwrap();

    let mut pipe = redis::pipe();
    pipe.cmd("HGETALL").arg("myhash");

    let res: Vec<HashMap<String, String>> = pipe.query_async(&mut con).await.unwrap();

    println!("{:?}", res);

    // This fails
    // The code is exactly the same, but we are using a multiplexed async connection
    let mut con = client.get_multiplexed_async_connection().await.unwrap();

    let mut pipe = redis::pipe();
    pipe.cmd("HGETALL").arg("myhash");

    let res: Vec<HashMap<String, String>> = pipe.query_async(&mut con).await.unwrap();

    println!("{:?}", res);
}
```

This outputs:

```
warning: use of deprecated method `redis::Client::get_async_connection`: aio::Connection is deprecated. Use client::get_multiplexed_async_connection instead.
 --> src/main.rs:6:26
  |
6 |     let mut con = client.get_async_connection().await.unwrap();
  |                          ^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default

warning: `redis-test` (bin "redis-test") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 1.66s
[...]

PR #1079, opened 2024-03-15
MultiplexedConnection: Separate response handling for pipeline.

nihohit:
https://github.com/redis-rs/redis-rs/issues/1078
This fixes a bug where pipeline responses for a single request that return a `Vec` as that single value aren't wrapped in another `Vec`.
Since this is the second bug introduced by https://github.com/redis-rs/redis-rs/commit/70dfb95713fbcfe944a9262189c2adf23c12977c,
while we still want to maintain the performance improvement for single requests introduced by same, the fix is to separate the logic of single responses from pipeline responses.

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -29,31 +29,41 @@ use tokio_util::codec::Decoder;
 // Senders which the result of a single request are sent through
 type PipelineOutput = oneshot::Sender<RedisResult<Value>>;
 
-struct InFlight {
-    output: PipelineOutput,
-    expected_response_count: usize,
-    current_response_count: usize,
-    buffer: Option<Value>,
-    first_err: Option<RedisError>,
+enum ResponseAggregate {
+    SingleCommand,
+    Pipeline {
+        expected_response_count: usize,
+        current_response_count: usize,
+        buffer: Vec<Value>,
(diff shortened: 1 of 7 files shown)


nihohit:
@jaymell another reason for the disambiguation between pipeline & single response is that the `Value::ServerError` change will introduce further complexity there, especially around ExecAbort errors, so I believe that this can help with that change, too.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -214,9 +210,21 @@ where
 
         match self_.sink_stream.start_send(input) {
             Ok(()) => {
-                self_
-                    .in_flight
-                    .push_back(InFlight::new(output, response_count));
+                let response_aggregate = match pipeline_response_count {
+                    Some(response_count) => ResponseAggregate::Pipeline {
```
jaymell, line 214: I wonder if this code should just go in a `new` method on `ResponseAggregate`

jaymell (approved):

jaymell:
Shall we yank 0.25.0/1 after this is released?

Let's make sure not to squash this when merging :pray:

nihohit:
> Shall we yank 0.25.0/1 after this is released?

I think so, although I can't say that I completely understand the ramifications of yanking a version. What do you think?

PR #1084, opened 2024-03-19
Wrap tests with modules.

nihohit:
This will allow us to run groups of async tests separately - `cargo test cluster_async` / c`argo test async`, etc.

### DIFF
--- a/redis/tests/test_async.rs
+++ b/redis/tests/test_async.rs
@@ -1,1059 +1,1068 @@
-use std::collections::HashMap;
-
-use futures::{prelude::*, StreamExt};
-use redis::{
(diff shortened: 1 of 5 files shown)


jaymell:
This is fine, I'm just wondering if there's a way we can break this into multiple commits to make the diff less horrendous.

nihohit:
I can split it into a commit for each file, but IMo it doesn't add any meaning or order.

Dav1dde:
> This is fine, I'm just wondering if there's a way we can break this into multiple commits to make the diff less horrendous.

Hide whitespace helps a lot:
image

jaymell:
My apologies -- I thought I had done so :facepalm:

jaymell (approved):

PR #1085, opened 2024-03-19
Add vscode settings to gitignore.

nihohit:


jaymell:
Personally I would rather we just a git-ignore entry for `.vscode`. That way we can have our own config and not have to worry about it overwriting / being overwritten by whatever's committed.

nihohit:
done.

.gitignore
```diff
@@ -2,3 +2,4 @@ build
 lib
 target
 .rust
+.vscode/settings.json
```
jaymell, line 5: I would just do `.vscode`

nihohit: great idea!

PR #1088, opened 2024-03-20
feat: add implementation of `ToRedisArgs`,`FromRedisValue` traits for `Arc<T>`,`Box<T>`,`Rc<T>`

xoac:
With these macros(kudos to serde crate) it will be very easy to add other std types like `Mutex`,`Rc`,`ArcWeak` 

I think this PR might contains breaking changes for some users because from now you need to specify output type. See `test_vec` function.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1976,70 +1976,81 @@ macro_rules! forwarded_impl {
 forwarded_impl!((T), Box<T>, Box::new);
 forwarded_impl!((T), std::sync::Arc<T>, std::sync::Arc::new);
 
-impl<T: FromRedisValue> FromRedisValue for Vec<T> {
-    fn from_redis_value(v: &Value) -> RedisResult<Self> {
-        match v {
-            // All binary data except u8 will try to parse into a single element vector.
-            // u8 has its own implementation of from_byte_vec.
-            Value::BulkString(bytes) => match FromRedisValue::from_byte_vec(bytes) {
-                Some(x) => Ok(x),
-                None => invalid_type_error!(
-                    v,
-                    format!("Conversion to {} failed.", std::any::type_name::<Self>())
-                ),
-            },
-            Value::Array(items) => FromRedisValue::from_redis_values(items),
-            Value::Set(ref items) => FromRedisValue::from_redis_values(items),
-            Value::Map(ref items) => {
-                let mut n: Vec<T> = vec![];
-                for item in items {
-                    match FromRedisValue::from_redis_value(&Value::Map(vec![item.clone()])) {
-                        Ok(v) => {
-                            n.push(v);
+/// Implement `FromRedisValue` for `$Type` (which should use the generic parameter `$T`).
+///
+/// The implementation parses the value into a vec, and then passes the value through `$convert`.
+/// If `$convert` is ommited, it defaults to `Into::into`.
+macro_rules! from_vec_from_redis_value {
+    (<$T:ident> $Type:ty) => {
+        from_vec_from_redis_value!(<$T> $Type; Into::into);
+    };
+
+    (<$T:ident> $Type:ty; $convert:expr) => {
+        impl<$T: FromRedisValue> FromRedisValue for $Type {
+            fn from_redis_value(v: &Value) -> RedisResult<$Type> {
+                match v {
+                    // All binary data except u8 will try to parse into a single element vector.
+                    // u8 has its own implementation of from_byte_vec.
+                    Value::BulkString(bytes) => match FromRedisValue::from_byte_vec(bytes) {
+                        Some(x) => Ok($convert(x)),
+                        None => invalid_type_error!(
+                            v,
+                            format!("Conversion to {} failed.", std::any::type_name::<$Type>())
+                        ),
+                    },
+                    Value::Array(items) => FromRedisValue::from_redis_values(items).map($convert),
+                    Value::Set(ref items) => FromRedisValue::from_redis_values(items).map($convert),
+                    Value::Map(ref items) => {
+                        let mut n: Vec<T> = vec![];
+                        for item in items {
+                            match FromRedisValue::from_redis_value(&Value::Map(vec![item.clone()])) {
+                                Ok(v) => {
+                                    n.push(v);
+                                }
+                                Err(e) => {
+                                    return Err(e);
+                                }
+                            }
                         }
-                        Err(e) => {
-                            return Err(e);
-                        }
+                        Ok($convert(n))
                     }
+                    Value::Nil => Ok($convert(Vec::new())),
+                    _ => invalid_type_error!(v, "Response type not vector compatible."),
                 }
-                Ok(n)
             }
-            Value::Nil => Ok(Vec::new()),
-            _ => invalid_type_error!(v, "Response type not vector compatible."),
-        }
-    }
-    fn from_owned_redis_value(v: Value) -> RedisResult<Self> {
-        match v {
-            // Binary data is parsed into a single-element vector, except
-            // for the element type `u8`, which directly consumes the entire
-            // array of bytes.
-            Value::BulkString(bytes) => FromRedisValue::from_owned_byte_vec(bytes),
-            Value::Array(items) => FromRedisValue::from_owned_redis_values(items),
-            Value::Set(items) => FromRedisValue::from_owned_redis_values(items),
-            Value::Map(items) => {
-                let mut n: Vec<T> = vec![];
-                for item in items {
-                    match FromRedisValue::from_owned_redis_value(Value::Map(vec![item])) {
-                        Ok(v) => {
-                            n.push(v);
+            fn from_owned_redis_value(v: Value) -> RedisResult<$Type> {
+                match v {
+                    // Binary data is parsed into a single-element vector, except
+                    // for the element type `u8`, which directly consumes the entire
+                    // array of bytes.
+                    Value::BulkString(bytes) => FromRedisValue::from_owned_byte_vec(bytes).map($convert),
+                    Value::Array(items) => FromRedisValue::from_owned_redis_values(items).map($convert),
+                    Value::Set(items) => FromRedisValue::from_owned_redis_values(items).map($convert),
+                    Value::Map(items) => {
+                        let mut n: Vec<T> = vec![];
+                        for item in items {
+                            match FromRedisValue::from_owned_redis_value(Value::Map(vec![item])) {
+                                Ok(v) => {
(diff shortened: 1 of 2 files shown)


redis/src/types.rs
```diff
@@ -1460,17 +1460,50 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
+deref_impl! {
+    <'a, T: ?Sized> ToRedisArgs for &'a mut T where T: ToRedisArgs
+}
+
+deref_impl! {
+    <T: ?Sized> ToRedisArgs for Box<T> where T: ToRedisArgs
+}
+
+deref_impl! {
+    /// Encoding a data structure containing `Arc` will encode a copy of
```
nihohit, line 1502: I don't understand this. encoding in this library always incurs a copy of the content, what are you clarifying?

redis/src/types.rs
```diff
@@ -1999,7 +2080,6 @@ macro_rules! from_vec_from_redis_value {
 
 from_vec_from_redis_value!(<T> Vec<T>);
 from_vec_from_redis_value!(<T> std::sync::Arc<[T]>);
-from_vec_from_redis_value!(<T> Box<[T]>; Vec::into_boxed_slice);
```
nihohit, line 2002: why prefer your implementation?

redis/src/types.rs
```diff
@@ -1925,6 +1958,54 @@ impl FromRedisValue for String {
...
+            }
+
+            fn from_owned_redis_value(v: Value) -> RedisResult<Self> {
+                Box::from_owned_redis_value(v).map(Into::into)
+            }
+        }
+    };
+}
+
+box_forwarded_impl!(T, std::sync::Arc<T>);
```
nihohit, line 2006: why can't this use `forwarded_impl` with `Arc::new`?

redis/tests/test_types.rs
```diff
@@ -77,6 +81,10 @@ fn test_i32() {
 
         let bad_i: Result<i32, _> = parse_mode.parse_redis_value(Value::SimpleString("42x".into()));
```
nihohit, line 82: minor - can you move `Value::SimpleString("42x".into())` to a variable, just to be clear that both calls use the same value?

redis/tests/test_types.rs
```diff
@@ -93,6 +101,33 @@ fn test_u32() {
...
+    for parse_mode in [RedisParseMode::Owned, RedisParseMode::Ref] {
+        let simple_string_exp = "Simple string".to_string();
+        let v = parse_mode.parse_redis_value(Value::SimpleString(simple_string_exp.clone()));
+        assert_eq!(v, Ok(Box::new(simple_string_exp.clone())));
+    }
+}
+
+#[test]
+fn test_parse_arc() {
+    use redis::Value;
```
nihohit, line 117: I think we can move these using statements out of the tests, to the top of the file.

redis/tests/test_types.rs
```diff
@@ -162,26 +197,26 @@ fn test_arc_slice() {
             Value::BulkString("2".into()),
             Value::BulkString("3".into()),
         ]));
-        assert_eq!(v, Ok(Arc::from(vec![1i32, 2, 3])));
```
nihohit, line 165: why is this change required?

nihohit (commented):
very impressive work, thanks!

redis/src/types.rs
xoac: I wanted to remove `from_vec_from_redis_value!` macro and use just `forwarded_impl!((T), std::sync::Arc<[T]>, std::sync::Arc::new);` but I wasn't able to make it compile. If you have suggestion how to make it compile this can benefit in one macro less.

redis/tests/test_types.rs
xoac: because you can have now `Arc<T>` and `Arc<[T]>` and you must give Rust a hint about type.

redis/tests/test_types.rs
xoac: Both of these values doesn't need to use the same value in this tests but refactor it anyway

redis/src/types.rs
nihohit: you can reuse the output, but this macro is for the inputs :)
`ToRedisArgs` - inputs
`FromRedisValue` - outputs

redis/src/types.rs
nihohit: I can try, but your macro doesn't cover `Vec`, so I don't see how we can avoid having 2 macros here.

redis/tests/test_types.rs
nihohit: this seems to break more tests than just this. if you remove the explicit support for `Arc<[T]>`, would it still be derived by the macros?

redis/src/types.rs
xoac: The `from_vec_from_redis_value!` will be used only for `Vec<T>` so no need to keep macro just direct impl for Vec.

redis/tests/test_types.rs
xoac: it will not be delivered because implementation for `Arc<[T]>` is to create `Vec<T>` first and then use [Arc::from](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-From%3CVec%3CT,+A%3E%3E-for-Arc%3C%5BT%5D,+A%3E) as far as I understand.

xoac: But I accidentally changed testing `Arc<[T]>` into `Arc<Vec<T>>` I fixed this with [`4f1d21e` (#1088)](https://github.com/redis-rs/redis-rs/pull/1088/commits/4f1d21ea4c086e368e7420063e2d5ad9e266fce6) commit

redis/src/types.rs
xoac: Was able to remove from_vec_from_redis_value! macro with [`f91f8e0` (#1088)](https://github.com/redis-rs/redis-rs/pull/1088/commits/f91f8e0e7fefbd4e070a9fccf48f16c385d9c2bd) commit

redis/src/types.rs
```diff
@@ -1925,81 +1954,92 @@ impl FromRedisValue for String {
...
-                                }
-                            }
+            fn from_owned_redis_value(v: Value) -> RedisResult<Self> {
+                FromRedisValue::from_owned_redis_value(v).map($func)
+            }
+        }
+    }
+}
+
+forwarded_impl!((T), Box<T>, Box::new);
```
nihohit, line 1976: why is it always `(T)` instead of `T`?

redis/src/types.rs
```diff
@@ -1460,17 +1460,46 @@ impl<T: ToRedisArgs> ToRedisArgs for Option<T> {
...
 }
 
-impl<T: ToRedisArgs> ToRedisArgs for &T {
-    fn write_redis_args<W>(&self, out: &mut W)
-    where
-        W: ?Sized + RedisWrite,
-    {
-        (*self).write_redis_args(out)
-    }
+macro_rules! deref_impl {
```
nihohit, line 1463: deref_to_write_redis_args_impl

redis/src/types.rs
```diff
@@ -1925,81 +1954,92 @@ impl FromRedisValue for String {
...
+            _ => invalid_type_error!(v, "Response type not vector compatible."),
         }
-    };
+    }
 }
 
-from_vec_from_redis_value!(<T> Vec<T>);
-from_vec_from_redis_value!(<T> std::sync::Arc<[T]>);
-from_vec_from_redis_value!(<T> Box<[T]>; Vec::into_boxed_slice);
+forwarded_impl!((T), Box<[T]>, Vec::into_boxed_slice);
```
nihohit, line 2039: move these to the other `forwarded_impl` declarations

redis/src/types.rs
```diff
@@ -1925,81 +1954,92 @@ impl FromRedisValue for String {
...
 
-/// Implement `FromRedisValue` for `$Type` (which should use the generic parameter `$T`).
-///
-/// The implementation parses the value into a vec, and then passes the value through `$convert`.
-/// If `$convert` is ommited, it defaults to `Into::into`.
-macro_rules! from_vec_from_redis_value {
-    (<$T:ident> $Type:ty) => {
-        from_vec_from_redis_value!(<$T> $Type; Into::into);
-    };
+macro_rules! forwarded_impl {
```
nihohit, line 1957: pointer_from_redis_value_impl

nihohit (approved):
LGTM. @jaymell WDYT?

xoac:
By removing `from_vec_from_redis_value!` it lose ability to know exact `type` when falling.. Currently I have no idea how to fix these

nihohit:
hi @xoac,
if you're still stuck onb this, you can always split this PR to 2 - one with ToRedisArgs, the other with `FromRedisArgs`. Alternatively, you can just remove the problematic implementation (`Arc<T>`, right?) from this PR, and solve that particular problem later.

xoac:
Yes, Just wanted to wait a little bit longer to check if maybe someone has better idea how to resolve issue (losing ability to return hint for type in error) that was added with [`f91f8e0` (#1088)](https://github.com/redis-rs/redis-rs/pull/1088/commits/f91f8e0e7fefbd4e070a9fccf48f16c385d9c2bd) The only solution from my side was to revert that commit what I did. 

Do you want to add more std types to be supported with above macros? Like `Mutx`, `Rc`, `RwMutex` etc?

nihohit:
please add Rc. I'm not sure whether there's a usecase for mutex types, so I'll wait for a request for that.

redis/src/types.rs
```diff
@@ -1975,6 +1979,7 @@ macro_rules! pointer_from_redis_value_impl {
 
 pointer_from_redis_value_impl!(T, Box<T>, Box::new);
 pointer_from_redis_value_impl!(T, std::sync::Arc<T>, std::sync::Arc::new);
+pointer_from_redis_value_impl!(T, std::rc::Rc<T>, std::rc::Rc::new);
```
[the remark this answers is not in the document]
xoac: added

xoac:
Anything left here from my side here to do?

jaymell:
This looks great! I am a bit lost by this comment, though:
> I think this PR might contains breaking changes for some users because from now you need to specify output type. See test_vec function.

jaymell (approved):

nihohit:
@jaymell see the changes in `test_arc_slice` (lines 204, 220, 222) - `Arc<[_]>` now needs to be explicitly typed, in order to disambiguate between it and `Arc<T>`. Not sure why this doesn't apply to `Box`.

PR #1093, opened 2024-03-24
Expose Errors in `Value`

nihohit:
1. Remove `InternalValue` and add the `ServerError` variant to `Value`.
2. Don't convert `ServerError`s in the parser to `RedisError`. In order to maintain the behavior of the high level APIs (query/_async), they now call `extract_error`.
3. Used the internal error in order to pipeline connection setup requests - this can remove some back-and-forth with the server, and thus reduce connection setup time.

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -15,7 +15,7 @@ use futures_util::{
     future::{Future, FutureExt},
     ready,
     sink::Sink,
-    stream::{self, Stream, StreamExt, TryStreamExt as _},
+    stream::{self, Stream, StreamExt},
 };
 use pin_project_lite::pin_project;
 use std::collections::VecDeque;
@@ -465,9 +465,7 @@ impl MultiplexedConnection {
         compile_error!("tokio-comp or async-std-comp features required for aio feature");
 
         let redis_connection_info = &connection_info.redis;
-        let codec = ValueCodec::default()
-            .framed(stream)
-            .and_then(|msg| async move { msg });
+        let codec = ValueCodec::default().framed(stream);
         let (mut pipeline, driver) = Pipeline::new(codec);
         let driver = boxed(driver);
         let pm = PushManager::default();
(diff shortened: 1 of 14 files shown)


nihohit:
@jaymell 
I see 2 design questions that this PR doesn't touch - 
1. There's a lot of duplication between `RedisError` and `ServerError`. Do we want to just make `ServerError` a variant of `RedisError`, instead of duplicating the different kinds?
2. the ExecAbort error kind is currently purely theoretical, because there's always another error before it - the transaction's execution is aborted because of an error which is returned before `exec` is called, and only once `exec` we get the `ExecAbort` error. Since ATM we only return the first received error from a transaction, it's never ExecAbort. Do we want to have a special variant for it, which will contain the triggering errors? This is yet another breaking change (for, potentially, another PR), but it can be informative for the user.

jaymell:
Are the last two commits needed here? Can they go in a separate PR?

nihohit:
> Can they go in a separate PR?

They can, but they're part of the reason for this PR, so I thought I'd add them, to demonstrate the benefits of this change. Removing.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -185,10 +190,23 @@ where
             }
             ResponseAggregate::Pipeline {
                 expected_response_count,
-                current_response_count,
+                skipped_response_count,
                 buffer,
                 first_err,
             } => {
+                if *skipped_response_count > 0 {
```
jaymell, line 197: This is a great change and makes the response handling logic here much less painful to grok. I wonder if some comments added elsewhere to the pipeline code might go a long way in making the pipe vs. transaction logic less confusing.

jaymell (approved):

redis/src/aio/multiplexed_connection.rs
nihohit: sure, got suggestions or places that you consider confusing? and do you want them as part of this PR?

jaymell: Nah, let's just get this in.

PR #1097, opened 2024-03-25
Add `invoke_script` to commands to allow for pipelining of scripts

Dav1dde:
Currently it's not possible to add a script to a pipeline, adds a new method `script` to the pipelines.

I also exposed `eval_cmd` and `load_cmd` in the script invocation.

### DIFF
--- a/redis/src/script.rs
+++ b/redis/src/script.rs
@@ -206,6 +206,7 @@ impl<'a> ScriptInvocation<'a> {
         Ok(hash)
     }
 
+    /// Returns a command to load the script.
     fn load_cmd(&self) -> Cmd {
         let mut cmd = cmd("SCRIPT");
         cmd.arg("LOAD").arg(self.script.code.as_bytes());
@@ -223,7 +224,8 @@ impl<'a> ScriptInvocation<'a> {
             + 4 /* Slots reserved for the length of keys. */
     }
 
-    fn eval_cmd(&self) -> Cmd {
+    /// Returns a command to evalute the command.
+    pub(crate) fn eval_cmd(&self) -> Cmd {
         let args_len = 3 + self.keys.len() + self.args.len();
         let mut cmd = Cmd::with_capacity(args_len, self.estimate_buflen());
         cmd.arg("EVALSHA")
--- a/redis/tests/test_script.rs
+++ b/redis/tests/test_script.rs
@@ -1,0 +1,68 @@
+#![cfg(feature = "script")]
+
+mod support;
+
+mod script {
+    use redis::ErrorKind;
+
+    use crate::support::*;
+
+    #[test]
+    fn test_script() {
+        let ctx = TestContext::new();
+        let mut con = ctx.connection();
+
+        let script = redis::Script::new(r"return {redis.call('GET', KEYS[1]), ARGV[1]}");
+
+        let _: () = redis::cmd("SET")
+            .arg("my_key")
+            .arg("foo")
+            .query(&mut con)
+            .unwrap();
+        let response = script.key("my_key").arg(42).invoke(&mut con);
+
+        assert_eq!(response, Ok(("foo".to_string(), 42)));
+    }
+
+    #[test]
+    fn test_script_load() {
+        let ctx = TestContext::new();
+        let mut con = ctx.connection();
+
+        let script = redis::Script::new("return 'Hello World'");
+
+        let hash = script.prepare_invoke().load(&mut con);
+
+        assert_eq!(hash, Ok(script.get_hash().to_string()));
+    }
+
+    #[test]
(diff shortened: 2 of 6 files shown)


redis/src/script.rs
```diff
@@ -206,12 +206,25 @@ impl<'a> ScriptInvocation<'a> {
         Ok(hash)
     }
 
-    fn load_cmd(&self) -> Cmd {
+    /// Returns a command to load the script.
+    pub fn load_cmd(&self) -> Cmd {
```
nihohit, line 210: why does this need to be a public method? the user has the actual script, this uses a single argument from the whole invocation.

redis/src/pipeline.rs
```diff
@@ -293,6 +293,16 @@ macro_rules! implement_pipeline_commands {
                 self.ignored_commands.clear();
             }
 
+            /// Adds a prepared script command to the pipeline.
+            ///
+            /// Shorthand for `add_command(invocation.eval_cmd())`.
+            #[inline]
+            #[cfg(feature = "script")]
```
nihohit, line 300: instead of adding it here, can you add it to the `implement_commands` macro invocation in commands/mod.rs? that way it will be added to multiple relevant objets.

nihohit (commented):
please add tests & public documentation.+

redis/src/script.rs
Dav1dde: I decided to make both methods public `load_cmd` and `eval_cmd` so they can be independently added to pipelines just through `add_command`. I can change them to be `pub(crate)` if you don't want them to be exposed.

In theory the `load_cmd` could also be a method of `Script`, I kept it here because all the execution (e.g. `load()`) is on the `ScriptInvocation`.

redis/src/pipeline.rs
Dav1dde: Sounds, good. I didn't initially add it to `implement_commands` because it is not an actual Redis command.

Dav1dde:
@nihohit Changed some stuff:

- Moved the existing tests together with the new ones into a separate file. 
- Kept the the `load_cmd` and `eval_cmd` methods private.
- Moved the `script` fn to the `implement_commands`.
- Updated the lib docs with an example.

redis/tests/test_script.rs
```diff
@@ -0,0 +1,70 @@
...
+    assert_eq!(hash, Ok(script.get_hash().to_string()));
+}
+
+#[test]
+#[cfg(feature = "script")]
+fn test_script_pipeline_no_autoload() {
+    let ctx = TestContext::new();
+    let mut con = ctx.connection();
+
+    redis::cmd("SCRIPT").arg("FLUSH").execute(&mut con);
```
nihohit, line 46: this is unnecessary - `TestContext` creates a new server.

redis/tests/test_script.rs
```diff
@@ -0,0 +1,68 @@
...
+    let script = redis::Script::new("return 'Hello World'");
+
+    let hash = script.prepare_invoke().load(&mut con);
+
+    assert_eq!(hash, Ok(script.get_hash().to_string()));
+}
+
+#[test]
+#[cfg(feature = "script")]
+fn test_script_pipeline_no_autoload() {
```
nihohit, line 42: test_script_that_is_not_loaded_fails_on_pipeline_invocation ?

nihohit (approved):
@jaymell WDYT?

redis/src/script.rs
Dav1dde: It can be still loaded by using the `ScriptInvocation`, currently using it like this:

```rs
    pub fn invoke(&self, con: &mut Connection<'_>) -> Result<Vec<CardinalityScriptResult>> {
        match self.pipe.query(con) {
            Ok(result) => Ok(result),
            Err(err) if err.kind() == redis::ErrorKind::NoScriptError => {
                self.script.load_redis(con)?;
                self.pipe
                    .query(con)
                    .map_err(relay_redis::RedisError::Redis)
                    .map_err(Into::into)
            }
            Err(err) => Err(relay_redis::RedisError::Redis(err).into()),
        }
    }
```

But it can't be added to the pipeline without it, but that can also be solved by adding another fn to the `implement_commands`.

Dav1dde:
Thanks for the review, updated with your comments and found some small doc issues I also fixed.

I am a bit unsure about the `script` function name since it does collide with [`SCRIPT`](https://redis.io/commands/script/).

nihohit:
> I am a bit unsure about the script function name since it does collide with [SCRIPT](https://redis.io/commands/script/).

maybe `invoke_script`?

nihohit:
@Dav1dde can you please rebase?

Dav1dde:
@nihohit squashed all the commits and rebased to latest main

PR #1100, opened 2024-03-25
Improve PushManager tests in sync connection

altanozlu:
This PR adds PushManager tests into existings PubSub tests.

### DIFF
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -642,6 +642,9 @@ mod basic {
 
         // Connection for subscriber api
         let mut pubsub_con = ctx.connection();
+        let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+        // Only useful when RESP3 is enabled
+        pubsub_con.get_push_manager().replace_sender(tx);
 
         // Barrier is used to make test thread wait to publish
         // until after the pubsub thread has subscribed.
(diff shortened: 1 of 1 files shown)


redis/tests/test_basic.rs
```diff
@@ -667,13 +670,50 @@ fn test_pubsub() {
     assert_eq!(con.publish("foo", 23), Ok(1));
 
     thread.join().expect("Something went wrong");
+    if ctx.protocol == ProtocolVersion::RESP3 {
```
nihohit, line 673: you can use this to remove the opt-in in the test
https://github.com/redis-rs/redis-rs/pull/1096/commits/62fe8a4e8141f8a668d47bf52faec0f8f0112842

altanozlu: It's nice to know about it, but I don't want to override the original test, it should run for both RESP3 and RESP2

redis/tests/test_basic.rs
```diff
@@ -640,6 +640,9 @@ fn test_pubsub() {
 
     // Connection for subscriber api
     let mut pubsub_con = ctx.connection();
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+    // Only useful when RESP3 is enabled
+    pubsub_con.get_push_manager().replace_sender(tx.clone());
```
nihohit, line 645: minor: why is tx cloned?

redis/tests/test_basic.rs
```diff
@@ -667,13 +670,50 @@ fn test_pubsub() {
...
 }
 
 #[test]
 fn test_pubsub_unsubscribe() {
     let ctx = TestContext::new();
     let mut con = ctx.connection();
 
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+    // Only useful when RESP3 is enabled
+    con.get_push_manager().replace_sender(tx.clone());
```
nihohit, line 716: same

redis/tests/test_basic.rs
nihohit, line 673: I'd prefer if this was in a test that didn't use a `PubSub` connection, but rather a test that demonstrated how this works on a regular connection.

altanozlu: Hi, I think this test should stay as is and there is already a test called `test_push_manager`. I don't think currently there will be usage for PushManager other than PubSub until Client Side Caching.

nihohit: part of the point of push manager is that it can be used with a regular connection, not only with a `PubSub` connection. Notice the distinction - I'm not talking about using the push manager to get pub-sub messages, but to use it with a different struct - `redis::Connection` instead of `redis::PubSub`. 
ATM there's no example on how to do that, not in tests and not in documentation. `test_push_manager` demonstrates how to use the push manager for client tracking, but not for pub-sub, which IMO is a more common usecase.

altanozlu: I have added simple test, but in future I want to add examples to examples folder

nihohit: The library's documentation is also a great place for examples :)

nihohit:
excellent! can you please rebase?

PR #1113, opened 2024-03-26
Update rustls + tokio-rustls + futures-rustls

nihohit:


nihohit:
@jaymell @djc 
I'm not very keen on all of the added indirect dependencies. Do we want to stay in an earlier version?

djc:
Just switch is back to the ring crypto provider via the Cargo features.

PR #1130, opened 2024-04-03
Fix issues that prevented cluster tests from running concurrently.

nihohit:
1. tests with mocks shared the same handle, so they dropped each other's state.
2. tests with real cluster shared the same ports, so the cluster couldn't run concurrently.

### DIFF
--- a/redis/tests/support/cluster.rs
+++ b/redis/tests/support/cluster.rs
@@ -74,8 +74,8 @@ fn port_in_use(addr: &str) -> bool {
 }
 
 pub struct RedisClusterConfiguration {
-    pub nodes: u16,
-    pub replicas: u16,
+    pub num_nodes: u16,
+    pub num_replicas: u16,
     pub modules: Vec<Module>,
     pub mtls_enabled: bool,
     pub ports: Vec<u16>,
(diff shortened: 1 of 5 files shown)


redis/tests/support/cluster.rs
```diff
@@ -72,6 +73,36 @@ fn port_in_use(addr: &str) -> bool {
     socket.connect(&socket_addr.into()).is_ok()
 }
 
+pub struct RedisClusterConfiguration {
+    pub nodes: u16,
```
jaymell, line 77:
```suggestion
    pub num_nodes: u16,
```

redis/tests/support/cluster.rs
```diff
@@ -72,6 +73,36 @@ fn port_in_use(addr: &str) -> bool {
...
+    pub replicas: u16,
```
jaymell, line 78:
```suggestion
    pub num_replicas: u16,
```

jaymell:
Any idea what's up with the test failure?

nihohit:
> Any idea what's up with the test failure?

No. I don't know if it's something introduced by this change, or the old flakeyness of the cluster tests. :(

jaymell (approved):

PR #1134, opened 2024-04-04
Make response_timeout Optional

zhixinwen:
If user uses `MultiplexedConnection::new` to create `MultiplexedConnection` without a response timeout, `std::time::Duration::MAX` is used. This would not do any timeout useful in practice.

In our usecase, we choose to do not provide `response_timeout` because we manage timeout ourselves in the upper layer. And during profile, we catch that the underlying `MultiplexedConnection` is also calling timeout which is not necessary.

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -361,7 +361,11 @@ where
             },
             None => match receiver.await {
                 Ok(result) => result.map_err(Some),
-                Err(_) => Err(None),
+                Err(_) => {
+                    // The `sender` was dropped which likely means that the stream part
+                    // failed for one reason or another
+                    Err(None)
+                }
             },
         }
     }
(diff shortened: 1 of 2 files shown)


redis/src/aio/multiplexed_connection.rs
```diff
@@ -348,14 +348,21 @@ where
...
-                // failed for one reason or another
-                Err(None)
-            }
-            Err(elapsed) => Err(Some(elapsed.into())),
+
+        match timeout {
+            Some(timeout) => match Runtime::locate().timeout(timeout, receiver).await {
+                Ok(Ok(result)) => result.map_err(Some),
+                Ok(Err(_)) => {
+                    // The `sender` was dropped which likely means that the stream part
```
nihohit, line 356: please copy the comment to line 363, or better yet, remove this repetition.

nihohit (approved):

nihohit:
Thanks for your contribution!

PR #1140, opened 2024-04-07
Cancel ongoing github actions on push.

nihohit:
https://docs.github.com/en/actions/using-jobs/using-concurrency

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,6 +10,10 @@ env:
   CARGO_TERM_COLOR: always
   REDIS_RS_REDIS_JSON_PATH: "/tmp/librejson.so"
 
+concurrency:
+  group: "${{ github.workflow }}-${{ github.head_ref || github.run_id || github.ref }}"
+  cancel-in-progress: true
+
 jobs:
   build:
 


.github/workflows/rust.yml
```diff
@@ -10,6 +10,10 @@ env:
   CARGO_TERM_COLOR: always
   REDIS_RS_REDIS_JSON_PATH: "/tmp/librejson.so"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
jaymell, line 14: Should we use `github.ref` instead of `github.head_ref`? If I'm reading this correctly, this wouldn't apply for multiple runs against `main`, for example.

nihohit: frankly, I don't know. I thought of using head_ref because it applies only to PRs, so I thought that would be a conservative first change. If you prefer `head`, I'll change.

jaymell: Yeah I think it's worth trying`${{ github.workflow }}-${{ github.ref }}` -- seems fine, and should hopefully kill redundant `main` jobs if we're merging multiple.

jaymell (approved):

nihohit:
I don't understand this :(
image

PR #1141, opened 2024-04-07
Update cluster_async router_command docs

joachimbulow:
Since `RoutingInfo` is not an `Option` type, docs should not hint at passing `None` as a parameter


nihohit (approved):

nihohit:
Thanks for the correction!

PR #1146, opened 2024-04-10
Fix compilation on Windows.

Yury-Fridlyand:
Fix compilation error occurs on Windows only. Looks like a mistake IMHO.
```
...
   Compiling redis v0.25.0 (C:\GitHub\redis-rs\redis)
error[E0412]: cannot find type `HashMap` in this scope
   --> C:\GitHub\redis-rs\redis\src\connection.rs:355:16
    |
355 |     let query: HashMap<_, _> = url.query_pairs().collect();
    |                ^^^^^^^ not found in this scope
    |
help: consider importing one of these items
    |
1   + use combine::lib::collections::HashMap;
    |
1   + use crate::types::HashMap;
    |
1   + use std::collections::HashMap;
    |

For more information about this error, try `rustc --explain E0412`.
error: could not compile `redis` (lib) due to previous error
```

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -11,13 +11,11 @@ use crate::cmd::{cmd, pipe, Cmd};
 use crate::parser::Parser;
 use crate::pipeline::Pipeline;
 use crate::types::{
-    from_redis_value, ErrorKind, FromRedisValue, PushKind, RedisError, RedisResult, ToRedisArgs,
-    Value,
+    from_redis_value, ErrorKind, FromRedisValue, HashMap, PushKind, RedisError, RedisResult,
+    ToRedisArgs, Value,
 };
 use crate::{from_owned_redis_value, ProtocolVersion};
 
-#[cfg(unix)]
-use crate::types::HashMap;
 #[cfg(unix)]
 use std::os::unix::net::UnixStream;
 use std::vec::IntoIter;


nihohit:
I can't reproduce this with either `cargo build` or `cargo build --all-features` on windows. I also suspect that something is setup wrong on your build, since `redis-rs\redis\src\connection.rs:355` is just `        },`, and implemeting the change adds a warning about an unused import.
please double-check yourself and update the PR.

barshaul:
> I can't reproduce this with either `cargo build` or `cargo build --all-features` on windows. I also suspect that something is setup wrong on your build, since `redis-rs\redis\src\connection.rs:355` is just ` },`, and implemeting the change adds a warning about an unused import. please double-check yourself and update the PR.

https://github.com/redis-rs/redis-rs/blob/be1c2c9db9ba1bc2e7214dfc8a21d12f9a80472d/redis/src/connection.rs#L351
I don't have Windows to reproduce it, but last month the use of HashMap was add to the 'url_to_tcp_connection_info' function by [this commit](https://github.com/redis-rs/redis-rs/commit/30d573072136a0d1bde9866c972949f86ae53b16), which isn't under `unix` feature flag, however the HashMap import is.
>    let query: HashMap<_, _> = url.query_pairs().collect();

nihohit:
@barshaul those changes aren't in the current codebase.

barshaul:
@nihohit  https://github.com/redis-rs/redis-rs/blob/main/redis/src/connection.rs#L353

nihohit:
you're right, I was wrong - I tested this without noticing I wasn't on main. Sorry for the confusion!

nihohit (approved):

PR #1149, opened 2024-04-14
Add un/subscribe commands to `aio::ConnectionManager`.

nihohit:
The last commit of this PR isn't strictly required for the PR, but it improves our test coverage of the connection manager, which is somewhat related.

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -1,10 +1,10 @@
 use super::RedisFuture;
-use crate::cmd::Cmd;
-use crate::push_manager::PushManager;
-use crate::types::{RedisError, RedisResult, Value};
 use crate::{
-    aio::{ConnectionLike, MultiplexedConnection, Runtime},
-    Client,
+    aio::{check_resp3, ConnectionLike, MultiplexedConnection, Runtime},
+    cmd,
+    push_manager::PushManager,
+    types::{RedisError, RedisResult, Value},
+    Client, Cmd, ToRedisArgs,
 };
 #[cfg(all(not(feature = "tokio-comp"), feature = "async-std-comp"))]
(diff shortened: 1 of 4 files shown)


altanozlu:
I really liked new macro and wrapper, nice.

redis/src/aio/connection_manager.rs
```diff
@@ -386,6 +386,44 @@ impl ConnectionManager {
         result
     }
 
+    /// Subscribes to a new channel.
+    /// It should be noted that the subscription will be removed on a disconnect, and must the re-subscribed.
```
jaymell, line 390:
```suggestion
    /// It should be noted that the subscription will be removed on a disconnect and must be re-subscribed.
```

redis/src/aio/connection_manager.rs
```diff
@@ -386,6 +386,44 @@ impl ConnectionManager {
...
+    pub async fn unsubscribe(&mut self, channel_name: impl ToRedisArgs) -> RedisResult<()> {
+        check_resp3!(self.client.connection_info.redis.protocol);
+        let mut cmd = cmd("UNSUBSCRIBE");
+        cmd.arg(channel_name);
+        cmd.query_async(self).await?;
+        Ok(())
+    }
+
+    /// Subscribes to a new channel with pattern.
+    /// It should be noted that the subscription will be removed on a disconnect, and must the re-subscribed.
```
jaymell, line 409:
```suggestion
    /// It should be noted that the subscription will be removed on a disconnect and must be re-subscribed.
```

redis/tests/test_async.rs
jaymell, line 1: Can we make sure that commit messages don't go past 72 characters per line? :nerd_face:

jaymell (approved):

PR #1152, opened 2024-04-17
fix #1150: change int types for expiry to `u64`

### LINKED ISSUE #1150: Expiration time integer types inconsistent with `SetExpiry`
Hello, I just noticed that the integer types for the expiration times were not updated for the [`SetExpiry`](https://docs.rs/redis/0.25.3/redis/enum.SetExpiry.html) enum, which still relies on `usize`s (see #589).

I believe it needs an update too, but I can't tell if these can accept negative values or not.

Thanks 😄

ahmadbky:
This fixes #1150.


nihohit (approved):

nihohit:
Thanks for the contribution!

PR #1160, opened 2024-04-23
Rework tls feature split to avoid unused dependencies

### LINKED ISSUE #1155: Is there a way to use tokio + rustls +webpki-roots ?
Hi!

I am looking for a way to use this crate with tokio and rustls (so basically `tokio-rustls-comp` feature) but I would like to use `webpki-roots` instead of `rustls-native-certs`. I want this specific setup because that is what I use in every other tls setup in my project and because my goal is to run my service in a very minimal environment (ideally without openssl and without local certificates).

redis-rs seem to support using `webpki-roots` only in a sync context but not with `tokio`. I do not think there is any specific reason behind that, it only seems to be an unwanted consequence of the way features are split and combined.

Could anyone please confirm that it is indeed not currently possible? And if so, do you think we could rework current features to make it possible?

axelfaure:
This is an attempt to rework the way tls related features are split in order to avoid depending on `rustls-native-certs` even when it is not actually used.

Fixes #1155

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,19 +85,27 @@ redis = { version = "0.25.3", features = ["async-std-native-tls-comp"] }
 To use `rustls`:
 
 ```
-redis = { version = "0.25.3", features = ["tls-rustls"] }
+redis = { version = "0.25.3", features = ["tls-rustls-native-certs"] }
+# or
+redis = { version = "0.25.3", features = ["tls-rustls-webpki-roots"] }
 
 # if you use tokio
-redis = { version = "0.25.3", features = ["tokio-rustls-comp"] }
+redis = { version = "0.25.3", features = ["tokio-rustls-native-certs-comp"] }
+# or
+redis = { version = "0.25.3", features = ["tokio-rustls-webpki-roots-comp"] }
 
 # if you use async-std
-redis = { version = "0.25.3", features = ["async-std-rustls-comp"] }
+redis = { version = "0.25.3", features = ["async-std-rustls-native-certs-comp"] }
+# or
+redis = { version = "0.25.3", features = ["async-std-rustls-webpki-roots-comp"] }
 ```
 
-With `rustls`, you can add the following feature flags on top of other feature flags to enable additional features:
+With `rustls`, you must choose between `native-certs` and `webpki-roots` variants of the tls features.
+The first ones rely on native root certificates (the one provided by your OS) while `webpki-roots` features use `webpki-roots` (Mozilla's root certificates) instead.
+
+With `rustls`, you can add the following feature flag on top of other feature flags to enable additional features:
 
 -   `tls-rustls-insecure`: Allow insecure TLS connections
--   `tls-rustls-webpki-roots`: Use `webpki-roots` (Mozilla's root certificates) instead of native root certificates
 
 then you should be able to connect to a redis instance using the `rediss://` URL scheme:
 
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -98,16 +98,6 @@ geospatial = []
 json = ["serde", "serde/derive", "serde_json"]
 cluster = ["crc16", "rand"]
 script = ["sha1_smol"]
-tls-native-tls = ["native-tls"]
-tls-rustls = ["rustls", "rustls-native-certs", "rustls-pemfile", "rustls-pki-types"]
-tls-rustls-insecure = ["tls-rustls"]
-tls-rustls-webpki-roots = ["tls-rustls", "webpki-roots"]
-async-std-comp = ["aio", "async-std"]
-async-std-native-tls-comp = ["async-std-comp", "async-native-tls", "tls-native-tls"]
-async-std-rustls-comp = ["async-std-comp", "futures-rustls", "tls-rustls"]
-tokio-comp = ["aio", "tokio/net"]
-tokio-native-tls-comp = ["tokio-comp", "tls-native-tls", "tokio-native-tls"]
-tokio-rustls-comp = ["tokio-comp", "tls-rustls", "tokio-rustls"]
 connection-manager = ["futures", "aio", "tokio-retry"]
 streams = []
 cluster-async = ["cluster", "futures", "futures-util", "log"]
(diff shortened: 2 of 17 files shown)


nihohit:
Hi @axelfaure, thanks for the contribution and sorry for the slow reply.
This looks great, but I'm not sure that having features that wrap multiple features is maintainable - we don't want to have a feature for each cartesian product of options. IMO we should strive to expose orthogonal features, so that the user could pick-and-choose freely.
 @jaymell WDYT?

nihohit:
I see that you wrote in the original issue "I couldn't decorrelate the async framework from the tls backend without creating way more unused dependencies." - can you please explain this?

axelfaure:
Hi @nihohit, I fully agree with you and as you mentioned in your second comment that was indeed my initial intention. 

Unfortunately I cannot have, let say, a `native-tls` feature on one hand and a `tokio` feature on the other without creating additional issue. This is because in that situation, I would need to express that `tokio-native-tls` dependency should only be present if both features are enabled. To the best of my knowledge this is not possible (I could do it in the rust code of course but not in the Cargo.toml). The same goes for `futures-rustls`, or `tokio-rustls`.

As far as I can tell, there seems to be 3 solutions to this problem:

1. The cartesian product as implemented in this PR,
2. Creating a lot of seemingly independent features that actually need to be correctly selected such as `tokio`, `rustls`, `webpki-root` and  `tokio-rustls. This would create a lot of features making it way more diffucult for the end users to understand what to do (and I am not really sure it would improve maintainability either)
3. Removing the need for these combined dependencies (`tokio-native-tls`, `futures-rustls`, `tokio-rustls`...). I am really not sure what they are needed for nor if it is possible to get rid of them without rewriting a lot of code. This could hypothetically be the best solution but it could also be quite difficult to achieve. Note that `sqlx` crate which does [expose orthogonal features](https://github.com/launchbadge/sqlx/blob/main/sqlx-core/Cargo.toml) does seem to do without such dependencies, so this might be the way to go in the long run.

nihohit:
Hi @axelfaure, I really apologize - this change requires a deep review, and ATM I just don't have the time for it. I haven't forgotten, but it will probably take some time before I can properly answer you.

axelfaure:
Hi, Don't worry, I totally get it. Take your time. 
Fixing the dependencies would be much nicer but in practice since the deps are not used, it does not have actual consequences on our binaries.

README.md
```diff
@@ -85,19 +85,27 @@ redis = { version = "0.25.3", features = ["async-std-native-tls-comp"] }
...
 
 # if you use async-std
-redis = { version = "0.25.3", features = ["async-std-rustls-comp"] }
+redis = { version = "0.25.3", features = ["async-std-rustls-native-certs-comp"] }
+# or
+redis = { version = "0.25.3", features = ["async-std-rustls-webpki-roots-comp"] }
 ```
 
-With `rustls`, you can add the following feature flags on top of other feature flags to enable additional features:
+With `rustls`, you must choose between `native-certs` and `webpki-roots` variants of the tls features.
```
nihohit, line 103: `native-certs` / `webpki-roots` - please put the full names.

redis/Cargo.toml
```diff
@@ -120,9 +110,39 @@ num-bigint = []
...
+async-std-comp = ["aio", "async-std"]
+async-std-native-tls-comp = ["async-std-comp", "async-native-tls", "tls-native-tls"]
+async-std-rustls-native-certs-comp = ["async-std-rustls-core", "tls-rustls-native-certs"]
+async-std-rustls-webpki-roots-comp = ["async-std-rustls-core", "tls-rustls-webpki-roots"]
+
+# tokio related features
+# At most one should be needed
+tokio-comp = ["aio", "tokio/net"]
+tokio-native-tls-comp = ["tokio-comp", "tls-native-tls", "tokio-native-tls"]
+tokio-rustls-native-certs-comp = ["tokio-rustls-core", "tls-rustls-native-certs"]
```
nihohit, line 130: instead of having these trees of features, can we just have the basic features, with compile-time errors if additional features aren't enabled?
for example, if you enable `async-std-rustls-core`, you need to also explicitly enable either `tls-rustls-webpki-roots` or `tls-rustls-native-certs`?

redis/Cargo.toml
```diff
@@ -120,9 +110,39 @@ num-bigint = []
...
+
+# Can be combined with any other feature that enables rustls
+tls-rustls-insecure = ["tls-rustls-core"]
+# mostly meant for internal use
+tls-rustls-core = ["rustls", "rustls-pemfile", "rustls-pki-types"]
+async-std-rustls-core = ["tls-rustls-core", "async-std-comp", "futures-rustls"]
+tokio-rustls-core = ["tls-rustls-core", "tokio-comp", "tokio-rustls"]
+
 # Deprecated features
 tls = ["tls-native-tls"] # use "tls-native-tls" instead
```
nihohit, line 141: can we add compile time warnings on deprecated features?

redis/Cargo.toml
```diff
@@ -120,9 +110,39 @@ num-bigint = []
...
+# At most one should be needed
+tokio-comp = ["aio", "tokio/net"]
+tokio-native-tls-comp = ["tokio-comp", "tls-native-tls", "tokio-native-tls"]
+tokio-rustls-native-certs-comp = ["tokio-rustls-core", "tls-rustls-native-certs"]
+tokio-rustls-webpki-roots-comp = ["tokio-rustls-core", "tls-rustls-webpki-roots"]
+
+# Can be combined with any other feature that enables rustls
+tls-rustls-insecure = ["tls-rustls-core"]
+# mostly meant for internal use
+tls-rustls-core = ["rustls", "rustls-pemfile", "rustls-pki-types"]
```
nihohit, line 136: I prefer that this be named `tls-rustls`, and add an explicit compile-time message that it requires additional features.

nihohit (commented):
sorry for taking such a long time to look at this change.
Excellent work, but I'm not sure that the right thing to do is to create features that represent multiple sub-features - IMO this might cause a combinatorical explosion down the line. We're replacing 10 features with 15 features + 3 deprecated features.

WDYT about creating minimal features, and failing compilation if the required dependencies aren't added?
@jaymell

jayvdb:
Hi @axelfaure , do you intend to continue working on this?

PR #1165, opened 2024-04-25
Cargo "redis-test-utils"

### LINKED ISSUE #1164: Question: How to use redis-rs to test a redis module written in rust?
Is there any example to follow?
If not, can anyone please provide directions?
Cheers

ashtul:
There is a need for a separate crate as mentioned in #1164.
The existing infrastructure can be used by developers of redis modules.

Fixes #1164

### DIFF
--- a/redis/tests/Cargo.toml
+++ b/redis/tests/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "redis-test-utils"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[lib]
+bench = false
+path = "support/mod.rs"
+
+[dependencies]
+redis = { version = "0.25.3", path = "../../redis" }
+socket2 = "0.5"
+rand = "0.8"
+tempfile = "=3.6.0"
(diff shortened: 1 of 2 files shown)


redis/tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "redis-test-utils"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[lib]
+bench = false
+path = "support/mod.rs"
```
jaymell, line 9: Is this correct?

jaymell:
I'm not sure how much we want this code exposed publicly :sweat_smile:, but I think this is potentially a great idea.

jaymell:
That said, my assumption here is that we'd want to move all the code currently under `redis/tests/support` into this module and have the main `redis` crate depend on this one.

nihohit:
@ashtul thanks for the contribution. As @jaymell said, please move all relevant files into the new crate, and also fix the lock file and add a new lock file for the new crate.

PR #1166, opened 2024-04-26
check tls mode before setting it in the call of certs()

MyBitterCoffee:
For ClusterClientBuilder, if we call tls(TlsMode::Insecure) first, a following certs() call will set builder_params.tls to Some(TlsMode::Secure) 
In this pr, we will check builder_params.tls  before setting it in fn certs()

### DIFF
--- a/redis/src/cluster_client.rs
+++ b/redis/src/cluster_client.rs
@@ -282,7 +282,10 @@ impl ClusterClientBuilder {
     /// If `root_cert` is not provided, then system root certificates are used instead.
     #[cfg(feature = "tls-rustls")]
     pub fn certs(mut self, certificates: TlsCertificates) -> ClusterClientBuilder {
-        self.builder_params.tls = Some(TlsMode::Secure);
+        if self.builder_params.tls.is_none() {
+            self.builder_params.tls = Some(TlsMode::Secure);
+        }
+
         self.builder_params.certs = Some(certificates);
         self
     }


jaymell:
Hi, thanks for the PR! The existing functionality here is definitely a bit awkward, but I am curious about your use case. Why would you supply certificate info but still require tls-mode to be insecure?

MyBitterCoffee:
> Hi, thanks for the PR! The existing functionality here is definitely a bit awkward, but I am curious about your use case. Why would you supply certificate info but still require tls-mode to be insecure?

Sometimes (for an example, in my dev/sit environment),  people want to use insecure certs to enable mTls.

jaymell:
> Sometimes (for an example, in my dev/sit environment), people want to use insecure certs to enable mTls.

What is an insecure cert from the perspective of mtls? My very limited understanding of mtls is that it doesn't work without each side trusting the other. In any case, maybe we shouldn't be setting `TlsMode::Secure` at all in the `certs` method. Seems like maybe that's a bad idea.

MyBitterCoffee:
> > Sometimes (for an example, in my dev/sit environment), people want to use insecure certs to enable mTls.
> 
> What is an insecure cert from the perspective of mtls? My very limited understanding of mtls is that it doesn't work without each side trusting the other. In any case, maybe we shouldn't be setting `TlsMode::Secure` at all in the `certs` method. Seems like maybe that's a bad idea.

1)
We have some environments. For some reasons(cost, security...), only the prod environment can access the RA and CA.
Some environments are partly isolated (we can access these environments through a jump server only, and the nodes in these environments cannot access any environments outside)
Each environment has a redis cluster deployed. All the redis clusters are tls-enabled.
In most of the environments, except for the prod environment, we try to use some insecure certs just for mtls. We do some restrict checks for the certs in the prod environment only.

2) Maybe remove the TlsMode::Secure setting in the certs method is a good idea.  
     If you guys decide to do so, this pr can be closed.

jaymell:
@MyBitterCoffee Can we leave this open for now?

nihohit:
> maybe we shouldn't be setting TlsMode::Secure at all in the certs method.

I think that would've been best initially, but right now it'll be a breaking change, so we probably want to avoid that. 
IMO this is a niche case, but there's no harm in allowing it.

jaymell:
> I think that would've been best initially, but right now it'll be a breaking change, so we probably want to avoid that.
> IMO this is a niche case, but there's no harm in allowing it.

Cool, let's get it in then!

jaymell (approved):

PR #1167, opened 2024-04-26
Timeouts for redis client gotten via sentinel

jrylander:
Add a redis::sentinel::Sentinel::SentinelClient::get_async_connection_with_config that accepts timeouts

I have added just a happy test for calling with long timeouts. I am not versed enough to add a test for actual timeouts but am not sure it is needed since it just ships the parameters on to redis client.

### DIFF
--- a/redis/src/client.rs
+++ b/redis/src/client.rs
@@ -66,6 +66,42 @@ impl Client {
     }
 }
 
+/// Options for creation of async connection
+pub struct AsyncConnectionConfig {
+    /// Maximum time to wait for a response from the server
+    response_timeout: Option<std::time::Duration>,
+    /// Maximum time to wait for a connection to be established
+    connection_timeout: Option<std::time::Duration>,
+}
+
+impl AsyncConnectionConfig {
+    /// Creates a new instance of the options with nothing set
+    pub fn new() -> Self {
+        Self {
+            response_timeout: None,
+            connection_timeout: None,
+        }
+    }
+
+    /// Sets the connection timeout
+    pub fn with_connection_timeout(mut self, connection_timeout: std::time::Duration) -> Self {
+        self.connection_timeout = Some(connection_timeout);
+        self
+    }
+
+    /// Sets the response timeout
+    pub fn with_response_timeout(mut self, response_timeout: std::time::Duration) -> Self {
+        self.response_timeout = Some(response_timeout);
+        self
+    }
+}
+
+impl Default for AsyncConnectionConfig {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 /// To enable async support you need to chose one of the supported runtimes and active its
 /// corresponding feature: `tokio-comp` or `async-std-comp`
 #[cfg(feature = "aio")]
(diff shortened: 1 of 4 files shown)


redis/src/sentinel.rs
```diff
@@ -742,6 +742,26 @@ impl SentinelClient {
...
+    pub timeouts: Option<AsyncConnectionOptionsForTimeout>,
+}
+
+/// Timeouts for async connections to redis
+#[cfg(feature = "aio")]
+#[cfg_attr(docsrs, doc(cfg(feature = "aio")))]
+#[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
+pub struct AsyncConnectionOptionsForTimeout {
+    /// Maximum time to wait for a response from the server
+    pub response_timeout: std::time::Duration,
```
nihohit, line 760: please move these fields into `AsyncConnectionOptions` - there's no need for a nested structure.
also, instead of making these fields public, please use the builder pattern, so that if we change the options' internal structure it won't be a breaking change.

jrylander: Will do

jrylander: Done

nihohit: Thanks, but sorry for not being explicit - can you split the with_timeouts function, in case a user is only interested in one timeout type?

jrylander:
Is this closer to your thinking? I am also guessing I should squash commits, right?

redis/src/client.rs
```diff
@@ -190,6 +232,59 @@ impl Client {
         }
     }
 
+    /// Returns an async connection from the client.
+    #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(any(feature = "tokio-comp", feature = "async-std-comp")))
+    )]
+    pub async fn get_multiplexed_async_connection_with_config(
```
nihohit, line 241: `get_multiplexed_async_connection_with_timeouts` should call this

nihohit: same here https://github.com/redis-rs/redis-rs/blob/3cf579fd0da327d018559f1981e6cc27d6fe9cb3/redis/src/aio/connection_manager.rs#L195C20-L195C66

redis/src/sentinel.rs
```diff
@@ -769,4 +770,17 @@ impl SentinelClient {
         let client = self.async_get_client().await?;
```
nihohit, line 770: minor: this can call `get_async_connection_with_config` with default config

nihohit (approved):
Excellent work, please fix & squash.

redis/src/sentinel.rs
jrylander: Good thoughts! Did squash as well. N.B. I am not able to run all tests.

redis/src/client.rs
jrylander: Done. Good thought

nihohit:
@jrylander looks like the change has some issues. Do you need help fixing this?

jrylander:
Ahhrg, my bad. I was on a clean computer and could not run tests. I am sorry!

nihohit (approved):
LGTM. please fix the formatting issue blocking the CI.

jrylander:
Ran cargo fmt --all

Issue #1153, opened 2024-04-18
Missing stream commands


urkle:
I am implementing a service that utilizes the streams in redis and am running into a number of items that are missing in redis-rs.

- XADD
   - [ ] NOMKSTREAM option
   - [ ] LIMIT option for MAXLEN
   - [ ] MINID option
- XAUTOCLAIM
   - [ ] fully missing
- XCLAIM
   - [ ] LASTID option
- XGROUP CREATE
   -  [ ] ENTRIESREAD option
- XGROUP CREATECONSUMER
   - [ ] fully missing
- XGROUP SETID
   -  [ ] ENTRIESREAD option
- XINFO STREAM
   - [ ] FULL option
   - [ ] FULL COUNT option
- XPENDING
    - [ ] IDLE arg
- XSETID
   - [ ] fully missing
- XTRIM
   - [ ] MINID
   - [ ] LIMIT count option

nihohit:
Would you like to open PRs for these?

urkle:
@nihohit I'm working on a local trait to get the ones I need implemented so I can get a PR together.

A few questions
1) should the options classes like StreamClaimOptions be `#[non_exhaustive]` (or not necessary since the internals are private). As I'll need to extend some of the like the StreamClaimOptions
2) can I create _options versions of some of these to make  more concise variant? (e.g. xadd_options)
3)

nihohit:
also, I'll save you time on common comments we give to PRs - 
1. please open separate PRs for each command, or at least for each set of commands. If you'll open 1 huge PR, I'll ask to break it down.
2. please add comprehensive sync & async tests for each command.

Thanks for the will to contribute :)

nihohit:
@StefanPalashev if you're looking for more missing APIs, there's this issue.

PR #1169, opened 2024-04-28
Add xautoclaim command support

urkle:
This adds a new command for the redis XAUTOCLAIM along with a Reply object to parse the response

This is one of the missing steam commands described in #1153

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1317,7 +1317,45 @@ implement_commands! {
             .arg(map)
     }
 
-
+    /// Perform a combined xpending and xclaim flow.
+    ///
+    /// ```no_run
+    /// use redis::{Connection,Commands,RedisResult};
+    /// use redis::streams::{StreamAutoClaimOptions, StreamAutoClaimReply};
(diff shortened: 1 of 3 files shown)


jaymell:
Hi -- it looks like you based this off #778? Can you actually pull in @kkharji's commit from their remote and add your changes on top in a separate commit?

urkle:
I ddn't know about that other PR. I should have searched first :)

but mine does handle more scenarios with my StreamAutoClaimReply struct than the other PR as I handle redis 6 and redis 7 responses as well as cleanly handling the JUSTID vs non JUSTID variants with that one struct.

urkle:
@jaymell which approach do you prefer for the StreamAutoClaimReply struct as that seems to be the main difference between my PR and  the PR by @kkharji

jaymell:
We clearly need the refactoring you've added here. Do you and @kkharji want to figure that out?

urkle:
@jaymell the other PR was closed by the author. So, what thoughts do you have on this PR so it can be merged?

jaymell (approved):
Thanks -- LGTM.

PR #1179, opened 2024-05-01
Update rustls + tokio-rustls + futures-rustls

nihohit:


nihohit:
@djc have I done this right?

redis/Cargo.toml
```diff
@@ -100,7 +100,7 @@ cluster = ["crc16", "rand"]
 script = ["sha1_smol"]
 tls-native-tls = ["native-tls"]
 tls-rustls = ["rustls", "rustls-native-certs", "rustls-pemfile", "rustls-pki-types"]
-tls-rustls-insecure = ["tls-rustls"]
+tls-rustls-insecure = ["tls-rustls", "rustls/ring"]
```
djc, line 103: This doesn't make sense, why did you pull in `rustls/ring` only for the `-insecure` variant?

redis/tests/support/mod.rs
```diff
@@ -244,6 +244,13 @@ impl RedisServer {
         modules: &[Module],
         spawner: F,
     ) -> RedisServer {
+        #[cfg(feature = "rustls")]
+        if rustls::crypto::CryptoProvider::get_default().is_none() {
+            rustls::crypto::ring::default_provider()
```
djc, line 249: The `ring` provider isn't enabled by default in rustls 0.23, because the default switched to aws-lc-rs. You can probably just use the aws-lc-rs default provider for this tests, but as a downstream user I'd like a way to disable the aws-lc-rs provider (which is enabled via rustls' default features) if I want to use ring instead.

nihohit: > but as a downstream user I'd like a way to disable the aws-lc-rs provider (which is enabled via rustls' default features) if I want to use ring instead.

how could we do that? I'd like to avoid the addition of new features.

nihohit: do you suggest avoiding pulling the default features?

redis/Cargo.toml
nihohit: because of this line that you've added - https://github.com/redis-rs/redis-rs/blob/3cf579fd0da327d018559f1981e6cc27d6fe9cb3/redis/src/connection.rs#L906

nihohit: can it be replaced with something generic?

redis/tests/support/mod.rs
djc: > > but as a downstream user I'd like a way to disable the aws-lc-rs provider (which is enabled via rustls' default features) if I want to use ring instead.
> 
> how could we do that? I'd like to avoid the addition of new features.

At the very least you should depend on rustls (and tokio-rustls) with `default-features = false`, yes. But then if you want to avoid panics you probably also want to have features that do enable one of the crypto providers.

nihohit: so, that's a backwards incompatible change? I'll leave it out of this PR, then.

djc: What exactly do you see as a backwards incompatible change? I think you can still make this all work out without backwards incompatible changes.

nihohit: Correct me if I'm wrong here, but if we add `default-features = false`, then we can't initialize a crypto provider here
https://github.com/redis-rs/redis-rs/blob/3cf579fd0da327d018559f1981e6cc27d6fe9cb3/redis/src/connection.rs#L906

and so this path will panic for a user that didn't initialize a crypto provider beforehand, right?

djc: You'd add `default-features = false` but also `features = [..]` to enable the some of the default features and pick a crypto provider.

redis/Cargo.toml
```diff
@@ -65,7 +65,7 @@ tokio-native-tls = { version = "0.3", optional = true }
 async-native-tls = { version = "0.5", optional = true }
 
 # Only needed for rustls
-rustls = { version = "0.23", optional = true }
+rustls = { version = "0.23", optional = true, default-features = false }
```
djc, line 68: You'll also want this for tokio-rustls.

nihohit:
@djc good to go?

redis/Cargo.toml
```diff
@@ -65,11 +65,11 @@ tokio-native-tls = { version = "0.3", optional = true }
 async-native-tls = { version = "0.5", optional = true }
 
 # Only needed for rustls
-rustls = { version = "0.22", optional = true }
+rustls = { version = "0.23", optional = true, default-features = false, features = ["ring"] }
```
djc, line 68: I'd remove `features = ["ring"]` from here and instead add `"tokio-rustls/ring"` to the features enabled via `tls-rustls`.

redis/Cargo.toml
```diff
@@ -65,11 +65,11 @@ tokio-native-tls = { version = "0.3", optional = true }
...
 
 # Only needed for rustls
-rustls = { version = "0.22", optional = true }
+rustls = { version = "0.23", optional = true, default-features = false }
 webpki-roots = { version = "0.26", optional = true }
 rustls-native-certs = { version = "0.7", optional = true }
-tokio-rustls = { version = "0.25", optional = true }
-futures-rustls = { version = "0.25", optional = true }
+tokio-rustls = { version = "0.26", optional = true, default-features = false }
+futures-rustls = { version = "0.26", optional = true }
```
djc, line 72: Sorry, I missed that there's also a futures-rustls dependency here -- that also needs the `default-features = false`.

djc (approved):

PR #1184, opened 2024-05-11
Correct the document of default feature flags

naskya:
As per `Cargo.toml`, `aio` is not enabled by default, and `keep-alive` is enabled:

https://github.com/redis-rs/redis-rs/blob/523b4ebe4ca052f97ea3db5b7792601092106397/redis/Cargo.toml#L94

However, the documentation for these flags is incorrect. This pull request fixes this error.

https://github.com/redis-rs/redis-rs/blob/523b4ebe4ca052f97ea3db5b7792601092106397/redis/src/lib.rs#L56

https://github.com/redis-rs/redis-rs/blob/523b4ebe4ca052f97ea3db5b7792601092106397/redis/src/lib.rs#L65

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -53,7 +53,7 @@
 //! if so desired.  Some of them are turned on by default.
 //!
 //! * `acl`: enables acl support (enabled by default)
-//! * `aio`: enables async IO support (enabled by default)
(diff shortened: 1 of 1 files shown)


jaymell (approved):
Thanks for the fix!

PR #1186, opened 2024-05-14
Handle errors even when out of retries.

nihohit:
Async cluster connections now can handle request errors even when the request shouldn't retry.
Before this change, topology refreshes and reconnects only happened on retries. This change ensures that they will happen regardless of retries.
This bug is evident when users create a connection with 0 retries - it will never reconnect or refresh topology.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -425,15 +425,18 @@ enum Next<C> {
         request: PendingRequest<C>,
     },
     Reconnect {
-        request: PendingRequest<C>,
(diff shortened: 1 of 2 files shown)


barshaul (commented):
If there's no more retries and we need to refresh the slots/connection, can we immediately return to the user with the response and spawn a different task to do the refreshment?

jaymell:
I would prefer we do _not_ merge this. Either #1188 or a PR based on the branch you've referenced previously would be preferable, IMO.

nihohit:
@jaymell - https://github.com/redis-rs/redis-rs/pull/1222

PR #1188, opened 2024-05-18
Refactor async-cluster error handling

jaymell:
*  Refactor async-cluster error handling
    Attempts to refactor request error handling to more clearly
    separate connection-wide issues from request-specific issues and
    also fix the problem identified in #1186 -- namely, that connection-
    wide issues will be addressed even if requests are not retried.

* Add tests
   Add tests written by @nihohit in #1186.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -290,6 +290,7 @@ enum Response {
     Multiple(Vec<Value>),
 }
 
+#[derive(Debug)]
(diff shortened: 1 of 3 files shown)


redis/src/cluster_async/mod.rs
```diff
@@ -1222,6 +1217,7 @@ where
     }
 }
 
+#[derive(Debug)]
```
jaymell, line 1220: Maybe this could be better named.

Issue #1190, opened 2024-05-23
multiplex connection doesn't recover from ErrorKind::IoError


shlomitda:
Version: 0.25.3

topology:
redis-client (multiplexed connection) -> HA proxy -> redis HA (multiple replicas)

Issue:
When there is no redis server available (temporarily) , HA proxy closes the connection with redis-client .

on redis-client side we get the error:

return Err(RedisError::from((
    crate::ErrorKind::IoError,
    "Multiplexed connection driver unexpectedly terminated",
)));

error returned:
ErrorRepr::WithDescription(crate::ErrorKind::IoError, "Multiplexed connection driver unexpectedly terminated")

Client doesn't recover from this error , and doesn't try to establish a new connection
because this error is not considered I/O error in reconnect_if_io_error!

Why this type of flow is not handled as ErrorRepr:IoError

Thanks in advance for your help!

nihohit:
The multiplexed connection doesn't recover from disconnects of any kind. If
you want a recovering connection, use 'connection_manager'.

shlomitda:
Hi
sorry it wasn't clear, we use connection_manager 
still doesn't recover

nihohit:
Thanks for the clarification.

Do you know what's the error that the HA proxy uses to close the connection? We have tests for a [closed server](https://github.com/redis-rs/redis-rs/blob/3cf579fd0da327d018559f1981e6cc27d6fe9cb3/redis/tests/test_async.rs#L960), so I want to understand the differences between the behavior of the proxy & the server.

The thing you can provide is code that will allow me to reproduce this issue, preferably within redis-rs' test suite, but even just the actual error that the proxy sends the connection is a good starting point.

nihohit:
@shlomitda can you pull 0.25.4 and check whether this keeps happening? a mitigation fix has been pushed.

shlomitda:
@nihohit 
issue was resolved 
thank you very much!

PR #1191, opened 2024-05-23
Fix explicit IoError not being recognized.

nihohit:
A RedisError that was created with `ErrorKind::IoError` instead of an actual `io::Error` will now be recognized as an I/O error when checking `is_io_error`. This is required for the correct behavior of `connection_manager`.

https://github.com/redis-rs/redis-rs/issues/1190


jaymell (approved):
:muscle:

PR #1192, opened 2024-05-24
Add  max_delay for every reconnect

tonynguyen-sotatek:


redis/src/connection.rs
```diff
@@ -211,6 +211,9 @@ pub struct ConnectionInfo {
 
     /// A boxed connection address for where to connect to.
     pub redis: RedisConnectionInfo,
+
+    /// In case of reconnection issues, the manager will retry reconnection
+    pub retry_strategy: Option<RetryStrategyInfo>,
```
nihohit, line 216: this isn't the place for this - if only one connection type uses this, it should be in that type's config, not here.

redis/src/aio/connection_manager.rs
```diff
@@ -123,12 +134,14 @@ impl ConnectionManager {
         exponent_base: u64,
         factor: u64,
         number_of_retries: usize,
+        max_delay: Option<u64>,
```
nihohit, line 137: instead of adding another field to the function, which is a breaking change, you should add another function taking that field.
and in order to limit the number of functions, please follow the example [in this PR](https://github.com/redis-rs/redis-rs/pull/1167) - that is,  add a configuration object, and a function using that configuration object, so that new fields won't break users.

nihohit:
as a rule of thumb, PRs should come either with a description or an issue explaining what is the goal of the PR and justifying the design chosen in order to achieve this goal.

redis/src/aio/connection_manager.rs
tonynguyen-sotatek: Thank you, i will update code with new PR

PR #1194, opened 2024-05-27
Add max retry delay for every reconnect

tonynguyen-sotatek:
Add a redis::aio::ConnectionManager::new_with_backoff_and_timeouts_with_max_delay  Apply a maximum delay. No retry delay will be longer than this `max_delay`.

Todo: I try to add some more cases

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -20,7 +20,7 @@ use tokio_retry::Retry;
 
 /// Apply a maximum delay. No retry delay will be longer than this `max_delay`.
 #[derive(Clone, Debug, Default)]
-pub struct ConnectionConfigInfo {
+pub struct ConnectionManagerConfig {
     /// The resulting duration is calculated by taking the base to the `n`-th power,
     /// where `n` denotes the number of past attempts.
     exponent_base: u64,
@@ -38,7 +38,7 @@ pub struct ConnectionConfigInfo {
     connection_timeout: std::time::Duration,
 }
 
-impl ConnectionConfigInfo {
+impl ConnectionManagerConfig {
     const DEFAULT_CONNECTION_RETRY_EXPONENT_BASE: u64 = 2;
     const DEFAULT_CONNECTION_RETRY_FACTOR: u64 = 100;
     const DEFAULT_NUMBER_OF_CONNECTION_RETRIESE: usize = 6;
@@ -58,37 +58,37 @@ impl ConnectionConfigInfo {
     }
 
     /// Sets the factor
-    pub fn factor(mut self, factor: u64) -> ConnectionConfigInfo {
+    pub fn factor(mut self, factor: u64) -> ConnectionManagerConfig {
         self.factor = factor;
         self
     }
 
     /// Sets the max_delay
-    pub fn max_delay(mut self, duration: u64) -> ConnectionConfigInfo {
-        self.max_delay = Some(duration);
+    pub fn max_delay(mut self, time: u64) -> ConnectionManagerConfig {
+        self.max_delay = Some(time);
         self
     }
 
     /// Sets the exponent_base
-    pub fn exponent_base(mut self, duration: u64) -> ConnectionConfigInfo {
-        self.exponent_base = duration;
+    pub fn exponent_base(mut self, base: u64) -> ConnectionManagerConfig {
+        self.exponent_base = base;
         self
     }
 
     /// Sets the number_of_retries
-    pub fn number_of_retries(mut self, amount: usize) -> ConnectionConfigInfo {
+    pub fn number_of_retries(mut self, amount: usize) -> ConnectionManagerConfig {
         self.number_of_retries = amount;
         self
     }
 
     /// Sets the response_timeout
-    pub fn response_timeout(mut self, duration: std::time::Duration) -> ConnectionConfigInfo {
+    pub fn response_timeout(mut self, duration: std::time::Duration) -> ConnectionManagerConfig {
         self.response_timeout = duration;
         self
     }
 
     /// Sets the response_timeout
-    pub fn connection_timeout(mut self, duration: std::time::Duration) -> ConnectionConfigInfo {
+    pub fn connection_timeout(mut self, duration: std::time::Duration) -> ConnectionManagerConfig {
         self.connection_timeout = duration;
         self
     }
@@ -173,15 +173,9 @@ impl ConnectionManager {
     /// This requires the `connection-manager` feature, which will also pull in
     /// the Tokio executor.
     pub async fn new(client: Client) -> RedisResult<Self> {
-        let config = ConnectionConfigInfo::new();
+        let config = ConnectionManagerConfig::new();
 
-        Self::new_with_backoff(
-            client,
-            config.exponent_base,
-            config.factor,
-            config.number_of_retries,
-        )
-        .await
+        Self::new_with_config(client, config).await
     }
 
     /// Connect to the server and store the connection inside the returned `ConnectionManager`.
@@ -228,14 +222,14 @@ impl ConnectionManager {
         response_timeout: std::time::Duration,
         connection_timeout: std::time::Duration,
     ) -> RedisResult<Self> {
-        let config = ConnectionConfigInfo::new()
+        let config = ConnectionManagerConfig::new()
             .exponent_base(exponent_base)
             .factor(factor)
             .number_of_retries(number_of_retries)
             .response_timeout(response_timeout)
             .connection_timeout(connection_timeout);
 
-        Self::new_with_backoff_and_timeouts_new_with_config(client, config).await
+        Self::new_with_config(client, config).await
     }
 
     /// Connect to the server and store the connection inside the returned `ConnectionManager`.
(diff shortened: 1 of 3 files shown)


redis/src/aio/connection_manager.rs
```diff
@@ -153,15 +207,53 @@ impl ConnectionManager {
...
+    ///
+    /// In case of reconnection issues, the manager will retry reconnection
+    /// number_of_retries times, with an exponentially increasing delay, calculated as
+    /// rand(0 .. factor * (exponent_base ^ current-try)).
+    ///
+    /// Apply a maximum delay. No retry delay will be longer than this `RetryStrategyInfo.max_delay` .
+    ///
+    /// The new connection will timeout operations after `response_timeout` has passed.
+    /// Each connection attempt to the server will timeout after `connection_timeout`.
+    pub async fn new_with_backoff_and_timeouts_with_max_delay(
```
nihohit, line 237: please use a single configuration object that contains both the retry & the timeouts. we don't add many fields, just a single field that represents the whole config of the connection manager.

accordingly, this funciton should be called `new_with_config`

redis/src/aio/connection_manager.rs
```diff
@@ -18,6 +18,62 @@ use std::sync::Arc;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
+/// Apply a maximum delay. No retry delay will be longer than this `max_delay`.
+#[derive(Clone, Debug, Default)]
+pub struct RetryStrategyInfo {
+    /// The resulting duration is calculated by taking the base to the `n`-th power,
+    /// where `n` denotes the number of past attempts.
+    pub exponent_base: u64,
```
nihohit, line 26: the fields of this struct should be private, in order to prevent breaking the user's code if fields are added.

redis/tests/test_async.rs
```diff
@@ -994,6 +994,56 @@ mod basic_async {
...
+            .factor(10000)
+            .max_delay(500);
+
+        let tempdir = tempfile::Builder::new()
+            .prefix("redis")
+            .tempdir()
+            .expect("failed to create tempdir");
+        let tls_files = build_keys_and_certs_for_tls(&tempdir);
+
+        let ctx = TestContext::with_tls(tls_files.clone(), false);
```
nihohit, line 1013: why build with tls instead of just a new context?

redis/tests/test_async.rs
```diff
@@ -994,6 +994,56 @@ mod basic_async {
...
+                    retry_strategy_info.clone(),
+                    std::time::Duration::MAX,
+                    std::time::Duration::MAX,
+                )
+                .await
+                .unwrap();
+            let server = ctx.server;
+            let addr = server.client_addr().clone();
+            let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+            manager.get_push_manager().replace_sender(tx.clone());
```
nihohit, line 1027: why bother with the push manager?

redis/tests/test_async.rs
```diff
@@ -994,6 +994,56 @@ mod basic_async {
         .unwrap();
     }
 
+    #[test]
+    #[cfg(feature = "connection-manager")]
+    fn test_connection_manager_reconnect_after_delay_with_max_delay() {
```
nihohit, line 999: this is just a happy-path test, right? As far as I see the max delay isn't tested here.
If so, I don't the added benefit of this test, and IMO there's no need to add it. You can change the existing test to use `new_with_config` instead.

nihohit (commented):
thanks!
There's no need to close the PR and open a new one for changes - just change, commit & push. the PR will be updated automatically.

redis/src/aio/connection_manager.rs
tonynguyen-sotatek: Is it possible to name it ConnectionConfigInfo?

redis/tests/test_async.rs
tonynguyen-sotatek: random delay in a range so I can't test the exact number, I try to add some more cases

nihohit: if this isn't easily tested, the tests can be omitted.

redis/src/aio/connection_manager.rs
```diff
@@ -18,6 +18,82 @@ use std::sync::Arc;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
+/// Apply a maximum delay. No retry delay will be longer than this `max_delay`.
```
nihohit, line 21: fix doc - this doesn't describe the whole struct

redis/src/aio/connection_manager.rs
```diff
@@ -18,6 +18,82 @@ use std::sync::Arc;
...
+    /// The resulting duration is calculated by taking the base to the `n`-th power,
+    /// where `n` denotes the number of past attempts.
+    exponent_base: u64,
+    /// A multiplicative factor that will be applied to the retry delay.
+    ///
+    /// For example, using a factor of `1000` will make each delay in units of seconds.
+    factor: u64,
+    /// number_of_retries times, with an exponentially increasing delay
+    number_of_retries: usize,
+    /// Apply a maximum delay. No retry delay will be longer than this 'duration' milliseconds.
```
nihohit, line 33: why not use Duration instead of u64?

nihohit (commented):
getting there :)

redis/src/aio/connection_manager.rs
tonynguyen-sotatek: we are using tokio_retry::strategy::ExponentialBackoff which is using u64
max_delay: maximum wait time per reconnection
connection_timeout: Each connection attempt to the server will timeout after `connection_timeout`.

redis/tests/test_async.rs
```diff
@@ -994,6 +994,53 @@ mod basic_async {
...
+        let tempdir = tempfile::Builder::new()
+            .prefix("redis")
+            .tempdir()
+            .expect("failed to create tempdir");
+        let tls_files = build_keys_and_certs_for_tls(&tempdir);
+
+        let ctx = TestContext::with_tls(tls_files.clone(), false);
+        block_on_all(async move {
+            let mut manager =
+                redis::aio::ConnectionManager::new_with_backoff_and_timeouts_new_with_config(
```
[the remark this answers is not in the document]
tonynguyen-sotatek: I'm trying to mock ConnectionManager's reconnection method. Can you help me?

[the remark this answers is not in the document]
nihohit: IMO it's not possible at the moment - the connection manager isn't generic, and you can't inject the logic. How would you mock this?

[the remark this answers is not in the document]
tonynguyen-sotatek: I want to mock the reconnect method, to make sure the params are correct when reconnecting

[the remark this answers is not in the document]
nihohit: How would you mock it? I've never used a mock in rust, so I don't know how to do that.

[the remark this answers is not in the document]
tonynguyen-sotatek: I don't see a solution so I'm trying to ask you if there is any solution. i will try 1 or 2 days for this case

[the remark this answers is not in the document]
nihohit: As far as I know it isn't possible at the moment, so just remove the test and change the existing tests to use the new initializer.

redis/src/aio/connection_manager.rs
nihohit, line 21: Hi, this is still the wrong documentation - it should be `configuration object for `ConnectionManager`., or something like it.

redis/src/aio/connection_manager.rs
```diff
@@ -18,6 +18,82 @@ use std::sync::Arc;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
+/// Apply a maximum delay. No retry delay will be longer than this `max_delay`.
+#[derive(Clone, Debug, Default)]
+pub struct ConnectionManagerConfig {
+    /// The resulting duration is calculated by taking the base to the `n`-th power,
```
nihohit, line 24: since these are private fields, this documentation should either be moved or copied to the relevant set_foo functions - ATM the user won't see these docs.

redis/src/aio/connection_manager.rs
```diff
@@ -18,6 +18,82 @@ use std::sync::Arc;
...
+    /// where `n` denotes the number of past attempts.
+    exponent_base: u64,
+    /// A multiplicative factor that will be applied to the retry delay.
+    ///
+    /// For example, using a factor of `1000` will make each delay in units of seconds.
+    factor: u64,
+    /// number_of_retries times, with an exponentially increasing delay
+    number_of_retries: usize,
+    /// Apply a maximum delay. No retry delay will be longer than this 'duration' milliseconds.
```
nihohit, line 33: Apply a maximum delay between connection attempts. The delay between attempts won't be longer than `max_delay` milliseconds.

nihohit (commented):
just needs to fix the docs, so that they'll be public & correct.

nihohit (approved):

nihohit:
Thanks for the contribution!

nihohit:
please fix the code - the CI fails.

nihohit:
please fix the build, and test that it builds before pushing.

nihohit:
Thanks for the contribution!

PR #1198, opened 2024-05-29
Fix typos

wutchzone:
I have noticed few typos.


nihohit (approved):
thanks for the corrections!

PR #1200, opened 2024-05-31
Log the server / cluster logfile on error.

nihohit:
This allows us to find errors that aren't visible
in stdout/stderr.

### DIFF
--- a/redis/tests/support/cluster.rs
+++ b/redis/tests/support/cluster.rs
@@ -204,6 +204,8 @@ impl RedisCluster {
                         let mut process = cmd.spawn().unwrap();
                         sleep(Duration::from_millis(50));
 
+                        let log_file_index = cmd.get_args().position(|arg|arg == "--logfile").unwrap() + 1;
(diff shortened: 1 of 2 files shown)


redis/tests/support/cluster.rs
```diff
@@ -216,8 +218,10 @@ impl RedisCluster {
                                     out.read_to_string(&mut str).unwrap();
                                     str
                                 });
+
+                                let log_file_contents = std::fs::read_to_string(log_file_path).unwrap();
```
jaymell, line 222: How verbose are these log files potentially? I get nervous any time I see a file get read into a string :sweat_smile:

jaymell (approved):

redis/tests/support/cluster.rs
nihohit: so far I've only seen 4 lines of text in a file. Remember that these are read only if the server fails initialization, not after the server has started running, so I assume it won't have a chance to grow.

PR #1202, opened 2024-05-31
Send retries in multi-node reconnect to new connection.

nihohit:
This fixes a situtation where a multi-node request is sent after a disconnect, and is repeatedly being sent to the old, disconnected connection. This is caused by multi-node requests being routed with a specific connection, instead of looking for a new connection after a disconnect.

https://github.com/amazon-contributing/redis-rs/issues/144

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -364,26 +364,31 @@ impl<C> RequestInfo<C> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


nihohit:
A simpler solution is to just remove the Connection variant - it's a minor optimization that saves fetching the connection multiple times, and might not be worth the potential for more bugs.

jaymell (approved):

PR #1217, opened 2024-06-06
Fix some typos in connection_manager.rs and client.rs

meierfra-ergon:
In my code, using the connection_manager, I would like to only set the timeouts but not the back off parameters on creation of the connection_manager. Therefore I suggest to add a new function `new_with_timeouts()` to the connection_manager and subsequently `get_connection_manager_with_timeouts()` to the client.

I also fixed some typos

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -32,16 +32,16 @@ pub struct ConnectionManagerConfig {
     number_of_retries: usize,
     /// Apply a maximum delay between connection attempts. The delay between attempts won't be longer than max_delay milliseconds.
     max_delay: Option<u64>,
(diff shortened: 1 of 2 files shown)


nihohit:
hi, please rebase over https://github.com/redis-rs/redis-rs/pull/1194 - I believe that PR answered your request.

meierfra-ergon:
ah yes with #1194 it is much more comfortable. So actually my new functions would not be needed. I updated the PR anyway, but I would understand if you would decline the new functions due to code bloat.

nihohit:
yes, IMO the new functions are unnecessary, but please keep the typo fixes, and rename the PR.

meierfra-ergon:
PR updated as requested

nihohit (approved):

PR #1219, opened 2024-06-06
Implement `ToRedisArgs` for `std::borrow::Cow`

caass:
In cases where `T: ToOwned, &T: ToRedisArgs, &T::Owned: ToRedisArgs` we can implement `ToRedisArgs` for `Cow<'_, T>` by delegating to the inner implementations. This gives support for `Cow<[u8]>` and `Cow<str>` out of the box, and allows downstream consumers to get `Cow<Newtype>` implementations for free, without needing to `match` in their code.

An alternative implementation would be to always call `Cow::as_ref`, which would relax the bound of `&T::Owned: ToRedisArgs`, but in cases where the implementation of `ToRedisArgs` is specialized for `&T::Owned` this is suboptimal. As an example, see [`String::as_bytes`][1] vs [`String::as_str`][2] followed by [`str::as_bytes`][3]; by using the specialized implementation of `String::to_redis_args`, we save some overhead.

[1]: https://doc.rust-lang.org/std/string/struct.String.html#method.as_bytes
[2]: https://doc.rust-lang.org/std/string/struct.String.html#method.as_str
[3]: https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::default::Default;
 use std::error;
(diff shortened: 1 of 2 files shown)


nihohit (approved):
Thanks for the contribution!

fosskers:
Just ran into the need for this and was about to implement it when I saw this PR. Thanks!

PR #1220, opened 2024-06-07
Implement connection manager with lazy connect

tonynguyen-sotatek:
Implement connection manager with lazy connect

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -18,6 +18,18 @@ use std::sync::Arc;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
+/// Policy defining how to combine multiple responses into one.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ConnectionStatus {
+    Wait,
+    // Reconnecting,  // will use it in the future
+    // Connecting, // will use it in the future
+    Connect,
+    // Ready, // will use it in the future
+    // Close, // will use it in the future
+    // End, // will use it in the future
+}
+
 /// A `ConnectionManager` is a proxy that wraps a [multiplexed
 /// connection][multiplexed-connection] and automatically reconnects to the
 /// server when necessary.
@@ -51,7 +63,7 @@ pub struct ConnectionManager {
     ///
     /// The `ArcSwap` is required to be able to replace the connection
     /// without making the `ConnectionManager` mutable.
-    connection: Arc<ArcSwap<SharedRedisFuture<MultiplexedConnection>>>,
+    connection: Option<Arc<ArcSwap<SharedRedisFuture<MultiplexedConnection>>>>,
 
     runtime: Runtime,
     retry_strategy: ExponentialBackoff,
@@ -59,6 +71,7 @@ pub struct ConnectionManager {
     response_timeout: std::time::Duration,
     connection_timeout: std::time::Duration,
     push_manager: PushManager,
+    connection_status: ConnectionStatus,
 }
 
 /// A `RedisResult` that can be cloned because `RedisError` is behind an `Arc`.
@@ -66,6 +79,13 @@ type CloneableRedisResult<T> = Result<T, Arc<RedisError>>;
 
 /// Type alias for a shared boxed future that will resolve to a `CloneableRedisResult`.
 type SharedRedisFuture<T> = Shared<BoxFuture<'static, CloneableRedisResult<T>>>;
+
+/// Handle a command result. If the connection was dropped, reconnect.
+macro_rules! connect_with_lazy_connect {
+    ($self:expr) => {
+        $self.new_connection_lazy_connect().await;
+    };
+}
 
 /// Handle a command result. If the connection was dropped, reconnect.
 macro_rules! reconnect_if_dropped {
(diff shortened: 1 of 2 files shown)


nihohit:
I see this is only in the draft stage. Please, before you mark this as ready for review, make sure that you add an explanation of what is the goal you're trying to achieve, and what's the general design guiding it.

xoac (approved):
LGTM

xoac:
@nihohit the goal of this PR is to make connection to redis optional on application startup. With lazy constructor connection is deferred until needed and if failed it's handled in single place the same way as it was broken after successful connection. 

This avoid create in application that use redis_rs crate special wrapper around redis-rs client that would implement reconnection (but only on startup because if connection fails after init successful connection then redis_rs can handle it). I hope you find this PR to be worth of merging!

nihohit:
Maybe I'm missing something, but wouldn't you get exactly the same effect by using `OnceLock::get_or_init`?

I wouldn't want to add a new API, with all relevant maintenance cost, if the standard library already contains a working alternative.
In any case, notice that the test you've added doesn't actually test the laziness of the initializer - you'd need to check the server whether a connection was made before & after a command is sent on the connection.

xoac:
> In any case, notice that the test you've added doesn't actually test the laziness of the initializer - you'd need to check the server whether a connection was made before & after a command is sent on the connection.

That is true. The test check if lazy connection can eventually connect to server. I though that laziness doesn't need to be checked because function `new_with_config_lazy` is synchronous and internally created feature do nothing until pulled. So the lazy "check is done" on compile time. Is that assumption of mine incorrect?

> I wouldn't want to add a new API, with all relevant maintenance cost
Any reason why this could add maintenance costs? It reuse your implementation that allow connection to be lazy (just wasn't exposed). Do you mean possible future refactor? 

> Maybe I'm missing something, but wouldn't you get exactly the same effect by using OnceLock::get_or_init?
-  You need re-implement `reconnect` function in your application with `OnceLock::get_or_init` and you don't have access to already implemented things like "back_off" from `ConnectionManager`
- Passing `ConnectionManager` is simple, on the other hand with OnceLocked you need either create global version (that you need to check in every place if is already initialized) or create own wrapper that share the same purpose as `ConnectionManager` --  reconnect to redis server when needed.
-  OnceLock is not async
- with OnceLock initialization that is done is separate thread, without extra code it's not possible to await for connection in different task (you can just check if OnceLock container was already initialized or not) 


example pseudo code
```rs
async fn data_loader(cache: &mut ConnectionManager, loader: Loader) -> i32 {

  let cache_result = cache.get("key").await();
  match cache_result {
    Ok(v) => return v;
    Err(err) => {
       // SINGLE place to handle all errors
       // here error is just ignored and 
      loader.get("key").await.unwrap()
    }
  }
}

fn main() {
  // `new_with_config` if you wish to handle first connection in special way
  let redis = ConnectionManager::new_with_config_lazy(...);
  let _ = data_loader(&mut redis, Loader::new()).await();  
}
```

Let me know if I correctly understand your suggestion for `OnceLock`. Also waiting your further opinion on this PR

redis/src/aio/connection_manager.rs
```diff
@@ -267,29 +287,28 @@ impl ConnectionManager {
...
+        let push_manager_cloned = push_manager.clone();
+        let connection = Self::new_connection(
             client.clone(),
             retry_strategy.clone(),
             config.number_of_retries,
             config.response_timeout,
             config.connection_timeout,
         )
-        .await?;
+        .map_err(Arc::new)
```
nihohit, line 298: why is the error mapped to `Arc`?

redis/src/client.rs
```diff
@@ -592,6 +592,35 @@ impl Client {
         crate::aio::ConnectionManager::new_with_config(self.clone(), config).await
     }
 
+    /// Returns an async [`ConnectionManager`][connection-manager] from the client.
```
nihohit, line 595: this documentation doesn't explain at all the difference of this initializer from the other one. The lazyness needs to be properly explained here.

nihohit:
Sorry it took a while to answer. The change is generally ok, and I'm not strongly opposed to it, but I do want to make sure that we're not adding code that can be rather easily replaced by the user by combining us with another crate. @jaymell thoughts on this?

> Any reason why this could add maintenance costs?

Every new public function, or even field, incurs potential maintenance cost. There could be bugs in the code you introduced. There would be feature requests for the code you introduced. EVERYTHING has a cost.

> You need re-implement reconnect function in your application with OnceLock::get_or_init and you don't have access to already implemented things like "back_off" from ConnectionManager

why?

> OnceLock is not async

Then use the async_once_cell crate - I believe [this is what you're looking for](https://docs.rs/async-once-cell/latest/async_once_cell/struct.Lazy.html).

> Passing ConnectionManager is simple, on the other hand with OnceLocked you need either create global version (that you need to check in every place if is already initialized) or create own wrapper that share the same purpose as ConnectionManager -- reconnect to redis server when needed

Or you could just wrap the Lazy oncelock with an `Arc`, and clone it around. You'd still need to await the connection, I agree that the ergonomics are worse, but it gets the job done.

nihohit:
@xoac are you still interested in pushing this change?

nihohit:
closed in favor of https://github.com/redis-rs/redis-rs/pull/2022

PR #1222, opened 2024-06-09
Cluster connection - Refactor response handling.

nihohit:
This PR is based over https://github.com/redis-rs/redis-rs/pull/1186

1. Adds a request.rs and routing.rs files to the async cluster connection, in order to move the additional logic outside of the huge mod.rs file.
2. Moves the error handling logic to a seprate function, and add tests for that function.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -769,6 +769,16 @@ where
                             return Err(err);
                         }
                         crate::types::RetryMethod::RetryImmediately => {}
+                        crate::types::RetryMethod::ReconnectFromInitialConnections => {
+                            // TODO - implement reconnect from initial connections
+                            if *self.auto_reconnect.borrow() {
+                                if let Ok(mut conn) = self.connect(&addr) {
+                                    if conn.check_connection() {
+                                        self.connections.borrow_mut().insert(addr, conn);
+                                    }
+                                }
+                            }
+                        }
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


jaymell:
Is the second commit intended, or was that just a missed fixup?

I'm honestly not sold on the need to restructure the files, but regardless -- can we break these changes into smaller PRs? I'm not confident I can review this as is. And I still don't like #1186 :laughing:

nihohit:
> Is the second commit intended, or was that just a missed fixup?

mistake

> I'm honestly not sold on the need to restructure the files

There's definitely no *need* - the code works as well in a single file :). But consider - cluster-async/mod.rs is currently at 1500~ LOC. If we kept all of the changes in this PR in a single file, it would get to ~1800 LOC. At some point the file just becomes very unwieldy.

> can we break these changes into smaller PRs?

Sure - https://github.com/redis-rs/redis-rs/pull/1222/commits/7db2abd4ce007c1ade932fa6ac6718f7356daa90 can easily be its own PR, either based on the current main, rebased on https://github.com/redis-rs/redis-rs/pull/1186 or on https://github.com/redis-rs/redis-rs/pull/1188. which one would you prefer?
That would definitely reduce the size of the diff.

jaymell:
> Sure - https://github.com/redis-rs/redis-rs/commit/7db2abd4ce007c1ade932fa6ac6718f7356daa90 can easily be its own PR, either based on the current main, rebased on https://github.com/redis-rs/redis-rs/pull/1186 or on https://github.com/redis-rs/redis-rs/pull/1188. which one would you prefer?

I think it would be great to move the restructuring into its own PR base on current main, so we can see the other code changes here in isolation.

nihohit:
https://github.com/redis-rs/redis-rs/pull/1226

nihohit:
rebased

jaymell (approved):

PR #1127, opened 2024-04-02
Remove unnecessary clones from pubsub codepaths.

nihohit:

PR #1224, opened 2024-06-14
Fix compilation break.

nihohit:
Introduced by https://github.com/redis-rs/redis-rs/pull/1096 not being rebased over https://github.com/redis-rs/redis-rs/pull/1127 when tested.


nihohit:
benchmark fails because main can't compile, so it can't compare results.
nightly fails due to a check that seems to be broken - https://github.com/rust-lang/rust/issues/123748#issuecomment-2167530990

Issue #1228, opened 2024-06-19
Warning about Rust 2024 with the newest Nightly Compiler: `this function depends on never type fallback being ()`


kamulos:
In my code I often use code that ignores the return value of operations like `set`:

```rust
use redis::{Commands, Connection, RedisError};

fn do_something(connection: &mut Connection) -> Result<(), RedisError> {
    connection.set("test", "test_data")?;
    Ok(())
}
```

Now with the newest nightly compiler I get the following warning:

```
warning: this function depends on never type fallback being `()`
 --> src/main.rs:3:1
  |
3 | pub fn do_something(connection: &mut Connection) -> Result<(), RedisError> {
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #123748 <https://github.com/rust-lang/rust/issues/123748>
  = help: specify the types explicitly
note: in edition 2024, the requirement `!: FromRedisValue` will fail
 --> src/main.rs:4:16
  |
4 |     connection.set("test", "test_data")?;
  |                ^^^
  = note: `#[warn(dependency_on_unit_never_type_fallback)]` on by default
```

Now I don't entirely understand what's going on here, but this warning seems to be fixable for me like this:

```diff
- connection.set("test", "test_data")?;
+ let _: () = connection.set("test", "test_data")?;
```

This is just not very convenient nor pretty. Am I doing something wrong? Is there any way the `redis` crate can mute this warning?

nihohit:
I'm looking into this, but ATM I'm not sure how to solve this on the crate side.

kamulos:
Hmm, just tried it again. I think this could basically be the following situation:

```rust
#![feature(never_type)]

trait Constructible {
    fn construct() -> Self;
}

impl Constructible for () {
    fn construct() -> Self {
        ()
    }
}

// impl Constructible for ! {
//     fn construct() -> Self {
//         loop {}
//     }
// }

fn generic_return<T: Constructible>() -> Result<T, ()> {
    Ok(T::construct())
}

fn main() -> Result<(), ()> {
    generic_return()?;
    Ok(())
}
```

If I enable the implementation for `!`, the warning goes away. Unfortunately I still don't understand it...

nihohit:
[...]

PR #1229, opened 2024-06-19
Fix nightly compilation warnings.

nihohit:
Adds missing generic declarations or return types for `()`, when until the 2024 edition ! was inferred. This fixes this warning:

```
warning: this function depends on never type fallback being `()`
 --> redis/examples/async-await.rs:4:7
  |
4 | async fn main() -> redis::RedisResult<()> {
  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #123748 <https://github.com/rust-lang/rust/issues/123748>
  = help: specify the types explicitly
note: in edition 2024, the requirement `!: FromRedisValue` will fail
 --> redis/examples/async-await.rs:8:9
  |
8 |     con.set("key1", b"foo").await?;
  |         ^^^
  = note: `-D dependency-on-unit-never-type-fallback` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(dependency_on_unit_never_type_fallback)]`
```

In order to make this easier, this also changes the generics used in `Cmd/Pipeline::query_async`/`Script::invoke_async`, so that the function declaration would only need to define a single generic value. This changed the usage from `cmd.query_async::<_, ()>(&mut conn).await;` to `cmd.query_async::<()>(&mut conn).await;`. This was done since the compiler doesn't always suggest to use `_` for the connection type, so the user might think they need to give a concrete connection type, which might be onerous to the user.
This can be reverted and replaced with `let _: () = cmd.query_async(&mut conn).await;`, as is done with other calls.

### DIFF
--- a/redis/src/pipeline.rs
+++ b/redis/src/pipeline.rs
@@ -196,7 +196,7 @@ impl Pipeline {
     /// ```rust,no_run
     /// # let client = redis::Client::open("redis://127.0.0.1/").unwrap();
     /// # let mut con = client.get_connection().unwrap();
-    /// let _ : () = redis::pipe().cmd("PING").query(&mut con).unwrap();
+    /// redis::pipe().cmd("PING").query::<()>(&mut con).unwrap();
     /// ```
     ///
     /// NOTE: A Pipeline object may be reused after `query()` with all the commands as were inserted
@@ -204,7 +204,18 @@ impl Pipeline {
     ///       it is necessary to call the `clear()` before inserting new commands.
     #[inline]
     pub fn execute(&self, con: &mut dyn ConnectionLike) {
-        self.query::<()>(con).unwrap();
+        self.exec(con).unwrap();
+    }
+
+    /// This is a shortcut to `query`, to avoid having to define generic bounds for `()`.
+    pub fn exec(&self, con: &mut dyn ConnectionLike) -> RedisResult<()> {
+        self.query::<()>(con)
+    }
+
+    /// This is a shortcut to `query_async`, to avoid having to define generic bounds for `()`.
+    #[cfg(feature = "aio")]
+    pub async fn exec_async(&self, con: &mut impl crate::aio::ConnectionLike) -> RedisResult<()> {
+        self.query_async::<()>(con).await
     }
 }
 
--- a/redis/src/pipeline.rs
+++ b/redis/src/pipeline.rs
@@ -203,16 +203,24 @@ impl Pipeline {
     ///       to them. In order to clear a Pipeline object with minimal memory released/allocated,
     ///       it is necessary to call the `clear()` before inserting new commands.
     #[inline]
+    #[deprecated(note = "Use Cmd::exec + unwrap, instead")]
     pub fn execute(&self, con: &mut dyn ConnectionLike) {
         self.exec(con).unwrap();
     }
 
-    /// This is a shortcut to `query`, to avoid having to define generic bounds for `()`.
+    /// This is an alternative to `query`` that can be used if you want to be able to handle a
+    /// command's success or failure but don't care about the command's response. For example,
+    /// this is useful for "SET" commands for which the response's content is not important.
+    /// It avoids the need to define generic bounds for ().
+    #[inline]
     pub fn exec(&self, con: &mut dyn ConnectionLike) -> RedisResult<()> {
         self.query::<()>(con)
     }
 
-    /// This is a shortcut to `query_async`, to avoid having to define generic bounds for `()`.
+    /// This is an alternative to `query_async` that can be used if you want to be able to handle a
+    /// command's success or failure but don't care about the command's response. For example,
+    /// this is useful for "SET" commands for which the response's content is not important.
+    /// It avoids the need to define generic bounds for ().
     #[cfg(feature = "aio")]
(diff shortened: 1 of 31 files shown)


jaymell:
Yowza

nihohit:
> Yowza

a-yup.

nihohit:
https://github.com/redis-rs/redis-rs/issues/1228

jaymell:
I just wonder if there's a better way we can solve this. Maybe different methods if you're not anticipating a result, that exxplicitly return `Result<()>`?

jaymell:
> We can definitely add a `call/_async` functions that return `RedisResult<()>`, or `exec/_async` (although in such a case I would consider deprecating `execute` in order to remove confusion between similarily named functions), 

Yep I think deprecating `execute` is definitely a good idea.

> but I'm not sure about how easy would it be to find these functions.

What do you mean exactly here?
 
> I'd also be happy to somehow solve the issue for all of the `set`s around, but I'm not sure what's the solution then.

What if we modify the macros, or add a separate one for write-only commands that calls `exec_async` instead of `query_async` / with a hard-coded return value of `RedisResult<()>`? I guess that could get tricky for variants of commands that potentially return things, ie `SET ... GET`, but we could probably handle most of the use cases.

nihohit:
> > but I'm not sure about how easy would it be to find these functions.

> What do you mean exactly here?

mostly discoverability. we'll need to add those functions to the docs, and explain when to use one function or the other.

> What if we modify the macros, or add a separate one for write-only commands that calls exec_async instead of query_async / with a hard-coded return value of RedisResult<()>?

If we modify the macros to change the return type, it's a breaking change. If anyone anywhere expects the result of `set` to be `String`, their code will be broken. If we add a new set of macros, then we'll need to guide users when to use one set or the other, not to mention adding verbosity in the case when a user imports both macros, and needs to explain to the compiler which `set` did they intend. 
And above that, we'd need to actually go through the code and understand which commands should be changed. There are quite a few set variants.

jaymell:
> If we modify the macros to change the return type, it's a breaking change. If anyone anywhere expects the result of set to be String, their code will be broken. If we add a new set of macros, then we'll need to guide users when to use one set or the other, not to mention adding verbosity in the case when a user imports both macros, and needs to explain to the compiler which set did they intend.
> And above that, we'd need to actually go through the code and understand which commands should be changed. There are quite a few set variants.

Yep, there are a lot of potential issues here, but I think there could be other solutions. The high-level command API in this library is already confusing to many people, and I feel like having to specify even more generics is going to make the problem worse.

nihohit:
@jaymell I added `exec/_async` (didn't deprecate `execute` yet - I'm undecided about that, since we have enough breaking changes ATM).
As for the macros system - I'm not sure what/how/where to do it in a way that will minimize the damage.

nihohit:
BTW, I think that we need nightly to generate docs, so we should fix this before the next release 🤦 .

jaymell:
> @jaymell I added exec/_async (didn't deprecate execute yet - I'm undecided about that, since we have enough breaking changes ATM).

I'll defer to you, but I'm totally in favor of deprecating here. What's one more breaking change? :shrug: More seriously, though, the `unwrap` in there is just bad from my perspective. I'd have advocated for this much earlier if I realized the command existed :rofl: :sob:

redis/src/pipeline.rs
```diff
@@ -193,15 +193,26 @@ impl Pipeline {
...
     /// NOTE: A Pipeline object may be reused after `query()` with all the commands as were inserted
     ///       to them. In order to clear a Pipeline object with minimal memory released/allocated,
     ///       it is necessary to call the `clear()` before inserting new commands.
     #[inline]
     pub fn execute(&self, con: &mut dyn ConnectionLike) {
-        self.query::<()>(con).unwrap();
+        self.exec(con).unwrap();
+    }
+
+    /// This is a shortcut to `query`, to avoid having to define generic bounds for `()`.
```
jaymell, line 207: I wonder if we can word this a bit more simply/descriptively, something like:

> This is an alternative to `query` that can be used if you want to be able to handle a command's success or failure but don't care about the command's response. For example, this is useful for "write" commands for which the response other is not important. It avoids the need to define generic bounds for `()`.

redis/src/pipeline.rs
```diff
@@ -193,15 +193,26 @@ impl Pipeline {
...
+    pub fn exec(&self, con: &mut dyn ConnectionLike) -> RedisResult<()> {
+        self.query::<()>(con)
+    }
+
+    /// This is a shortcut to `query_async`, to avoid having to define generic bounds for `()`.
```
jaymell, line 212: ... and same here of course.

jaymell:
>  As for the macros system - I'm not sure what/how/where to do it in a way that will minimize the damage.

I'm not sure either. I'm even imagining something like a set of commands with `_no_return` appended to the name, but I don't think that's much of an improvement. I think what you've got here is about as good as we're going to get for now.

jaymell (approved):

nihohit:
@ckcr4lyf you're right. fixed that.

Issue #1201, opened 2024-05-31
add ability to specify count for scan match


alekspickle:
Right now the only thing scan match accepts is a pattern, and while for simplicity it's great, it would be nice to be able to add count to force scan to return more results.
```

127.0.0.1:6379> set kek1 nope
OK
127.0.0.1:6379> set kek2 nope
OK
127.0.0.1:6379> set kek3 nope
OK
127.0.0.1:6379> set kek4 nope
OK
127.0.0.1:6379> scan 0 match kek*
1) "42"
2) 1) "kek3"
127.0.0.1:6379> scan 0 match kek* count 100
1) "0"
2) 1) "kek3"
   2) "kek4"
   3) "kek1"
   4) "kek2"
127.0.0.1:6379>

```

I can sketch a PR but would be glad to hear an opinion on what can it look like
I guess the most obvious path is to add new `scan_match_n`

nihohit:
if we add a function for each option, we'll get 7 different functions - scan, scan_match, scan_match_count, scan_match_count_type, scan_match_type, scan_count_type, scan_type.
IMO it's better to just use `scan_options`, like `set_options`, and use an options struct that defines the requested behavior.

nihohit:
closed, since PR was merged.

PR #1231, opened 2024-06-20
Add scan_options

alekspickle:
Solves #1201 

Add new scan_options that receives cursor, pattern and count as arguments.

Also added deprecation warning for scan_match because in my eyes it is unusable without count anyway, but if you think we should leave it, I'll remove it.

### DIFF
--- a/redis/src/commands/macros.rs
+++ b/redis/src/commands/macros.rs
@@ -54,7 +54,16 @@ macro_rules! implement_commands {
                 c.iter(self)
             }
 
+            /// Incrementally iterate the keys space with options.
+            #[inline]
+            fn scan_options<RV: FromRedisValue>(&mut self, opts: ScanOptions) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("SCAN");
+                c.cursor_arg(opts.cursor).arg(opts);
+                c.iter(self)
+            }
+
             /// Incrementally iterate the keys space for keys matching a pattern.
+            #[deprecated(since="0.26.0", note="please use `scan_options` instead")]
             #[inline]
             fn scan_match<P: ToRedisArgs, RV: FromRedisValue>(&mut self, pattern: P) -> RedisResult<Iter<'_, RV>> {
                 let mut c = cmd("SCAN");
@@ -178,7 +187,16 @@ macro_rules! implement_commands {
                 Box::pin(async move { c.iter_async(self).await })
             }
 
+            /// Incrementally iterate the keys space with options.
+            #[inline]
+            fn scan_options<RV: FromRedisValue>(&mut self, opts: ScanOptions) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
+                let mut c = cmd("SCAN");
+                c.cursor_arg(opts.cursor).arg(opts);
+                Box::pin(async move { c.iter_async(self).await })
+            }
+
             /// Incrementally iterate set elements for elements matching a pattern.
+            #[deprecated(since="0.26.0", note="please use `scan_options` instead")]
             #[inline]
             fn scan_match<P: ToRedisArgs, RV: FromRedisValue>(&mut self, pattern: P) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
                 let mut c = cmd("SCAN");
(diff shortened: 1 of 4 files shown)


redis/src/commands/macros.rs
```diff
@@ -54,7 +54,16 @@ macro_rules! implement_commands {
                 c.iter(self)
             }
 
+            /// Incrementally iterate the keys space with options.
+            #[inline]
+            fn scan_options<RV: FromRedisValue>(&mut self, opts: ScanOptions) -> RedisResult<Iter<'_, RV>> {
```
jaymell, line 59: Hi, thanks for the PR! 

I think the cursor parameter should probably not be part of options, since it's required. In other commands with the same pattern, the options struct is for optional arguments.

alekspickle: hmm, ok, I can hardcode it to 0, I am not that profound redis user and I just thought that there are use cases when one would want to iterate using the cursor returned from redis

nihohit: The `Cmd::iter/_async` functions return an iterator that updates the cursor internally, so there's no need to set it externally.

redis/src/commands/mod.rs
```diff
@@ -2166,6 +2166,73 @@ impl PubSubCommands for Connection {
...
+#[derive(Default)]
+pub struct ScanOptions {
+    cursor: u64,
+    pattern: Option<String>,
+    count: Option<usize>,
+}
+
+impl ScanOptions {
+    /// Limit the results to the first N matching items.
+    pub fn count(mut self, n: usize) -> Self {
```
nihohit, line 2197: please use `set_X` instead of `X` for the options function names. I know that we're not consistent on this, I try to improve that in the library.

redis/tests/test_async.rs
```diff
@@ -638,6 +638,23 @@ mod basic_async {
         );
     }
 
+    #[tokio::test]
+    async fn test_scan_with_options_works() {
+        let ctx = TestContext::new();
+        let mut con = ctx.multiplexed_async_connection().await.unwrap();
+        for i in 0..20usize {
+            let _: () = con.append(format!("test/{i}"), i).await.unwrap();
```
nihohit, line 646: please also add values that don't have `test/` prefix, in order to verify that only `test/` values are returned, and then also test that without setting `pattern` they are returned.

redis/tests/test_async.rs
```diff
@@ -638,6 +638,23 @@ mod basic_async {
         );
     }
 
+    #[tokio::test]
+    async fn test_scan_with_options_works() {
```
nihohit, line 642: please also add a sync test

nihohit (commented):
thanks, looking good!

redis/src/commands/mod.rs
alekspickle: how about `with_*`? its a general convention in rust libs for builder pattern: `Options::default().with_field(a)`

nihohit: sure

redis/tests/test_async.rs
```diff
@@ -644,15 +644,20 @@ mod basic_async {
         let mut con = ctx.multiplexed_async_connection().await.unwrap();
         for i in 0..20usize {
             let _: () = con.append(format!("test/{i}"), i).await.unwrap();
+            let _: () = con.append(format!("other/{i}"), i).await.unwrap();
         }
-        let opts = ScanOptions::default().count(20).pattern("test/*");
+        let opts = ScanOptions::default().set_count(20).set_pattern("test/*");
         let values = con.scan_options::<String>(opts).await.unwrap();
         let values: Vec<_> = timeout(Duration::from_millis(100), values.collect())
```
nihohit, line 651: just to make the tests clear - add here the doc "scan with pattern", and below "scan without pattern", just so it'll be easy to parse what's the difference between the 2 queries.

redis/tests/test_basic.rs
```diff
@@ -414,6 +414,24 @@ mod basic {
...
 
+    #[test]
+    fn test_scan_with_options_works() {
+        let ctx = TestContext::new();
+        let mut con = ctx.connection();
+        for i in 0..20usize {
+            let _: () = con.append(format!("test/{i}"), i).unwrap();
+            let _: () = con.append(format!("other/{i}"), i).unwrap();
+        }
+        let opts = ScanOptions::default().set_count(20).set_pattern("test/*");
```
nihohit, line 425: same

nihohit (approved):
Thanks for the contribution!

PR #1232, opened 2024-06-21
Add support of HASH expiration commands

git-hulk:
Redis 7.4 will support the HASH expired, for commands can refer to: https://redis.io/docs/latest/commands/?group=hash

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -2,8 +2,8 @@ use crate::cmd::{cmd, Cmd, Iter};
 use crate::connection::{Connection, ConnectionLike, Msg};
 use crate::pipeline::Pipeline;
 use crate::types::{
-    ExistenceCheck, Expiry, FromRedisValue, NumericBehavior, RedisResult, RedisWrite, SetExpiry,
-    ToRedisArgs,
+    ExistenceCheck, ExpireOption, Expiry, FromRedisValue, NumericBehavior, RedisResult, RedisWrite,
+    SetExpiry, ToRedisArgs,
 };
 
 #[macro_use]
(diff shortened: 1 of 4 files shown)


redis/src/commands/mod.rs
```diff
@@ -400,6 +404,51 @@ implement_commands! {
         cmd("HEXISTS").arg(key).arg(field)
     }
 
+    /// Get one or more fields TTL in seconds.
+    fn httl<K: ToRedisArgs, F: ToRedisArgs>(key: K, fields: &'a [F]) {
```
nihohit, line 408: I believe that here, and below, `[fields](fields: &'a [F])` can be replaced by `field: F`, since slices implement ToRedisArgs.

git-hulk: @nihohit Thanks for your reply. I use slices here since those commands need to know how many fields are sent. So want to know how I can get the length using the `ToRedisArgs`? 

I'm a very newbie to Rust, so forgive me if asked silly questions.

nihohit: you're right, I missed that.

@jaymell do you think we should add to `ToRedisArgs` a length function, instead of `is_single_arg`?

git-hulk: @jaymell How do you think about this proposal? I will change it if it sounds good.

git-hulk: @jaymell What nihohit means should be to remove `is_single_arg` and replace it with the `len` function.

jaymell: Ok, I guess we could do that, but we should probably put that in a separate PR and get that in first.

git-hulk: ok, thanks.

nihohit (approved):

nihohit:
Thanks for the contribution!

PR #1235, opened 2024-06-24
Add support of EXPIRETIME/PEXPIRETIME command

git-hulk:
See https://redis.io/docs/latest/commands/expiretime/


redis/tests/test_basic.rs
```diff
@@ -1494,6 +1494,21 @@ mod basic {
         assert_args!(&opts, "EX", "1000");
     }
 
+    #[test]
```
jaymell, line 1497: I think the test would be stronger if you used `SET` w/ `EXAT`/`PXAT` and then check to make sure the times match with the new commands.

git-hulk: @jaymell Thanks for your review, will change it. And the CI failure is that Redis 7.2 doesn't support those commands yet, will check the version before running tests.

jaymell (approved):
LGTM -- thanks!

PR #1250, opened 2024-07-01
Use pipelines to setup connections

nihohit:
This change speeds up the setup phase of each new connection, by pipelining all of the commands, instead of sending them one after the other.

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -920,71 +920,195 @@ pub(crate) fn create_rustls_config(
     }
 }
 
-fn connect_auth(
-    con: &mut Connection,
+fn authenticate_cmd(
     connection_info: &RedisConnectionInfo,
+    check_username: bool,
     password: &str,
-) -> RedisResult<()> {
+) -> Cmd {
     let mut command = cmd("AUTH");
-    if let Some(username) = &connection_info.username {
-        command.arg(username);
-    }
-    let err = match command.arg(password).query::<Value>(con) {
-        Ok(Value::Okay) => return Ok(()),
-        Ok(_) => {
-            fail!((
-                ErrorKind::ResponseError,
-                "Redis server refused to authenticate, returns Ok() != Value::Okay"
-            ));
+    if check_username {
+        if let Some(username) = &connection_info.username {
+            command.arg(username);
         }
-        Err(e) => e,
-    };
-    let err_msg = err.detail().ok_or((
-        ErrorKind::AuthenticationFailed,
-        "Password authentication failed",
-    ))?;
-    if !err_msg.contains("wrong number of arguments for 'auth' command") {
-        fail!((
-            ErrorKind::AuthenticationFailed,
-            "Password authentication failed",
-        ));
     }
-
-    // fallback to AUTH version <= 5
-    let mut command = cmd("AUTH");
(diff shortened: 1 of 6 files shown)


jaymell:
Good job of minimizing code diffs between sync and async, even if the code is a bit convoluted for it.

redis/src/connection.rs
```diff
@@ -918,71 +918,195 @@ pub(crate) fn create_rustls_config(
...
 ) -> RedisResult<Connection> {
-    let con = ActualConnection::new(&connection_info.addr, timeout)?;
+    let con: ActualConnection = ActualConnection::new(&connection_info.addr, timeout)?;
     setup_connection(con, &connection_info.redis)
 }
 
-#[cfg(not(feature = "disable-client-setinfo"))]
-pub(crate) fn client_set_info_pipeline() -> Pipeline {
-    let mut pipeline = crate::pipe();
+pub(crate) struct ConnectionSetupComponents {
```
jaymell, line 944: I get the point of the struct, but I wonder if we can somehow live w/out it? Maybe at least shorten the field names to make it less noticeable :joy:

jaymell (approved):

redis/src/connection.rs
nihohit: I like the explicitly long names :(
Maybe 
```rs
enum CmdIndex {
  Resp2Auth,
  Resp3Auth,
  SelectDb
}

type  ConnectionSetupComponents = HashMap<CmdIndex, usize>;
```
?

jaymell: I would just make the fields shorter.
```
pub(crate) struct ConnectionSetupComponents {
    resp3_auth_cmd_idx: Option<usize>,
    resp2_auth_cmd_idx: Option<usize>,
    select_cmd_idx: Option<usize>,
}
```

redis/src/connection.rs
```diff
@@ -918,71 +918,195 @@ pub(crate) fn create_rustls_config(
...
+                    "Redis server refused to switch database",
+                )
+                    .into()),
+            };
+        }
+    }
+    Ok(())
+}
+
+pub(crate) fn check_connection_setup(
```
jaymell, line 1088: Why not have this function destructure the `ConnectionSetupComponents` itself and selectively call the `check...` methods? It would make the `check` methods less awkward if they were just evaluating the particular response when relevant.

Again, I think the functionality is fine, but some of these functions are just unnecessarily awkward and therefore harder to grok.

nihohit:
@jaymell tell me if you feel it's simpler this way.

redis/src/connection.rs
jaymell: Maybe just pass a value directly to `resp_2/3_auth`?. These functions stand a better chance of reuse if they're more generic. Something like this? 

```rust
fn check_resp3_auth(results: &Value) -> RedisResult<()> {
   match results {
        Value::ServerError(err) =>  Err(get_resp3_hello_command_error(err.clone().into())),
        _ => Ok(())
    }
}

// ...

    if let Some(resp3_response) = resp3_auth_cmd_idx.and_then(|idx| results.get(idx)) {
        check_resp3_auth(&resp_response)?
```

Or even just ditch check_resp2/3_auth entirely? Again, I know it's nitpicky. Just some thoughts.

Issue #1252, opened 2024-07-04
Connection Timeout leads to data corruption


Dav1dde:
A connection is not invalidated when the connection timeout is reached. This is especially problematic when combined with a connection pool.

Scenario:

- Get connection from pool
- Set connection timeout 
- Execute `GET key1`
- Request times out, connection is returned back to the pool
- Server produces response for `GET key1`
- Get the same connection from the pool
- Execute `GET key2`
- Client returns value from `GET key1` request

Currently the connection is only invalidated on EOF: https://github.com/redis-rs/redis-rs/blob/255f7c2eb5f4cad6a295f8d5b1f9a9f1aafb2fe1/redis/src/connection.rs#L1244-L1274

I am not quite sure in which circumstances the connection can be kept open, but to be safe I think the best option is to always close the connection on an `IoError`.

Simple example (enable `disable-client-setinfo` feature):

```rs
use std::{
    io::{BufRead, BufReader, Write},
    net::TcpListener,
    thread,
    time::Duration,
};

use anyhow::Result;
use redis::{Commands, ConnectionLike, Value};

fn read_command(mut reader: impl BufRead) -> Result<(String, String)> {
    let mut discard = String::new();
    reader.read_line(&mut discard)?;
    reader.read_line(&mut discard)?;

    let mut command = String::new();
    reader.read_line(&mut command)?;
    command.truncate(command.len() - 2);

    reader.read_line(&mut discard)?;

    let mut arg = String::new();
    reader.read_line(&mut arg)?;
    arg.truncate(arg.len() - 2);

    Ok((command, arg))
}

fn fake_redis() -> Result<()> {
    let listener = TcpListener::bind("127.0.0.1:7777")?;

    for stream in listener.incoming() {
        let mut stream = stream?;
        println!("Got connection from {:?}", stream.peer_addr()?);

        let mut reader = BufReader::new(stream.try_clone()?);
        let (command, arg) = read_command(&mut reader)?;

        match (&*command, &*arg) {
            ("GET", "key1") => {
                thread::sleep(Duration::from_secs(1));
                stream.write_all(b"$4\r\nkey1\r\n")?;
            }
            ("GET", "key2") => {
                stream.write_all(b"$4\r\nkey2\r\n")?;
            }
            _ => println!("Invalid command {command} {arg}"),
        }
    }

    Ok(())
}

fn main() -> Result<()> {
    std::thread::spawn(fake_redis);

[...]

PR #1253, opened 2024-07-04
Close connection on read timeout

Dav1dde:
Potential fix for #1252.

This only deals with the timeout case, not other IO errors. For clustered clients, a timeout does not cause a retry, the user probably wanted to limit the time it takes until the request returns.

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -1242,9 +1242,10 @@ impl Connection {
         };
         // shutdown connection on protocol error
         if let Err(e) = &result {
-            let shutdown = match e.as_io_error() {
-                Some(e) => e.kind() == io::ErrorKind::UnexpectedEof,
-                None => false,
+            let shutdown = match e.as_io_error().map(|e| e.kind()) {
+                Some(io::ErrorKind::UnexpectedEof) => true,
+                Some(io::ErrorKind::WouldBlock) => true,
+                _ => false,
             };
             if shutdown {
                 // Notify the PushManager that the connection was lost
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -1242,11 +1242,14 @@ impl Connection {
         };
         // shutdown connection on protocol error
         if let Err(e) = &result {
-            let shutdown = match e.as_io_error().map(|e| e.kind()) {
-                Some(io::ErrorKind::UnexpectedEof) => true,
-                Some(io::ErrorKind::WouldBlock) => true,
-                Some(io::ErrorKind::TimedOut) => true,
-                _ => false,
+            let shutdown = match e.as_io_error() {
+                Some(e) => matches!(
+                    e.kind(),
+                    io::ErrorKind::UnexpectedEof
+                        | io::ErrorKind::WouldBlock
+                        | io::ErrorKind::TimedOut
+                ),
+                None => false,
             };
             if shutdown {
                 // Notify the PushManager that the connection was lost
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -1245,6 +1245,7 @@ impl Connection {
             let shutdown = match e.as_io_error().map(|e| e.kind()) {
                 Some(io::ErrorKind::UnexpectedEof) => true,
                 Some(io::ErrorKind::WouldBlock) => true,
+                Some(io::ErrorKind::TimedOut) => true,
                 _ => false,
             };
             if shutdown {
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1029,6 +1029,7 @@ impl RedisError {
                     io::ErrorKind::BrokenPipe => RetryMethod::Reconnect,
                     io::ErrorKind::UnexpectedEof => RetryMethod::Reconnect,
 
+                    io::ErrorKind::WouldBlock => RetryMethod::NoRetry,
                     io::ErrorKind::PermissionDenied => RetryMethod::NoRetry,
                     io::ErrorKind::Unsupported => RetryMethod::NoRetry,
 
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1029,6 +1029,7 @@ impl RedisError {
                     io::ErrorKind::BrokenPipe => RetryMethod::Reconnect,
                     io::ErrorKind::UnexpectedEof => RetryMethod::Reconnect,
 
+                    io::ErrorKind::TimedOut => RetryMethod::NoRetry,
                     io::ErrorKind::WouldBlock => RetryMethod::NoRetry,
                     io::ErrorKind::PermissionDenied => RetryMethod::NoRetry,
                     io::ErrorKind::Unsupported => RetryMethod::NoRetry,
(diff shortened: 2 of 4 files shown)


nihohit:
why fail on "would block"? It just means that the socket is currently busy. I'm also not sure about timeouts - they're perfectly safe in a connection that maintains its internal state.

I'm not convinced that there isn't a simpler solution in this situation. For example, maybe we can just maintain an internal counter of the requests and replies, and drop answers that don't have matching requests.

Dav1dde:
Fail on `WOULDBLOCK` because that's the behaviour the user intended by setting the timeout.

> I'm not convinced that there isn't a simpler solution in this situation. For example, maybe we can just maintain an internal counter of the requests and replies, and drop answers that don't have matching requests.

That's the alternative the connection internally has a buffer and tracks requests, it needs the buffer because the read can fail in the middle of a response (unlikely but possible). Extra work for something the user isn't interested in anymore, but avoids the reconnect.

> but it imparis the async connections.

In which way, the code only touches the sync parts?

nihohit:
> Fail on WOULDBLOCK because that's the behaviour the user intended by setting the timeout.

Are you certain that WOULDBLOCK only happens by user settings? I think it can also happen if the OS can't push more bytes to the socket's buffer.

> In which way, the code only touches the sync parts?

you're touching types.rs, which affect all connections.

> That's the alternative the connection internally tracks requests/responses and bytes read, it needs to track bytes because the read can fail in the middle of a response (unlikely but possible). Extra work for something the user isn't interested in anymore, but avoids the reconnect.

You're right, if the timeout can happen mid-message it requires more work.

So, we need to find the simplest solution that doesn't affect async connections.

Dav1dde:
> Are you certain that WOULDBLOCK only happens by user settings? 

I went by the `recv` man page and the [`TcpStream` docs](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_read_timeout).

> Platforms may return a different error code whenever a read times out as a result of setting this option. For example Unix typically returns an error of the kind [WouldBlock](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.WouldBlock), but Windows may return [TimedOut](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.TimedOut).

But assuming it can happen for other reasons, at least in the sync code path, it would lead to the same problem.

> I think it can also happen if the OS can't push more bytes to the socket's buffer.

Since we're only reading in this codepath, that should be fine.

> you're touching types.rs, which affect all connections.

Right, I missed that, thanks!

Dav1dde:
I was looking through the code trying to find where best to make the distinction between async and sync. From what it looks like, the async code uses timeouts on futures, not directly through the socket (which makes sense, the socket is non blocking here anyways). So the change should have no effect on the async codepath.

I haven't really found a good spot how to deal with this though, maybe you have some pointers for me?

nihohit:
> So the change should have no effect on the async codepath.

Sadly, that's not true - you can get "internal" timeouts in the async cluster connection, or the connection manager. Any change to general error handling is a change that affects those types, too.

Why isn't the change to the connection sufficient?

Dav1dde:
> You're the one experiencing the issue, and frankly, I don't have production experience with the sync connection, your guess is as good as mine :). Why isn't the change to the connection sufficient?

You mean why the cluster impl also needs a change? The cluster impl shouldn't retry a timeout, the user set the timeout for a reason and that should be honored. Also some operations shouldn't be retried at all, like `INCR`, retrying it will just double count, this is even worse for Redis scripts.

> Not sure what you mean here. Are you claiming that any WOULDBLOCK error means that a connection is in a broken state? Because I don't thisnk that's true for async connections.

I don't think you'd ever get a `WOULDBLOCK` from the async connection, tokio doesn't have a way of setting a read timeout on the stream/socket in the first place, you'd always use a `tokio::time::Timeout`. The socket itself is will be set to nonblocking and the runtime/event loop should deal with this.

nihohit:
> Also some operations shouldn't be retried at all, like INCR, retrying it will just double count, this is even worse for Redis scripts.

Yes, this is an ongoing source of debate - the operations should be retried if they failed *before* reaching the server, but not after. ATM we don't have a way to distinguish that.

> The cluster impl shouldn't retry a timeout, the user set the timeout for a reason and that should be honored.

fair enough. @jaymell, thoughts?

> I don't think you'd ever get a WOULDBLOCK from the async connection, tokio doesn't have a way of setting a read timeout on the stream/socket in the first place, you'd always use a tokio::time::Timeout. The socket itself is will be set to nonblocking and the runtime/event loop should deal with this.

Tokio isn't the only supported runtime, and as mentioned before, WOULDBLOCK can be sent by the OS for nonblocking sockets, regardless of timeouts - I'm not sure how it would be filtered by the runtime. I'd want concrete data on this before changing the behavior.

Dav1dde:
Running this PR in production solved all the inconsistency errors we saw, but there are still very very few `EAGAIN` errors, I am wondering if that is also caused by the timeout, according to `man recv` this would make sense: ` Enables nonblocking operation; if the operation would  block, the call fails with the error EAGAIN or EWOULDBLOCK. `

> Tokio isn't the only supported runtime, and as mentioned before, WOULDBLOCK can be sent by the OS for nonblocking sockets, regardless of timeouts - I'm not sure how it would be filtered by the runtime. I'd want concrete data on this before changing the behavior.

That's when the future becomes pending, but I am no expert in this and just going off general understanding. I dug a bit through tokio and you can see it explicitly being [handled](https://docs.rs/tokio/1.38.0/src/tokio/io/poll_evented.rs.html#200-202) (codepath `AsyncRead` leads to) and [documented](https://docs.rs/tokio/1.38.0/tokio/net/struct.TcpStream.html#method.poll_read_ready).

nihohit:
Hi, sorry that I was less available. I would like to examine a solution with less splash damage potential - how about https://github.com/redis-rs/redis-rs/pull/1290?

Dav1dde:
Hey, thanks for picking this up.

I was going to report back some findings and conclusions and never got around. Ee've been running this fork now for about a month in production and it solved all of the initial issues we observed.

I also had some more thoughts about this. Whenever `WouldBlock` or `TimeOut` is propagated to the level where this condition now matches, the connection is in an uncertain state, because it is simply not clear how much data was written or read from the connection before failure, the underlying layer has to correctly book keep this and if it does it shouldn't propagate the errors in the first place. Curious if you think that makes sense.

Additionally what I noticed was, we're still getting a (comparatively) very very tiny amount of `Resource temporarily unavailable (os error 11)` errors from the Redis client. This doesn't seem to lead to any (observed) data corruption like the initially reported issue though (maybe not related at all).

I'll check out your PR now!

nihohit:
> Additionally what I noticed was, we're still getting a (comparatively) very very tiny amount of Resource temporarily unavailable (os error 11) errors from the Redis client. This doesn't seem to lead to any (observed) data corruption like the initially reported issue though (maybe not related at all).

Well, this is what fails the new tests in my PR on CI (but not locally), so I'd like to understand what's the cause of this.

Dav1dde:
Superseded by #1290

PR #1265, opened 2024-07-19
Remove tokio dependency from non-aio build.

nihohit:
This is done by changing the sync Connection's push sender to the std sender instead of tokio sender.
The decision to use the tokio sender in the sync implementation came from the previous design, where all connections used the same PushManager, and so had to use the same channel type.

https://github.com/redis-rs/redis-rs/pull/898#discussion_r1297473079

based over https://github.com/redis-rs/redis-rs/pull/1251

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -14,7 +14,6 @@ use std::io;
 use std::ops::Deref;
 use std::str::{from_utf8, Utf8Error};
 use std::string::FromUtf8Error;
-use tokio::sync::mpsc;
 
 macro_rules! invalid_type_error {
     ($v:expr, $det:expr) => {{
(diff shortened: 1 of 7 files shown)


redis/src/types.rs
```diff
@@ -2562,4 +2561,7 @@ pub struct PushInfo {
     pub data: Vec<Value>,
 }
 
-pub(crate) type PushSender = mpsc::UnboundedSender<PushInfo>;
+#[cfg(feature = "aio")]
+pub(crate) type AsyncPushSender = tokio::sync::mpsc::UnboundedSender<PushInfo>;
+
+pub(crate) type SyncPushSender = std::sync::mpsc::SyncSender<PushInfo>;
```
jaymell, line 2567: Why have you opted for `SyncSender` instead of `Sender`?

nihohit: Because I misread the documentation, and thought that the fact that the channel is asynchronous means it uses `async` 🤦

jaymell (approved):
:sunglasses:

redis/src/types.rs
jaymell: Random, but I've often thought about making a similar change [here](https://github.com/redis-rs/redis-rs/blob/main/redis/src/aio/multiplexed_connection.rs#L343)

nihohit: yup, I'm not sure why the original writer used a buffered channel there, or what would be the ramifications of changing it.

PR #1270, opened 2024-07-28
update MultiplexedConnection

zh-jq:
see each commits for the changes

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -4,7 +4,9 @@ use crate::cmd::Cmd;
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use crate::parser::ValueCodec;
 use crate::types::{AsyncPushSender, RedisError, RedisFuture, RedisResult, Value};
-use crate::{cmd, AsyncConnectionConfig, ConnectionInfo, ProtocolVersion, PushInfo, ToRedisArgs};
+use crate::{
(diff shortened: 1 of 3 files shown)


nihohit:
for what are the 2nd & 3rd commits needed?
The canonical way to create a connection is using `Client::get_multiplexed_async_connection_with_config`, why another public function is needed?
And why do you need to replace the connection type?

zh-jq:
> for what are the 2nd & 3rd commits needed? The canonical way to create a connection is using `Client::get_multiplexed_async_connection_with_config`, why another public function is needed? And why do you need to replace the connection type?

The 2nd commit is that, only the `RedisConnectionInfo` is needed within `MultiplexedConnection`, as it's constructed from a `stream` which should consume the Address part.

I thought it was by design to be able to use custom connections (proxy, custom tls handeshake) by constructing `MultiplexedConnection` directly, and I need one to also support pub/sub, so I just changed the previous `pub(crate)` one to be `pub`.

nihohit:
There's work on providing a new Pub/Sub implementation - https://github.com/redis-rs/redis-rs/pull/1144 if you ahve any recommendations / requirements, you're welcome to add them to that PR.

Why do you need to create the multiplexed connection directly, instead of through `Client`?

zh-jq:
> Why do you need to create the multiplexed connection directly, instead of through `Client`?

I need a custom tcp/tls connect implementation to support HTTP/Socks5 Proxy, or to support more TLS backends or just more handshake params.

nihohit (approved):

nihohit:
That makes sense. Thanks for the contribution :)

Issue #1272, opened 2024-08-01
bug: Compilation error when compiling `sentinel` feature without `aio` feature, due to a configured-out struct being used.


EmilyMatt:
The following line seems to be to blame
https://github.com/redis-rs/redis-rs/blob/6c8fa824ceaeaac7eed39d05d5df253aa8e7baf8/redis/src/sentinel.rs#L115

PR #1274, opened 2024-08-01
CI: Build major features without aio.

nihohit:
This will catch some compilation issues. For example, #1272,
and will warn about fields that are only used in async features.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -238,7 +238,7 @@ where
             connections: RefCell::new(HashMap::new()),
             slots: RefCell::new(SlotMap::new(cluster_params.read_from_replicas)),
             auto_reconnect: RefCell::new(true),
-            read_timeout: RefCell::new(None),
+            read_timeout: RefCell::new(Some(cluster_params.response_timeout)),
             write_timeout: RefCell::new(None),
             initial_nodes: initial_nodes.to_vec(),
             cluster_params,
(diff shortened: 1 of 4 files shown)


redis/src/cluster.rs
```diff
@@ -232,7 +232,7 @@ where
             connections: RefCell::new(HashMap::new()),
             slots: RefCell::new(SlotMap::new(cluster_params.read_from_replicas)),
             auto_reconnect: RefCell::new(true),
-            read_timeout: RefCell::new(None),
+            read_timeout: RefCell::new(Some(cluster_params.response_timeout)),
```
jaymell, line 235: Is this needed for this PR? If not can we move to a separate commit?

nihohit: I'll split the commits. This was added because the split build raised warnings about unused fields.

nihohit: done. in the long term, we should probably use https://github.com/TheBevyFlock/flag-frenzy, https://crates.io/crates/cargo-hack, or https://github.com/frewsxcv/cargo-all-features to properly test this.

jaymell: Ah, found it :laughing: Should this be re-based on the other PR?

nihohit: actually, this PR can be closed in favor of the other PR, since it gives a more through check.

Issue #1288, opened 2024-08-09
When building a ClusterClient using urls, the url param protocol=resp3 has no effect.


dorentus:
```rust
let nodes = vec![
  "redis://127.0.0.1:7000?protocol=resp3",
];

let cluster_client = ClusterClientBuilder::new(nodes).build();
```

`cluster_client.cluster_params.protocol` is excepted to be `ProtocolVersion::RESP3` as specified in url params. However it's `ProtocolVersion::RESP2` here.

----

This is because `ClusterClientBuilder` intializes `builder_params` using defaults [here](https://github.com/redis-rs/redis-rs/blob/main/redis/src/cluster_client.rs#L134).

And then inside method `pub fn build(self) -> RedisResult<ClusterClient>`, neither the `builder_params.protocol` property is updated using `first_node.redis.protocol` as other properties do, nor the `redis.protocol` property of each node is checked to be the same among all initial nodes.

Instead, each node's `redis.protocol` is overwritten to `cluster_params.protocol` [here](https://github.com/redis-rs/redis-rs/blob/main/redis/src/cluster_client.rs#L215), which is, unless `use_protocol` been called, the default value `ProtocolVersion::RESP2` initialized in `new`.

nihohit:
indeed, and this is also inconsistent with the behavior for passwords &usernames.

PR #1289, opened 2024-08-09
Align verification of protocol & TLS during cluster creation.

nihohit:
Like username & password, Protocol & TLS will now be copied from the first node if not explicitly set, and then verified that the same value is set for all other initial nodes.

https://github.com/redis-rs/redis-rs/issues/1288

### DIFF
--- a/redis/src/cluster_client.rs
+++ b/redis/src/cluster_client.rs
@@ -213,7 +213,7 @@ impl ClusterClientBuilder {
             if protocol.is_some() && Some(node.redis.protocol) != protocol {
                 return Err(RedisError::from((
                     ErrorKind::InvalidClientConfig,
-                    "Cannot use different password among initial nodes.",
+                    "Cannot use different protocol among initial nodes.",
                 )));
             }
 
--- a/redis/src/cluster_client.rs
+++ b/redis/src/cluster_client.rs
@@ -30,7 +30,7 @@ struct BuilderParams {
     retries_configuration: RetryParams,
     connection_timeout: Option<Duration>,
     response_timeout: Option<Duration>,
-    protocol: ProtocolVersion,
+    protocol: Option<ProtocolVersion>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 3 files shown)


dorentus (approved):

redis/src/cluster_client.rs
```diff
@@ -212,12 +210,23 @@ impl ClusterClientBuilder {
                     "Cannot use different username among initial nodes.",
                 )));
             }
-            node.redis.protocol = cluster_params.protocol;
-            nodes.push(node);
+            if protocol.is_some() && Some(node.redis.protocol) != protocol {
+                return Err(RedisError::from((
+                    ErrorKind::InvalidClientConfig,
+                    "Cannot use different password among initial nodes.",
```
jaymell, line 216:
```suggestion
                    "Cannot use different protocol among initial nodes.",
```

redis/src/cluster_client.rs
```diff
@@ -177,23 +177,21 @@ impl ClusterClientBuilder {
...
+            cluster_params.protocol = Some(first_node.redis.protocol);
+            cluster_params.protocol
+        } else {
+            None
+        };
 
-        let mut nodes = Vec::with_capacity(initial_nodes.len());
-        for mut node in initial_nodes {
+        // verify that the initial nodes match the cluster client's configuration.
+        for node in initial_nodes.iter() {
```
jaymell, line 194:
```suggestion
        for node in &initial_nodes {
```

Just a nit -- looks cleaner to me. Ignore if you think otherwise.

redis/src/cluster_client.rs
```diff
@@ -177,23 +177,21 @@ impl ClusterClientBuilder {
...
+        let protocol = if cluster_params.protocol.is_none() {
+            cluster_params.protocol = Some(first_node.redis.protocol);
+            cluster_params.protocol
+        } else {
+            None
+        };
 
-        let mut nodes = Vec::with_capacity(initial_nodes.len());
-        for mut node in initial_nodes {
+        // verify that the initial nodes match the cluster client's configuration.
```
jaymell, line 193:
```suggestion
        // Verify that the initial nodes match the cluster client's configuration.
```

jaymell (approved):
Looks good!

PR #1293, opened 2024-08-29
MpxConnection: Allow pipelines to continue after error.

nihohit:
Non-transaction pipelines can still complete, even if one of the results is an error.

This might be used down the line in optimizing some workflows that current send multiple separate requests.


jaymell (approved):
:sunglasses:

Issue #331, opened 2020-06-05
Support client-side caching


Terkwood:
We would like to support [client-side caching](https://redis.io/topics/client-side-caching), made available in Redis 6.0.

This is a subset of the effort to support the new features exposed in Redis 6.0.  See #312.

Currently this ticket is a stub.  Please feel free to share technical details needed to satisfy this goal and we'll work them into the ticket description.

Issue #312, opened 2020-05-01
Plan for Redis 6 supports


weihanglo:
Redis just released 6.0.0 GA yesterday. There are many interesting features need client-side supports, such as 

- [x] #305 TLS
- [x] #329 RESP3 protocol
- [x] #330 Access Control Lists (ACL)
- [ ] #331 Client-side caching
- [x] #333 STRALGO LCS 

What is the plan and the roadmap to support Redis 6? I would love to do some works.

Edited: This issue would collect update information about Redis 6 major features in its best efforts.

seanpianka:
Does this crate work with 6.0.x, even though there is no client-side support for the listed features?

badboy:
> Does this crate work with 6.0.x, even though there is no client-side support for the listed features? Or do users of this crate have to ensure they're using 5.x?

No, the old protocol still works and thus you can use this crate against a newer Redis just fine. You won't get support for some newer features and we might lack nice wrappers for newer commands.

> What is the plan and the roadmap to support Redis 6? I would love to do some works.

I personally have no plan to write specific support for Redis 6 features, both because I have no requirement for it and also not enough time for redis-rs.
If someone wants to work on this it's appreciated (I can't guarantee quick review though, but maybe it's then time to expand maintainership of redis-rs)

Terkwood:
I plan to open individual tickets to track the issues mentioned that aren't TLS (that one looks close to merge). Please let me know if you have any objections to that. It will help break down the effort into smaller chunks of work.

Terkwood:
Please see #329, #330, and #331:  stubbed issues raised as sub-tasks for this overall effort.  We welcome comments on the design and implementation detail that will help advance those individual tasks.

Terkwood:
And with #333 (STRALGO LCS), we should have tickets representing all the new features in Redis 6.

Terkwood:
We found in #345 that the CI environment is running Redis 5.  This causes tests that use new features from Redis 6 to fail when run by Travis. 

For now I would recommend that we make all new commands specific to Redis 6 _opt-in_, and _not_ part of the default feature set.  This will let us continue to make quick progress on developing the new features.  We could then explore the feasibility of upgrading the CI environment from Redis 5 to Redis 6, separately. 

This is a bit of an ugly compromise!  But solving CI issues isn't something that we can expect any individual contributor to be very excited about, so I think this is the best way to maintain our forward momentum on Redis 6 feature development.

[...]

Issue #826, opened 2023-04-08
Planning Client Side Caching


altanozlu:
After https://github.com/redis-rs/redis-rs/pull/757 gets merged i'd like to create pr for client side caching.

- It needs LRU/LFU cache with custom ttl for each key, since every key may have different ttl and cache shouldn't store stale values.
- It should be behind a feature & connection option
- It should be opt in for each command (every command will send MULTI, CLIENT CACHING YES, PTTL <KEY>, CMD, EXEC)
- It should support sending multiple command with cache

example pipeline:
```rust
let (k1, k2) : (i32, i32) = redis::pipe()
    .cmd("SET").arg("key_1").arg(42).ignore()
    .cmd("SET").arg("key_2").arg(43).ignore()
    .cmd("GET").arg("key_1").cache() // this will enable opt in cache mode for this command
    .cmd("GET").arg("key_2").query(&mut con)?;
```
example single command:
```rust
let mut c = Cmd::new();
let cm = c.arg("GET").arg("k1").cache();
let num: i32 = cm.query(&mut con).unwrap();
assert_eq!(num,1);
```

Also we can have `con.cache_get("my_key")` with `implement_commands` macro

altanozlu:
@nihohit @jaymell I would like to get your opinion.

nihohit:
👍 
Definitely a must-have IMO.
the redis support for client-side caching:
https://redis.io/docs/manual/client-side-caching/

I would start with implementing only one type of client-side caching offered by Redis. From the addition of the `cache` command I assume that you're using the Opt-in caching model.

> It should be behind a feature & connection option

I would use the decorator design pattern - the cache is a wrapper around a connection, implementing the ConnectionLike trait by passing commands to the internal connection.

> It should be opt in for each command (every command will send MULTI, CLIENT CACHING YES, PTTL , CMD, EXEC)

I don't think that the addition of `MULTI` is required for every command. You just need to call `CLIENT CACHING YES` before the next command.

Open questions:
1. How would you implement this for RESP2 connections? IMO initially supporting RESP3 is reasonable.
2. is the `ignore` addition required? What does it mean?
3. How would you modify Cmd to handle the `cache` command?
4. What's the cache datastructure? How would it handle connection cloning for MultiplexedConnection, for example?
5. How would you know when to look in the cache? That is, how will you filter commands that can be short-circuited and those that can't?
6. How will you limit the amount of memory used by the cache? Just limiting the number of keys means that you might hold several large values.

[...]

PR #1296, opened 2024-08-31
Add client side caching support for MultiplexedConnection

altanozlu:
I've cleared more over #1089 and added changes under `cache-aio` feature.
I believe it's ready to review.

Benchmark results:
```
group                       with_cache                               without_cache
-----                       ----------                               -------------
0.1-100-10000/thread-1      1.00       3.1±0.17s 320.0 KElem/sec     1.06       3.2±0.14s 302.7 KElem/sec
0.1-100-10000/thread-16     1.20   888.4±64.88ms 1099.3 KElem/sec    1.00  740.5±114.53ms 1318.8 KElem/sec
0.1-100-10000/thread-32     1.08   898.6±56.34ms 1086.8 KElem/sec    1.00   834.8±18.18ms 1169.8 KElem/sec
0.1-100-10000/thread-4      1.07  844.0±103.16ms 1157.1 KElem/sec    1.00   786.9±26.11ms 1241.0 KElem/sec
0.5-1-100000/thread-1       1.00   334.1±20.54ms 292.3 KElem/sec     1.02   339.4±30.71ms 287.7 KElem/sec
0.5-1-100000/thread-16      1.00    104.1±6.89ms 938.1 KElem/sec     1.01    104.8±7.10ms 931.7 KElem/sec
0.5-1-100000/thread-32      1.04    112.4±4.39ms 868.7 KElem/sec     1.00    108.1±3.03ms 903.4 KElem/sec
0.5-1-100000/thread-4       1.00    111.6±5.54ms 875.0 KElem/sec     1.06   118.4±10.85ms 825.1 KElem/sec
0.5-100-10000/thread-1      1.00       2.0±0.10s 487.7 KElem/sec     1.62       3.2±0.22s 302.0 KElem/sec
0.5-100-10000/thread-16     1.00   796.1±94.15ms 1226.7 KElem/sec    1.01   802.1±84.05ms 1217.5 KElem/sec
0.5-100-10000/thread-32     1.00   839.1±64.25ms 1163.8 KElem/sec    1.02   854.3±38.28ms 1143.1 KElem/sec
0.5-100-10000/thread-4      1.01   892.7±34.85ms 1094.0 KElem/sec    1.00   886.9±26.67ms 1101.1 KElem/sec
0.5-5-100000/thread-1       1.00  1126.0±118.10ms 433.7 KElem/sec    1.46  1643.4±164.08ms 297.1 KElem/sec
0.5-5-100000/thread-16      1.16   571.1±17.98ms 854.9 KElem/sec     1.00   491.3±16.86ms 993.8 KElem/sec
0.5-5-100000/thread-32      1.14   558.6±34.31ms 874.1 KElem/sec     1.00   489.6±11.13ms 997.4 KElem/sec
0.5-5-100000/thread-4       1.23   594.2±33.80ms 821.8 KElem/sec     1.00   482.0±15.11ms 1013.0 KElem/sec
0.9-100-10000/thread-1      1.00   692.4±73.42ms 1410.4 KElem/sec    4.86       3.4±0.14s 289.9 KElem/sec
0.9-100-10000/thread-16     1.00   273.8±18.87ms  3.5 MElem/sec      3.04  831.5±101.61ms 1174.5 KElem/sec
0.9-100-10000/thread-32     1.00    279.5±6.85ms  3.4 MElem/sec      3.13   873.9±38.05ms 1117.5 KElem/sec
0.9-100-10000/thread-4      1.00    230.0±4.80ms  4.1 MElem/sec      4.08   938.2±51.10ms 1040.8 KElem/sec
0.99-100-10000/thread-1     1.00    274.7±2.60ms  3.5 MElem/sec      11.73      3.2±0.22s 303.1 KElem/sec
0.99-100-10000/thread-16    1.00     69.4±4.17ms 13.7 MElem/sec      11.47  796.5±99.59ms 1226.1 KElem/sec
0.99-100-10000/thread-32    1.00     63.7±1.30ms 15.0 MElem/sec      13.83  881.5±41.41ms 1107.9 KElem/sec
0.99-100-10000/thread-4     1.00    149.8±6.65ms  6.4 MElem/sec      6.15   921.3±66.03ms 1060.0 KElem/sec
```

Commands used for benchmark:
```shell
RUSTFLAGS="-C target-cpu=native" ENABLE_CLIENT_SIDE_CACHE=false PROTOCOL=RESP3 cargo bench --all-features --bench bench_cache -- --save-baseline without_cache
RUSTFLAGS="-C target-cpu=native" ENABLE_CLIENT_SIDE_CACHE=true PROTOCOL=RESP3 cargo bench --all-features --bench bench_cache -- --save-baseline with_cache
critcmp without_cache with_cache
```
Testbench: 
- Ryzen 9950X(16C 32T) with linux 6.11.8-300.fc41.x86_64
- Redis server v=7.9.225 sha=0a24bfec:0 malloc=jemalloc-5.3.0 bits=64 build=3995de5f5cc4075a
- benchmark connects to redis server via localhost

Benchmark explanation:
`0.99-100-10000/thread-4` means 
- for each key 99% of requests will be GET and 1% of requests will be SET
- there will be 100 commands for each key
- there will be 10000 different keys
- tokio thread will have only 4 worker thread

So it means there will be 1 million requests and 10 thousands of them will be SET others are GET.

As far as I see when workload is write heavy it will be around 10% slower than non cache variant, but when it's mostly/equally read , caching will improve the performance. Having 13x performance gain for 99% read workload is nice :) Of course this is not free, client side caching will use more cpu.

I'm open to better benchmark methods 

fixes https://github.com/redis-rs/redis-rs/issues/331  https://github.com/redis-rs/redis-rs/issues/312


redis/src/connection.rs
```diff
@@ -1014,6 +1024,31 @@ pub(crate) fn connection_setup_pipeline(
...
+    cache_config: crate::caching::CacheConfig,
+    pipeline: &mut Pipeline,
+) {
+    if connection_info.protocol == ProtocolVersion::RESP2 {
+        return;
+    }
+    match cache_config.mode {
+        crate::caching::CacheMode::None => {}
+        crate::caching::CacheMode::All => {
+            pipeline.cmd("CLIENT").arg("TRACKING").arg("ON").ignore();
```
nihohit, line 1039: I prefer that caching won't be silently ignored. If the user turns on caching with a client level <6, they should receive an informative error message.

redis/src/cmd.rs
```diff
@@ -601,6 +630,63 @@ impl Cmd {
...
+
+    /// It marks command as triable for client side caching.
+    /// Useful when `CacheMode` is set to `CacheMode::OptIn`
+    #[cfg(feature = "cache-aio")]
+    pub fn cache(&mut self) -> &mut Cmd {
+        self.cache = Some(CommandCacheOptions {
+            client_side_ttl: None,
+        });
+        self
+    }
```
nihohit, line 642: nit: add empty line between methods. same elsewhere

redis/src/cmd.rs
```diff
@@ -601,6 +630,63 @@ impl Cmd {
...
+    /// Useful when `CacheMode` is set to `CacheMode::OptIn`
+    #[cfg(feature = "cache-aio")]
+    pub fn cache_with_client_side_ttl(&mut self, client_side_ttl: Duration) -> &mut Cmd {
+        self.cache = Some(CommandCacheOptions {
+            client_side_ttl: Some(client_side_ttl),
+        });
+        self
+    }
+    #[cfg(feature = "cache-aio")]
+    #[inline]
```
nihohit, line 653: why inline this?

redis/src/cmd.rs
```diff
@@ -601,6 +630,63 @@ impl Cmd {
...
+                if let Some(key) = cache_key {
+                    return Some(CommandCacheInformationByRef {
+                        client_side_ttl: self.cache.as_ref().and_then(|x| x.client_side_ttl),
+                        redis_key: key,
+                        cmd: if is_mget { x } else { self.data.as_slice() },
+                        is_mget,
+                    });
+                }
+            } else {
+                return None;
```
nihohit, line 678: nit: I think these 2 else clauses can be omitted.

redis/src/cmd.rs
```diff
@@ -601,6 +630,63 @@ impl Cmd {
...
+            if let Some(command_name) = self.arg_idx(0) {
+                let cache_key = match command_name {
+                    b"GET" | b"HEXISTS" | b"HGET" | b"HMGET" | b"HGETALL" | b"PTTL" => {
+                        Some(self.arg_idx(1).unwrap())
+                    }
+                    b"MGET" => {
+                        is_mget = true;
+                        Some(x)
+                    }
+                    _ => return None,
```
nihohit, line 667: why other commands can't be cached?

redis/src/caching/statistics.rs
```diff
@@ -0,0 +1,49 @@
...
+pub struct CacheStatistics {
+    /// Count of how many cache hits occurred.
+    pub hit: usize,
+    /// Count of how many cache misses occurred.
+    pub miss: usize,
+    /// Count of how many cache item is invalidated via [`crate::types::PushKind::Invalidate`] or LRU.
+    pub invalidate: usize,
+    /// Count of commands sent when caching is enabled, only increased when key is cached.
+    pub sent_command_count: usize,
+}
```
nihohit, line 15: add empty line between scopes

redis/src/caching/statistics.rs
```diff
@@ -0,0 +1,49 @@
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+
+/// CacheStatistics holds statistics generated by Client Side Caching.
+#[derive(Debug)]
+pub struct CacheStatistics {
```
nihohit, line 6: Maybe `FinalizedStatistics`? I want a name that clarifies this was computed from `Statistics` and is its final form.

redis/src/caching/config.rs
```diff
@@ -0,0 +1,49 @@
...
+    None,
+    /// Try to cache every command.
+    All,
+    /// Only try to cache commands which explicitly called cache methods.
+    OptIn,
+}
+
+/// Holds configuration to be used by [`crate::caching::CacheManager`]
+#[derive(Clone, Debug, Copy)]
+pub struct CacheConfig {
```
nihohit, line 17: please use the builder pattern instead of exposing internal fields. That way we can refactor the class without breaking user code.

redis/src/caching/config.rs
```diff
@@ -0,0 +1,49 @@
+use std::num::NonZeroUsize;
+use std::time::Duration;
+
+/// Defines CacheMode of CacheConfig
+#[derive(Clone, PartialEq, Debug, Copy)]
+pub enum CacheMode {
+    /// Don't cache
+    None,
```
nihohit, line 8: why is there a `None` mode? wouldn't a disabled cache just be a `let cache_config: Option<CacheMode> = None;`?

redis/src/caching/config.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub mode: CacheMode,
+    /// `cache_size` defines size of [`crate::caching::cache_manager::LRUCacheShard`].
+    pub cache_size: NonZeroUsize,
+    /// `default_client_ttl` is the maximum amount for holding redis item in the cache.
+    pub default_client_ttl: Duration,
+}
+
+impl CacheConfig {
+    /// Enables cache with 10_000 LRU size and 30 minutes of client ttl
+    pub fn enabled() -> Self {
```
nihohit, line 28: this looks like a `Default` implementation. the name doesn't show that this is just a proposed default.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -88,14 +96,26 @@ impl Debug for Pipeline {
...
+#[cfg(feature = "cache-aio")]
 pin_project! {
-    struct PipelineSink<T> {
-        #[pin]
-        sink_stream: T,
-        in_flight: VecDeque<InFlight>,
-        error: Option<RedisError>,
-        push_sender: Option<AsyncPushSender>,
-    }
+struct PipelineSink<T> {
```
nihohit, line 101: please fix indentation.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -88,14 +96,26 @@ impl Debug for Pipeline {
...
+    #[pin]
+    sink_stream: T,
+    in_flight: VecDeque<InFlight>,
+    error: Option<RedisError>,
+    push_sender: Option<AsyncPushSender>,
+    cache_manager: CacheManager,
+}
+}
```
nihohit, line 109: add empty line

redis/src/aio/multiplexed_connection.rs
```diff
@@ -164,8 +190,19 @@ where
         let result = match result {
             // If this push message isn't a reply, we'll pass it as-is to the push manager and stop iterating
             Ok(Value::Push { kind, data }) if !kind.has_reply() => {
+                #[cfg(feature = "cache-aio")]
+                if kind == PushKind::Invalidate {
+                    if let Some(Value::Array(redis_key)) = data.first() {
```
nihohit, line 195: I assume that this logic will be repeated in other connection types. It probably shouldn't exist here.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -164,8 +190,19 @@ where
...
+                            if let Ok(redis_key) =
+                                crate::FromRedisValue::from_redis_value(redis_key)
+                            {
+                                self_.cache_manager.invalidate(&redis_key)
+                            }
+                        }
+                    }
+                }
                 send_push(self_.push_sender, PushInfo { kind, data });
-
```
nihohit, line 168: why did you remove all of these empty lines?

redis/src/aio/multiplexed_connection.rs
```diff
@@ -520,15 +580,180 @@ impl MultiplexedConnection {
         self.response_timeout = Some(timeout);
     }
 
+    #[cfg(feature = "cache-aio")]
+    async fn handle_cached_mget<'a>(
+        &mut self,
+        cmd: &Cmd,
+        ci: &CommandCacheInformationByRef<'a>,
+    ) -> RedisResult<Value> {
+        let mut response: Vec<Value> = vec![];
```
nihohit, line 589: please document what this function does. Its huge and unclear.

redis/src/cmd.rs
```diff
@@ -601,6 +630,63 @@ impl Cmd {
...
+                };
+                if let Some(key) = cache_key {
+                    return Some(CommandCacheInformationByRef {
+                        client_side_ttl: self.cache.as_ref().and_then(|x| x.client_side_ttl),
+                        redis_key: key,
+                        cmd: if is_mget { x } else { self.data.as_slice() },
+                        is_mget,
```
nihohit, line 674: mget isn't the only multi-key command. for example, there's mset.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -520,15 +580,180 @@ impl MultiplexedConnection {
...
+                if cmd.has_opt_in_cache() {
+                    cmd.compute_cache_information()
+                } else {
+                    None
+                }
+            }
+            CacheMode::None => None,
+        };
+        #[cfg(feature = "cache-aio")]
+        let (cmd_bytes, cmd_count, notifier) = if let Some(ci) = &cache_information {
```
nihohit, line 696: there's a lot of code here and above which isn't multiplexer specific, and seems to also be relevant for other connections - anything which involves creating new commands, for example. Please move that to the cache module.

nihohit (commented):
Great start!
I didn't go over all of the code yet, so these are just the initial comments.

redis/src/cmd.rs
altanozlu: My bad, it grew too much for this

redis/src/cmd.rs
altanozlu: but MSET is not something can be cached, it will trigger invalidation. There can be also JSON.MGET but I'm not sure if there are other ones for client side caching.

redis/src/cmd.rs
altanozlu: Only read commands *should* be cached, I can prepare extensive list but this is kinda MVP.

redis/src/caching/config.rs
altanozlu: Hmm I will refactor CacheConfig part in general, that might be better

redis/src/caching/statistics.rs
altanozlu: I chose `CacheStatistics` because `Statistics` isn't a public struct and users don't have to know about it. So `FinalizedStatistics` might bring more confusion.

redis/src/caching/cache_manager.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+struct ShardedLRU {
+    shards: Vec<std::sync::Mutex<LRUCacheShard>>,
+}
+impl ShardedLRU {
+    const MAX_SHARD_SIZE: usize = 32;
+    fn new(total_key_size: NonZeroUsize) -> Self {
+        // If total cache size is smaller than max shard size then it won't use sharding.
+        let (shard_size, per_lru_size) = if total_key_size.get() >= Self::MAX_SHARD_SIZE {
```
nihohit, line 37: is `shard_size` actually `shard_count`? otherwise I don't understand what's the difference between `shard_size` and `per_lru_size`.

redis/src/caching/cache_manager.rs
```diff
@@ -0,0 +1,226 @@
...
+use tokio::sync::watch::{channel, Receiver, Sender};
+
+type RedisCmd = Vec<u8>;
+type RedisKey = Vec<u8>;
+type LRUCacheShard = LruCache<RedisKey, CacheItem>;
+
+struct CacheCmdEntry {
+    cmd: RedisCmd,
+    value: Option<Value>,
+    receiver: Receiver<Value>,
```
nihohit, line 22: what is this? what purpose does it serve?

redis/src/caching/cache_manager.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+struct CacheItem {
+    ttl: Instant,
```
nihohit, line 26: nit: time to live is a duration. if this is an instant, it's `expiry_time`, or some similar name.

redis/src/caching/cache_manager.rs
```diff
@@ -0,0 +1,226 @@
...
+            cache_config,
+        }
+    }
+
+    pub(crate) async fn get_with_guard_new<'a>(
+        &self,
+        cache_information: &CommandCacheInformationByRef<'a>,
+    ) -> Result<Option<Value>, Sender<Value>> {
+        let mut receiver = {
+            let mut lru_cache_better = self
```
nihohit, line 81: what do you mean by better?

redis/src/caching/cache_manager.rs
```diff
@@ -0,0 +1,226 @@
...
+                .lru
+                .get_shard(cache_information.redis_key)
+                .lock()
+                .unwrap();
+            if let Some(ch) = lru_cache_better.get_mut(cache_information.redis_key) {
+                // We have found redis key in cache, but we need to make sure redis cmd is also here.
```
nihohit, line 87: `is also here` - what do you mean by "here"?

redis/src/caching/cache_manager.rs
```diff
@@ -0,0 +1,226 @@
...
+        &self,
+        cache_information: &CommandCacheInformationByRef<'a>,
+    ) -> Result<Option<Value>, Sender<Value>> {
+        let mut receiver = {
+            let mut lru_cache_better = self
+                .lru
+                .get_shard(cache_information.redis_key)
+                .lock()
+                .unwrap();
+            if let Some(ch) = lru_cache_better.get_mut(cache_information.redis_key) {
```
nihohit, line 86: this function is fairly large, mostly because of large conditional split. Can it be torn apart to smaller, clearly named functions?

redis/src/caching/cache_manager.rs
altanozlu: If multiple request is trying to access same key with same command, only one request will send request to server, other requests will wait this  `receiver`
```
if receiver.changed().await.is_ok() {
                self.statistics.increase_hit(1);
                return Ok(Some(receiver.borrow().clone()));
           }
```

redis/src/caching/cache_manager.rs
altanozlu: oops, will be removed (it's coming from an old patch)

redis/src/caching/cache_manager.rs
altanozlu: I've updated the documentation.

redis/src/caching/config.rs
altanozlu: I've stopped exposing internal field and added setter methods.

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    block_on_all(async move {
+        let mut con = ctx.multiplexed_async_connection_tokio_with_cache().await?;
+        let val: Option<String> = redis::cmd("GET")
+            .arg("key_1")
+            .query_async(&mut con)
+            .await
+            .unwrap();
+        assert_eq!(val, None);
+        {
```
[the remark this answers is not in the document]
altanozlu: It's a precaution I use when there is multiple rebinding to the same name, I can remove them.

redis/src/caching/cache_manager.rs
altanozlu: I've changed it and now it should be clear.

redis/src/caching/cache_manager.rs
altanozlu: I will work on this

redis/src/cmd.rs
altanozlu: Hmm NOLOOP option was never on my plans, I will create an issue about it.
there was an early issue about it when it was mostly draft (https://github.com/redis-rs/redis-rs/issues/826)

redis/src/caching/config.rs
altanozlu: With changes to cacheconfig default is now enabled.

redis/src/aio/multiplexed_connection.rs
altanozlu: I've documented and refactored it, I believe it's better now, can you check it again please ?

redis/src/connection.rs
altanozlu: Now it will raise an error

redis/src/caching/cache_manager.rs
altanozlu: I've split the function, it required new enums for handling the logic

redis/src/aio/multiplexed_connection.rs
altanozlu: I've moved some but not all. Is this enough ? I prefer not moving everything to cache module when potential overlap is unknown.

redis/src/cmd.rs
altanozlu: https://github.com/redis-rs/redis-rs/issues/1316

redis/src/cmd.rs
altanozlu: Also Added JSON.MGET to here but didn't implement it yet.

redis/tests/support/mod.rs
```diff
@@ -549,6 +550,39 @@ impl TestContext {
         self.client.get_multiplexed_tokio_connection().await
     }
 
+    #[cfg(all(feature = "tokio-comp", feature = "cache-aio"))]
+    pub fn multiplexed_async_connection_tokio_with_cache(
```
nihohit, line 554: why is this needed? I see that everywhere you use `multiplexed_async_connection_tokio_with_cache_config` it's with the default cache.

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,307 @@
...
+use redis::caching::CacheConfig;
+use redis::{AsyncCommands, RedisError, RedisFuture};
+use std::time::Duration;
+
+use crate::support::*;
+
+mod support;
+
+#[test]
+fn test_cache() {
```
nihohit, line 12: I'm thinking about implementing a mock multiplexed connection, would that make the statistics redundant?

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,307 @@
...
+        for i in 0..1000 {
+            let mut c = con.clone();
+            v.push(async move {
+                let x: RedisFuture<Option<u32>> = c.get("KEY");
+                tokio::time::timeout(Duration::from_micros(i), x).await
+            })
+        }
+        futures::future::join_all(v).await;
+        let s = con.get_cache_statistics();
+        assert_eq!(s.miss, 1000);
```
nihohit, line 140: why is this?
shouldn't some of the concurrent operations hit the cache, like in line 41?

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,307 @@
...
+        let con = ctx
+            .multiplexed_async_connection_tokio_with_cache_config(CacheConfig::default())
+            .await?;
+
+        let mut v = vec![];
+        for i in 0..1000 {
+            let mut c = con.clone();
+            v.push(async move {
+                let x: RedisFuture<Option<u32>> = c.get("KEY");
+                tokio::time::timeout(Duration::from_micros(i), x).await
```
nihohit, line 135: why is the timeout needed? just for safety?

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    block_on_all(async move {
+        let mut con = ctx
+            .multiplexed_async_connection_tokio_with_cache_config(CacheConfig::default())
+            .await?;
+        let _: () = con.set("KEY", "77").await?;
+        let mut v = vec![];
+        for _ in 0..5 {
+            let mut c = con.clone();
+            v.push(tokio::spawn(async move {
```
nihohit, line 163: why are the spawns needed?

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,307 @@
...
+        futures::future::join_all(v).await;
+        let s = con.get_cache_statistics();
+        assert_eq!(s.miss, 1000);
+        Ok::<_, RedisError>(())
+    })
+    .unwrap();
+}
+
+#[test]
+fn test_cache_guards_and_different_types() {
```
nihohit, line 147: test_cache_is_not_target_type_dependent?

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,307 @@
...
+            let mut c = con.clone();
+            v.push(tokio::spawn(async move {
+                let x: Option<u8> = c.get("KEY").await.unwrap();
+                assert_eq!(x, None);
+            }));
+        }
+        futures::future::join_all(v).await;
+        let s = con.get_cache_statistics();
+        // It should get value once and miss then it should use cache.
+        assert_eq!(s.hit, 14);
```
nihohit, line 225: why here the requests hit the cache and in `test_cache_drops_and_guards` they don't?

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,307 @@
...
+            let _: Option<u32> = con.get("KEY").await.unwrap();
+        }
+        join_handle.await.unwrap();
+        Ok::<_, RedisError>(())
+    })
+    .unwrap();
+}
+
+#[test]
+fn test_cache_mget_verify() {
```
nihohit, line 271: I'm having a hard time understanding the purpose of all of the tests. The names are very terse, and require knowledge of the internal implementation in order to understand what they're actually testing.

nihohit (commented):
Sorry that this is taking forever. I'm trying to review this in chunks when I have the time, but the last couple of weeks have been intensive, and the chunks are large :(

nihohit:
I briefly scanned through the multiplexed connection changes, and noticed these broad issues:
1. caching is only used when sending single commands, not pipelines.
2. there are very few usage of `self` that aren't `self.cache_manager` in the logic surrounding caching. That implies that most of the logic there could be moved to the cache manager, and tested without relying on the TCP connection.
3. Couldn't the interface between the connection and the cache manager be something like 
* pass command/pipeline to cache manager.
* cache manager returns an enum with one of these variants:
- `Value` for cmd / `Vec<Value>` for pipeline, if all values are cached
- `Cmd`/`Pipeline`, in case no value was cached. This can either be the passed command, or a new pipeline generated to enable caching.
- `(Vec<Value>, Cmd, Instructions)` / `<Vec<Value>, pipeline, Instructions)` in case some of the values are cached and some aren't. `Instructions` is an object explaining how to interleave the returned values.
* connection sends the received command, and updates the cache with its result.

This means that the cache manager needs to only expose 2 functions to the connection (not including the functions exposed to the pipeline) 
1. get_cached_value_or_command
2. set_command_result 
WDYT? am I missing something?

altanozlu:
> I briefly scanned through the multiplexed connection changes, and noticed these broad issues:
> 
>     1. caching is only used when sending single commands, not pipelines.
> 
>     2. there are very few usage of `self` that aren't `self.cache_manager` in the logic surrounding caching. That implies that most of the logic there could be moved to the cache manager, and tested without relying on the TCP connection.
> 
>     3. Couldn't the interface between the connection and the cache manager be something like
> 
> 
>     * pass command/pipeline to cache manager.
> 
>     * cache manager returns an enum with one of these variants:
> 
> 
>     * `Value` for cmd / `Vec<Value>` for pipeline, if all values are cached
> 
>     * `Cmd`/`Pipeline`, in case no value was cached. This can either be the passed command, or a new pipeline generated to enable caching.
> 
>     * `(Vec<Value>, Cmd, Instructions)` / `<Vec<Value>, pipeline, Instructions)` in case some of the values are cached and some aren't. `Instructions` is an object explaining how to interleave the returned values.
> 
> 
>     * connection sends the received command, and updates the cache with its result.
> 
> 
> This means that the cache manager needs to only expose 2 functions to the connection (not including the functions exposed to the pipeline)
> 
>     1. get_cached_value_or_command
> 
>     2. set_command_result
>        WDYT? am I missing something?

I had few try for managing everything (at least for MGET) in cache manager and just using mpx for executing commands, I've eventually scrapped them because return part became a complex enum and cachemanager became state manager. It felt like an overkill. If you are also okay with complex enum passed between mpx and cachemanager, I would like to work on it again.

About first point (single commands) I didn't want to deal with pipelines for first iteration, if it's you see something as a *must* have, I'd like to work on it, it may just take more time...

redis/src/caching/statistics.rs
```diff
@@ -0,0 +1,53 @@
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+
+/// CacheStatistics holds statistics generated by Client Side Caching.
+#[derive(Debug, Default)]
+pub struct CacheStatistics {
```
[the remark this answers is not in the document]
altanozlu: For sure it makes easier to test things, but my main goal for this to be used by users. As an user if I have enabled client side caching in my application, I have no idea how well it works rather than seeing latency decrease etc. But by using CacheStatistics I can generate metrics from it.

redis/tests/test_cache.rs
altanozlu: What is being tested here is 1000 future is trying to get same key, and one of them will be able to send the query and others will hold guard until that query is completed. Since the one tries to send query will timeout others will stop holding guard and query the server instead. 1 for first miss from first future sends the query 999 miss from futures awaited the guard and dropped it.

I guess this one requires more wording on what it does.

redis/tests/test_cache.rs
altanozlu: CacheManager stores `Value` and a `Value::SimpleString` can be converted into `String` or `i32`. This pattern also aligns with normal command usage.

redis/tests/test_cache.rs
altanozlu: For more reliability in tests, to make sure it doesn't rely in order of commands. Can be removed but I see no harm.

redis/tests/test_cache.rs
altanozlu: Answered below.

redis/tests/test_cache.rs
altanozlu: I will add more documentation and make them clearer.

redis/tests/test_cache.rs
nihohit: yes, sorry - that wasn't at all clear from the test :)

redis/tests/test_cache.rs
nihohit: this is fine - just add a comment explaining this, please.

redis/tests/test_cache.rs
nihohit: right, I suggest what I think is a clearer test name. sorry I wasnt' clear :)

nihohit:
> About first point (single commands) I didn't want to deal with pipelines for first iteration

That's a good question. If we didn't expose the statistics, then I'd say that as far as the users are concerned, it doesn't matter. But because we do expose statistics, I think that it would be confusing for the users if some requests changed statistics and some didn't. 
I definitely wouldn't want to ship a version without pipeline support, but maybe we could merge the initial PR and then either finish with pipeline support before releasing a version, or hide the caching API, release a version, and continue on pipelines.

>  If you are also okay with complex enum passed between mpx and cachemanager, I would like to work on it again.

generally I'm ok with complex enums, but do you have a measure of how complex the enum will need to be?

altanozlu:
> > About first point (single commands) I didn't want to deal with pipelines for first iteration
> 
> That's a good question. If we didn't expose the statistics, then I'd say that as far as the users are concerned, it doesn't matter. But because we do expose statistics, I think that it would be confusing for the users if some requests changed statistics and some didn't. I definitely wouldn't want to ship a version without pipeline support, but maybe we could merge the initial PR and then either finish with pipeline support before releasing a version, or hide the caching API, release a version, and continue on pipelines.
> 
> > If you are also okay with complex enum passed between mpx and cachemanager, I would like to work on it again.
> 
> generally I'm ok with complex enums, but do you have a measure of how complex the enum will need to be?

If pipeline support is important for the release then I can batch them up. I won't be able to work on this for next few weeks, I will publish revised patch in november.

nihohit:
Thanks, you're doing awesome work!
If you think that pipeline support is more of the same, than we can finish with the current design and then implement the pipeline on top of it. I think that you were right to limit the initial work to single commands, but as a feature, it isn't complete without pipelines IMO.

altanozlu:
Hey! I've rebased over main and changed few things:

- Splitted LRU related code from CacheManager into sharded_lru.rs
- Got rid of guard logic in order to reduce complexity, I want to add back again but not in this PR.
- Created CacheablePipeline and CacheableCommand to store states for requests.
- Unified cache and cmd creation logic for cmd/pipeline under caching/cache_manager.rs , caching/cmd.rs instead of using mpx directly.
- Updated tests for rstest


This is not final commit, I'm planning to add more tests and Optout support.
Until then 👋🏻

redis/tests/test_cache.rs
altanozlu: I want to test with real connection, because how client side caching works is really about how redis server responds. This way we might be able to see any issue comes up new valkey,redis versions.

altanozlu:
Updated benchmark results, I'm planning to add more benchmarks in upcoming PRs, @nihohit can I get a review ?

redis/src/aio/multiplexed_connection.rs
```diff
@@ -556,6 +646,16 @@ impl MultiplexedConnection {
             _ => Ok(vec![value]),
         }
     }
+
+    /// Gets [`crate::caching::CacheStatistics`] for current connection.
+    #[cfg(feature = "cache-aio")]
+    pub fn get_cache_statistics(&self) -> CacheStatistics {
```
nihohit, line 652: nit: this should probably be an `Option<CacheStatistics>`, so that `None` is returned if there's no cache manager. (and adjust docs)

redis/src/cmd.rs
```diff
@@ -601,6 +615,38 @@ impl Cmd {
     pub fn is_no_response(&self) -> bool {
         self.no_response
     }
+
+    /// It marks command as triable for client side caching.
+    /// Useful when `CacheMode` is set to `CacheMode::OptIn`
+    #[cfg(feature = "cache-aio")]
+    pub fn cache(&mut self) -> &mut Cmd {
```
nihohit, line 622: IMO this should take `client_side_ttl` and remove `cache_with_client_side_ttl`.
Do you think that future iterations might take more arguments? should we create some config struct for per-command caching, in order to maintain forward compatibility.

redis/src/pipeline.rs
```diff
@@ -350,3 +352,24 @@ macro_rules! implement_pipeline_commands {
...
+    pub fn cache(&mut self) -> &mut Self {
+        let cmd = self.get_last_command();
+        cmd.cache();
+        self
+    }
+
+    /// It marks command as triable for client side caching with custom Client Side TTL for cache.
+    /// Useful when `CacheMode` is set to `CacheMode::OptIn`
+    #[cfg(feature = "cache-aio")]
+    pub fn cache_with_client_side_ttl(&mut self, client_side_ttl: Duration) -> &mut Self {
```
nihohit, line 370: same as for `Cmd` re: not having 2 functions.

redis/src/caching/cache_manager.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+#[derive(Clone)]
+pub(crate) struct CacheManager {
+    lru: Arc<ShardedLRU>,
+    pub(crate) cache_config: CacheConfig,
+}
+
+impl CacheManager {
+    pub fn new(cache_config: CacheConfig) -> Self {
```
nihohit, line 26: nit: the `pub` functions here should be `pub(crate)`

redis/src/caching/config.rs
```diff
@@ -0,0 +1,60 @@
+use std::num::NonZeroUsize;
+use std::time::Duration;
+
+/// Defines CacheMode of CacheConfig
```
nihohit, line 4: Defines the behavior of the cache regarding which commands should be cached.

redis/src/caching/config.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Clone, PartialEq, Debug, Copy)]
+pub enum CacheMode {
+    /// Try to cache every command.
```
nihohit, line 7: Try to cache every command. This saves the user a call of [Cmd::cache].

redis/src/caching/config.rs
```diff
@@ -0,0 +1,60 @@
...
+    All,
+    /// Only try to cache commands which explicitly called cache methods.
+    /// If OptIn is used `cache` or `cache_with_client_side_ttl` functions
```
nihohit, line 10: use [Cmd::cache] to link to the function.

redis/src/caching/config.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// must be used on [`crate::Cmd`] for caching to work with  the command.
+    OptIn,
+}
+
+/// Options for managing client side caching
```
nihohit, line 15: Configuration for client side caching.

redis/src/caching/config.rs
```diff
@@ -0,0 +1,60 @@
...
+        self
+    }
+
+    /// Sets maximum key count for the cache.
+    pub fn set_size(mut self, size: NonZeroUsize) -> Self {
+        self.size = size;
+        self
+    }
+
+    ///  Sets the default client side TTL when it's not set with `cache_with_client_side_ttl` per command.
```
nihohit, line 45: try to avoid acronyms in documentation, or at least add the expanded version - 
Sets the default client-side Time To Live (TTL) for cached values, when a TTL isn't explicitly passed on a `Cmd`.

nihohit (commented):
public API is good. I'll try to get another round on the `caching` module. One thing that I already noticed is that there are no unit tests there.

nihohit:
BTW, the benchmark is for a local server, right? I assume that the number will be better for the cache once you have to pay for network wait, too.

altanozlu:
> BTW, the benchmark is for a local server, right? I assume that the number will be better for the cache once you have to pay for network wait, too.

Indeed I'm planning to use `tc` for simulating 1ms latency environment.
Will look into rest of the review tomorrow

redis/src/cmd.rs
altanozlu: Creating option struct and having one function to take it makes more sense because currently this function also enables for OptIn mode and once OptOut is introduced there will be need for another function.

altanozlu:
Test coverage was already 88% with existing tests, now I've also added unit tests for areas tests didn't cover.
Also applied changes in the review

altanozlu:
I will dig into more about PipelineResponseExpectation for now just pushed to pass CI

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,438 @@
...
+            // Caching must work with both MGET and GET
+            let _: Vec<Option<String>> = redis::cmd("MGET")
+                .arg("key_1")
+                .arg("key_2")
+                .query_async(&mut con)
+                .await?;
+            assert_hit(&con, 0);
+            assert_miss(&con, 2);
+
+            let _: Vec<Option<String>> = redis::cmd("MGET")
```
nihohit, line 100: can this check for keys 1,2, &3, but in the order of 1,3,2?
Just to verify that the caching mechanism is not in some way order dependent

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,438 @@
...
+                .query_async(&mut con)
+                .await?;
+
+            assert_eq!(k1, 42);
+            assert_eq!(k2, 43);
+            // There are 2 miss for key_1, key_2 used with MGET
+            assert_hit(&con, 0);
+            assert_miss(&con, 2);
+
+            let ((k1, k2), k3, k4): ((i32, i32), i32, Option<i32>) = get_pipe(atomic)
```
nihohit, line 178: nit: k1-4 implies *key* 1-4, but you're only testing 2 keys. maybe res1-4.

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,438 @@
...
+                .cmd("MGET")
+                .arg(&["key_1", "key_2"])
+                .cmd("GET")
+                .arg("key_1")
+                .cmd("GET")
```
nihohit, line 183: can this be moved one command earlier, just to verify that the mechanism isn't order-dependent?

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,438 @@
...
+        },
+        runtime,
+    )
+    .unwrap();
+}
+
+#[rstest]
+#[case::tokio(RuntimeType::Tokio)]
+#[cfg_attr(feature = "async-std-comp", case::async_std(RuntimeType::AsyncStd))]
+fn test_cache_opt_in(#[case] runtime: RuntimeType) {
```
nihohit, line 205: it's unclear from the name of the test how this is different from `test_cache_basic` with opt_in=true

redis/tests/test_cache.rs
```diff
@@ -0,0 +1,438 @@
...
+        },
+        runtime,
+    )
+    .unwrap();
+}
+
+#[rstest]
+#[case::tokio(RuntimeType::Tokio)]
+#[cfg_attr(feature = "async-std-comp", case::async_std(RuntimeType::AsyncStd))]
+fn test_cache_pipeline_with_opt_in(
```
nihohit, line 275: same

nihohit (commented):
love the tests!

nihohit:
a strategic question - I don't think that we should ship this as part of 0.28 - that version will concentrate on pubsub, and I don't want to burden it with another large feature.
Will we wait with merging this until after the version ships? merge it but somehow disable/hide the feautre?
I don't want to use another long living feature branch - that was pretty painful with RESP3 support. WDYT?

altanozlu:
using cfg and putting it behind unstable flag like tokio does as temporary measure could work but doesn't look nice. So it's better to merge it after releasing 0.28. And maybe cut out ` add pubsub support to async cluster - TBD` from 0.28 plans to make things faster ? Let me know if you know better way to hide a feature.

redis/tests/test_cache.rs
altanozlu: Parts with `// Since cache is not used, hit should still be 0` and `// GET is not marked with cache(), there should be no MISS/HIT`  which is important to check OptIn works as opt in

redis/tests/test_cache.rs
altanozlu: I've put MGET in between them

redis/tests/test_cache.rs
altanozlu: Changed the order also added SET and asserts for values.

redis/tests/test_cache.rs
altanozlu: Changed them to r* also there is something similar in test_cache_pipeline_partial_opt_in, I will keep it unresolved for now.

redis/tests/test_cache.rs
altanozlu: Tried my best to rename it but I'm bad at this :/

nihohit:
no, that's the only way I know - using configuration flags.

wojciechbator:
You could still push it behind the flag in 0.28 and add a note of experimental support for client side caching. Then, in next release you could add it as a part of the main lib, and that'd still be a major introduction for most of users. A benefit of this would be a feedback from early users and less headache later.

redis/Cargo.toml
```diff
@@ -121,6 +123,8 @@ sentinel = ["dep:rand"]
 tcp_nodelay = []
 num-bigint = []
 disable-client-setinfo = []
+hashbrown = ["dep:hashbrown"]
```
altanozlu, line 126: It looks like it's removed in https://github.com/redis-rs/redis-rs/commit/a871419f53381dab429e596c6dc50b2d09bd711c#diff-0024a66d093c343a9c17108e393c8ebb89213d963deaeba6a7547ab7c799304dL126

but somehow it got here via rebase

altanozlu:
@nihohit I'll push another patch soon regarding hashbrown and adding experimental clause in docs. Then it's good to go.

redis/src/lib.rs
```diff
@@ -101,6 +101,7 @@
 //! * `uuid`: enables type conversion to UUID (optional)
 //! * `sentinel`: enables high-level interfaces for communication with Redis sentinels (optional)
 //! * `json`: enables high-level interfaces for communication with the JSON module (optional)
+//! * `cache-aio`: enables **experimental** client side caching for MultiplexedConnection and ConnectionManager (optional)
```
nihohit, line 104: still not for connection manager :(

redis/src/cmd.rs
```diff
@@ -601,6 +659,20 @@ impl Cmd {
...
+    #[cfg(feature = "cache-aio")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "cache-aio")))]
+    pub fn set_cache_config(&mut self, command_cache_config: CommandCacheConfig) -> &mut Cmd {
+        self.cache = Some(command_cache_config);
+        self
+    }
+
+    #[cfg(feature = "cache-aio")]
+    #[inline]
+    pub(crate) fn get_cache(&self) -> &Option<CommandCacheConfig> {
```
nihohit, line 673: get_cache_config

redis/src/caching/config.rs
```diff
@@ -0,0 +1,61 @@
+use std::num::NonZeroUsize;
+use std::time::Duration;
+
+/// Defines the behavior of the cache regarding which commands should be cached.
+#[derive(Clone, PartialEq, Debug, Copy)]
+pub enum CacheMode {
+    /// Try to cache every command. This is equivalent to calling [crate::Cmd::set_cache_config]/[crate::Pipeline::set_cache_config] on every command.
+    /// Disabling individual commands for caching with this [CacheMode::All] will not work.
```
nihohit, line 8: "Disabling individual commands for caching" - what does this mean? Do you mean "Disabling caching for individual commands"?

redis/src/caching/config.rs
```diff
@@ -0,0 +1,61 @@
+use std::num::NonZeroUsize;
+use std::time::Duration;
+
+/// Defines the behavior of the cache regarding which commands should be cached.
+#[derive(Clone, PartialEq, Debug, Copy)]
+pub enum CacheMode {
+    /// Try to cache every command. This is equivalent to calling [crate::Cmd::set_cache_config]/[crate::Pipeline::set_cache_config] on every command.
```
nihohit, line 7: Try to cache the key & value for every relevant command? I think that I'm missing a document that explains what does caching actually mean - it's not commands that are cached, it's keys from sent commands, right?
I can write the explanation, if you prefer.

redis/src/caching/config.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Disabling individual commands for caching with this [CacheMode::All] will not work.
+    All,
+    /// Only try to cache commands which explicitly enabled caching.
```
nihohit, line 10: Only cache commands* that are explicitly marked by the user for caching?

or "keys from commands"

redis/src/caching/config.rs
```diff
@@ -0,0 +1,61 @@
...
+        self
+    }
+
+    /// Sets maximum key count for the cache.
+    pub fn set_size(mut self, size: NonZeroUsize) -> Self {
+        self.size = size;
+        self
+    }
+
+    ///  Sets the default client side time to live (TTL) for cached values, when a TTL isn't explicitly passed to [crate::Cmd]/[crate::Pipeline].
```
nihohit, line 46: The doc here should explain how to explicitly pass a TTL to a command. Also, since `CacheMode::All` is like setting a cache config on every command, which TTL take precedence when using that cache mode?

redis/src/caching/mod.rs
```diff
@@ -0,0 +1,18 @@
+//! This module provides **experimental** support for Server-assisted client-side caching in Redis.
+//!
+//! This implementation of Server-assisted client-side cache, only works with RESP3 connections.
+//! Some Redis providers might not enable this feature (e.g. Google Cloud Memorystore).
```
nihohit, line 4: lt's not name `Memorystore` here, since this will require us to update the docs if they change their behavior.

redis/src/caching/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+//! Commands that are cached will be kept in memory, commands won't be deleted from memory immediately,
+//! instead they will be removed by LRU or when same key is requested again.
```
nihohit, line 6: by LRU, by their Time To Live (TTL), or when ...

nihohit (approved):
still didn't have time to look at the internal implementation, but aside from the doc corrections, I think this is ready to merge, and we'll iterate on the internal implementation from main.
Very through work, well done!

redis/src/lib.rs
altanozlu: ah :(

redis/src/caching/config.rs
altanozlu: It's actually both key and command, when you try `GET mouse` it will create new CacheItem and put it on LRU using key `mouse` then it will write `GET A` information and it's response under CacheItem's value_list (Vec). When you use another command like `STRLEN mouse` it will not use cached one because `STRLEN mouse` is not cached `GET mouse` is. Holding all command instead of  cmd name makes sense for cases like `GETRANGE mouse 0 3`.

altanozlu: I will write more specific information in documentation.

redis/src/caching/config.rs
altanozlu: > The doc here should explain how to explicitly pass a TTL to a command.
will do it with next patch

> Also, since CacheMode::All is like setting a cache config on every command, which TTL take precedence when using that cache mode?

`   ///  Sets the default client side time to live (TTL) for cached values, when a TTL isn't explicitly passed to [crate::Cmd]/[crate::Pipeline].`
It actually says that if user isn't setting it explicitly, default client side ttl will be used. I will reword this bit

redis/src/caching/config.rs
nihohit: thank you :)

nihohit: what's the rationale behind this? I would like to cache `GET mouse`, and get the cached value from `MGET mouse` too.

redis/src/caching/mod.rs
altanozlu: I'm changing the wording and also adding redis invalidation.

redis/src/caching/config.rs
altanozlu: for MGET/ JSON.MGET it works like you expected, they have special case and they will be saved&checked as their singular form GET/JSON.GET.

you can check `test_cache_mget` for this and I can add more unit tests in future PRs regarding this.

redis/src/caching/config.rs
altanozlu: Changed also here

nihohit:
But what's the reason for this? Why cache the value for mget but not for
strlen?

redis/src/caching/config.rs
nihohit: But what's the reason for this? Why cache the value for mget but not for
strlen?

altanozlu: If `STRLEN A` is sent then result of it will be cached but `GET A` won't use that cached information, because their command is different and it make sense because otherwise you can get response of `STRLEN A` when you send `GET A` which is broken. MGET is just a safe optimization.
If you think this still doesn't make sense can you please explain what  should be cached for STRLEN exactly?

nihohit: oh, ok, I see - it's because the response of STRLEN is the computed value in the server. Sorry, this flew over my head 🤦.

nihohit:
Excellent work @altanozlu. Anything to add before we merge this?

altanozlu:
> Excellent work @altanozlu. Anything to add before we merge this?

Currently none, there are some changes regarding benchmark but I would like to do it in next iterations.

nihohit:
Then it's in! Hurrah!

Issue #1294, opened 2024-08-30
r2d2 pooling for SentinelClient


smf8:
Hi, 

Currently, there is no support for SentinelClient in `r2d2.rs`. Adding 
```rust
#[cfg(feature = "sentinel")]
impl_manage_connection!(
    crate::sentinel::SentinelClient,
    crate::Connection
);
```
will fix the compile issue for using `r2d2::Pool<sentinel::SentinelClient>`. Is there any runtime issue for using r2d2 pool with this approach?

nihohit:
from a brief look, it looks feasible. would you like to open a PR?

smf8:
I think we can close it as completed :)

PR #1297, opened 2024-09-01
Add r2d2 support for SentinelClient

smf8:
Hi, I apologize beforehand for I'm a newbie in Rust.

Regarding #1294,  I've introduced a new struct wrapper around `SentinelClient` to make it possible to use it with the macro used in `r2d2.rs`.

The usage will be something like this: 
```rust
pub struct PooledSentinelClient;

impl PooledSentinelClient {
    pub fn new_pool(
        cfg: &crate::config::RedisSentinelConfig,
        server_type: &str,
    ) -> anyhow::Result<Pool<sentinel::LockedSentinelClient>> {
        let server_type = match server_type {
            "master" => SentinelServerType::Master,
            "slave" => SentinelServerType::Replica,
            _ => bail!("invalid server type provided: {}", cfg.sentinel.master_name),
        };

        let sentinel_client = SentinelClient::build(
            vec![format!("redis://{}", cfg.address)],
            cfg.sentinel.master_name.clone(),
            None,
            server_type,
        )?;

        let pool = Pool::builder()
            .min_idle(Some(cfg.min_idle_conns))
            .max_size(cfg.max_pool_size)
            .idle_timeout(Some(cfg.idle_timeout))
            .connection_timeout(cfg.pool_timeout)
            .build(LockedSentinelClient::new(sentinel_client))?;

        Ok(pool)
    }
}
```

And to use the pool:
```
        let mut conn = pool.get().unwrap();
        let val: String = conn.get("key").unwrap();
```

What do you think about it?

### DIFF
--- a/redis/src/r2d2.rs
+++ b/redis/src/r2d2.rs
@@ -1,4 +1,4 @@
-use crate::sentinel::{LockedSentinelClient};
+use crate::sentinel::LockedSentinelClient;
 use crate::{ConnectionLike, RedisError};
 use std::io;
 
--- a/redis/src/r2d2.rs
+++ b/redis/src/r2d2.rs
@@ -1,6 +1,6 @@
-use std::io;
-
+use crate::sentinel::{LockedSentinelClient};
 use crate::{ConnectionLike, RedisError};
+use std::io;
 
 macro_rules! impl_manage_connection {
     ($client:ty, $connection:ty) => {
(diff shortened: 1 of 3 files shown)


nihohit:
Looks great, well done!

Can you please add a test to `test_sentinel.rs` to check this?

smf8:
Hi @nihohit

I'm not sure if this test covers what is expected, If you have any special case in mind, tell me and I'll try to add it.

Thank you very much.

nihohit (approved):
Excellent work, thank you very much!

Issue #1307, opened 2024-09-03
`ConnectionManagerConfig::default` should return the default


feelingsonice:
I think it's counter-intuitive that the `ConnectionManagerConfig::default` doesn't actually return the default, but you have to use `ConnectionManagerConfig::new` instead. This caused some confusion for me. A very trivial change - just a suggestion :)

jaymell:
I think that's a totally reasonable suggestion.

PR #1308, opened 2024-09-04
Default for ConnectionManagerConfig

feelingsonice:
https://github.com/redis-rs/redis-rs/issues/1307#issue-2503994237

Any where else this needs to be done? @jaymell


nihohit (approved):

jaymell:
Thanks!

> Any where else this needs to be done?

Not sure offhand. Feel free to raise other PRs, though.

jaymell:
This doesn't break semver, right? :thinking:

nihohit:
Arguably it does, but IMO the default were probably unusable, so this is the right decision.

Issue #1324, opened 2024-09-14
error compling crate without sentinel feature


FireMasterK:
This is a regression in 0.27.0 from 0.26.1.

Here is how I define the dependency in `Cargo.toml`

```
redis = { version = "0.27.0", features = ["tokio-comp", "connection-manager", "r2d2"] }
```

Compilation error logs:

```
Compiling redis v0.27.0
error[E0432]: unresolved import `crate::sentinel`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba[15](https://github.com/FireMasterK/market-handler/actions/runs/10852313646/job/30118074533?pr=80#step:5:16)001f/redis-0.27.0/src/r2d2.rs:1:12
    |
1   | use crate::sentinel::LockedSentinelClient;
    |            ^^^^^^^^ could not find `sentinel` in the crate root
    |
note: found an item that was configured out
   --> /home/runner/.cargo/registry/src/index.crates.io-6f[17](https://github.com/FireMasterK/market-handler/actions/runs/10852313646/job/30118074533?pr=80#step:5:18)d22bba15001f/redis-0.27.0/src/lib.rs:607:9
    |
607 | pub mod sentinel;
    |         ^^^^^^^^
note: the item is gated behind the `sentinel` feature
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/redis-0.27.0/src/lib.rs:605:7
    |
605 | #[cfg(feature = "sentinel")]
    |       ^^^^^^^^^^^^^^^^^^^^
```

Edit: #1323 seems to fix this?

nihohit:
Thanks for reporting this. I released a version with a fix.

PR #1328, opened 2024-09-14
Add CI action to test whether feature combinations compile.

nihohit:
This should prevent issues such as https://github.com/redis-rs/redis-rs/issues/1324


redis/Cargo.toml
```diff
@@ -38,26 +38,26 @@ combine = { version = "4.6", default-features = false, features = ["std"] }
 
 # Only needed for AIO
 bytes = { version = "1", optional = true }
-futures-util = { version = "0.3.15", default-features = false, optional = true }
```
jaymell, line 41: Is there a reason to do this? Offhand it seems to me like the patch versions specified in here are useful from a documentation perspective if nothing else. If we don't actually know for sure that the `major.minor.0` versions of these crates will actually compile with the code here, it seems better to not go back and indicate as such now.

redis/Cargo.toml
```diff
@@ -99,7 +99,7 @@ json = ["serde", "serde/derive", "serde_json"]
 cluster = ["crc16", "rand"]
 script = ["sha1_smol"]
 tls-native-tls = ["native-tls"]
-tls-rustls = ["rustls", "rustls/ring", "rustls-native-certs", "rustls-pemfile", "rustls-pki-types"]
+tls-rustls = ["rustls/std", "rustls/ring", "rustls-native-certs", "rustls-pemfile", "rustls-pki-types"]
```
jaymell, line 102: Why is this required?

redis/src/cluster.rs
```diff
@@ -238,7 +238,7 @@ where
             connections: RefCell::new(HashMap::new()),
             slots: RefCell::new(SlotMap::new(cluster_params.read_from_replicas)),
             auto_reconnect: RefCell::new(true),
-            read_timeout: RefCell::new(None),
+            read_timeout: RefCell::new(cluster_params.response_timeout),
```
jaymell, line 241: Do we actually need `read_timeout`? Should it just be removed in favor of `cluster_params.response_timeout`?

redis/src/cluster.rs
```diff
@@ -410,7 +410,7 @@ where
         let params = self.cluster_params.clone();
         let info = get_connection_info(node, params)?;
 
-        let mut conn = C::connect(info, None)?;
```
jaymell, line 413: If I'm reading this right, we effectively don't use connection timeout settings in this client? This seems like a potentially impactful to merit a more detailed commit message/possibly a separate PR?

redis/src/cluster_client.rs
```diff
@@ -82,8 +82,8 @@ pub(crate) struct ClusterParams {
     pub(crate) tls: Option<TlsMode>,
     pub(crate) retry_params: RetryParams,
     pub(crate) tls_params: Option<TlsConnParams>,
-    pub(crate) connection_timeout: Duration,
-    pub(crate) response_timeout: Duration,
+    pub(crate) connection_timeout: Option<Duration>,
```
jaymell, line 85: Sorry, I vaguely recall we discussed this elsewhere, but remind me why we're doing this? Can we fix the warnings some other way? It seems to me like things are more clear if the defaults are defined here rather than in the connect method.

redis/Cargo.toml
nihohit: Doesn't this limit users of the crate? The dependency resolver has to use this specific version of the dependency, and if they have another crate pointing at a different patch version, the dependency is compiled twice?

redis/src/cluster.rs
nihohit: I guess we can remove it - I tried to avoid a larger change.

redis/src/cluster.rs
nihohit: I'll open this as a separate PR, then.

redis/src/cluster_client.rs
nihohit: this is done in order to bridge the gap between sync & async clients. ATM the sync client doesn't use these fields, so if we want to use these fields in the sync clients, either we introduce new defaults to the sync client or change the fields to `Option` and set the defaults in the async client.

redis/Cargo.toml
jaymell: I don't think it's limiting, though. I think it ensures that only versions >= what's actually been tested are used, which should generally result in fewer old bugs.

redis/src/cluster.rs
jaymell: Ok, avoiding the larger change is probably fine then, or maybe worth considering in the separate PR.

redis/src/cluster_client.rs
jaymell: We could also just potentially add an `allow_unused` that's specific to  `!cluster_async` feature?

nihohit: we could, but wouldn't it be better if these fields actually did what they say they do?

jaymell: Yep, I agree. I think adding defaults for the sync client is a good idea.

nihohit: just making sure - you want to apply the same defaults we currently have on the async client to the sync client?

jaymell: Yes, albeit in a separate PR. Alternatively, I'm ok with just allowing the fields to be unused, which although goofy, is effectively what we currently have, right?

nihohit: right.
https://github.com/redis-rs/redis-rs/pull/1333

redis/Cargo.toml
nihohit: removed.

redis/Cargo.toml
nihohit: removed

jaymell: I was referring to line 102. I still see the change.

nihohit: Sorry, I was wrong. This is needed in order to build the crate with only the `tls-rustls` flag. You can try it locally.

jaymell (approved):
:sunglasses:

PR #1332, opened 2024-09-16
Add support for [TYPE type] in SCAN commands

Reiuji-ch:
Redis supports adding a `[TYPE type]` subcommand to SCAN since 6.0.0
Similar to a `[MATCH pattern]`, this will filter the returned values, but does not actually reduce the amount of work done by Redis. The TYPE subcommand is only supported on SCAN. SSCAN, HSCAN and ZSCAN does not support it. 

This PR adds basic support for the `[TYPE type]` subcommand. It simply adds a `with_type` method to the existing `ScanOptions`. Currently it takes an arbitrary string, which should be a Redis type, e.g. "string" or "hash". 

This PR should be fully backwards compatible.

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -2262,7 +2262,7 @@ impl ScanOptions {
         self
     }
     
-    // Limit the results to those with the given Redis type
+    /// Limit the results to those with the given Redis type
     pub fn with_type(mut self, t: impl Into<String>) -> Self {
         self.scan_type = Some(t.into());
         self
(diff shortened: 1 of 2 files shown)


nihohit (approved):
Thanks for the contribution!

nihohit:
@Reiuji-ch please add the missing documentation

Reiuji-ch:
@nihohit Fixed

Reiuji-ch:
@nihohit I forgot the formatter hates whitespace. This should fix the cargo fmt check as well

nihohit:
Thanks again for the contribution :)

PR #1333, opened 2024-09-17
Align default timeouts on cluster client.

nihohit:
The sync cluster client now uses the same default timeouts as the async client. Additionally, now `ClusterClient::response_timeout` and `ClusterClient::connection_timeout` will actually do something on the sync cluster client.


jaymell (approved):
:sunglasses:

Issue #1006, opened 2023-12-12
`TlsConnParams` is opaque: Intentional?


aumetra:
When attempting to update the implementation of [`deadpool-redis`](https://github.com/bikeshedder/deadpool/pull/288) to `redis` v0.24, we noticed that the [`TlsConnParams`](https://docs.rs/redis/0.24.0/redis/enum.ConnectionAddr.html#variant.TcpTls.field.tls_params) struct is fully opaque.

Is this intentional? This seems a little like an oversight, especially since all other types associated with this feature are [exposed publicly](https://github.com/redis-rs/redis-rs/blob/3e6dc2ef611060a665d8e5d4c14bd8618517fac2/redis/src/lib.rs#L468-L469)

jaymell:
Can you explain the use case this is breaking for you? The biggest issue at the moment is that the fields in that struct are `rustls`-specific, which we definitely don't want to expose if/when we support native in the future.

aumetra:
> Can you explain the use case this is breaking for you?

Well, it's not particularly breaking, it's just a question if this is truly intended.

Background: `deadpool-redis` creates a full duplicate of the configuration struct so it can derive the `serde` traits on it.

With this field being opaque, this would technically create a discrepancy between the vendored configuration struct and the actual configuration struct.

This is more of a question issue, to ask whether this was intended, since it would be a breaking change in our case to add this field after the fact (while making the struct public would be a non-breaking change).

aumetra:
Technically another solution would be if the `redis` crate would implement the `serde` traits for the configuration struct, but that's more of an out-there request, especially since that would impose one of the two issues:

Either one has to commit to stability of the serialized representation (at which point the fields might as well be public), or the representation is explicitly opted out of semver, which would then create issues for downstream users that want to rely on the serialized representation across patch versions.

jaymell:
Another option could be to store raw `Vec<u8>`s in `ConnectionAddr` and deal with the tls-crate specifics elsewhere, though I'm not really clear offhand on what kind of inefficiencies that might introduce. I'm not opposed to a bit of refactoring here though it's not something we're likely to work on any time soon. PRs are welcome, though.

PR #1335, opened 2024-09-19
Allow configuring Sentinel with custom TLS certificates

ergonjomeier:
We would like to be able to configure the TLS certificates that are used with Sentinel. Similar functionality was already included in https://github.com/redis-rs/redis-rs/pull/858, but it was reverted and was said to better be included in a separate PR.

There are two connections for which we would like to configure the certificates:

1. Client <-> Redis. This is done in 2954990dcee076d9d3fbca1b771e12d05379be24 by adding a field `certs` to `SentinelNodeConnectionInfo`. This is a breaking change but I think it's quite hard to include this functionality without any breaking changes. For example, `Sentinel` offers several `pub fn`, such as `master_for`, which would need some access to the certificates. If we don't want to change the signature of these function we would need to duplicate them. I haven't thought this all the way through, but feel like the solution in this PR is much cleaner.
2. Client <-> Sentinel. This is done in 20248c160d73631d89a20fa38e6cf0c711e5bfb2 by adding a new function `SentinelClient::build_with_certs`, which patches the nodes in `params` with the certificates. The reason why this is not possible in the first place ist because `TlsConnParam` is opaque (https://github.com/redis-rs/redis-rs/issues/1006).

### DIFF
--- a/redis/src/sentinel.rs
+++ b/redis/src/sentinel.rs
@@ -119,9 +119,9 @@
 //! builder = builder.client_to_sentinel_password(String::from("password1"));
 //! builder = builder.client_to_sentinel_tls_mode(redis::TlsMode::Insecure);
 //!
-//! builder = builder.sentinel_to_redis_username(String::from("username2"));
-//! builder = builder.sentinel_to_redis_password(String::from("password2"));
-//! builder = builder.sentinel_to_redis_tls_mode(redis::TlsMode::Secure);
+//! builder = builder.client_to_redis_username(String::from("username2"));
+//! builder = builder.client_to_redis_password(String::from("password2"));
+//! builder = builder.client_to_redis_tls_mode(redis::TlsMode::Secure);
 //!
 //!
 //! let client = builder.build().unwrap();
@@ -870,11 +870,14 @@ struct BuilderConnectionParams {
 }
 
 /// Used to configure and build a [`SentinelClient`].
+/// There are two connections that can be configured independently
+/// 1. The connection towards the redis nodes (configured via `set_client_to_redis_..` functions)
+/// 2. The connection towards the sentinel nodes (configure via `set_client_to_sentinel_..` functions)
 pub struct SentinelClientBuilder {
     sentinels: Vec<ConnectionAddr>,
     service_name: String,
     server_type: SentinelServerType,
-    sentinel_to_redis_params: BuilderConnectionParams,
+    client_to_redis_params: BuilderConnectionParams,
     client_to_sentinel_params: BuilderConnectionParams,
 }
 
@@ -892,7 +895,7 @@ impl SentinelClientBuilder {
             sentinels: sentinels.into_iter().collect::<Vec<_>>(),
             service_name,
             server_type,
-            sentinel_to_redis_params: BuilderConnectionParams {
+            client_to_redis_params: BuilderConnectionParams {
                 tls_mode: None,
                 db: None,
                 username: None,
@@ -913,28 +916,28 @@ impl SentinelClientBuilder {
 
     /// Creates a new `SentinelClient` from the parameters
     pub fn build(mut self) -> RedisResult<SentinelClient> {
-        let mut sentinel_to_redis_connection_info = RedisConnectionInfo::default();
+        let mut client_to_redis_connection_info = RedisConnectionInfo::default();
 
-        if let Some(db) = self.sentinel_to_redis_params.db {
-            sentinel_to_redis_connection_info.db = db;
+        if let Some(db) = self.client_to_redis_params.db {
+            client_to_redis_connection_info.db = db;
         }
 
-        if let Some(username) = self.sentinel_to_redis_params.username {
-            sentinel_to_redis_connection_info.username = Some(username);
+        if let Some(username) = self.client_to_redis_params.username {
+            client_to_redis_connection_info.username = Some(username);
         }
 
-        if let Some(password) = self.sentinel_to_redis_params.password {
-            sentinel_to_redis_connection_info.password = Some(password);
+        if let Some(password) = self.client_to_redis_params.password {
+            client_to_redis_connection_info.password = Some(password);
         }
 
-        if let Some(protocol) = self.sentinel_to_redis_params.protocol {
-            sentinel_to_redis_connection_info.protocol = protocol;
+        if let Some(protocol) = self.client_to_redis_params.protocol {
+            client_to_redis_connection_info.protocol = protocol;
         }
 
-        let sentinel_to_redis_connection_info = SentinelNodeConnectionInfo {
-            tls_mode: self.sentinel_to_redis_params.tls_mode,
-            redis_connection_info: Some(sentinel_to_redis_connection_info),
-            certs: self.sentinel_to_redis_params.certificates,
+        let client_to_redis_connection_info = SentinelNodeConnectionInfo {
+            tls_mode: self.client_to_redis_params.tls_mode,
+            redis_connection_info: Some(client_to_redis_connection_info),
+            certs: self.client_to_redis_params.certificates,
         };
 
         for sentinel in &mut self.sentinels {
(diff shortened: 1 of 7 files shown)


redis/src/sentinel.rs
```diff
@@ -151,7 +163,10 @@ impl SentinelNodeConnectionInfo {
                 host: ip,
                 port,
                 insecure: false,
+                #[cfg(not(feature = "tls-rustls"))]
                 tls_params: None,
+                #[cfg(feature = "tls-rustls")]
+                tls_params: Some(retrieve_tls_certificates(self.certs.clone().unwrap()).unwrap()),
```
nihohit, line 169: let's not use `unwrap` here, and instead match on `self.certs.clone()`, and return error on a failed result of `retrieve_tls_certificates`.

redis/src/sentinel.rs
```diff
@@ -141,6 +149,10 @@ pub struct SentinelNodeConnectionInfo {
 
     /// The Redis specific/connection independent information to be used.
     pub redis_connection_info: Option<RedisConnectionInfo>,
+
+    /// Certificates to be used, or none if we want to use certificates from local truststore
+    #[cfg(feature = "tls-rustls")]
+    pub certs: Option<TlsCertificates>,
 }
 
 impl SentinelNodeConnectionInfo {
```
nihohit, line 158: please add setters for all of `SentinelNodeConnectionInfo`'s fields.
We should've done it beforehand, sorry :(

redis/src/sentinel.rs
```diff
@@ -714,6 +731,27 @@ impl LockedSentinelClient {
...
+
+    for node in params.iter_mut() {
+        if let crate::ConnectionAddr::TcpTls {
+            host: _,
+            port: _,
+            insecure: _,
+            ref mut tls_params,
+        } = node.addr
+        {
+            *tls_params = Some(new_tls_params.clone());
```
nihohit, line 749: what if `tls_params` is already set with some value? this should probably return an error.

nihohit (commented):
thanks for the contribution!

@jaymell, you've done most of the TLS reviews so far - any additional comments?

redis/src/sentinel.rs
jaymell: Do we actually need setters? I'm not clear what it benefits here.

jaymell:
> @jaymell, you've done most of the TLS reviews so far - any additional comments?

I think this mostly looks good, but if we're going to introduce breaking changes here, we might consider some larger refactors. There's a lot that makes me confused/grumpy when I go back and look at the implementation of the Sentinel clients. Let me think on it a bit more.

redis/src/sentinel.rs
nihohit: adding a public field to a public struct is a breaking change, since it breaks all explicit initializations. Setters should be used instead of public fields, so that the `Foo::new().set_bar(bar)` structure is never broken.

jaymell: I'm not a huge fan of patterns that require mutability. Builders are the right choice here (and for all the clients).

nihohit: ah, sorry. yes, I meant the builder pattern. my bad :)

ergonjomeier:
Thanks a lot for the reviews! If we want to do a larger refactoring, one could consider a builder-Pattern for the `SentinelClient`, similar to `ClusterClientBuilder`. This would make it easier to introduce new functionality in a non-breaking way in the future. The initial introduction of a `SentinelClientBuilder` might still require some breaking changes, though.

ergonjona:
@jaymell, I would like to continue working on this PR my colleague @ergonjomeier started.
Have you had time to look into the implementation of the sentinel client?

I could address the creation of `SentinelNodeConnectionInfo` first. It was suggested to implement a builder, but I'm not sure if a simple constructor would suffice.
My colleague also suggested to implement a builder for the sentinel client. With a `SentinelClientBuilder` `SentinelNodeConnectionInfo` could become a private struct.

What are your thoughts?

nihohit:
Hi, sorry it took so long to get back to you. I think that ATM using a builder is a sound strategy. It will, sadly, be a breaking change, but I don't see a way to expand those structs without one.

redis/src/sentinel.rs
```diff
@@ -807,3 +837,251 @@ impl SentinelClient {
...
+        SentinelClient::build(
+            sentinels,
+            self.service_name,
+            Some(sentinel_to_redis_connection_info),
+            self.server_type,
+        )
+    }
+
+    /// Set tls mode for the connection between sentinels and redis nodes
+    pub fn sentinel_to_redis_tls_mode(mut self, tls_mode: TlsMode) -> SentinelClientBuilder {
```
nihohit, line 1005: I know that the crate is inconsistent on this, I'm trying to make it better. please use `set_` on all of these setters.

redis/src/sentinel.rs
```diff
@@ -141,17 +149,27 @@ pub struct SentinelNodeConnectionInfo {
 
     /// The Redis specific/connection independent information to be used.
     pub redis_connection_info: Option<RedisConnectionInfo>,
+
+    /// Certificates to be used, or none if we want to use certificates from local truststore
+    #[cfg(feature = "tls-rustls")]
+    pub certs: Option<TlsCertificates>,
```
nihohit, line 155: this is the only breaking change that I see, and it too is avoidable.
pushed a commit to demonstrate how to avoid this:
https://github.com/nihohit/redis-rs/commit/0b7a901b6f14cc4f90a4eb2442a3c23be500480f

redis/src/sentinel.rs
```diff
@@ -807,3 +837,251 @@ impl SentinelClient {
...
+pub struct SentinelClientBuilder {
+    sentinels: Vec<ConnectionAddr>,
+    service_name: String,
+    server_type: SentinelServerType,
+    sentinel_to_redis_params: BuilderConnectionParams,
+    client_to_sentinel_params: BuilderConnectionParams,
+}
+
+impl SentinelClientBuilder {
+    /// Creates a new `SentinelClientBuilder`
```
nihohit, line 860: please expand the documentation to explain the arguments here.

redis/src/sentinel.rs
```diff
@@ -807,3 +837,251 @@ impl SentinelClient {
...
+    }
+
+    /// Set tls mode for the connection between sentinels and redis nodes
+    pub fn sentinel_to_redis_tls_mode(mut self, tls_mode: TlsMode) -> SentinelClientBuilder {
+        self.sentinel_to_redis_params.tls_mode = Some(tls_mode);
+        self
+    }
+
+    /// Set db for the connection between sentinels and redis nodes
+    pub fn sentinel_to_redis_db(mut self, db: i64) -> SentinelClientBuilder {
```
nihohit, line 1011: is the `db` argument used in sentinel connections?

nihohit, line 1011: the struct documentation should explain the difference between `sentinel_to_redis` values and `client_to_sentinel` values.

nihohit (commented):
great job!

I think that this design is robust, and doesn't break backwards compatibility. What's missing is to add documentation & example to the top of the `sentinel.rs` file, and to add or modify an existing test to use the builder.

redis/src/sentinel.rs
ergonjomeier: Unfortunately, I am not sure whether the proposed solution solves the issue. I am wondering about the following two things:

- Isn't there still a breaking change, but now it's in `SentinelClient::build`, as we add an additional argument to the function?
- It seems the certificates for the client <-> redis connection are not properly propagated anymore (the `certificates` field of `client_to_sentinel_params` is unused)

ergonjomeier:
Thanks a lot for the review! I am currently working on the improvements. I have done a few renames, and added documentation. I will add a test as well. As for making the whole change non-breaking, I am unsure (see https://github.com/redis-rs/redis-rs/pull/1335#discussion_r1939667181)

redis/src/sentinel.rs
nihohit: I think that it isn't. let's drop this.

nihohit:
@ergonjomeier Thanks for the effort :)

redis/src/sentinel.rs
ergonjomeier: Sorry! Yes, you are absolutely right, I think i misread some parts of the code. I have implemented your suggestion, the only remaining breaking change I see is that a `certificates` field is added to the `Sentinel` struct, which is also `pub`.

nihohit:
@ergonjomeier  Thanks for implementing the changes!

> the only remaining breaking change I see is that a certificates field is added to the Sentinel struct, which is also pub.

Hi, I don't see that addition. Am I missing something? Also, the CI is broken - can you please fix the build & lint issues?

ergonjomeier:
I am currently working on fixing the CI. Still WIP though, just pushed a change to see tomorrow whether the build went through.

As for the breaking change, you are right. I missed that the fields of the `struct Sentinel` themselves are not public.

nihohit:
@ergonjomeier just missing the last lints :)

ergonjomeier:
Ah damn! I thought I ran the linter. Anway, should be ok now :)

nihohit (approved):

nihohit:
Thanks for the contribution, and sorry it took so long :)

ergonjomeier:
Glad I could help! No worries, I took my time as well :) Cheers

PR #1339, opened 2024-09-20
Improve docs.

nihohit:
Most of the changes here involve taking a large block of test and splitting off the first line. docs.rs shows only the first paragraph of a doc in an upper scope as blurb, and these types had outlandishly long blurbs.

### DIFF
--- a/redis/src/sentinel.rs
+++ b/redis/src/sentinel.rs
@@ -694,9 +694,9 @@ pub enum SentinelServerType {
     Replica,
 }
 
-/// LockedSentinelClient is a wrapper around SentinelClient since
-/// it SentinelClient requires &mut ref for get_connection()
-/// and we can't use it like this inside the r2d2 Manager
+/// LockedSentinelClient is a wrapper around SentinelClient usable in r2d2.
+// [internal comment]: this was added since SentinelClient requires &mut ref for get_connection()
+// and we can't use it like this inside the r2d2 Manager, which requires a shared reference.
 #[cfg(feature = "r2d2")]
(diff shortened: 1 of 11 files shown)


redis/src/sentinel.rs
```diff
@@ -694,9 +694,7 @@ pub enum SentinelServerType {
     Replica,
 }
 
-/// LockedSentinelClient is a wrapper around SentinelClient since
```
jaymell, line 697: I think it's probably worth preserving the comment about the `&mut` ref (though just as a comment, not in the docs). It would be great to change that in the future :grin:

redis/src/sentinel.rs
```diff
@@ -714,10 +712,10 @@ impl LockedSentinelClient {
     }
 }
 
-/// An alternative to the Client type which creates connections from clients created
-/// on-demand based on information fetched from the sentinels. Uses the Sentinel type
-/// internally. This is basic an utility to help make it easier to use sentinels but
-/// with an interface similar to the client (`get_connection` and
+/// An utility wrapping `Sentinel` with an interface similar to [Client].
```
jaymell, line 715:
```suggestion
/// A utility wrapping `Sentinel` with an interface similar to [Client].
```

redis/src/sentinel.rs
```diff
@@ -714,10 +712,10 @@ impl LockedSentinelClient {
...
+///
+/// Uses the Sentinel type internally. This is an utility to help make it easier
```
jaymell, line 717:
```suggestion
/// Uses the Sentinel type internally. This is a utility to help make it easier
```

jaymell (approved):

Issue #639, opened 2022-06-24
PubSub should support PING and other commands redis allows


tugtugtug:
According to https://redis.io/docs/manual/pubsub/#:~:text=The%20commands%20that%20are%20allowed,the%20mode%20with%20Ctrl-C%20.
A subscriber connection can still ping the server. 

However, the current implementation seems to have purposely disabled such functions. We rely on it for server end connection management.

Can we please allow pings for the PubSub connections?

---- EDIT ---

The code seems extremely confusing with the `pubsub` boolean field. The `pubsub` does not seem to be ever set when PubSet is in use and intentionally so, as
https://github.com/redis-rs/redis-rs/blob/624a59a34b3b205c44e26a0391be1757d54a7e10/src/connection.rs#L718,
so technically, this `pubsub` flag is only used if a previous PubSub connection failed to unsubscribe all channels. Maybe name it `pending_unsubscribe` to be more accurate.

Since the PubSub takes the connection and hides it as an internal member, other than the exposed interfaces, the connection really is only accessible after either the PubSub is dropped or is transformed into a connection again.
So, is it possible to just provide a new function for PubSub to send ping? I'm unfamiliar with the reader, so not sure if that would break if one creates the on_message stream (for the async implementation), then sends a ping, as that now requires multiplexing of the responses.

djc:
Unfortunately I don't have a lot of capacity to allocate to this crate. I think it would make sense to add a `ping()` API to `PubSub` and I would be happy to review a PR implementing it, but I won't be able to work on it much (nor do have I time right now to investigate how it should work exactly).

nihohit:
https://github.com/redis-rs/redis-rs/pull/898

nihohit:
If you're using Redis/Valkey v6 or newer, you can use the RESP3 pubsub in order to use a regular connection with pubsub - https://docs.rs/redis/latest/redis/#resp3-async-pubsub

nihohit:
implemented in https://github.com/redis-rs/redis-rs/pull/1341 .

PR #1341, opened 2024-09-23
Add ping support to commands & pubsubs.

nihohit:
RESP2 allows pubsub connections to send PINGs on the connection without breaking from the pub/sub contract. This change adds `ping` and `ping_message` methods to all pubsub implementations, so that they could send the pings.
For the sync pubsub connection this requires saving responses from the server which aren't pubsub messages, instead of just returning completion from user requests.

https://github.com/redis-rs/redis-rs/issues/639

### DIFF
--- a/redis/src/aio/pubsub.rs
+++ b/redis/src/aio/pubsub.rs
@@ -5,7 +5,7 @@ use crate::connection::{
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use crate::parser::ValueCodec;
 use crate::types::{closed_connection_error, RedisError, RedisResult, Value};
-use crate::{cmd, Msg, RedisConnectionInfo, ToRedisArgs};
+use crate::{cmd, from_owned_redis_value, FromRedisValue, Msg, RedisConnectionInfo, ToRedisArgs};
 use ::tokio::{
     io::{AsyncRead, AsyncWrite},
     sync::oneshot,
@@ -28,12 +28,12 @@ use tokio_util::codec::Decoder;
 use super::SharedHandleContainer;
 
 // A signal that a un/subscribe request has completed.
-type RequestCompletedSignal = oneshot::Sender<RedisResult<()>>;
+type RequestResultSender = oneshot::Sender<RedisResult<Value>>;
 
 // A single message sent through the pipeline
 struct PipelineMessage {
     input: Vec<u8>,
-    output: RequestCompletedSignal,
+    output: RequestResultSender,
 }
 
 /// The sink part of a split async Pubsub.
@@ -81,7 +81,7 @@ pin_project! {
         #[pin]
         sink_stream: T,
         // The requests that were sent and are awaiting a response.
-        in_flight: VecDeque<RequestCompletedSignal>,
+        in_flight: VecDeque<RequestResultSender>,
         // A sender for the push messages received from the server.
         sender: UnboundedSender<Msg>,
     }
@@ -128,10 +128,10 @@ where
                 if let Some(Value::BulkString(kind)) = value.first() {
                     if matches!(
                         kind.as_slice(),
-                        b"subscribe" | b"psubscribe" | b"unsubscribe" | b"punsubscribe"
+                        b"subscribe" | b"psubscribe" | b"unsubscribe" | b"punsubscribe" | b"pong"
                     ) {
                         if let Some(entry) = self_.in_flight.pop_front() {
-                            let _ = entry.send(Ok(()));
+                            let _ = entry.send(Ok(Value::Array(value)));
                         };
                         return Ok(());
                     }
@@ -148,7 +148,7 @@ where
             Ok(Value::Push { kind, data }) => {
                 if kind.has_reply() {
                     if let Some(entry) = self_.in_flight.pop_front() {
-                        let _ = entry.send(Ok(()));
+                        let _ = entry.send(Ok(Value::Push { kind, data }));
                     };
                     return Ok(());
                 }
@@ -165,7 +165,7 @@ where
 
             _ => {
                 if let Some(entry) = self_.in_flight.pop_front() {
-                    let _ = entry.send(result.map(|_| ()));
+                    let _ = entry.send(result);
                     Ok(())
                 } else {
                     Err(())
@@ -278,7 +278,7 @@ impl PubSubSink {
         (PubSubSink { sender }, f)
     }
 
-    async fn send_recv(&mut self, input: Vec<u8>) -> Result<(), RedisError> {
+    async fn send_recv(&mut self, input: Vec<u8>) -> Result<Value, RedisError> {
         let (sender, receiver) = oneshot::channel();
 
         self.sender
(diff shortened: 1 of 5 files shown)


avifenesh:
As someone who is not following the development of the library constantly, I would appreciate some overview in the pr conversation page to have an easy life while reviewing.
An even better approach would be to add comments on the code changes (not in the code, here, in the PR).
This way I would be able to spend more time actually reviewing and less time getting to know the context.

nihohit:
@avifenesh you're right, added a PR description.

altanozlu (commented):
You can also add testing or example for PING commands without PubSub because RESP3&RESP2 behaves same when PubSub is not enabled

avifenesh:
> @avifenesh you're right, added a PR description.

I will get to it this weekend, crazy days

nihohit:
> You can also add testing or example for PING commands without PubSub

done

altanozlu (approved):

redis/src/aio/pubsub.rs
```diff
@@ -297,28 +297,44 @@ impl PubSubSink {
...
-        self.send_recv(cmd.get_packed_command()).await
+        self.send_recv(cmd.get_packed_command()).await.map(|_| ())
+    }
+
+    /// Sends a ping with a message to the server
+    pub async fn ping_message<T: FromRedisValue>(
+        &mut self,
+        message: impl ToRedisArgs,
+    ) -> RedisResult<T> {
+        let mut cmd = cmd("PING");
```
avifenesh, line 329: Is there a reason to use two lines with `mut` and not `cmd("PING").arg(message);`?
If the goal is readability, the extra verbosity needed is taking `cmd.get_packed_command()).await?` out, and giving the result as a parameter to `from_owned_redis_value`. 
Current style complicating the code, IDK what you expect from `cmd.get_packed_command()).await?`, the parameter itself is too long and not fun to look at.
Giving it a name will enhance the ability to understand what is the value you want to pass.
There is a hidden assumption that the next developer working on this code already knows all the functions and their results, while that's not the case if I would want to add some code to this part.

redis/src/aio/pubsub.rs
```diff
@@ -297,28 +297,44 @@ impl PubSubSink {
...
+        cmd.arg(message);
+        from_owned_redis_value(self.send_recv(cmd.get_packed_command()).await?)
+    }
+
+    /// Sends a ping to the server
+    pub async fn ping<T: FromRedisValue>(&mut self) -> RedisResult<T> {
+        let cmd = cmd("PING");
+        from_owned_redis_value(self.send_recv(cmd.get_packed_command()).await?)
```
avifenesh, line 337: Same

redis/src/aio/pubsub.rs
```diff
@@ -297,28 +297,44 @@ impl PubSubSink {
...
 
     /// Unsubscribes from channel pattern.
     pub async fn punsubscribe(&mut self, channel_pattern: impl ToRedisArgs) -> RedisResult<()> {
         let mut cmd = cmd("PUNSUBSCRIBE");
         cmd.arg(channel_pattern);
-        self.send_recv(cmd.get_packed_command()).await
+        self.send_recv(cmd.get_packed_command()).await.map(|_| ())
+    }
+
+    /// Sends a ping with a message to the server
```
avifenesh, line 324: Can't you use builder style here and do `ping().message()` or just accept optional message? (I prefer the first)
Common in rust, and align with the real command name. Both are `ping` command.
Moreover, not a strong reason, but if tomorrow there's another optional parameter, you just add another option to the builder.

redis/src/aio/pubsub.rs
```diff
@@ -424,6 +440,19 @@ impl PubSub {
         self.sink.punsubscribe(channel_pattern).await
     }
 
+    /// Sends a ping to the server
```
avifenesh, line 443: same

redis/src/connection.rs
```diff
@@ -1667,39 +1667,61 @@ impl<'a> PubSub<'a> {
...
+                return Ok(response);
             }
-            response = self.con.recv_response()?;
         }
     }
 
     /// Subscribes to a new channel.
     pub fn subscribe<T: ToRedisArgs>(&mut self, channel: T) -> RedisResult<()> {
-        self.cache_messages_until_received_response(cmd("SUBSCRIBE").arg(channel))
+        self.cache_messages_until_received_response(cmd("SUBSCRIBE").arg(channel), true)?;
```
avifenesh, line 1692: Why all pub/sub are returning Ok and not the response? It is valuable information.

avifenesh:
@nihohit Left comments. Will review the tests after comments resolved/dismissed and tests aligned with the comments addressing decisions.
🌷

redis/src/connection.rs
nihohit: That's the current behavior, which I'm not changing. Changing the return type of the function is a breaking change.

redis/src/aio/pubsub.rs
nihohit: I went with how the previous commands are written, no big reason.

`let cmd = cmd("PING").arg(message);` doesn't compile, because it keeps a reference to a temp value.
I can do `let cmd = cmd("PING").arg(message).get_packed_command();`

redis/src/aio/pubsub.rs
nihohit: there isn't a builder pattern for commands ATM, and given how long the ping command was unchanged, I'll be very susprised if this will be changed any time soon.
If we had ` ping()` object, what would it return? how would the object be executed?

nihohit:
@avifenesh answered & pushed

nihohit:
@avifenesh any additional comments?

redis/tests/test_basic.rs
```diff
@@ -915,6 +926,72 @@ mod basic {
...
+        use std::sync::{Arc, Barrier};
+        let ctx = TestContext::new();
+        let mut con = ctx.connection();
+
+        // Connection for subscriber api
+        let mut pubsub_con = ctx.connection();
+
+        // Barrier is used to make test thread wait to publish
+        // until after the pubsub thread has subscribed.
+        let publish_barrier = Arc::new(Barrier::new(2));
```
avifenesh, line 940: nit
It's not wrong, but `Barrier` seems like an untraditional solution, why not `mspc`?

avifenesh:
> @avifenesh any additional comments?

Just one not crucial comment, other than that LGTM.

avifenesh (approved):

redis/tests/test_basic.rs
nihohit: In general, barriers are pretty traditiontal - if anything, out of Rust they're much more common than channels. :)

But more relevant to this specific test - notice that the barriers are used for 2-directional blocking. The barriers block either thread until the other thread reaches the barrier. If I wanted to implement this with channels, I'd have complex code that needs to account for the order in which the threads reached the point of waiting. So Barrier is the abstraction that better fits the problem domain for this test.

PR #1343, opened 2024-09-26
default rustls to aws-lc-rs (instead of ring)

### LINKED ISSUE #1342: support rustls aws-lc-rs
Currently ring is hardcoded for rustls,
this gives however problems for projects that user other dependencies
which use the default `aws-lc-rs`...

GlenDC:
There are now `-ring` feature variants for all rustls features.

This does change the default for redis, but I think it's the right choice as rustls has chosen the default to be aws-lc-rs, and so not following that makes it impossible for users to use a crate that might use redis as a dependency itself... This instead follows the default while still supporting ring just the same.

Closes #1342


nihohit:
I would like to avoid the scenario where every crypto provider supported by rustls adds another feature to redis. Instead maybe we should add to `TlsConnParams` a field with a dyn crypto provider, or maybe even just the `WebPkiSupportedAlgorithms` that we get from `signature_verification_algorithms`?
We wrap it in an option and add it using the builder pattern, and if it wasn't set, we'll use `ring` for default provider.

WDYT @jaymell @djc ?

GlenDC:
Whatever the case... for Rustls please use aws-lc-rs as the default...

And I'm fine in going for a different approach once you align on what that might be.. Just let me know and I'll go for it. Might need guidance though if it's different than this.

Honestly I couldn't care less about aws-lc-rs... however... rustls choose it as default (instead of ring)...
That choice of rustls basically forces you as well to make that default because otherwise complex enough rust
projects get into a scenario where they have to fork libraries like this one in order to make it compile... (which is my scenario already more times than I dare to count... It's a mess, but it becomes less of a mess if the defaults just align...)

GlenDC:
To put some context, in case it comes as news. When `redis` crate was made, `ring` was the default crypto provider, and I think even the only one. However at some point this year (I think, or was it last year, cannot remember), `aws-lc-rs` was promoted to being the default rustls crypto provider...

In that context, an even simpler change would be to _only_ use `aws-lc-rs` in rustls features of redis, as that is anyway the default, and if you combine it in a project with other dependencies that also make use of `rustls` you anyway get in a scenario where dependencies just use the default (aws-lc-rs) and you have no choice but to use it given they might not even expose a feature to tweak the crypto provider... So honestly my PR here does already more than most libs even allow you to do.

Also. If you are ever open to it I would even be willing (in a separate ticket and PR) to help make redis sans-io compatible. Because while there is value in providing this transport layer out of the box, it does mean that in a project with its own tcp / tls stack, you now have yet another stack in the same process... While redis could just operate sans-io at its core, and only bring the tcp/tls parts when the features are enabled, requiring you to bring your own transport otherwise.

djc:
> I would like to avoid the scenario where every crypto provider supported by rustls adds another feature to redis. Instead maybe we should add to `TlsConnParams` a field with a dyn crypto provider, or maybe even just the `WebPkiSupportedAlgorithms` that we get from `signature_verification_algorithms`? We wrap it in an option and add it using the builder pattern, and if it wasn't set, we'll use `ring` for default provider.
> 
> WDYT @jaymell @djc ?

Do you want this change to be semver-compatible or not? If you do, IMO you can't change the crypto provider default for your downstream users so the default should be the same. Adding an `Arc<CryptoProvider>` to the `TlsConnParams` seems like a reasonable way forward to me, but your plan of setting a default doesn't quite work if you want to give downstream users control over which dependencies are pulled in (otherwise you will at least need some way of disabling the default crypto provider dependency).

GlenDC:
Honestly the more I think about it the more I come to conclude that libraries such as redis should really be sans-io. They can still provide an implementation in a module when feature gated of course.

Would require a new major version though.

can the maintainers discuss this idea and approach internally. If you agree I don’t mind spear diving this endeavour as a new PR and close this one.

Also any pointers, ideas and hint of possible skeletons would be greatly appreciated as I’m still a foreigner in this codebase. But even if all you do is give me the greenlight, that is fine as well.

nihohit:
Thanks for the suggestions!
I've never designed nor implemented a sans-io library, and even though I like the idea of it, I can't say whether it's something that's worth a new major version, or how much of an interface change it will require. In any case I wouldn't close this PR, since I assume that any such redesign would take a long while, even just the research & discussion part, so why not have a short-term fix while we chew this over?

> Adding an Arc<CryptoProvider> to the TlsConnParams seems like a reasonable way forward to me, but your plan of setting a default doesn't quite work if you want to give downstream users control over which dependencies are pulled in (otherwise you will at least need some way of disabling the default crypto provider dependency).

Since the dependencies only affect compilation time, I think that this is the right direction for us. Not breaking semver at the cost  of forcing our consumers to pull some dependencies sounds worth it.

> Also any pointers, ideas and hint of possible skeletons would be greatly appreciated as I’m still a foreigner in this codebase. But even if all you do is give me the greenlight, that is fine as well.

If you'd like to give the codebase a look and draft a rough design, that would be great. I would start by looking at the functions for connection creation - 

https://github.com/redis-rs/redis-rs/blob/d4f9d0a17294a4a9ca7bd8a15f28c9540c5c790c/redis/src/connection.rs#L579
https://github.com/redis-rs/redis-rs/blob/d4f9d0a17294a4a9ca7bd8a15f28c9540c5c790c/redis/src/aio/connection.rs#L451

Consider creating a dedicated issue for the discussion on implementing sans-io. I would prefer that the discussion here would center on a more immediate fix.

kamulos:
I am a bit worried about `ring` at the moment: in the last days it went to unmaintained status, changed hands, changed hands back.

What I would like of `redis` is just to get out of the way and not force `ring` on me. `rustls` will have the `aws-lc` feature enabled anyways in my workspace, so there is no need for `redis` to have an opinion on that matter.

Would it make sense to go the `reqwest` route and leave everything the same, just add a `-no-provider` feature, that enables neither crypto backend and leaves the responsibility to the user to have something enabled?

nihohit:
Sure @kamulos, that sounds reasonable.

nihohit:
@GlenDC can you please check if https://github.com/redis-rs/redis-rs/pull/1632 solves your issue?

nihohit:
Ok, then I'll close this PR, and I'll try to find time for https://github.com/redis-rs/redis-rs/pull/1514.

PR #1346, opened 2024-09-27
Further limit parser recursion

jaymell:
Ensure we enforce max recursion depth across all types, not just arrays.


nihohit (approved):

nihohit:
was this caught in real life or the fuzzer?

PR #1360, opened 2024-10-03
Add lastid option to xclaim

urkle:
Adds the lastid option to xclaim

https://redis.io/docs/latest/commands/xclaim/

This is one of the items in #1153

Behavior of LASTID is documented here:
https://github.com/redis/redis/blob/unstable/src/t_stream.c#L3123

(docs are missing on official redis docs https://redis.io/docs/latest/commands/xclaim/ )

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1563,7 +1563,7 @@ implement_commands! {
     /// ```text
     /// XCLAIM <key> <group> <consumer> <min-idle-time> <ID-1> <ID-2>
     ///     [IDLE <milliseconds>] [TIME <mstime>] [RETRYCOUNT <count>]
-    ///     [FORCE] [JUSTID]
+    ///     [FORCE] [JUSTID] [LASTID <lastid>]
     /// ```
     #[cfg(feature = "streams")]
     #[cfg_attr(docsrs, doc(cfg(feature = "streams")))]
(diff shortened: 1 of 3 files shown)


nihohit:
is the last id necessarilly a string? if not, let's make it a `Vec<u8>`, or other byte representation.

urkle:
@nihohit it is a stream-id so the same data type as other "stream" IDs. (two u64 numbers separated by -).

I can actually change the lastid fn to accept impl Into<String> or even impl ToRedisArgs as we have other "ID" params. Plus String is used for Stream IDs in response objects for the stream API.

So I can do this (keeping lastid an Option<String> but maybe we add docs clarifying the format?
```rust
    /// Set `LASTID <lastid>` cmd arg.
    pub fn with_lastid(mut self, lastid: impl Into<String>) -> Self {
        self.lastid = Some(lastid.into());
        self
    }
```

If we make it ToRedisArgs though it could end up accepting multiple arguments in. (which is actually something that can happen in lots of places right now)..  e.g. we don't have a ToRedisArg  single arg enforcement anywhere.

e.g. lastid would have to be a Vec<Vec<u8>> OR I have to add error handling somewhere if multiple args are passed through.

Let me know which direction to go on this.. As this will also impact #1361 which also has an options builder that accepts a stream ID (for the MINID trim option)

nihohit:
yeah, looks like `impl Into<String>` is the best option.

urkle:
@nihohit ok.. I have adjusted the code to use impl Into<String>

nihohit (approved):

nihohit:
Thanks for the contribution!

PR #1361, opened 2024-10-05
Add xadd_options and xtrim_options

urkle:
- these support the full command set for the commands

- nomkstream on xadd
- minid/limit for trimming

This implements more functionality for #1153

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1423,17 +1423,29 @@ implement_commands! {
 
     /// Add a stream message with options.
     ///
+    /// Items can be any list type, e.g.
+    /// ```rust
+    /// // static items
+    /// let items = &[("key", "val"), ("key2", "val2")];
+    /// # use std::collections::BTreeMap;
+    /// // A map (Can be BTreeMap, HashMap, etc)
+    /// let mut map: BTreeMap<&str, &str> = BTreeMap::new();
+    /// map.insert("ab", "cd");
+    /// map.insert("ef", "gh");
+    /// map.insert("ij", "kl");
+    /// ```
+    ///
     /// ```text
     /// XADD key [NOMKSTREAM] [<MAXLEN|MINID> [~|=] threshold [LIMIT count]] <* | ID> field value [field value] ...
     /// ```
     #[cfg(feature = "streams")]
     #[cfg_attr(docsrs, doc(cfg(feature = "streams")))]
     fn xadd_options<
-        K: ToRedisArgs, ID: ToRedisArgs, F: ToRedisArgs, V: ToRedisArgs
+        K: ToRedisArgs, ID: ToRedisArgs, I: ToRedisArgs
     >(
         key: K,
         id: ID,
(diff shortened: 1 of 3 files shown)


redis/tests/test_streams.rs
```diff
@@ -421,6 +421,62 @@ fn test_xadd_maxlen_map() {
...
+    let mut con = ctx.connection();
+
+    // NoMKStream will cause this command to fail
+    let result: RedisResult<String> = con.xadd_options(
+        "k1",
+        "*",
+        &[("h", "w")],
+        &StreamAddOptions::default().nomkstream(),
+    );
+    assert!(result.is_err());
```
nihohit, line 436: can you assert on the error kind?

redis/src/commands/mod.rs
```diff
@@ -1420,6 +1420,53 @@ implement_commands! {
...
 
+
+    /// Add a stream message with options.
+    ///
+    /// ```text
+    /// XADD key [NOMKSTREAM] [<MAXLEN|MINID> [~|=] threshold [LIMIT count]] <* | ID> field value [field value] ...
+    /// ```
+    #[cfg(feature = "streams")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "streams")))]
+    fn xadd_options<
```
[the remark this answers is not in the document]
urkle: I was following the existing pattern.

but we probably do not need both at all do we.  Let me add more conclusive tests for that scenario (Sending a hashmap in and the manual item list).

redis/tests/test_streams.rs
urkle: Sure thing!

redis/src/commands/mod.rs
```diff
@@ -1420,6 +1420,53 @@ implement_commands! {
...
+        K: ToRedisArgs, ID: ToRedisArgs, F: ToRedisArgs, V: ToRedisArgs
+    >(
+        key: K,
+        id: ID,
+        items: &'a [(F, V)],
+        options: &'a streams::StreamAddOptions
```
urkle, line 1437: @nihohit I have a question about this here.   currently there is an inconsistency in how options params are passed in. Sometimes they are passed by ref, and other times they are passed as owned.  What is the preference for the new methods I am adding in here (and what the API should morph to going forward)?

(examples of fns using owned objects for options are geo_radius* or xclaim_options)

redis/src/commands/mod.rs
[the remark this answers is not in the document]
nihohit: I think that was for historical reasons - we didn't have a simple way to use slices until recently.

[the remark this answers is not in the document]
urkle: I figured,  I de-duplicated the two functions so we have only `xadd_options`

redis/src/commands/mod.rs
nihohit: Let's use references here. I don't know why some options are owned 🤷

nihohit (approved):

PR #1363, opened 2024-10-07
Move server creation logic to redis-test

nihohit:
This change allows users to easily create local redis servers, for use in testing their own modules.

This stems from https://github.com/redis-rs/redis-rs/issues/1164 - this change doesn't solve the entire request, but it's a first step towards it.


nihohit:
@altanozlu WDYT?

altanozlu (approved):
I see no harm in this patch, although didn't test it locally. 
CI looks good.

PR #1372, opened 2024-10-14
Allow disabling timeouts in `ConnectionManager`

dcreager:
The connection and response timeouts in `MultiplexedConnection` are truly optional: they are wrapped in `Option`, and if they are `None`, the runtime's `timeout` function is never called at all.

`ConnectionManager` currently mimics an optional timeout by using `Duration::MAX` as the default timeout value.  The runtime's `timeout` function is called, but with an incredibly large duration.

In normal operation, this would not be a problem, since the timeouts will not fire unless the process runs for ~half a trillion years.  But it breaks test cases that use Tokio's [pausable clock][].  If the task invoking Redis is parked (e.g. while waiting for the Redis command to complete), Tokio will [auto-advance the clock][] to the next sleep point, allowing the timeout to fire.

This patch updates the `ConnectionManager` timeouts to be consistent with the `MultiplexedConnection` timeouts, so that the runtime's `timeout` function is only called if you set connection and/or response timeouts.

[pausable clock]: https://docs.rs/tokio/latest/tokio/time/fn.pause.html
[auto-advance the clock]: https://thomask.sdf.org/blog/2022/12/28/auto-advance-vs-time-guards.html

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -134,7 +134,7 @@ futures-time = "3"
 criterion = "0.4"
 partial-io = { version = "0.5", features = ["tokio", "quickcheck1"] }
 quickcheck = "1.0.3"
-tokio = { version = "1", features = ["rt", "macros", "rt-multi-thread", "time"] }
+tokio = { version = "1", features = ["rt", "macros", "rt-multi-thread", "test-util", "time"] }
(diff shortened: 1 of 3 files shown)


nihohit (approved):
Thanks for the contribution!

PR #1373, opened 2024-10-14
implement command `DRYRUN`

ArtemIsmagilov:
- doc valkey command https://valkey.io/commands/acl-dryrun/
- doc redis command https://redis.io/docs/latest/commands/acl-dryrun/

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1112,7 +1112,7 @@ implement_commands! {
     /// Simulate the execution of a given command by a given user.
     #[cfg(feature = "acl")]
     #[cfg_attr(docsrs, doc(cfg(feature = "acl")))]
-    fn acl_dryrun<K: ToRedisArgs, C: ToRedisArgs, A: ToRedisArgs>(username: K, command: C, args: &'a [A]) {
+    fn acl_dryrun<K: ToRedisArgs, C: ToRedisArgs, A: ToRedisArgs>(username: K, command: C, args: A) {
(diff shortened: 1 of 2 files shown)


redis/src/commands/mod.rs
```diff
@@ -1109,6 +1109,13 @@ implement_commands! {
         cmd("ACL").arg("DELUSER").arg(usernames)
     }
 
+    /// Simulate the execution of a given command by a given user.
+    #[cfg(feature = "acl")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "acl")))]
+    fn acl_dryrun<P: ToRedisArgs>(username: P, command: P, args: &'a [P]) {
```
nihohit, line 1115: please give separate types for the arguments, to allow the user more freedom, and make sure they all implement ToRedisArgs

ArtemIsmagilov: Thank you for commenting on my draft. I'll fix it now.

ArtemIsmagilov: There may be one problem here though. Your Github Action checks for redis versions 7 and 6. However, this command is available from version 7. Well, let me correct the command first.

nihohit: You can check the db's version and exit realty if it doesn't support the command. I'm not next to a computer right now, but look for is_major_version in tests

ArtemIsmagilov: Thanks!

nihohit (approved):

PR #1374, opened 2024-10-16
implement high level command `AUTH`

ArtemIsmagilov:
Hi. High-level authentication command is implemented. True, you have authentication at the moment of connection creation and that's great! I also want to be able to authenticate by a reserved user.
AUTH - https://valkey.io/commands/auth/
Test example - https://valkey.io/topics/acl/

### DIFF
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -2058,4 +2058,38 @@ mod basic {
         let mut con = ctx.connection();
         let _num: i64 = con.client_id().unwrap();
     }
+
+    #[test]
+    fn test_auth_command() {
+        let ctx = TestContext::new();
+        let mut con = ctx.connection();
+
+        redis::cmd("ACL")
+            .arg("SETUSER")
+            .arg("alice")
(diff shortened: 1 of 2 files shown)


redis/tests/test_basic.rs
```diff
@@ -2015,4 +2015,23 @@ mod basic {
...
+        redis::cmd("ACL")
+            .arg("SETUSER")
+            .arg("alice")
+            .arg("on")
+            .arg(">p1pp0")
+            .arg("+ACL")
+            .exec(&mut con)
+            .unwrap();
+
+        assert_eq!(con.auth(Some("alice"), "p1pp0"), Ok(()));
```
nihohit, line 2033: please also test the None user scenario.

ArtemIsmagilov: Done

nihohit: Using the `None` option is cumbersome, and requires adding a meaningless type. This requires a more ergonomic solution.

PR #1377, opened 2024-10-16
impl high level command `ECHO`

ArtemIsmagilov:
redis/valkey https://valkey.io/commands/echo/


nihohit:
why do we need to add this, if we have PING?

ArtemIsmagilov:
I'm just going by protocol. You can cancel this. The teams are really similar in the result.

nihohit:
Let's not add commands without a concrete usecase.

PR #1378, opened 2024-10-16
impl CLIENT HELP

ArtemIsmagilov:
https://valkey.io/commands/client-help/


nihohit:
why would someone need this command in a redis client? what's the usage scenario?

ArtemIsmagilov:
Hi, thanks a lot for the hints at the beginning. I plan to implement all possible commands from redis/valkey for your library, if you allow.

Despite the fact that I practice in Rust, nevertheless, the commands implement more or less the necessary logic for applications)

Perhaps many commands will never be used, but why don't we make them in advance before we need them? This is a rhetorical question)

PR #1380, opened 2024-10-17
impl CLIENT GETNAME

ArtemIsmagilov:
doc link https://valkey.io/commands/client-getname/


nihohit (approved):

ArtemIsmagilov:
It seems we have problems with formatting check. It would be nice to fix this. I'm not sure if this should go into the main branch after errors. I'll try to fix the formatting if possible.

Thanks again for the support and help.

PR #1383, opened 2024-10-17
impl QUIT

ArtemIsmagilov:
deprecated, but not deleted `quit`. https://valkey.io/commands/quit/

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1044,6 +1044,11 @@ implement_commands! {
         cmd("OBJECT").arg("REFCOUNT").arg(key)
     }
 
+    /// Ask the server to close the connection.
+    fn quit<>() {
+        Cmd::new().arg("QUIT")
+    }
+
     // ACL commands
 
     /// When Redis is configured to use an ACL file (with the aclfile
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -2015,4 +2015,12 @@ mod basic {
             .unwrap();
         assert!(info.contains("db=5"));
     }
+
+    #[test]
+    fn test_quit() {
+        let ctx = TestContext::new();
+        let mut con = ctx.connection();
+        let _: () = con.quit().unwrap();
+        assert_eq!(con.check_connection(), false);
+    }
 }


nihohit:
"Note: Clients should not use this command. Instead, clients should simply close the connection when they're not used anymore. "

What's the usage scenario for this?

ArtemIsmagilov:
The goal is simple, to provide all possible commands that are available)

Sometimes it makes sense to close the connection explicitly for clarity. Why not?)

However, if you think that there are commands that are not worth implementing in your library, please write back.

ArtemIsmagilov:
I would like to add that you have implemented a lot of data handling, but not all the general commands. And in general, the answer is the same for all new commands - I want to add all the features at a high level.

It's nice that this is done in the simplest way. And even at a low level with cmd, it is also easy to do!

For example, the Python client seems to actually implement all the commands. It's cool when a library can provide everything you need to work with it)

I would like to add that you have implemented a lot of data handling, but not all the general commands. And in general, the answer is the same for all new commands - I want to add all the capabilities at a high level.

It's nice that this is done in the simplest way. And even at a low level with cmd, it is also easy to do!

For example, the Python client seems to actually implement all the commands. It's cool when a library can provide everything you need to work with it)

However, as the owner of the repository, can you warn me which commands you consider not worth implementing?

ArtemIsmagilov:
In python I have the ability to close a file via close, via the context manager, or not close it at all and the interpreter will close it itself.

This is close to me. However, it needs to be supported and tested. If you think that this approach is unnecessary, then you are also right. As the author of this library, the decision is yours. I am simply implementing the protocol.

nihohit:
> It's cool when a library can provide everything you need to work with it)

do you need the QUIT command to work with the library? I'm pretty sure that it's a command that was meant to be used by library implementors, and now is deprecated - so what's the concrete usage here?

please don't open PRs for commands that you don't see a concrete usage for. If you want to look at commands that would be helpful to implement, you can start with issues about missing commands.

ArtemIsmagilov:
Ok, I understand you.

ArtemIsmagilov:
I can understand you. However, the official python client supports this command, even if it is not recommended.
https://github.com/redis/redis-py/blob/d6ddb0d66bc4bd28c99d9cf148c9085117cc8ce9/redis/commands/core.py#L1214

PR #1384, opened 2024-10-17
impl RESET

ArtemIsmagilov:
https://valkey.io/commands/reset/

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1044,6 +1044,12 @@ implement_commands! {
         cmd("OBJECT").arg("REFCOUNT").arg(key)
     }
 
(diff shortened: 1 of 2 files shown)


nihohit:
This is a dangerous command, that allows the user to reset things that are tracked in the client side - for example, this would break the in-progress client-side caching implementation.
Why do you need this?

ArtemIsmagilov:
I just implement the commands that are specified in the documentation. I think that the drop table commands are also dangerous, but they exist and they are necessary.
Let me finish the high-level commands, I'm a Rust enthusiast now)

nihohit:
Sure, please add commands that are necessary. When/ Why is this command necessary?

PR #1385, opened 2024-10-18
feat: new API `query_raw` and `query_v2` to handle partial responses …

Abioy:
…from a pipeline

Pre-discussion can be found here https://github.com/redis-rs/redis-rs/pull/1057#issuecomment-2411069879

### DIFF
--- a/redis-test/src/lib.rs
+++ b/redis-test/src/lib.rs
@@ -245,6 +245,40 @@ impl ConnectionLike for MockRedisConnection {
         next_cmd.responses
     }
 
+    fn req_packed_commands_raw(
+        &mut self,
+        cmd: &[u8],
+        _offset: usize,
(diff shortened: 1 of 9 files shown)


nihohit:
Thanks for the contribution!
Can you please explain here or in comments why are the changes in the `types` file needed?
And this change requires significantly more documentation and tests, in order to properly convey when the additional API is required. 
Additionally, what's the expected behavior for transactions? Does the new API even make sense for use with transactions?

Abioy:
> Can you please explain here or in comments why are the changes in the `types` file needed?

nihohit:
Hi, I think that this change might be relevant for the `1.0.x` branch - its the branch for backwards incompatible changes.

nihohit:

PR #1387, opened 2024-10-18
suggestion to use default fmt

ArtemIsmagilov:
I would like to suggest using formatting by default. 
It is a pity that you can't format macros yet. This would help make macros clearer.

### DIFF
--- a/redis/src/cluster_async/routing.rs
+++ b/redis/src/cluster_async/routing.rs
@@ -144,8 +144,10 @@ mod pipeline_routing_tests {
         pipeline
             .get("foo") // route to replica of slot 12182
             .add_command(cmd("FLUSHALL")) // route to all masters
-            .add_command(cmd("EVAL"))// route randomly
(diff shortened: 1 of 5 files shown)


redis/src/cluster_async/routing.rs
```diff
@@ -144,8 +144,10 @@ mod pipeline_routing_tests {
         pipeline
             .get("foo") // route to replica of slot 12182
             .add_command(cmd("FLUSHALL")) // route to all masters
-            .add_command(cmd("EVAL"))// route randomly
-            .cmd("CONFIG").arg("GET").arg("timeout") // unkeyed command
+            .add_command(cmd("EVAL")) // route randomly
+            .cmd("CONFIG")
```
nihohit, line 148: adding a command per-line in pipelines is clearer than this, which doesn't separate commands from args. Is there a way to prevent this?

ArtemIsmagilov: Of course, you can add parameters to the config. I will finish this PR.

PR #1391, opened 2024-10-21
impl command SELECT

ArtemIsmagilov:
I want to select logical databases at a high level. The documentation for the command can be found here.
- https://valkey.io/commands/select/
- https://redis.io/docs/latest/commands/select/

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -169,6 +169,11 @@ implement_commands! {
         cmd("MSET").arg(items)
     }
 
+    /// Select the logical database having the specified zero-based numeric index.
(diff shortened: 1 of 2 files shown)


nihohit:
Listen, I appreciate the effort that you're making, but please - focus on commands that aren't already covered by the logic of the crate, and that don't break existing code when used (this breaks the `get_db` function). As I said before, you can start with commands that have outstanding issues requesting them.

ArtemIsmagilov:
Sorry for taking up your time with inappropriate functionality. Thank you for this review, I'm starting to understand the problems with my straightforward solution.

PR #1407, opened 2024-11-01
ConnectionManager: reconnect on disconnect pushes.

nihohit:
This change allows the ConnectionManager to reconnect not only when receiving errors on requests, but also when receiving disconnect pushes on the push channel from the internal connection.

Based over https://github.com/redis-rs/redis-rs/pull/1392

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -1,18 +1,20 @@
-use super::{AsyncPushSender, RedisFuture};
+use super::{AsyncPushSender, RedisFuture, SharedHandleContainer};
 use crate::{
     aio::{check_resp3, ConnectionLike, MultiplexedConnection, Runtime},
     cmd,
     types::{RedisError, RedisResult, Value},
-    AsyncConnectionConfig, Client, Cmd, ToRedisArgs,
+    AsyncConnectionConfig, Client, Cmd, PushInfo, PushKind, ToRedisArgs,
 };
 use arc_swap::ArcSwap;
 use backon::{ExponentialBuilder, Retryable};
 use futures::{
+    channel::oneshot,
     future::{self, Shared},
     FutureExt,
 };
 use futures_util::future::BoxFuture;
 use std::sync::Arc;
+use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
 
 /// The configuration for reconnect mechanism and request timing for the [ConnectionManager]
 #[derive(Clone)]
@@ -127,6 +129,9 @@ impl ConnectionManagerConfig {
     ///
     /// The sender can be a channel, or an arbitrary function that handles [crate::PushInfo] values.
     /// This will fail client creation if the connection isn't configured for RESP3 communications via the [crate::RedisConnectionInfo::protocol] field.
+    /// Setting this will mean that the connection manager actively listens to updates from the
+    /// server, and so it will cause the manager to reconnect after a disconnection, even if the manager was unused at
+    /// the time of the disconnect.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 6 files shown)


nihohit:
@avifenesh, another one.

altanozlu:
Currently when there is an error on CM (Connection Manager), user will know about it but by receiving an error, with this change when push sender is given user might get an error or not, depending on how fast it triggers re-connect. That behaviour might is flaky and users might rely on wrong things.
If this is okay then I would like to ask, is it possible to handle it without using PushInfo::Disconnection so there will be no need for user setting up push sender for this feature and it can be the default behaviour.

nihohit:
> when push sender is given user might get an error or not, depending on how fast it triggers re-connect. That behaviour might is flaky and users might rely on wrong things.

Can you give an example of a potentially problematic chain of events?

> If this is okay then I would like to ask, is it possible to handle it without using PushInfo::Disconnection so there will be no need for user setting up push sender for this feature and it can be the default behaviour.

You're asking whether the connection manager can automatically reconnect without setting a push manager?

altanozlu:
> > when push sender is given user might get an error or not, depending on how fast it triggers re-connect. That behaviour might is flaky and users might rely on wrong things.
> 
> Can you give an example of a potentially problematic chain of events?
> 
Testing things in local and somehow it always reconnect and nothing is noticed
I can't come up with one now, but auto reconnect is best effort and if user sends a command when auto reconnect didn't happen yet, command will result in an error, maybe it should be documented somewhere ?

> > If this is okay then I would like to ask, is it possible to handle it without using PushInfo::Disconnection so there will be no need for user setting up push sender for this feature and it can be the default behaviour.
> 
> You're asking whether the connection manager can automatically reconnect without setting a push manager?

I wanted to know if it's feasible to have reconnections without user providing push sender.

nihohit:
> I wanted to know if it's feasible to have reconnections without user providing push sender.

Well, I pushed a commit doing that, but it seems to fail for async_std runtime, when using RusTLS. not sure why.

altanozlu:
> > I wanted to know if it's feasible to have reconnections without user providing push sender.
> 
> Well, I pushed a commit doing that, but it seems to fail for async_std runtime, when using RusTLS. not sure why.

It's something related to tls, for me tokio also hangs with:
```
RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp+tls RUST_BACKTRACE=1 cargo test basic_async::manager_should_reconnect_without_actions_if_push_sender_is_set::case_1_tokio --locked -p redis --all-feature
```
which is weird because test shouldn't even run something important because there is `if ctx.protocol == ProtocolVersion::RESP2 {` check

as far as I see `TestContext::new` hangs

nihohit:
> as far as I see TestContext::new hangs

That would be weird, because this doesn't hang in any other test.

nihohit:
in any case, this doesn't reproduce locally, but seems consistent in CI. annoying :(

nihohit:
@altanozlu tests seem to pass now.

redis/src/aio/connection_manager.rs
```diff
@@ -454,8 +511,9 @@ impl ConnectionManager {
     pub async fn subscribe(&mut self, channel_name: impl ToRedisArgs) -> RedisResult<()> {
         check_resp3!(self.client.connection_info.redis.protocol);
         let mut cmd = cmd("SUBSCRIBE");
-        cmd.arg(channel_name);
+        cmd.arg(&channel_name);
         cmd.exec_async(self).await?;
+
```
altanozlu, line 516: nitpick: unnecessary new line

altanozlu (approved):
LGTM except one little newline which you can dismiss 🙌🏻

PR #1408, opened 2024-11-02
ConnectionManager: Support automatic resubscription.

nihohit:
This change adds a configuration to allow the ConnectionManager to automatically resubscribe on pubsub channels after reconnection.
This is left as an opt-in configuration in order not to break backwards-compatibility.

Based over https://github.com/redis-rs/redis-rs/pull/1407 & https://github.com/redis-rs/redis-rs/pull/1392

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -39,7 +39,7 @@ pub struct ConnectionManagerConfig {
     /// sender channel for push values
     push_sender: Option<Arc<dyn AsyncPushSender>>,
     /// if true, the manager should resubscribe automatically to all pubsub channels after reconnect.
-    resubsrcibe_automatically: bool,
+    resubscribe_automatically: bool,
 }
 
 impl std::fmt::Debug for ConnectionManagerConfig {
@@ -52,7 +52,7 @@ impl std::fmt::Debug for ConnectionManagerConfig {
             response_timeout,
             connection_timeout,
             push_sender,
-            resubsrcibe_automatically,
+            resubscribe_automatically,
         } = &self;
         f.debug_struct("ConnectionManagerConfig")
             .field("exponent_base", &exponent_base)
@@ -61,7 +61,7 @@ impl std::fmt::Debug for ConnectionManagerConfig {
             .field("max_delay", &max_delay)
             .field("response_timeout", &response_timeout)
             .field("connection_timeout", &connection_timeout)
-            .field("resubsrcibe_automatically", &resubsrcibe_automatically)
+            .field("resubscribe_automatically", &resubscribe_automatically)
             .field(
                 "push_sender",
                 if push_sender.is_some() {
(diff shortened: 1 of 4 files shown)


redis/src/aio/connection_manager.rs
```diff
@@ -271,10 +289,17 @@ impl ConnectionManager {
         // Create a MultiplexedConnection and wait for it to be established
         let runtime = Runtime::locate();
 
-        let mut retry_strategy =
-            ExponentialBackoff::from_millis(config.exponent_base).factor(config.factor);
+        if config.resubsrcibe_automatically && config.push_sender.is_none() {
+            return Err((crate::ErrorKind::ClientError, "Cannot set resubsrcibe_automatically without setting a push sender to receive messages.").into());
```
altanozlu, line 293: nitpick: resubsrcibe_automatically
```suggestion
            return Err((crate::ErrorKind::ClientError, "Cannot set resubscribe_automatically without setting a push sender to receive messages.").into());
```

redis/src/aio/connection_manager.rs
```diff
@@ -37,6 +38,8 @@ pub struct ConnectionManagerConfig {
     connection_timeout: Option<std::time::Duration>,
     /// sender channel for push values
     push_sender: Option<AsyncPushSender>,
+    /// if true, the manager should resubscribe automatically to all pubsub channels after reconnect.
+    resubsrcibe_automatically: bool,
```
altanozlu, line 42: same here

redis/src/aio/connection_manager.rs
```diff
@@ -113,10 +127,27 @@ impl Default for ConnectionManagerConfig {
...
+    /// The connection future.
+    ///
+    /// The `ArcSwap` is required to be able to replace the connection
+    /// without making the `ConnectionManager` mutable.
+    connection: ArcSwap<SharedRedisFuture<MultiplexedConnection>>,
+
+    runtime: Runtime,
+    retry_strategy: ExponentialBuilder,
+    connection_config: AsyncConnectionConfig,
+    subscription_tracker: Option<Mutex<SubscriptionTracker>>,
```
altanozlu, line 147: As far as I see it's also possible to use standard mutex, there is no need to hold guard across await points.



ref: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

nihohit: true, but I've learned that standard mutexes could stil cause a runtime to slow down, if they're stopping multiple tasks.

altanozlu: Oh, I didn't know can you point me to a reference, if you can reach it easily ?

altanozlu: (feel free to resolve the review)

altanozlu (approved):

redis/src/aio/connection_manager.rs
nihohit: it's a very simple issue - instead of yielding to another task, the runtime's thread is blocked on the mutex. , thus blocking all tasks on that thread. while I believe this isn't an issue here, since the mutex would be held very shortly, I prefer not to optimize in a way that is potentially harmful to the users.

Issue #1043, opened 2024-01-30
Backwards incompatible changes to consider


nihohit:
The purpose of this list is to raise major design decisions which we should consider for a major backwards incompatible version, such as 1.0. Everyone is welcome to comment on these, or offer more suggestions.

## Connections:
- [x] Reduce `ConnectionInfo`'s visibility - this hampers our ability to add fields with structs that we don't want to be public. We could expose a builder, instead, or  rely on redis URLs. Sample issue - https://github.com/redis-rs/redis-rs/issues/1006
- [ ] async connections that work with shared references -
https://github.com/redis-rs/redis-rs/issues/1236#issuecomment-2194827216

## Parsing and values:
- [x] Make `from_redis_value` take the value, instead of a reference. Today we have to maintain both `from_redis_value` and `from_owned_redis_value`. We should remove `from_owned_redis_value`, have `from_redis_value` take an owned value. and reduce the duplication. If the user wants to convert by reference, they can just clone the value.
- [ ] Similarly, `redis::Value` can contain better clonable values - replace `String` with `Arc<String>` or [`ArcStr`](https://crates.io/crates/arcstr), or maybe just `Bytes` objects. This will reduce the cost of cloning values, and if we replace `String` and `Vec<u8>` with `Bytes` it can even allow for zero-copy parsing.
- [ ] Similarly, consider using [redis-protocol](https://crates.io/crates/redis-protocol) for parsing, value. It looks like it allows for zero-copy parsing already.

## Async
- [x] ~Currently the library supports both async-std and tokio as runtimes, and allows compilation with both features. We should consider whether we want to go with full runtime agnostic and also support `smol`.~ (implemented in https://github.com/redis-rs/redis-rs/pull/1606)
- [x] ~Alternatively, if there's no adoption of `async-std`, we can drop that support and reduce the async implementation complexity.~ (marked as deprecated in https://github.com/redis-rs/redis-rs/pull/1606)
- [ ] Orthogonally, we should consider whether we allow compiling with both the async-std and tokio features, or whether these features should be mutually exclusive.

kamulos:
Some things we encountered in the past:

- For Connections like `MultiplexedConnection` or `cluster_async::ClusterConnection` it would be great if executing a command took `&self` instead of `&mut self`. They should have interior mutability anyway (?). This was previously discussed in #638 .

- ~There is no good way to handle pipelines where some of the commands succeed and others fail. This was previously discussed in #813 and #746 .~

[...]

Issue #773, opened 2023-02-07
Change (async) iterator item to Result<T>


PhotonQuantum:
Almost all (async) iterators returned by this crate silently ignore errors and yield None, terminating the iteration. This can make it difficult to handle errors robustly, and can confuse users into thinking that the iterator has finished successfully when in fact it has not. This has already caused problems such as https://github.com/redis-rs/redis-rs/issues/66 and https://github.com/redis-rs/redis-rs/issues/759.

I suggest changing the item yielded by the iterator to `Result<T>` instead of `T`. This change would mean that the consumer of the iterator could easily handle error cases and decide whether to continue the iteration.

The `Iter` and `AsyncIter` structs need to be updated, and there are  `aio::PubSub::on_message` and `aio::PubSub::into_on_message`. However, this is not an exhaustive list, and there may be other places that need to be changed.

Finally, the proposed change alters the public API and is a breaking one. A version bump is required and we should inform users of this change in the release note.

ckcr4lyf:
I think this is pretty useful, I was super confused why my iteration was ending so early. Turns out I had assumed type `String`, but some of my keys have raw bytes. 

Pulled out wireshark and was able to figure out that must be it!

PR #1416, opened 2024-11-17
Make iterator return an Option<RedisResult>

ckcr4lyf:
This is an example PR of how it would look like if the iterator returned a `RedisResult`, instead of silently failing and returning None if it failed to parse the next set of keys.

I faced this when I was trying to iterate over the keyspace of my DB to perform a migration, but the target DB had way less keys! I realized that the iterator was incorrectly "ending" early.

For me the current workaround is to just use `Vec<u8>` instead of `String` since in redis everything should be able to be represented as raw bytes, especially keys.

Hopefully this can help prevent other application logic bugs.

Some related issues:
* https://github.com/redis-rs/redis-rs/issues/773
* https://github.com/redis-rs/redis-rs/issues/66

This is obviously backwards incompatible, as noted in: https://github.com/redis-rs/redis-rs/issues/1043#issuecomment-2367507019

I tried to make a unit test to illustrate how this would work, and had to patch some existing tests as well. I'm new to rust, so my style may not be the most idiomatic :-)

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -88,6 +88,37 @@ pub struct Cmd {
 
 /// Represents a redis iterator.
 pub struct Iter<'a, T: FromRedisValue> {
+    checked_iter: CheckedIter<'a, T>,
+}
+
+impl<T: FromRedisValue> Iterator for Iter<'_, T> {
+    type Item = T;
+
+    #[inline]
+    fn next(&mut self) -> Option<T> {
+        // we need to do this in a loop until we produce at least one item
+        // or we find the actual end of the iteration.  This is necessary
+        // because with filtering an iterator it is possible that a whole
+        // chunk is not matching the pattern and thus yielding empty results.
+        loop {
+            // iterate over the CheckedIter, but keep existing behavior, i.e.
+            // if there is an error, just silently return `None`
+            if let Some(v) = self.checked_iter.next() {
+                if let Ok(v) = v {
+                    return Some(v);
+                }
+
+                return None;
+            };
+
+            // nothing left in the checked_iter
+            return None;
+        }
+    }
+}
+
+/// Represents a safe(r) redis iterator.
+pub struct CheckedIter<'a, T: FromRedisValue> {
     batch: std::vec::IntoIter<Value>,
     cursor: u64,
     con: &'a mut (dyn ConnectionLike + 'a),
@@ -95,7 +126,7 @@ pub struct Iter<'a, T: FromRedisValue> {
     _phantom: PhantomData<T>,
 }
 
-impl<T: FromRedisValue> Iterator for Iter<'_, T> {
+impl<T: FromRedisValue> Iterator for CheckedIter<'_, T> {
     type Item = RedisResult<T>;
 
     #[inline]
@@ -542,6 +573,18 @@ impl Cmd {
     /// tuple of cursor and list).
     #[inline]
     pub fn iter<T: FromRedisValue>(self, con: &mut dyn ConnectionLike) -> RedisResult<Iter<'_, T>> {
+        Ok(Iter {
+            checked_iter: self.checked_iter(con)?,
+        })
+    }
+
+    /// Similar to `iter()` but does not silently fail and return None if a value can't be parsed
+    /// (i.e. allows iterating to next possible value)
+    #[inline]
+    pub fn checked_iter<T: FromRedisValue>(
+        self,
+        con: &mut dyn ConnectionLike,
+    ) -> RedisResult<CheckedIter<'_, T>> {
         let rv = con.req_command(&self)?;
 
         let (cursor, batch) = if rv.looks_like_cursor() {
@@ -550,7 +593,7 @@ impl Cmd {
             (0, from_owned_redis_value(rv)?)
         };
 
-        Ok(Iter {
+        Ok(CheckedIter {
             batch: batch.into_iter(),
             cursor,
             con,
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -8,11 +8,11 @@ use futures_util::{
 use std::pin::Pin;
 #[cfg(feature = "cache-aio")]
 use std::time::Duration;
-use std::{fmt, io};
+use std::{fmt, io, marker::PhantomData};
 
-use crate::connection::ConnectionLike;
 use crate::pipeline::Pipeline;
 use crate::types::{from_owned_redis_value, FromRedisValue, RedisResult, RedisWrite, ToRedisArgs};
+use crate::{connection::ConnectionLike, Value};
 
 /// An argument to a redis command
 #[derive(Clone)]
@@ -88,32 +88,40 @@ pub struct Cmd {
 
 /// Represents a redis iterator.
 pub struct Iter<'a, T: FromRedisValue> {
-    batch: std::vec::IntoIter<T>,
+    batch: std::vec::IntoIter<Value>,
     cursor: u64,
     con: &'a mut (dyn ConnectionLike + 'a),
     cmd: Cmd,
+    _phantom: PhantomData<T>,
 }
 
 impl<T: FromRedisValue> Iterator for Iter<'_, T> {
-    type Item = T;
+    type Item = RedisResult<T>;
 
     #[inline]
-    fn next(&mut self) -> Option<T> {
+    fn next(&mut self) -> Option<RedisResult<T>> {
         // we need to do this in a loop until we produce at least one item
         // or we find the actual end of the iteration.  This is necessary
         // because with filtering an iterator it is possible that a whole
         // chunk is not matching the pattern and thus yielding empty results.
         loop {
             if let Some(v) = self.batch.next() {
-                return Some(v);
+                return Some(T::from_owned_redis_value(v));
             };
             if self.cursor == 0 {
                 return None;
             }
 
             let pcmd = self.cmd.get_packed_command_with_cursor(self.cursor)?;
-            let rv = self.con.req_packed_command(&pcmd).ok()?;
-            let (cur, batch): (u64, Vec<T>) = from_owned_redis_value(rv).ok()?;
+            let rv = match self.con.req_packed_command(&pcmd) {
+                Ok(v) => v,
+                Err(e) => return Some(Err(e)),
+            };
+
+            let (cur, batch): (u64, Vec<Value>) = match from_owned_redis_value(rv) {
+                Ok(v) => v,
+                Err(e) => return Some(Err(e)),
+            };
 
             self.cursor = cur;
             self.batch = batch.into_iter();
@@ -537,7 +545,7 @@ impl Cmd {
         let rv = con.req_command(&self)?;
 
         let (cursor, batch) = if rv.looks_like_cursor() {
-            from_owned_redis_value::<(u64, Vec<T>)>(rv)?
+            from_owned_redis_value::<(u64, Vec<Value>)>(rv)?
         } else {
             (0, from_owned_redis_value(rv)?)
         };
@@ -547,6 +555,7 @@ impl Cmd {
             cursor,
             con,
             cmd: self,
+            _phantom: PhantomData,
         })
     }
 
--- a/redis/examples/basic.rs
+++ b/redis/examples/basic.rs
@@ -62,13 +62,7 @@ fn do_show_scanning(con: &mut redis::Connection) -> redis::RedisResult<()> {
     // as a simple exercise we just sum up the iterator.  Since the fold
     // method carries an initial value we do not need to define the
     // type of the iterator, rust will figure "int" out for us.
-
-    let mut iterator = cmd.iter::<i32>(con)?;
-
-    let mut sum: i32 = 0;
-    while let Some(Ok(x)) = iterator.next() {
-        sum += x;
-    }
+    let sum: i32 = cmd.iter::<i32>(con)?.sum();
 
     println!("The sum of all numbers in the set 0-1000: {sum}");
 
--- a/redis/examples/basic.rs
+++ b/redis/examples/basic.rs
@@ -62,7 +62,13 @@ fn do_show_scanning(con: &mut redis::Connection) -> redis::RedisResult<()> {
     // as a simple exercise we just sum up the iterator.  Since the fold
     // method carries an initial value we do not need to define the
     // type of the iterator, rust will figure "int" out for us.
-    let sum: i32 = cmd.iter::<i32>(con)?.sum();
+
+    let mut iterator = cmd.iter::<i32>(con)?;
+
+    let mut sum: i32 = 0;
+    while let Some(Ok(x)) = iterator.next() {
+        sum += x;
+    }
 
     println!("The sum of all numbers in the set 0-1000: {sum}");
 
--- a/redis/src/commands/macros.rs
+++ b/redis/src/commands/macros.rs
@@ -63,6 +63,14 @@ macro_rules! implement_commands {
                 c.iter(self)
             }
 
+            /// Incrementally iterate the keys space safely, with options.
+            #[inline]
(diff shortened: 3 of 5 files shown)


redis/src/cmd.rs
```diff
@@ -41,25 +41,32 @@ pub struct Iter<'a, T: FromRedisValue> {
...
-            let (cur, batch): (u64, Vec<T>) = from_owned_redis_value(rv).ok()?;
+            let rv = match self.con.req_packed_command(&pcmd) {
+                Ok(v) => v,
+                Err(e) => return Some(Err(e)),
+            };
+
+            let (cur, batch): (u64, Vec<T>) = match from_owned_redis_value(rv) {
+                Ok(v) => v,
+                Err(e) => return Some(Err(e)),
+            };
```
ckcr4lyf, line 69: This is the main change, instead of the `.ok?` silently returning `None` when an error was encountered (implying end of the iterator), it will now return the error inside `Some`

ckcr4lyf:
Some other stuff to note:

* This does not change AsyncIterator, since this is kind of an example MR I thought I'd keep the scope small.

* If a caller ignores the Error, there is a risk of infinite loop: the cursor isn't updated, so someone can keep calling `.next()` with the same result (this is why I need to `break` in my test)

A possible idea I have to fix this: if we get an error from the two fallible functions, then we set some internal state `self.status = some_kind_of_redis_error` , and then in calls to `.next()` we'd check this and return it first. This does not truly combat the infinite loop scenario though, if someone is counting keys like:

```rs
while let Some(x) = iter.next() {
 count += 1;
}
```

would never end. I do not know a great solution to this, since by rust language specifics an iterator MUST return an `Option`. 

One idea would be if the "internal state" is error, then this implies the caller didn't handle the `RedisResult` correctly, and we can just panic.

redis/src/cmd.rs
```diff
@@ -41,25 +41,32 @@ pub struct Iter<'a, T: FromRedisValue> {
...
 
             let pcmd = self.cmd.get_packed_command_with_cursor(self.cursor)?;
-            let rv = self.con.req_packed_command(&pcmd).ok()?;
-            let (cur, batch): (u64, Vec<T>) = from_owned_redis_value(rv).ok()?;
+            let rv = match self.con.req_packed_command(&pcmd) {
+                Ok(v) => v,
+                Err(e) => return Some(Err(e)),
+            };
+
+            let (cur, batch): (u64, Vec<T>) = match from_owned_redis_value(rv) {
```
nihohit, line 66: if you'll save `batch` as `Vec<RedisResult<T>>`, I believe that the infinite loop problem will be solved. Each returned value will be its own result value, and failures to parse won't block the iterator.

nihohit (commented):
while we can't change this now, due to backwards compatibility, we can add this as a `safe_iter` and `async_safe_iter` (name bikeshedding aside), and point this out in the docs. 
if we'll implement `iter` and `async_iter` using the safe versions, that will also save on code duplication.

ckcr4lyf:
> we can add this as a `safe_iter`

I guess this means we can't change `.scan()` either, but instead introduce `.safe_scan()`?

redis/src/types.rs
```diff
@@ -2354,6 +2354,12 @@ impl FromRedisValue for Value {
     }
 }
 
+impl<T: FromRedisValue> FromRedisValue for RedisResult<T> {
```
nihohit, line 2357: why is this needed?

redis/src/cmd.rs
```diff
@@ -458,7 +465,7 @@ impl Cmd {
         let rv = con.req_command(&self)?;
 
         let (cursor, batch) = if rv.looks_like_cursor() {
-            from_owned_redis_value::<(u64, Vec<T>)>(rv)?
+            from_owned_redis_value::<(u64, Vec<RedisResult<T>>)>(rv)?
```
nihohit, line 468: instead of this, keep `from_owned_redis_value::<(u64, Vec<T>)>` and invert the result - Ok(vec) -> vec of Ok, and Err(err) -> vec![err].

nihohit:
> I guess this means we can't change .scan() either, but instead introduce .safe_scan()?

yup. we should also bikeshed the name a bit - maybe it should be `checked_scan/iter/async_iter`? WDYT?

redis/src/cmd.rs
ckcr4lyf: I was basically trying to implement this suggestion: https://github.com/redis-rs/redis-rs/pull/1416#discussion_r1845546102

redis/src/types.rs
ckcr4lyf: (see comment below)

redis/src/cmd.rs
ckcr4lyf: nvm, I was looking at the wrong diff

ckcr4lyf: >  invert the result - Ok(vec) -> vec of Ok, and Err(err) -> vec![err].

I'm not entirely sure what this means. Also I think if we change this, then when any one of the elements of the vec fails to parse into the type `T` , it will return an error and the whole operation will fail.

redis/src/types.rs
ckcr4lyf: It's basically needed to the thingy you mentioned below, otherwise it'll lead to this error:

```
error[E0277]: the trait bound `Result<T, types::RedisError>: FromRedisValue` is not satisfied
    --> redis/src/cmd.rs:468:38
     |
468  |             from_owned_redis_value::<(u64, Vec<RedisResult<T>>)>(rv)?
     |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `FromRedisValue` is not implemented for `Result<T, types::RedisError>`, which is required by `(u64, Vec<Result<T, types::RedisError>>): FromRedisValue`
     |
```

I think by implementing this, if we have a `Vec<RedisResult>` , then when parsing a bulk (i.e. array) reply, it will fail on the individual items, otherwise all the elements will end up failing.

redis/src/cmd.rs
nihohit:
```rs


/// Represents a redis iterator.
pub struct Iter<'a, T: FromRedisValue> {
    batch: std::vec::IntoIter<Value>,
    cursor: u64,
    con: &'a mut (dyn ConnectionLike + 'a),
    cmd: Cmd,
    _phantom: PhantomData<T>,
}

impl<'a, T: FromRedisValue> Iterator for Iter<'a, T> {
    type Item = RedisResult<T>;

    #[inline]
    fn next(&mut self) -> Option<RedisResult<T>> {
        // we need to do this in a loop until we produce at least one item
        // or we find the actual end of the iteration.  This is necessary
        // because with filtering an iterator it is possible that a whole
        // chunk is not matching the pattern and thus yielding empty results.
        loop {
            if let Some(v) = self.batch.next() {
                return Some(T::from_owned_redis_value(v));
            };
            if self.cursor == 0 {
                return None;
            }

            let pcmd = self.cmd.get_packed_command_with_cursor(self.cursor)?;
            let rv = match self.con.req_packed_command(&pcmd) {
                Ok(v) => v,
                Err(e) => return Some(Err(e)),
            };

            let (cur, batch): (u64, Vec<Value>) = match from_owned_redis_value(rv) {
                Ok(v) => v,
                Err(e) => return Some(Err(e)),
            };

            self.cursor = cur;
            self.batch = batch.into_iter();
        }
    }
}

ckcr4lyf: Hey @nihohit , sorry for the delay. I'm not too familiar with rust iterators, and I tried to fight the compiler a bit to implement the suggestion. I'm not sure if I understood correctly, but what I did seems to undo the prior suggestion (https://github.com/redis-rs/redis-rs/pull/1416#discussion_r1845546102) and bring back the infinite loop

ckcr4lyf: i.e. once the invalid UTF-8 is returned, it keeps scanning the same position:
image

ckcr4lyf: Unless `Value` means something else in the context of the suggestion.

nihohit: and I don't think you've implemented my suggestion - you keep pre-converting the values, so your batch is still a `Vec<T>` instead of `Vec<Value>`. I appreciate that you made the effort, Rust isn't always easy, but changing a type from my suggestion should've been a warning that you're doing something wrong :).
see my branch - https://github.com/nihohit/redis-rs/tree/safe-iterator, where the tests that fail are those that you modified, and still need to be solved. `test_scanning_error` passes.

nihohit: That's the spirit :)

redis/src/types.rs
ckcr4lyf: Removed

redis/src/cmd.rs
ckcr4lyf: I think I've done it correctly this time, though just for iter right now (not async iter)

ckcr4lyf:
> yup. we should also bikeshed the name a bit - maybe it should be `checked_scan/iter/async_iter`? WDYT?

`checked_` seemes like a good prefix yeah!

nihohit:
looks good, excellent work :)

remaining work:

1. move this to be a separate function/struct, in order not to break existing behavior.
2. add tests.
3. add async iter  (doesn't have to be in this PR)
4. see if the unchecked iterator can be implemented using the checked iterator (doesn't have to be in this PR)

nihohit:
Hi @ckcr4lyf , a kind reminder of this PR. Do you still want to move it forward?

ckcr4lyf:
> Hi @ckcr4lyf , a kind reminder of this PR. Do you still want to move it forward?

Sorry I totally forgot. A couple of qyestions

> 1. move this to be a separate function/struct, in order not to break existing behavior.
Sounds good.

> 2. add tests.
Should I copy over all the existing tests for the old iterator, and just also add the new test which illustrates the new behavior?

nihohit:
> Should I copy over all the existing tests for the old iterator, and just also add the new test which illustrates the new behavior?

No need to copy all of the tests, just those that seem the most pertinent.

threema-donat:
Is there anything I can do to help to push this issue forward? I missed this PR and implemented it for `Iter` and `AsyncIter`, so I already have some context.

ckcr4lyf:
> Is there anything I can do to help to push this issue forward? I missed this PR and implemented it for `Iter` and `AsyncIter`, so I already have some context.

No, it's entirely on me. Let me try and push some changes right away, trying to move it into a separate function / struct. But could you share the branch / diff of what you've done? If you've already basically implemented it, no point duping the work

threema-donat:
Thank you very much for your good work! I didn't want to increase the pressure but offer my help.

You can find my current WIP [here](https://github.com/redis-rs/redis-rs/compare/main...threema-donat:redis-rs:main), but I don't know if it will help you much since I didn't make the changes backward compatible and the code isn't reviewed yet.

ckcr4lyf:
Reopening, I think it got auto closed when i was fixing merge conflicts

ckcr4lyf:
Ive tried to implement `CheckedIterator` , and user it to replace the existing `Iter` implementation. However some tests are failing, likely due to error handling. Let me try and fix those.

threema-donat:
I'm quite concerned about the current usage of the iterators in the wild, since a lot of problems could be out there that won't be noticed when this change is made in a backwards-compatible way.

One suggestion to improve this situation would be, that the functions returning the flawed iterators and the iterators themselves are gated with a feature flag (e.g. `unchecked_iterators`). This feature could either be in the default features or not. This way, the users of the library could decide themselves if they want to keep the current behavior or not.

If this is applied, I would suggest to have the feature not in the default features, and mention in the release notes that the current behavior can be restored with enabling the feature. This way, the users of the library are made aware of the flawed iterators without too much migration burden.

@nihohit what do you think about that?

nihohit:
@threema-donat the proper behavior is to allow users to opt into breaking existing code, not force them to opt out of existing code. If we are to use features, then the default would be to keep the existing iterator.

Sorry, you've chosen the wrong time to argue for breaking backwards compatibility. I'm currently struggling with upgrading a crate that broke its API, and I'm not keen on making others share this joy :)

IMO a more forgiving mechanism is to use deprecation. We can deprecate the existing iterator with a note that explains the issues in using it, and that guides the users towards using the checked iterator. That way the users can either switch to the safer API, or simply add an attribute allowing for continued use of the existing iterator, or just ignore the warning.

threema-donat:
@nihohit In general, I would agree with you. But due to the fact that the existing code is not save to use, my personal opinion is that opting in makes more sense and is a reasonable (in terms of upgrade work) heads-up for people using that code.

I thought about a deprecation, too, but since you probably want to use `scan_match` in the `1.0` version, deprecating it now in favor of `safe_scan_match` will make upgrading to a future `1.0` version kind of strange.

Sorry for the discussion in a PR, I simply wanted to make a suggestion how to mitigate the potential of a user (I was thinking of future me) using one of the unsafe functions by accident.

nihohit:
deprecation can be done in combination with a feature - you can do something like 
```rs
#[cfg(not(feature = "safe_iterators"))]
#[deprecate(note = "explanation here")]
struct Iter {... }

#[cfg(feature = "safe_iterators")]
struct Iter { ... }

threema-donat:
Ah, right. This would be a very good solution for the problem. If I understand this correctly,  `safe_scan_match` wouldn't even be necessary, right? Since the implementation of `scan_match` could simply be different if the `safe_iterators` feature is enabled.

nihohit:
> If I understand this correctly, safe_scan_match wouldn't even be necessary, right?

Well, removing it would be another breaking change :). I try to do those only if there's a concrete reason, not just to remove code.

altanozlu:
> deprecation can be done in combination with a feature - you can do something like
> 
> ```rust
> #[cfg(not(feature = "safe_iterators"))]
> #[deprecate(note = "explanation here")]
> struct Iter {... }
> 
> #[cfg(feature = "safe_iterators")]
> struct Iter { ... }
> ```
> 
> But also, we can always just make `safe_scan_match` an inlined call to `scan_match` in v1.0.
> 
> @altanozlu, as usual your opinion is greatly appreciated :)

Why not keep two `Iter` and `CheckedIter` without any feature but making `iter` function changeable by feature ? 
If `safe_iterators`/`checked_iterators` feature is enabled `iter` will return `CheckedIter` instead of `Iter`. During `1.0` release, feature can be no-op and old behavior can be removed. This can cause some problems before `1.0` in deep dependency chains, but it's also good for progressive change.

nihohit:
That is indeed the minimally disruptive pattern, but it means that people who don't know to opt into the feature might lose values without knowing.

nihohit:
I think that we'll go with deprecation + new feature. Also, I think that the iterator shouldn't return `RedisResult`, but rather `Result<T, Value>`, so that in case of conversion failure the user still receives the returned value.
Anyone wants to pick up the implementation? @threema-donat @ckcr4lyf ?

ckcr4lyf:
> I think that we'll go with deprecation + new feature. Also, I think that the iterator shouldn't return `RedisResult`, but rather `Result<T, Value>`, so that in case of conversion failure the user still receives the returned value.
> Anyone wants to pick up the implementation? @threema-donat @ckcr4lyf ?

Let me give it a quick go today or Monday.

I'm still struggling a bit with macros and parsing, so if I'm stuck I might ask you guys to do it :pray:

threema-donat:
Sounds good to me! I also think the feature + deprecation is a good compromise. Thanks @ckcr4lyf for implementing; if you're stuck I would happily take over the MR.


Regarding the deprecation: The Makefile has currently failure on warnings enabled. Would you be OK with allowing deprecation warnings there (or do you know of a better way)? e.g. `RUSTFLAGS="-D warnings -A deprecated"`

nihohit:
The tests should enable the feature, so they won't receive the deprecation warning.

nihohit:
closing in favor of https://github.com/redis-rs/redis-rs/pull/1641

PR #1423, opened 2024-11-26
Describe as "for Redis"

zuiderkwast:
To avoid trademark issues, describe this as a library "for Redis" rather than a "Redis library".


jaymell (approved):
This certainly seems harmless to me :smile: Should we adopt the same phrasing in [`lib.rs`](https://github.com/redis-rs/redis-rs/blob/main/redis/src/lib.rs#L1) as well?

zuiderkwast:
I don't know if it helps though or if the problem is the crate name itself. IANAL.

Issue #1435, opened 2024-12-02
Better connection design for pooling


musjj:
I was trying to use [deadpool](https://github.com/bikeshedder/deadpool) with redis-rs, but got stuck when trying to use pubsub with it: https://github.com/bikeshedder/deadpool/issues/226#issuecomment-2464925279.

It's also a problem for bb8: https://github.com/djc/bb8/issues/70

It looks like that the way the connection objects are designed makes it hard for third-parties to integrate with it. Quoting the deadpool maintainer @bikeshedder:

> I just checked the documentation and [example code](https://github.com/redis-rs/redis-rs/blob/main/redis/examples/async-pub-sub.rs) from the current `redis` version and it seams that `PubSub` now works very different from previous versions. Instead of calling `into_pubsub` you create a dedicated `PubSub` object via the [Client::get_async_pubsub](https://docs.rs/redis/latest/redis/struct.Client.html#method.get_async_pubsub) method.
> 
> After a bit more digging it is possible to convert a `Connection` into a `PubSub` but there doesn't seam to be a way to get back the original connection once it's been converted into a `PubSub` object.
> 
> I currently see only one way for this to work:
> 
>     * Create a separate pool implementation just for `PubSub` connections. e.g. `deadpool_redis::PubSubPool` which creates `PubSub` connections.
> 
> 
> It really is a bummer that the `redis` crate makes a distinction between `Connection`, `PubSub` and `Monitor` which are all different types of connections which can't be interchanged.
> 
> It's weird as the `MultiplexedConnection` even provides methods for (un)subscribing but there doesn't seam to be a way to access the published messages: https://docs.rs/redis/latest/redis/aio/struct.MultiplexedConnection.html

I'm not really sure what's the best way forward. Maybe we can have a unified connection object, instead of splitting them? What do you think?

nihohit:
> It really is a bummer that the redis crate makes a distinction between Connection, PubSub and Monitor which are all different types of connections which can't be interchanged.

The problem is that the Redis RESP2 makes this distinction. Once a connection is made into a pubsub connection, sending messages on it breaks the pubsub. The way to overcome it is to use RESP3 pubsub, as @bikeshedder mentioned. I don't see a way to overcome this distinction between types for RESP2 - suggestions are welcome.

> but [PushInfo](https://docs.rs/redis/latest/redis/struct.PushInfo.html) which still requires a bit of parsing to be actually usable.

What's the issue you're experiencing? What interface would you like to see?

[...]

PR #1436, opened 2024-12-02
Clarify RESP3 subscription documentation.

nihohit:
This hopefully helps clarify some of the issues raised in https://github.com/redis-rs/redis-rs/issues/1435


bikeshedder (approved):
This clarifies the API. Until now I didn't even know (or bothered) what RESP3 was about. Knowing that those are two different serialization formats which behave slightly different it makes a lot more sense now.

PR #1437, opened 2024-12-02
Simplify parsing PushInfo into a concrete value.

nihohit:


nihohit:
Hopefully this fixes issues raised in https://github.com/redis-rs/redis-rs/pull/1435

bikeshedder:
I just missed the [Msg::from_push_info](https://docs.rs/redis/latest/redis/struct.Msg.html#method.from_push_info) method. :facepalm: 

Looking at the code I wonder why you opted for those implementations rather than implementing the `From` and `TryFrom` traits for conversion between those types?

nihohit:
mostly for consistency with the existing code, but also I worry that implementing TryFrom for every type that implements FromRedisValue might cause definitions collisions, so I try to avoid wildcard implementations.

redis/src/types.rs
```diff
@@ -2691,6 +2693,24 @@ pub struct PushInfo {
     pub data: Vec<Value>,
 }
 
+impl PushInfo {
+    /// Attempts to parse the push message's payload into a specific type.
+    pub fn try_from_redis_value<T>(mut self) -> RedisResult<T>
```
altanozlu, line 2698: IMHO it doesn't make much sense to have it, since it tries from singular value. Some PushKinds (e.g. *Message) has multiple data in it.
Or you can change documentation of function about this.

PR #1440, opened 2024-12-03
Increase MSRV to 1.72.

nihohit:


redis/Cargo.toml
```diff
@@ -8,7 +8,7 @@ repository = "https://github.com/redis-rs/redis-rs"
 documentation = "https://docs.rs/redis"
 license = "BSD-3-Clause"
 edition = "2021"
-rust-version = "1.70"
+rust-version = "1.72"
```
altanozlu, line 11: Let's also bump redis-test to 1.72

altanozlu (changes requested):
I don't see any reason not to have same MSRV

PR #1444, opened 2024-12-06
Run tests concurrently.

nihohit:


nihohit:
this reduces the test runtime from ~50 minutes to ~15 mintues, without seemingly increasing flakyness.

PR #1449, opened 2024-12-16
Async Cluster: Add pubsub support.

nihohit:
This change allows users to set a push sender on a cluster connection, subcscribe & unsubscribe from channels, channel patterns, and sharded channels, and receive pubsub messages.
The cluster connection will automatically resubscribe when a connection will disconnect, or after a topology change.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -89,9 +89,11 @@ use crate::{
         Slot, SlotMap,
     },
     cluster_topology::parse_slots,
+    cmd,
+    subscription_tracker::SubscriptionTracker,
     types::closed_connection_error,
-    Cmd, ConnectionInfo, ErrorKind, IntoConnectionInfo, RedisError, RedisFuture, RedisResult,
-    Value,
+    AsyncConnectionConfig, Cmd, ConnectionInfo, ErrorKind, IntoConnectionInfo, RedisError,
+    RedisFuture, RedisResult, ToRedisArgs, Value,
 };
 
 use futures_sink::Sink;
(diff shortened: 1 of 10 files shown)


nihohit:
@avifenesh, @altanozlu , I'll apprecaite your feedback.

altanozlu (commented):
I'll take a better look at and test it at weekend

redis/src/cluster_async/mod.rs
```diff
@@ -199,6 +201,81 @@ where
...
+
+    /// Subscribes to a new channel.
+    ///
+    /// Updates from the sender will be sent on the push sender that was passed to the manager.
+    /// If the manager was configured without a push sender, the connection won't be able to pass messages back to the user..
+    ///
+    /// This method is only available when the connection is using RESP3 protocol, and will return an error otherwise.
+    /// It should be noted that the subscription will be automatically resubscribed after disconnections, so the user might
+    /// receive additional pushes with [crate::PushKind::SSubcribe], later after the subscription completed.
+    pub async fn subscribe(&mut self, channel_name: impl ToRedisArgs) -> RedisResult<()> {
```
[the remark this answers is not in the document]
nihohit: done

redis/src/cluster_async/mod.rs
```diff
@@ -287,11 +365,17 @@ where
         cluster_params: ClusterParams,
     ) -> RedisResult<Self> {
         let connections = Self::create_initial_connections(initial_nodes, &cluster_params).await?;
+        let subscription_tracker = if cluster_params.async_push_sender.is_some() {
```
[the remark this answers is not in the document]
nihohit: sorry, I don't understand what you're asking for.

.github/workflows/rust.yml
```diff
@@ -25,12 +25,6 @@ jobs:
         config:
           [
             # Different DB cases:
-            {
```
altanozlu, line 28: Hey why it's dropping Redis 6 tests ?

redis/src/cluster_async/mod.rs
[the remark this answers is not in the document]
altanozlu: Sorry it's my confusion.

.github/workflows/rust.yml
nihohit: the commit message is pretty clear - 
"Redis 6 doesn't support SSUBCSRIBE, and so fails on cluster-async-pubsub tests."

we can add a check for DB version in the test, but since redis 6 is soon out of support, I'm fine with removing the tests.

nihohit: decided to keep testing redis 6, and try to opt out of using ssubscribe when testing those servers.

redis/src/cluster_async/mod.rs
```diff
@@ -340,6 +440,33 @@ where
         Ok(connections)
     }
 
+    fn resubscribe(&self) {
```
altanozlu, line 443: I don't have much knowledge about this part, what happens when any of additional commands created here fails for some reason ? Also is there any consistent way to fail sharded subscription command ?

nihohit: failure will be handled internally like any other failure - with retries, reconnects, and topology refreshes if needed.

> Also is there any consistent way to fail sharded subscription command ?

Do you mean, send it to another node? no.

altanozlu: I just wanted to know if there is something specific which could leave cluster connection in infinite retry loop.

nihohit: Theoretically the cluster connection can enter an infinite retry loop in a split brain situation, but I don't think that infinite retries are possible when faced with a healthy server.

nihohit:
@avifenesh, @altanozlu any last comments?
I'd really like this week to merge this, ship 0.28, and start working on 0.29 & clientside caching.

redis/src/cluster_async/mod.rs
```diff
@@ -199,6 +211,87 @@ where
...
+    /// Subscribes to a new channel.
+    ///
+    /// Updates from the sender will be sent on the push sender that was passed to the manager.
+    /// If the manager was configured without a push sender, the connection won't be able to pass messages back to the user..
+    ///
+    /// This method is only available when the connection is using RESP3 protocol, and will return an error otherwise.
+    /// It should be noted that the subscription will be automatically resubscribed after disconnections, so the user might
+    /// receive additional pushes with [crate::PushKind::SSubcribe], later after the subscription completed.
+    pub async fn subscribe(&mut self, channel_name: impl ToRedisArgs) -> RedisResult<()> {
+        check_resp3!(self.state.protocol);
```
avifenesh, line 224: Is it a normal behavior of the library to check for the user the protocol and return an in-house built error, instead of the server error?
Is it something that will be aligned for any command with differences?
What is the difference from version compatibility, why diff is not being checked for version?
Is it worth the overhead and maintenance?

redis/src/cluster_async/mod.rs
```diff
@@ -212,6 +305,7 @@ struct InnerCore<C> {
     cluster_params: ClusterParams,
     pending_requests: Mutex<Vec<PendingRequest<C>>>,
     initial_nodes: Vec<ConnectionInfo>,
+    subscription_tracker: Option<Mutex<SubscriptionTracker>>,
```
avifenesh, line 308: Why `Mutex`? Why not `RWLock`? How heavy is the writing compared to general usage?

redis/src/cluster_async/mod.rs
```diff
@@ -1133,13 +1302,24 @@ where
...
     let response_timeout = params.response_timeout;
+    let push_sender = params.async_push_sender.clone();
     let info = get_connection_info(node, params)?;
-    let mut conn: C = C::connect(info, response_timeout, connection_timeout).await?;
-    check_connection(&mut conn).await?;
-    if read_from_replicas {
-        // If READONLY is sent to primary nodes, it will have no effect
-        crate::cmd("READONLY").exec_async(&mut conn).await?;
+    let mut config = AsyncConnectionConfig::default()
+        .set_connection_timeout(connection_timeout)
```
avifenesh, line 1308: I think the convention is `with_...`

redis/src/cluster_async/mod.rs
nihohit: you won't receive a server error if you send a "subscribe" request on a RESP2 connection. Instead the connection will enter a pub/sub state, which will break once you send other requests on the connection. The RESP3 check is made in order to verify that the connection can be used to receive pubsub messages without breaking when sending requests.

> Is it something that will be aligned for any command with differences?

I don't understand the question. Commands with what differences?

> why diff is not being checked for version?

What diff?

> Is it worth the overhead and maintenance?

The overhead of a single check is literally an integer comparison and I don't see meaningful potential maintenance cost. On the other hand, without this we can get correctness errors, so yes - definitely worth it.

redis/src/cluster_async/mod.rs
nihohit: this potentially writes on every request, and reads on disconnects. There are more writes than reads, so an RwLock isn't optimized for this scenario.

redis/src/cluster_async/mod.rs
nihohit: the function exists, we won't break backwards compatibility for convention.

redis/src/cluster_async/mod.rs
avifenesh: I meant difference between behaviors, but your answer explained it.

avifenesh (approved):
LGTM

altanozlu (changes requested):

redis/tests/test_cluster_async.rs
```diff
@@ -2296,6 +2296,541 @@ mod cluster_async {
...
+            let response = conn
+                .route_command(
+                    cmd("INFO").arg("server"),
+                    RoutingInfo::SingleNode(SingleNodeRoutingInfo::Random),
+                )
+                .await
+                .unwrap();
+            let res = from_owned_redis_value::<String>(response).unwrap();
+            println!("res: {:?}, {}", res, res.contains("redis_version:6"));
+            true
```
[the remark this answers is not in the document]
nihohit: I'm an idiot 😅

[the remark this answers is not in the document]
altanozlu: It happens, no worries, approving it 👌🏻

altanozlu (approved):

PR #1451, opened 2024-12-18
Remove async_trait dependency.

nihohit:


altanozlu:
1.74 build complains, maybe you need to bump to 1.75?
https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html#async-fn-and-return-position-impl-trait-in-traits

nihohit:
> 1.74 build complains, maybe you need to bump to 1.75?

I know, I wanted to double check that 1.75 is the lowest viable version :)

PR #1454, opened 2024-12-24
Expose RetryMethod to consumers

Braedon-Wooding-Displayr:
While clusters do implement retrying if you are using the non-cluster API you would need to do this by hand, which makes the RetryMethod API very useful.

I've tried to add appropriate comments onto things but if you think it needs extra clarification feel free to let me know and I'll update the PR.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -778,13 +778,22 @@ impl fmt::Debug for RedisError {
     }
 }
 
-pub(crate) enum RetryMethod {
+/// What method should be used if retrying this request.
+#[non_exhaustive]
+pub enum RetryMethod {
+    /// Create a fresh connection, since the current connection is no longer usable.
     Reconnect,
+    /// Don't retry, this is a permanent error.
     NoRetry,
+    /// Retry immediately, this doesn't require a wait.
     RetryImmediately,
+    /// Retry after sleeping to avoid overloading the external service.
     WaitAndRetry,
+    /// The key has moved to a different node but we have to ask which node, this is only relevant for clusters.
     AskRedirect,
+    /// The key has moved to a different node, this is only relevant for clusters.
     MovedRedirect,
+    /// Reconnect the initial connection to the master cluster, this is only relevant for clusters.
     ReconnectFromInitialConnections,
 }
 
@@ -998,7 +1007,13 @@ impl RedisError {
         Self { repr }
     }
 
-    pub(crate) fn retry_method(&self) -> RetryMethod {
+    /// Specifies what method (if any) should be used to retry this request.
+    ///
+    /// If you are using the cluster api retrying of requests is already handled by the library.
+    ///
+    /// This isn't precise, and internally the library uses multiple other considerations rather
+    /// than just the error kind on when to retry.
+    pub fn retry_method(&self) -> RetryMethod {
         match self.kind() {
             ErrorKind::Moved => RetryMethod::MovedRedirect,
             ErrorKind::Ask => RetryMethod::AskRedirect,


redis/src/types.rs
```diff
@@ -778,13 +778,21 @@ impl fmt::Debug for RedisError {
...
+    /// Create a fresh connection.
     Reconnect,
+    /// Don't retry, this is a permanent error.
     NoRetry,
+    /// Retry immediately, this doesn't require a wait since it's an IO level issue.
     RetryImmediately,
+    /// Retry after sleeping to avoid overloading the external service.
     WaitAndRetry,
+    /// The key has moved to a different node but we have to ask which node.
     AskRedirect,
```
Braedon-Wooding-Displayr, line 792: I'm also okay if you want to hide these behind feature flags?  I just didn't because the actual ErrorKinds themselves aren't (unlike Json which is).

redis/src/types.rs
```diff
@@ -778,13 +778,21 @@ impl fmt::Debug for RedisError {
...
+pub enum RetryMethod {
+    /// Create a fresh connection.
     Reconnect,
+    /// Don't retry, this is a permanent error.
     NoRetry,
+    /// Retry immediately, this doesn't require a wait since it's an IO level issue.
     RetryImmediately,
+    /// Retry after sleeping to avoid overloading the external service.
     WaitAndRetry,
+    /// The key has moved to a different node but we have to ask which node.
```
nihohit, line 791: relevant only for cluster

redis/src/types.rs
```diff
@@ -778,13 +778,21 @@ impl fmt::Debug for RedisError {
...
     AskRedirect,
+    /// The key has moved to a different node.
```
nihohit, line 793: relevant only for cluster

redis/src/types.rs
```diff
@@ -778,13 +778,21 @@ impl fmt::Debug for RedisError {
...
 }
 
-pub(crate) enum RetryMethod {
+/// What method should be used if retrying this request.
+pub enum RetryMethod {
+    /// Create a fresh connection.
     Reconnect,
+    /// Don't retry, this is a permanent error.
     NoRetry,
+    /// Retry immediately, this doesn't require a wait since it's an IO level issue.
```
nihohit, line 787: remove the IO level stuff, or clarify that ti's an example, not every case of this value.

redis/src/types.rs
```diff
@@ -778,13 +778,21 @@ impl fmt::Debug for RedisError {
     }
 }
 
-pub(crate) enum RetryMethod {
+/// What method should be used if retrying this request.
+pub enum RetryMethod {
+    /// Create a fresh connection.
```
nihohit, line 783: clarify that the current connection is no longer usable.

redis/src/types.rs
```diff
@@ -998,7 +1006,10 @@ impl RedisError {
         Self { repr }
     }
 
-    pub(crate) fn retry_method(&self) -> RetryMethod {
+    /// Specifies what method (if any) should be used to retry this request.
```
nihohit, line 1009: docs should clarify that this isn't precise, but just rough guidelines / recommendations, and that the library uses more considerations internally than just the error kind.

nihohit (commented):
the issue with this is that this makes adding a retry method a breaking change, so that needs to be handled, too. maybe using non-exhaustive?

Braedon-Wooding-Displayr:
All comments you made should now be resolved @nihohit

nihohit (approved):

Issue #1389, opened 2024-10-19
Features like `tokio-comp` or `async-std-comp` must be enabled when `aio` is enabled 


SamuelMarks:
Otherwise you get errors like:
```
error[E0433]: failed to resolve: use of undeclared type `ValueCodec`
   --> cargo/registry/src/index.crates.io-6f17d22bba15001f/redis-0.27.5/src/aio/connection.rs:373:9
    |
373 |         ValueCodec::default()
    |         ^^^^^^^^^^ use of undeclared type `ValueCodec`
```

nihohit:
Yes, see `compile_error!("tokio-comp or async-std-comp features required for aio feature");`. 
What is the behavior you expect to see, if you build the async code without any runtime support?

Backwards compatiblity aside, `aio` should be a private feature. Users shouldn't enable it.

PR #1464, opened 2025-01-01
Improve features and TLS docs.

nihohit:
The purpose of this change is to make the documentation
around features clearer, and to prevent mistakes such
https://github.com/redis-rs/redis-rs/issues/1389


nihohit:
@altanozlu thoughts?

altanozlu (approved):
LGTM.
Also in the future I think it make sense to move features like `tcp_nodelay`, `disable-client-setinfo`,`keep-alive` into connection configuration etc.

nihohit:
Sure, do you want to open a pr with the suggestion?

Issue #1322, opened 2024-09-13
Implement FromRedisValue Trait for ! (never type)


xinchengxx:
# Implement `FromRedisValue` Trait for `!` (never type)

## Problem Description

With the upcoming changes in the Rust compiler, particularly the resolution of this [issue](https://github.com/rust-lang/rust/issues/123748), the `!` (never type) will no longer automatically convert to the `()` type. This change will cause some common Redis operations to fail compilation in the 2024 edition and beyond.

For example, the following code may not compile in the future:

```rust
conn.srem(k, v).await?;
```

The Rust compiler will infer the return type of this line as `Result<!, E>`, but since `!: redis::FromRedisValue` is not implemented, it will result in a compilation error.

## Current Situation

It's worth noting that this issue is already starting to affect users. As of Rust 1.81.0, Clippy is raising warnings about this problem. This early warning indicates that the issue is imminent and needs to be addressed soon to ensure a smooth transition for redis-rs users.

## Current Workaround

Currently, to make the code compile and silence Clippy warnings, we must explicitly specify the return type as `()`:

```rust
let _: () = conn.srem(k, v).await?;
```

While this workaround solves the problem, it's inconvenient and reduces code readability.

## Proposed Solution

To ensure redis-rs maintains good compatibility and user experience in future Rust versions, I suggest implementing the `FromRedisValue` Trait for `!` (never type). This will ensure that existing code continues to work in the 2024 edition and beyond without requiring extensive modifications.

## Potential Impact

Implementing this feature will:
1. Improve redis-rs compatibility with future Rust versions
2. Maintain the simplicity and readability of existing code
3. Reduce the need for code modifications when users upgrade to newer Rust versions
4. Eliminate current Clippy warnings in Rust 1.81.0 and later versions

## Additional Information

- Rust issue link: https://github.com/rust-lang/rust/issues/123748
- This change is expected to take effect in the 2024 edition (Due to release at 2025.2)
- Clippy in Rust 1.81.0 is already warning about this issue

Thank you for considering this suggestion! Please let me know if you need any further information or have any questions.

nihohit:
Is this suggestion even possible? Specifically, in stable Rust, not with the nightly compiler.
You're more than welcome to contribute a solution.

nihohit:
also, it should be noted, for future reader - another current workaround is to use `exec/exec_async`.

clbarnes:
[...]

PR #1480, opened 2025-01-09
Add new set of typed commands

MurrayGroves:
A solution to #1228 and #1322 , as described by @nihohit.

Introduce two new traits `TypedCommands` and `AsyncTypedCommands` that behave like their non-typed counterparts, but return specifically opinionated types instead.

The return types are defined in the existing place in `commands/mod.rs` that defines the inputs for `Commands` and `AsyncCommands`. The existing macros for generating `Commands` and `AsyncCommands` will ignore the return type, while the new ones for the typed variants will include those types in the generated methods. Additionally, if a return type is omitted, it will default to `redis::Value`.

### DIFF
--- a/redis/src/commands/macros.rs
+++ b/redis/src/commands/macros.rs
@@ -1,6 +1,6 @@
 // Generate implementation for function skeleton, we use this for `AsyncTypedCommands` because we want to be able to handle having a return type specified or unspecified with a fallback
-#[allow(unused_macros)]
-macro_rules! implement_command {
+#[cfg(feature = "aio")]
+macro_rules! implement_command_async {
 	// If return type is specified in the input skeleton, then we will return it in the generated function (note match rule `$rettype:ty`)
 	(
         $lifetime: lifetime
@@ -36,6 +36,46 @@ macro_rules! implement_command {
 		) -> crate::types::RedisFuture<'a, crate::Value>
 		{
 			Box::pin(async move { ($body).query_async(self).await })
+		}
+	}
+}
+
+macro_rules! implement_command_sync {
+	// If return type is specified in the input skeleton, then we will return it in the generated function (note match rule `$rettype:ty`)
+	(
+        $lifetime: lifetime
+		$(#[$attr:meta])+
+		fn $name:ident<$($tyargs:ident : $ty:ident),*>(
+			$($argname:ident: $argty:ty),*) $body:block $rettype:ty
+    ) => {
+		$(#[$attr])*
+		#[inline]
+		#[allow(clippy::extra_unused_lifetimes, clippy::needless_lifetimes)]
+		fn $name<$lifetime, $($tyargs: $ty + Send + Sync + $lifetime,)*>(
+			& $lifetime mut self
+			$(, $argname: $argty)*
+		) -> RedisResult<$rettype>
+
+		{
+			Cmd::$name($($argname),*).query(self)
+		}
+	};
+	// If no return type is specified in the input skeleton, we default to returning `redis::Value` in the generated function (note lack of match rule `$rettype:ty`)
+	(
+        $lifetime: lifetime
+		$(#[$attr:meta])+
+		fn $name:ident<$($tyargs:ident : $ty:ident),*>(
+			$($argname:ident: $argty:ty),*) $body:block
+    ) => {
+		$(#[$attr])*
+		#[inline]
+		#[allow(clippy::extra_unused_lifetimes, clippy::needless_lifetimes)]
+		fn $name<$lifetime, $($tyargs: $ty + Send + Sync + $lifetime,)*>(
+			& $lifetime mut self
+			$(, $argname: $argty)*
+		) -> RedisResult<crate::Value>
+		{
+			Cmd::$name($($argname),*).query(self)
 		}
 	}
 }
@@ -299,12 +339,107 @@ macro_rules! implement_commands {
             }
         }
 
+        /// Implements common redis commands.
+        /// The return types are concrete and opinionated. If you want to choose the return type you should use the `Commands` trait.
+        pub trait TypedCommands : ConnectionLike+Sized {
+            $(
+				implement_command_sync! {
+					$lifetime
+					$(#[$attr])*
+					fn $name<$($tyargs: $ty),*>(
+						$($argname: $argty),*
+					)
+
+					{
+						$body
+					} $($rettype)?
+				}
+            )*
+
+            /// Incrementally iterate the keys space.
+            #[inline]
+            fn scan<RV: FromRedisValue>(&mut self) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("SCAN");
+                c.cursor_arg(0);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate the keys space with options.
+            #[inline]
+            fn scan_options<RV: FromRedisValue>(&mut self, opts: ScanOptions) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("SCAN");
+                c.cursor_arg(0).arg(opts);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate the keys space for keys matching a pattern.
+            #[inline]
+            fn scan_match<P: ToRedisArgs, RV: FromRedisValue>(&mut self, pattern: P) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("SCAN");
+                c.cursor_arg(0).arg("MATCH").arg(pattern);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate hash fields and associated values.
+            #[inline]
+            fn hscan<K: ToRedisArgs, RV: FromRedisValue>(&mut self, key: K) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("HSCAN");
+                c.arg(key).cursor_arg(0);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate hash fields and associated values for
+            /// field names matching a pattern.
+            #[inline]
+            fn hscan_match<K: ToRedisArgs, P: ToRedisArgs, RV: FromRedisValue>
+                    (&mut self, key: K, pattern: P) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("HSCAN");
+                c.arg(key).cursor_arg(0).arg("MATCH").arg(pattern);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate set elements.
+            #[inline]
+            fn sscan<K: ToRedisArgs, RV: FromRedisValue>(&mut self, key: K) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("SSCAN");
+                c.arg(key).cursor_arg(0);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate set elements for elements matching a pattern.
+            #[inline]
+            fn sscan_match<K: ToRedisArgs, P: ToRedisArgs, RV: FromRedisValue>
+                    (&mut self, key: K, pattern: P) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("SSCAN");
+                c.arg(key).cursor_arg(0).arg("MATCH").arg(pattern);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate sorted set elements.
+            #[inline]
+            fn zscan<K: ToRedisArgs, RV: FromRedisValue>(&mut self, key: K) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("ZSCAN");
+                c.arg(key).cursor_arg(0);
+                c.iter(self)
+            }
+
+            /// Incrementally iterate sorted set elements for elements matching a pattern.
+            #[inline]
+            fn zscan_match<K: ToRedisArgs, P: ToRedisArgs, RV: FromRedisValue>
+                    (&mut self, key: K, pattern: P) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("ZSCAN");
+                c.arg(key).cursor_arg(0).arg("MATCH").arg(pattern);
+                c.iter(self)
+            }
+        }
+
+
 		/// Implements common redis commands over asynchronous connections.
-		/// The return types are concrete and opinionated. If you want to choose the return type you should use the `AsyncCommands` trait.
+        /// The return types are concrete and opinionated. If you want to choose the return type you should use the `AsyncCommands` trait.
 		#[cfg(feature = "aio")]
         pub trait AsyncTypedCommands : crate::aio::ConnectionLike + Send + Sized {
             $(
-				implement_command! {
+				implement_command_async! {
 					$lifetime
 					$(#[$attr])*
 					fn $name<$($tyargs: $ty),*>(
@@ -350,7 +485,7 @@ macro_rules! implement_commands {
             }
 
             /// Incrementally iterate hash fields and associated values for
-			/// field names matching a pattern.
+            /// field names matching a pattern.
             #[inline]
             fn hscan_match<K: ToRedisArgs, P: ToRedisArgs, RV: FromRedisValue>
                     (&mut self, key: K, pattern: P) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
--- a/redis/src/commands/macros.rs
+++ b/redis/src/commands/macros.rs
@@ -1,4 +1,5 @@
 // Generate implementation for function skeleton, we use this for `AsyncTypedCommands` because we want to be able to handle having a return type specified or unspecified with a fallback
+#[allow(unused_macros)]
 macro_rules! implement_command {
 	// If return type is specified in the input skeleton, then we will return it in the generated function (note match rule `$rettype:ty`)
 	(
--- a/redis/src/commands/macros.rs
+++ b/redis/src/commands/macros.rs
@@ -1,5 +1,6 @@
-#[allow(unused_macros)]
+// Generate implementation for function skeleton, we use this for `AsyncTypedCommands` because we want to be able to handle having a return type specified or unspecified with a fallback
 macro_rules! implement_command {
+	// If return type is specified in the input skeleton, then we will return it in the generated function (note match rule `$rettype:ty`)
 	(
         $lifetime: lifetime
 		$(#[$attr:meta])+
@@ -18,6 +19,7 @@ macro_rules! implement_command {
 			Box::pin(async move { ($body).query_async(self).await })
 		}
 	};
+	// If no return type is specified in the input skeleton, we default to returning `redis::Value` in the generated function (note lack of match rule `$rettype:ty`)
 	(
         $lifetime: lifetime
 		$(#[$attr:meta])+
--- a/redis/src/commands/macros.rs
+++ b/redis/src/commands/macros.rs
@@ -11,10 +11,10 @@ macro_rules! implement_command_async {
 		$(#[$attr])*
 		#[inline]
 		#[allow(clippy::extra_unused_lifetimes, clippy::needless_lifetimes)]
-		fn $name<$lifetime, $($tyargs: $ty + Send + Sync + $lifetime,)*>(
+		fn $name<$lifetime, RV: FromRedisValue, $($tyargs: $ty + Send + Sync + $lifetime,)*>(
 			& $lifetime mut self
 			$(, $argname: $argty)*
-		) -> crate::types::RedisFuture<'a, crate::Value>
+		) -> crate::types::RedisFuture<'a, RV>
 		{
 			Box::pin(async move { ($body).query_async(self).await })
 		}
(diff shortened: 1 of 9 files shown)


nihohit:
@MurrayGroves brilliant, thanks!

I recommend that, in order to speed up the review process, you start with only a few commands, so we can agree on the interface and behavior before you implement a whole slew of commands, and then need to make a massive change if there are any comments.
Maybe 1-2 commands with a simple typed return (`SET`/`GET` are popular), and 1 command that returns a complex type?

MurrayGroves:
@nihohit I typed a few commands, and there's some example usages in `examples/async-typed.rs`.

redis/src/commands/mod.rs
```diff
@@ -138,17 +138,17 @@ implement_commands! {
...
 
     /// Get the value of a key.  If key is a vec this becomes an `MGET`.
-    fn get<K: ToRedisArgs>(key: K) {
+    fn get<K: ToRedisArgs>(key: K) -> Option<String> {
         cmd(if key.num_of_args() <= 1 { "GET" } else { "MGET" }).arg(key)
     }
 
     /// Get values of keys
-    fn mget<K: ToRedisArgs>(key: K){
+    fn mget<K: ToRedisArgs>(key: K) -> Vec<String> {
```
nihohit, line 146: IIRC This should be `Vec<Option<String>>`, because you need to reflect the fact that some key can be missing.

redis/src/commands/mod.rs
```diff
@@ -2603,4 +2606,4 @@ pub fn resp3_hello(connection_info: &RedisConnectionInfo) -> Cmd {
     }
 
     hello_cmd
-}
+}
```
nihohit, line 2609: nit: please restore the last empty line

redis/src/types.rs
```diff
@@ -2731,3 +2731,46 @@ pub(crate) type SyncPushSender = std::sync::mpsc::Sender<PushInfo>;
 pub(crate) fn closed_connection_error() -> RedisError {
     RedisError::from(io::Error::from(io::ErrorKind::BrokenPipe))
 }
+
+/// Possible types of value held in Redis, from https://redis.io/docs/latest/commands/type/
+#[derive(Debug, Clone, PartialEq)]
+pub enum ValueType {
```
nihohit, line 2737: this could be an issue once the server adds more types (for example, when using modules). either leave room for an unknown type by adding a variant holding `String`, or drop this and just return `String` from `key_type`

redis/src/types.rs
```diff
@@ -2731,3 +2731,46 @@ pub(crate) type SyncPushSender = std::sync::mpsc::Sender<PushInfo>;
...
+				"set" => Ok(ValueType::Set),
+				"zset" => Ok(ValueType::ZSet),
+				"hash" => Ok(ValueType::Hash),
+				"stream" => Ok(ValueType::Stream),
+				_ => invalid_type_error!(s, "Unknown value type")
+			},
+			_ => invalid_type_error!(v, "Unknown value type")
+		}
+	}
+}
```
nihohit, line 2776: same

redis/src/commands/macros.rs
```diff
@@ -1,10 +1,49 @@
+#[allow(unused_macros)]
+macro_rules! implement_command {
+	(
+        $lifetime: lifetime
```
nihohit, line 4: can you add documentation here and below, explaining these derivation rules?

nihohit (approved):
Excellent work!
The documentation of the crate will also need to be updated, and so will the tests.

nihohit:
@altanozlu WDYT?

redis/src/commands/macros.rs
MurrayGroves: Added, let me know if it's unclear

redis/src/types.rs
MurrayGroves: Added `Unknown(String)` variant.

redis/src/commands/mod.rs
MurrayGroves: Oops, fixed.

redis/src/commands/macros.rs
```diff
@@ -247,6 +288,101 @@ macro_rules! implement_commands {
...
+
+					{
+						$body
+					} $($rettype)?
+				}
+            )*
+
+            /// Incrementally iterate the keys space.
+            #[inline]
+            fn scan<RV: FromRedisValue>(&mut self) -> crate::types::RedisFuture<crate::cmd::AsyncIter<'_, RV>> {
```
nihohit, line 321: would be nice to save the repetition of this code, too. got any ideas? the code differs between sync & async only in the last line, so maybe we can add a macro that adds the right iterator type

MurrayGroves: Agreed on saving repetition, adding a macro seems like a reasonable solution to me.

altanozlu:
@nihohit  I've totally missed this MR, will look into this soon

altanozlu:
I really liked the work thanks for the contribution, I don't have much to add @nihohit 's comments. Just one question I believe there will also be `TypedCommands` as a sync variant right ?

MurrayGroves:
> I really liked the work thanks for the contribution, I don't have much to add @nihohit 's comments. Just one question I believe there will also be `TypedCommands` as a sync variant right ?

@altanozlu yep, just figured I would get Async working first.

nihohit:
@MurrayGroves I'd be happy to move forward with this PR. Do you need any kind of help with this?

MurrayGroves:
@nihohit sorry I've been pretty busy, hopefully I can get some work done on it tonight.

nihohit:
@MurrayGroves I can't run the build, because of diffs. can you please rebase?

MurrayGroves:
@nihohit rebased 😀

nihohit:
@MurrayGroves looks like compilation fails due to warnings. Do you need help with running compilation locally, in order to test that changes compile?

nihohit:
The reason for the failure is that you only use expand the types in async code, so when `make test` compiles without async feature flags, the type is unused. You'll need to either create sync typed commands, or move the imports and declarations of `IntegerReplyOrNoOp` to under a `#[cfg(feature = "aio")]` flag.

MurrayGroves:
@nihohit of course, that makes perfect sense. Apologies for the mistakes, I'm new to library development, so I haven't developed anything with optional features before. Added a macro for the generation of typed sync commands, and removed the attributes for ignoring the dead code warnings.

nihohit:
No worries. You jumped into the deep end with this change, there's no wonder it takes time and effort to make it really work :)

MurrayGroves:
@nihohit are you happy with how I've done things? If so, I've got time now to type the rest of the commands, and I can also update the documentation. You also mentioned updating tests, what sort of things should I be testing?

redis/src/commands/macros.rs
```diff
@@ -1,10 +1,92 @@
...
+		fn $name<$lifetime, $($tyargs: $ty + Send + Sync + $lifetime,)*>(
+			& $lifetime mut self
+			$(, $argname: $argty)*
+		) -> crate::types::RedisFuture<'a, $rettype>
+
+		{
+			Box::pin(async move { ($body).query_async(self).await })
+		}
+	};
+	// If no return type is specified in the input skeleton, we default to returning `redis::Value` in the generated function (note lack of match rule `$rettype:ty`)
```
nihohit, line 23: do we want the fallback? wouldn't it be better to require that values be defined, so that new functions won't accidentally be added without a type?
How can we prevent such mistakes?

redis/src/commands/macros.rs
```diff
@@ -257,6 +339,196 @@ macro_rules! implement_commands {
...
+
+					{
+						$body
+					} $($rettype)?
+				}
+            )*
+
+            /// Incrementally iterate the keys space.
+            #[inline]
+            fn scan<RV: FromRedisValue>(&mut self) -> RedisResult<Iter<'_, RV>> {
```
nihohit, line 361: why are return value-generic functions part of the trait? just for completion?

redis/src/commands/macros.rs
```diff
@@ -257,6 +339,196 @@ macro_rules! implement_commands {
...
+            fn zscan_match<K: ToRedisArgs, P: ToRedisArgs, RV: FromRedisValue>
+                    (&mut self, key: K, pattern: P) -> RedisResult<Iter<'_, RV>> {
+                let mut c = cmd("ZSCAN");
+                c.arg(key).cursor_arg(0).arg("MATCH").arg(pattern);
+                c.iter(self)
+            }
+        }
+
+
+		/// Implements common redis commands over asynchronous connections.
```
nihohit, line 437: nit: indentation

redis/src/commands/macros.rs
```diff
@@ -257,6 +339,196 @@ macro_rules! implement_commands {
...
+        /// The return types are concrete and opinionated. If you want to choose the return type you should use the `AsyncCommands` trait.
+		#[cfg(feature = "aio")]
```
nihohit, line 439: nit: indentation

redis/src/commands/mod.rs
```diff
@@ -412,72 +442,99 @@ implement_commands! {
...
 
     /// Get one or more fields' TTL in milliseconds.
-    fn hpttl<K: ToRedisArgs, F: ToRedisArgs>(key: K, fields: F) {
+    fn hpttl<K: ToRedisArgs, F: ToRedisArgs>(key: K, fields: F) -> IntegerReplyOrNoOp {
         cmd("HPTTL").arg(key).arg("FIELDS").arg(fields.num_of_args()).arg(fields)
     }
 
     /// Set one or more fields' time to live in seconds.
-    fn hexpire<K: ToRedisArgs, F: ToRedisArgs>(key: K, seconds: i64, opt: ExpireOption, fields: F) {
+    /// Returns an array where each element corresponds to the field at the same index in the fields argument.
```
nihohit, line 460: I'm not sure that we need to provide the server behavior here and in other functions - we're documenting the behavior of another project, which might change without updating this code. wdyt?

nihohit (approved):
I very much love this change

redis/src/commands/macros.rs
nihohit: any thoughts about this? I don't think that this is a hard requirement, but I think that there's a risk of changing the behavior of only one trait, and breaking things.

nihohit:
> You also mentioned updating tests, what sort of things should I be testing?

I don't think that tests should be added, just that existing tests should use the new traits & types.

redis/src/commands/macros.rs
MurrayGroves: Good point, my initial thinking was that there might be some functions where there isn't a solid opinionated choice for return type so we might still want to leave it as `RedisValue`. I can adapt the macros so that the return type is a required parameter, and then it won't compile without a return type provided.

redis/src/commands/macros.rs
MurrayGroves: Yeah just for completion, I wanted to reduce the chances that someone would need to use both traits as that is quite annoying because the function names overlap, so you have to specify the trait for every single command if you're using both traits.

redis/src/commands/mod.rs
MurrayGroves: That makes sense to me, probably a good compromise is to include a link to the Redis docs for each command?

redis/src/commands/macros.rs
nihohit: and worst case scenario, the iterators initially won't be implemented with concrete types.

redis/src/commands/macros.rs
nihohit: Now I'm unsure whether these should be generic, or return `redis::Value`. Maybe because its almost certain that RV isn't `()` or `!`, this isn't a big issue.

redis/src/commands/mod.rs
nihohit: yes, definitely.

nihohit:
And just to make sure - I know that I give are a lot of comments. Please understand - I think that this change is very important, and I want it to be correct the first time around. I really appreciate the effort that your work here, and I don't want you to feel that I underestimate your effort or take it for granted.

MurrayGroves:
> And just to make sure - I know that I give are a lot of comments. Please understand - I think that this change is very important, and I want it to be correct the first time around. I really appreciate the effort that your work here, and I don't want you to feel that I underestimate your effort or take it for granted.

Don't worry, I totally understand. I am very much a proponent of give as much feedback as possible, and doing things right the first time - even if doing things "wrong" means misplaced indentation or something else that may seem inconsequential, it's still important.

redis/src/commands/macros.rs
MurrayGroves: I think w.r.t this and the conversation below, I think it's a good idea to explicitly return `redis::Value` where it's unclear what other return type makes sense in most cases. I'll remove the fallback value from the macro, and we can say that functions should be explicitly defined as having a return type of `redis::Value` in cases where there isn't a definitive type that should be returned?

nihohit: I'm not sure about this. We have 2 problematic cases - 
1. Functions that are defined in `implement_command_sync / implement_command_async`, and don't have a clear return type
2. Iterator functions that are manually implemented in each trait.

I'm sorry that I'm inconsistent about this, but I'm still undecided whether they should return a constant type (where `redis::Value` is the only type that makes sense) or should remain generic and require definition by the user. The first option ensures that the return value of each new function is considered when it is added, but the second option is probably the better user experience, since `redis::Value` is almost always used immediately with a conversion function.
@altanozlu thoughts?

MurrayGroves: What about some way to explicitly "type" a function as having a generic return? Since we're on the macro-level we don't actually need to have the return type in the function definitions be an actual Rust type. We could require some value be placed after `-> ` but in the case where it is `GENERIC` (or some other pre-defined identifier) the macro could generate a generic function instead of a specifically typed one.

nihohit: Oh, if we can do that, that would be great! We'll just require *some* marker of intention, and that should be sufficient.

MurrayGroves:
@nihohit do you have any thoughts on how we should be typing commands which return some negative integer or zero integer for some form of failure, and everything else is the return value e.g. [linsert](https://redis.io/docs/latest/commands/linsert/). I had implemented an enum `IntegerReplyOrNoOp`, but I didn't realise at the time that different functions use different integers to mean different things. One possibility is making a new type for each of those different serialisations, but that seems like a bad idea to me.

nihohit:
> do you have any thoughts on how we should be typing commands which return some negative integer or zero integer for some form of failure, and everything else is the return value e.g. [linsert](https://redis.io/docs/latest/commands/linsert/).

If there are multiple commands with their different integer meaning, we should just return integer, unless we want to strongly type every command, which sound like a lot of work.

> I'm currently having issues running make test - the sentinel tests fail with Connection refused. Is there some other setup step I'm missing?

This is weird, and no - assuming the redis server installed on your machine is reasonably new (v6 or later) the tests should pass as-is. Do the tests also fail when you run them from `main` branch without any changes?

MurrayGroves:
> If there are multiple commands with their different integer meaning, we should just return integer, unless we want to strongly type every command, which sound like a lot of work.

Now that I've gone over all the commands, it seems there are actually only two different formats, so I think it should be reasonable to have two different return types? I'm not sure what a good type name for the two would be, because the end user doesn't really care that there are two different formats.

 I spent a bit of time trying to figure out a way to have them all return one type, but with different rules to convert from the integer reply to the enum. However I couldn't get anything working.
 
> This is weird, and no - assuming the redis server installed on your machine is reasonably new (v6 or later) the tests should pass as-is. Do the tests also fail when you run them from `main` branch without any changes?

Tests pass on the version of main on my fork, but not on the version of main on this repo :thinking:

nihohit:
> Tests pass on the version of main on my fork, but not on the version of main on this repo 🤔

sounds like `git bisect` is the tool for this mystery :)

MurrayGroves:
The tests fail as of [this commit](https://github.com/redis-rs/redis-rs/commit/4bd1b4b901197a0bd9c360815fd49d222c63ad81). Further digging shows that commenting out the call to `get_major_version` in `redis-test/src/server.rs` (and dependent functionality) stops the tests failing. Keeping the function in, and only increasing the `6` in the conditional on line 176 to prevent the debug command being enabled does not fix the test. It seems that just the calling of `get_major_version` causes this issue. Considering that all this function does is run `redis-server -v`, I honestly have no idea how that could be causing this. Hopefully you have some idea?

nihohit:
what happens when you run `redis-server -v` locally?

MurrayGroves:
```failed check for replica role - current role: [bulk-string('"slave"'), bulk-string('"127.0.0.1"'), int(37321), bulk-string('"handshake"'), int(-1)]                                                                                         
failed check for replica role - current role: [bulk-string('"slave"'), bulk-string('"127.0.0.1"'), int(37321), bulk-string('"handshake"'), int(-1)]                                                                                         
failed check for replica role - current role: [bulk-string('"slave"'), bulk-string('"127.0.0.1"'), int(37321), bulk-string('"handshake"'), int(-1)]                                                                                         
failed check for replica role - current role: [bulk-string('"slave"'), bulk-string('"127.0.0.1"'), int(45495), bulk-string('"handshake"'), int(-1)]                                                                                         
                                                                                                                                                                                                                                            
thread 'test_sentinel_client' panicked at redis/tests/test_sentinel.rs:207:14:                                                                                                                                                              
called `Result::unwrap()` on an `Err` value: Connection refused (os error 111)                                                                                                                                                              
stack backtrace:                                                                                                                                                                                                                            
   0: rust_begin_unwind                                                                                                                                                                                                                     
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs:695:5                                                                                                                                          
   1: core::panicking::panic_fmt                                                                                                                                                                                                            
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/panicking.rs:75:14                                                                                                                                         
   2: core::result::unwrap_failed                                                                                                                                                                                                           
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/result.rs:1704:5                                                                                                                                           
   3: core::result::Result<T,E>::unwrap                                                                                                                                                                                                     
             at /home/murrax2/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1109:23                                                                                                    
   4: test_sentinel::test_sentinel_client                                                                                                                                                                                                   
             at ./tests/test_sentinel.rs:204:24                                                                                                                                                                                             
   5: test_sentinel::test_sentinel_client::{{closure}}                                                                                                                                                                                      
             at ./tests/test_sentinel.rs:200:26                                                                                                                                                                                             
   6: core::ops::function::FnOnce::call_once                                                                                                                                                                                                
             at /home/murrax2/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5                                                                                                
   7: core::ops::function::FnOnce::call_once                                                                                                                                                                                                
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ops/function.rs:250:5                                                                                                                                      
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
test test_sentinel_client ... FAILED

failures:

failures:
    test_sentinel_client

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 20 filtered out; finished in 0.59s

  TRY 3 FAIL [   0.596s] redis::test_sentinel test_sentinel_client
```

Occurs  when `get_major_version` is called, and doesn't when it is not.

nihohit:
That version isn't in our testing matrix - [I'll run a test on it, to see if there's a bug there](https://github.com/redis-rs/redis-rs/pull/1633). In the meanwhile, try installing redis/valkey 7.2.4, or another version from our testing matrix. Those should work.

nihohit:
Ok, double weird - the issue is that the call to ROLE returns `handshake` as the state of replication. That state doesn't appear either in the [redis ](https://redis.io/docs/latest/commands/role/)or [valkey ](https://valkey.io/commands/role/) ROLE docs. Are you using some modified DB?

MurrayGroves:
I was just using the latest version of `extra/redis` on the AUR. I uninstalled it and tried with the latest Valkey - which also didn't work. So I built 7.2.4 from source, which also has the same issue!

nihohit:
WAT.
Could it be that you have multiple installations of redis-server on your device? If you call `where redis-server` do you get the path to which you installed it?

MurrayGroves:
I only get `/usr/bin/redis-server`, which is where I put it after I built it. When I had it uninstalled, `redis-server` would give the standard "no such command" error.

nihohit:
Sorry, then I simply don't know how/why this happens.

MurrayGroves:
One of life's mysteries then! I'll just disable the sentinel tests locally I guess 🤔

MurrayGroves:
Added a link to the Redis docs for each command. Also, Clippy complains about me using uppercase for the `GENERIC` type placeholder. I chose uppercase to indicate that it was a special case, and not a normal type. Should I suppress the warning, or change it to `Generic` like Clippy requests?

nihohit:
> change it to Generic like Clippy requests

this.

nihohit:
@MurrayGroves tests are green, but this is still in draft mode. What's the next step? is this ready for review?

MurrayGroves:
- Write a macro for the iterator commands so we don't have the same logic in four places
- Add a bit to the main docs page basically just saying that these exist and you can use them
- Adapt a few tests over to use these

I think that's everything?

nihohit:
I think so.

MurrayGroves:
Rebased to fix lints, added missing `use` that my auto-formatter removed because I had the (locally broken) test which used it commented out

nihohit:
Hi, sorry that I didn't get to this PR in the last week. I'll try to go over again during the weekend. 
One thing that I noticed is that you didn't add docs in lib.rs - those are the docs that appear in docs.rs, which, from my experience, is where people to go in order to understand how to use the library.

MurrayGroves:
@nihohit build fixed!

MurrayGroves:
Happy to help, and thank you for all your comments and patience!

PR #1485, opened 2025-01-13
Add `transaction_async` function

alekece:
# Description
This PR aims to add the asynchronous version of the `redis::transaction` function.
To do so, I duplicated the implementation of `redis::transaction` and adapted it with the asynchronous constraints:
The function takes a `redis::aio::ConnectionLike` instead of`redis::ConnectionLike` and requires this argument to be cloneable (and avoid any problem with lifetime). The underlying connection should be `MultiplexedConnection` or `ConnectionManager` which are both no costly to clone.

### DIFF
--- a/redis/src/aio/connection.rs
+++ b/redis/src/aio/connection.rs
@@ -4,7 +4,8 @@
 use super::async_std;
 use super::ConnectionLike;
 use super::{setup_connection, AsyncStream, RedisRuntime};
-use crate::cmd::{cmd, Cmd};
+use crate::cmd::{cmd, pipe, Cmd};
+use crate::pipeline::Pipeline;
 use crate::connection::{
     resp2_is_pub_sub_state_cleared, resp3_is_pub_sub_state_cleared, ConnectionAddr, ConnectionInfo,
     Msg, RedisConnectionInfo,
@@ -24,6 +25,7 @@ use futures_util::{
     future::FutureExt,
     stream::{Stream, StreamExt},
 };
+use std::future::Future;
 use std::net::SocketAddr;
 use std::pin::Pin;
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
@@ -501,4 +503,36 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
             )))
         }
     })
+}
+
+/// This function, akin to `redis::transaction`, simplifies transaction management slightly, but asynchronously.
+pub async fn transaction_async<
+    C: ConnectionLike + Clone,
+    K: ToRedisArgs,
+    T: FromRedisValue,
+    F: Fn(C, Pipeline) -> Fut,
+    Fut: Future<Output = Result<Option<T>, RedisError>>,
+>(
+    mut connection: C,
+    keys: &[K],
+    func: F,
+) -> Result<T, RedisError> {
+    loop {
+        cmd("WATCH")
+            .arg(keys)
+            .exec_async(&mut connection)
+            .await?;
+
+        let mut p = pipe();
+        let response = func(connection.clone(), p.atomic().to_owned()).await?;
+
+        match response {
+            None => continue,
+            Some(response) => {
+                cmd("UNWATCH").exec_async(&mut connection).await?;
+
+                return Ok(response);
+            }
+        }
+    }
 }
--- a/redis/tests/test_async.rs
+++ b/redis/tests/test_async.rs
@@ -438,6 +438,37 @@ mod basic_async {
         );
     }
 
+    #[rstest]
+    #[case::tokio(RuntimeType::Tokio)]
+    #[cfg_attr(feature = "async-std-comp", case::async_std(RuntimeType::AsyncStd))]
+    fn test_transaction_async(#[case] runtime: RuntimeType) {
+        test_with_all_connection_types(
+            |con| async move {
+                let res: Vec<usize> = redis::aio::transaction_async(
+                    con,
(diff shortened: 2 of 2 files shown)


alekece:
@hatemMn

HatemMn (commented):
@alekece

HatemMn:
@nihohit  Here is an update for this older PR : https://github.com/redis-rs/redis-rs/pull/904

Can you please review the proposed changes ?

Thank you

HatemMn (approved):
Amazing !

nihohit:
The main issue I see is that because this is run on an async connection that could be used from other threads, there's nothing that promises you that external commands would be interleaved between the `WATCH`, the pipe and the `UNWATCH`, and thus potentially breaking the assumptions of this function.
I recommend investigating changing `F: Fn(C, Pipeline) -> Fut,` to `F: Fn(Pipeline) -> Pipeline` - letting the user only populate the pipeline, and then inside the `transaction_async` function building a pipeline that begins with `WATCH` and `MULTI`, and after the user populates it, ends with `EXEC` and `UNWATCH`, and querying the connection with that pipeline.

alekece:
@nihohit 
Taking into consideration your point, is this implementation correct ?

```rust
pub async fn transaction_async<
    C: ConnectionLike,
    K: ToRedisArgs,
    T: FromRedisValue,
    F: FnMut(&mut Pipeline),
>(
    connection: &mut C,
    keys: &[K],
    mut prepare: F,
) -> Result<T, RedisError> {
    loop {
        cmd("WATCH").arg(keys).exec_async(connection).await?;

        let p = pipe();

        prepare(pipe().atomic());

        let response = p.query_async(connection).await;

        if matches!(response, Ok(None)) {
            trace!("Transaction failed, retrying");
            continue;
        }

        cmd("UNWATCH").exec_async(connection).await?;

        // At this point, the response can only be `Ok(Some(T))` or `Err(RedisError)`.
        return response.map(Option::unwrap);
    }
}

```

There is two points I want to discuss before going further:
- The `transaction_async` function signature differs from the `transaction` one and it should be great to unify it.
- By taking only a pipeline for now, the user cannot (using this `transaction_async` function) watch some keys, perform some reading commands (like getting a counter) and populate the pipeline with it (see example below):
```rust
redis::cmd("WATCH").arg(key).exec_async(&mut conn).await?;
let counter = conn.get(key).await?;
redis::pipe().atomic().set(counter+1).query_async(&mut conn).await?;
// then unwatch
```
which is possible with my previous `transaction_async` implementation:
```rust
redis::aio::transaction_async(conn, &[key], |mut conn, mut pipe| async move {
   let counter = conn.get(key).await?;
   pipe.set(counter+1).query_async(conn).await
})
.await
```

nihohit:
> is this implementation correct ?

It isn't. Consider 2 concurrent calls to `transaction_async` - one might call `WATCH` and await, while the other completes and calls `UNWATCH`, thus implicitly cancelling the first transaction. Alternatively, both call `WATCH` with different keys, and then a change to the keys from the first transaction cancels the second transaction, and vice-versa, leading to a deadlock.

nihohit:
> The transaction_async function signature differs from the transaction one and it should be great to unify it.

That's a less important than having a function that does what it should, safely. If you can achieve that, great, but I doubt you will be able to do that.

> By taking only a pipeline for now, the user cannot (using this transaction_async function) watch some keys, perform some reading commands (like getting a counter) and populate the pipeline with it (see example below):

Right, replace `let counter = conn.get(key).await?;` with an arbitrary action on the connection. Is your proposal safe when ANYTHING can happen then?

nihohit:
Hi, sorry that it took me so long to get back to this, @alekece. I was convinced that this is the only working design, so if you want to rebase we can get this moving forward again. The only missing piece is to document the danger of using concurrent transactions on clones of the same multiplexed connections.

redis/src/aio/connection.rs
```diff
@@ -502,3 +505,88 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+>(
+    mut connection: C,
+    keys: &[K],
+    func: F,
+) -> Result<T, RedisError> {
+    loop {
+        cmd("WATCH").arg(keys).exec_async(&mut connection).await?;
+
+        let mut p = pipe();
+        let response = func(connection.clone(), p.atomic().to_owned()).await;
```
nihohit, line 580: instead of using an atomic pipeline, you should manually add before the closure the "MULTI" command, and after the closure the "EXEC" and "UNWATCH" commands.
That will save waiting for another back-and-forth between the client and the server for the UNWATCH.

redis/src/aio/connection.rs
```diff
@@ -502,3 +505,88 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+
+        if matches!(response, Ok(None)) {
```
nihohit, line 582: use let bindings or matches, to save the later unwrap.

nihohit (approved):

nihohit:
also needs testing for cluster connections.

LockedThread:
Any updates on this?

redis/src/aio/connection.rs
nihohit: this doesn't work.

nihohit:
@LockedThread I opened a rebased PR, based on the good work here - https://github.com/redis-rs/redis-rs/pull/1969
comments are welcome :)

nihohit:
https://github.com/redis-rs/redis-rs/pull/1969 was merged

PR #1487, opened 2025-01-13
feat(cluster): add `get(_async)?_connection_with_config`

Totodore:
This will allow to change some configuration fields based on the `ClusterConfig` type when creating the connection.
It is inspired from the `AsyncConnectionConfig` type for standalone async redis.

Here are the fields that can be set when calling `get_async_connection_with_config` and `get_connection_with_config`:
```rust
pub struct ClusterConfig {
    pub(crate) connection_timeout: Option<Duration>,
    pub(crate) response_timeout: Option<Duration>,
    #[cfg(feature = "cluster-async")]
    pub(crate) async_push_sender: Option<std::sync::Arc<dyn crate::aio::AsyncPushSender>>,
}
 ```

It a subset of the `ClusterParams` type and overwrite its fields if there are set on `ClusterConfig`.

* Fixes issue #1486

### DIFF
--- a/redis/src/cluster_client.rs
+++ b/redis/src/cluster_client.rs
@@ -121,6 +121,18 @@ impl ClusterParams {
             async_push_sender: value.async_push_sender,
         })
     }
+    fn apply_config(&mut self, config: cluster::ClusterConfig) {
+        if let Some(connection_timeout) = config.connection_timeout {
+            self.connection_timeout = connection_timeout;
+        }
+        if let Some(response_timeout) = config.response_timeout {
+            self.response_timeout = response_timeout;
+        }
+        #[cfg(feature = "cluster-async")]
+        if let Some(async_push_sender) = config.async_push_sender {
+            self.async_push_sender = Some(async_push_sender);
+        }
+    }
 }
 
 /// Used to configure and build a [`ClusterClient`].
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: redis/src/cluster_client.rs [hunk 1]
```diff
@@ -121,6 +121,18 @@ impl ClusterParams {
             async_push_sender: value.async_push_sender,
         })
     }
+    fn apply_config(&mut self, config: cluster::ClusterConfig) {
```
nihohit, line 124: nit: add empty line.

also, change to 
```rs
fn with_config(&self, config: cluster::ClusterConfig) -> Self {
  let new = self.clone();
   ....
```
Less room for error, IMO.

redis/src/cluster_client.rs
```diff
@@ -434,6 +446,21 @@ impl ClusterClient {
         cluster::ClusterConnection::new(self.cluster_params.clone(), self.initial_nodes.clone())
     }
 
+    /// Creates new connections to Redis Cluster nodes with a custom config and returns a
+    /// [`cluster_async::ClusterConnection`].
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if there is a failure while creating connections or slots.
+    pub fn get_connection_with_config(
```
nihohit, line 455: if the config is for the creation of async connections, why add this?,

Totodore: These fields:
```rust
    pub(crate) connection_timeout: Option<Duration>,
    pub(crate) response_timeout: Option<Duration>,
```
can also be applied  to a sync connection AFAIK. So I was thinking that it would be nice to also have this to match the async version. However I won't use it myself so I can remove this part.

redis/src/cluster_client.rs
Totodore: I think it makes more sense and it matches the builder pattern to take it by value and let the caller be responsible of the cloning or not : 
```rust
fn with_config(mut self, config: cluster::ClusterConfig) -> Self {
```

Are you ok with that? If not, I'll go in your direction.

redis/src/cluster_client.rs
nihohit: keep it, but adjust the config's documentation to remove the async part.

redis/src/cluster_client.rs
nihohit: that's fine too. It's specifically `& mut` which I try to avoid :)

Totodore:
> can you add or change an existing test to use this API, just to reduce the chance of accidental breakage?

I added a `pub_sub_subscription_with_config` test.

altanozlu (approved):

nihohit (approved):

nihohit:
Thanks for the contribution!

nihohit:
Thanks, this was released in 0.28.2

PR #1494, opened 2025-01-24
Fix RetryMethod Import

Braedon-Wooding-Displayr:
Noticed that while the `retry_method()` function was exported the type wasn't, this was due to it being missing from `lib.rs` it seems.

Could this be included in a minor release in version 28?


nihohit (approved):

nihohit:
Thanks, this was released in 0.28.2

PR #1497, opened 2025-01-27
feat: Include the reason for one connection failure if cluster connect fails

Marwes:
The error message if all cluster connections fail currently throws away all detail about the specific connection attempts. The `RedisError` type isn't great about dealing with nested errors but I think it would be preferable to at least display one error reason, though ideally we might want to spell out all failures.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -35,7 +35,7 @@
 //!     let client = ClusterClient::new(nodes).unwrap();
 //!     let mut connection = client.get_async_connection().await.unwrap();
 //!     let key = "test";
-//!     
+//!
 //!     redis::pipe()
 //!         .rpush(key, "123").ignore()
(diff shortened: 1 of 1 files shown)


nihohit (approved):
Thanks for the contribution!

I think that this is a fine solution, but just for the sake of discussion, let's consider some simple alternatives - 
we can add a RedisError variant that contains sub-errors, or just an arbitrary string, which will be the concatenation of all error messages. Which options is preferable?

ogrman:
As someone who works with the PR author I don't think it matters much to us whether or not it has it's own variant or not, we just want to be able to see a reason for the failure. To be it seems unlikely that we'd ever want to base our behavior on which error variant we're seeing. I guess a concatenation of errors would be useful for cases where more than one initial node is used (which we do not).

nihohit:
well, if it doesn't matter, then we'll just merge instead of waiting longer. Thanks for the contribution :)

PR #1498, opened 2025-01-28
Update rustls-native-certs.

nihohit:


nihohit:
@djc is this the correct way to handle errors here?

redis/src/connection.rs
```diff
@@ -884,8 +884,14 @@ pub(crate) fn create_rustls_config(
...
-    for cert in load_native_certs()? {
-        root_store.add(cert)?;
+    {
+        let certificate_result = load_native_certs();
+        if let Some(error) = certificate_result.errors.pop() {
+            return Err(error);
+        }
+        for cert in certificate_result.certs() {
+            root_store.add(cert)?;
+        }
```
djc, line 894: There are basically two reasonable approaches:

- Yield an error if rustls-native-certs found one or errors
- Yield an error if rustls-native-certs found less than x `certs`

(Also, the best solution is probably to drop the direct dependency on rustls-native-certs in favor of rustls-platform-verifier -- see the notice in the [README](https://github.com/rustls/rustls-native-certs).)

nihohit: I agree that using the platform verifier is better, but I don't see a way to use it with the[ supplied TLS params](https://github.com/redis-rs/redis-rs/blob/e2eb1ff1ab06c011bcd41dd0cf9b7ef7d9c3b409/redis/src/connection.rs#L894).

djc: Do you mean how to apply the `RootCertStore` that you build from the `TlsCertificates`? You can't directly, but you should be able to convert the `TlsCertficates::root_cert` into a `CertificateDer` which you can pass to [`Verifier::new_with_extra_roots()`](https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots).

nihohit: I couldn't find such a conversion interface.
Frankly, I just don't understand the different parts there, and I'm afraid to break something.

djc: CertificateDer implements From<Vec<u8>>, and also has a from_slice() method, is that what you meant?

Happy to advise on rustls stuff, given that all of it is private API we can always iterate if something breaks.

djc: (That is conversion from DER, if you need conversion from PEM you can use the PemObject trait like I did in my other PR.)

nihohit: Thanks for the offer!
So, what I don't understand -
1. Do I even need to create the `RootCertStore`? I assume that using the verifier means that the
```rs
for cert in load_native_certs()? {
    root_store.add(cert)?;
}
```
is no longer needed, but what about
```rs
#[cfg(feature = "tls-rustls-webpki-roots")]
root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
```
?

2. If I use the `CertificateDer` from `ClientTlsParams::client_cert_chain` to create the verifier, how do I sue the `ClientTlsParams::client_key`?

3. The only thing I see that I can get from the RootCertStore are distinguished names, which I assume that aren't convertible to CertificateDer. How do I use the user-supplied `tls_params.root_cert_store`?

In short, I don't really understand how to use the verifier, given the inputs I have - optional root stre, client cert chain & private key chain from the user, and `webpki_roots::TLS_SERVER_ROOTS`. 

The current code:

```rs
pub(crate) fn create_rustls_config(
    insecure: bool,
    tls_params: Option<TlsConnParams>,
) -> RedisResult<rustls::ClientConfig> {
  use crate::tls::ClientTlsParams;

  #[allow(unused_mut)]
  let mut root_store = RootCertStore::empty();
  #[cfg(feature = "tls-rustls-webpki-roots")]
  root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
  #[cfg(all(
      feature = "tls-rustls",
      not(feature = "tls-native-tls"),
      not(feature = "tls-rustls-webpki-roots")
  ))]
  for cert in load_native_certs()? {
      root_store.add(cert)?;
  }

  let config = rustls::ClientConfig::builder();
  let config = if let Some(tls_params) = tls_params {
      let config_builder =
          config.with_root_certificates(tls_params.root_cert_store.unwrap_or(root_store));

      if let Some(ClientTlsParams {
          client_cert_chain: client_cert,
          client_key,
      }) = tls_params.client_tls_params
      {
          config_builder
              .with_client_auth_cert(client_cert, client_key)
              .map_err(|err| {
                  RedisError::from((
                      "Unable to build client with TLS parameters provided.",
                      err.to_string(),
                  ))
              })?
      } else {
          config_builder.with_no_client_auth()
      }
  } else {
      config
          .with_root_certificates(root_store)
          .with_no_client_auth()
  };

djc: 1. In principle you don't need a `RootCertStore` when you use rustls-platform-verifier. If you want to implement additivity with (for example) webpki-roots, you want to use the [`Verifier::new_with_extra_roots()`](https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots) API.

2. The verifier is about verifying the peer's certificate. If client authentication is enabled, this is about supplying client credentials to the peer -- which doesn't have anything to do with the verifier. So you always call either `with_no_client_auth()` or `with_client_auth_cert()` **after** configuring verification.

3. See the `new_with_extra_roots()` API referenced above.

Hope that helps -- sorry for the delay.

PR #1512, opened 2025-02-04
Hm/review fixes

HatemMn:
Difference with previous work : 

- Updated F: Fn(C, Pipeline) -> Fut, to F: Fn(Pipeline) -> Pipeline like the maintainer asked
- Manually performed the wrapping the .atomic() function does, otherwise the DB will return an error of a conflicting MULTI / WATCH
- discard the intermediary potential outputs of the transaction
In more simple terms, this is the actual response : 
```
[ok, ok, simple-string("QUEUED"), simple-string("QUEUED"), simple-string("QUEUED"), simple-string("QUEUED"), array([ok, ok, bulk-string('"42"'), bulk-string('"21"')]), ok]
```

We only want that last array

- update docs, tests, etc

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -127,7 +127,7 @@ cache-aio = ["aio", "dep:lru"]
 tls = ["tls-native-tls"] # use "tls-native-tls" instead
 async-std-tls-comp = ["async-std-native-tls-comp"] # use "async-std-native-tls-comp" instead
 # Instead of specifying "aio", use either "tokio-comp" or "async-std-comp".
-aio = ["bytes", "dep:pin-project-lite", "dep:futures-util", "dep:tokio", "tokio/io-util", "dep:tokio-util", "tokio-util/codec", "combine/tokio"]
+aio = ["bytes", "dep:pin-project-lite", "dep:futures-util", "dep:tokio", "tokio/io-util", "dep:tokio-util", "tokio-util/codec", "combine/tokio", "dep:log"]
 
 [dev-dependencies]
 assert_approx_eq = "1.0"
--- a/redis/src/aio/connection.rs
+++ b/redis/src/aio/connection.rs
@@ -5,13 +5,13 @@ use super::async_std;
 use super::ConnectionLike;
 use super::{setup_connection, AsyncStream, RedisRuntime};
 use crate::cmd::{cmd, pipe, Cmd};
-use crate::pipeline::Pipeline;
 use crate::connection::{
     resp2_is_pub_sub_state_cleared, resp3_is_pub_sub_state_cleared, ConnectionAddr, ConnectionInfo,
     Msg, RedisConnectionInfo,
 };
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use crate::parser::ValueCodec;
+use crate::pipeline::Pipeline;
 use crate::types::{ErrorKind, FromRedisValue, RedisError, RedisFuture, RedisResult, Value};
 use crate::{from_owned_redis_value, ProtocolVersion, ToRedisArgs};
 #[cfg(all(not(feature = "tokio-comp"), feature = "async-std-comp"))]
@@ -30,6 +30,7 @@ use std::net::SocketAddr;
 use std::pin::Pin;
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use tokio_util::codec::Decoder;
+use log::trace;
 
 /// Represents a stateful redis TCP connection.
 #[deprecated(note = "aio::Connection is deprecated. Use aio::MultiplexedConnection instead.")]
@@ -511,7 +512,62 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
     })
 }
 
-/// This function, akin to `redis::transaction`, simplifies transaction management slightly, but asynchronously.
+/// Executes a Redis transaction asynchronously by automatically watching keys and running
+/// a transaction loop until it succeeds. Similar to the synchronous [`transaction`](crate::transaction)
+/// function but for async execution.
+///
+/// The provided closure may be executed multiple times if the transaction fails due to
+/// watched keys being modified between WATCH and EXEC. Any side effects in the closure
+/// should account for possible multiple executions.
+///
+/// # Arguments
+///
+/// * `connection` - A connection implementing [`ConnectionLike`](crate::aio::ConnectionLike) + Clone
+/// * `keys` - Array of keys to watch for changes during the transaction
+/// * `func` - Closure that implements the transaction logic. Will receive a cloned connection
+///   and an atomic pipeline. Should return `Result<Option<T>, RedisError>` where:
+///   - `None` indicates the transaction should be retried
+///   - `Some(T)` contains the final result if successful
+///   - `Err` aborts the transaction
+///
+/// # Type Parameters
+///
+/// * `C` - Connection type implementing ConnectionLike + Clone
+/// * `K` - Key type implementing ToRedisArgs
+/// * `T` - Return value type implementing FromRedisValue
+/// * `F` - Closure type taking (C, Pipeline) and returning Fut
+/// * `Fut` - Future returning Result<Option<T>, RedisError>
+///
+/// # Examples
+///
+/// ```rust,no_run
+/// use redis::{AsyncCommands, RedisResult, pipe};
+///
+/// async fn increment(con: redis::aio::ConnectionManager) -> RedisResult<isize> {
+///     let key = "my_counter";
+///     redis::aio::transaction_async(con, &[key], |mut con, mut pipe| async move {
+///         //  Get the current value
+///         let val: isize = con.get(key).await?;
+///         //   Create pipeline that increments and returns new value
+///         pipe.set(key, val + 1)
+///             .ignore()
+///             .get(key)
+///             .query_async(&mut con)
+///             .await
+///     })
+///     .await
+/// }
+/// ```
+///
+/// # Notes
+///
+/// - The closure provided to this function may be executed multiple times if the
+///   watched keys are modified by other clients between the WATCH and EXEC commands
+/// - Side effects in the closure should be idempotent since they may occur multiple times
+/// - The transaction is automatically rolled back if the closure returns an error
+/// - Returns the final result once the transaction succeeds
+///
+/// For more details on Redis transactions, see the [Redis documentation](https://redis.io/topics/transactions)
 pub async fn transaction_async<
     C: ConnectionLike + Clone,
     K: ToRedisArgs,
@@ -524,21 +580,19 @@ pub async fn transaction_async<
     func: F,
 ) -> Result<T, RedisError> {
     loop {
-        cmd("WATCH")
-            .arg(keys)
-            .exec_async(&mut connection)
-            .await?;
+        cmd("WATCH").arg(keys).exec_async(&mut connection).await?;
 
         let mut p = pipe();
-        let response = func(connection.clone(), p.atomic().to_owned()).await?;
+        let response = func(connection.clone(), p.atomic().to_owned()).await;
 
-        match response {
-            None => continue,
-            Some(response) => {
-                cmd("UNWATCH").exec_async(&mut connection).await?;
-
-                return Ok(response);
-            }
+        if matches!(response, Ok(None)) {
+            trace!("Transaction failed, retrying");
+            continue;
         }
+
+        cmd("UNWATCH").exec_async(&mut connection).await?;
+
+        // At this point, the response can only be `Ok(Some(T))` or `Err(RedisError)`.
+        return response.map(Option::unwrap);
     }
 }
--- a/redis/src/aio/connection.rs
+++ b/redis/src/aio/connection.rs
@@ -25,12 +25,11 @@ use futures_util::{
     future::FutureExt,
     stream::{Stream, StreamExt},
 };
-use std::future::Future;
+use log::trace;
 use std::net::SocketAddr;
 use std::pin::Pin;
 #[cfg(any(feature = "tokio-comp", feature = "async-std-comp"))]
 use tokio_util::codec::Decoder;
-use log::trace;
 
 /// Represents a stateful redis TCP connection.
 #[deprecated(note = "aio::Connection is deprecated. Use aio::MultiplexedConnection instead.")]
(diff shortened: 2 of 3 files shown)


redis/src/aio/connection.rs
```diff
@@ -24,6 +25,7 @@ use futures_util::{
     future::FutureExt,
     stream::{Stream, StreamExt},
 };
+use log::trace;
```
nihohit, line 28: please add a feature flag over this, I prefer not to add another dependency to `aio`.

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+/// function but for async execution.
+///
+/// The provided closure may be executed multiple times if the transaction fails due to
+/// watched keys being modified between WATCH and EXEC. Any side effects in the closure
+/// should account for possible multiple executions.
+///
+/// # Arguments
+///
+/// * `connection` - A connection implementing [`ConnectionLike`](crate::aio::ConnectionLike) + Clone
+/// * `keys` - Array of keys to watch for changes during the transaction
```
nihohit, line 525: nit: watched_keys

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+    mut connection: Connection,
+    keys: &[K],
+    pipeline: &mut Pipeline,
+) -> Result<T, RedisError> {
+    let mut transaction_pipeline = pipe(); // a long lived value that will be used to build the transaction pipeline
+    transaction_pipeline.cmd("WATCH").arg(keys).cmd("MULTI");
+
+    std::mem::take(&mut pipeline.commands) // consume the pipeline and give into_iter() ownership of the commands attribute
+        .into_iter()
+        .filter(|cmd| cmd.data != b"UNWATCH")
```
nihohit, line 584: I prefer to return an error here, instead of silently ignoring this

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+///
+/// The provided closure may be executed multiple times if the transaction fails due to
+/// watched keys being modified between WATCH and EXEC. Any side effects in the closure
+/// should account for possible multiple executions.
+///
+/// # Arguments
+///
+/// * `connection` - A connection implementing [`ConnectionLike`](crate::aio::ConnectionLike) + Clone
+/// * `keys` - Array of keys to watch for changes during the transaction
+/// * `func` - Closure that implements the transaction logic. Will receive a cloned connection
```
nihohit, line 526: please update the docs with the new values

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+///   - `None` indicates the transaction should be retried
+///   - `Some(T)` contains the final result if successful
+///   - `Err` aborts the transaction
+///
+/// # Type Parameters
+///
+/// * `C` - Connection type implementing ConnectionLike + Clone
+/// * `K` - Key type implementing ToRedisArgs
+/// * `T` - Return value type implementing FromRedisValue
+/// * `F` - Closure type taking (C, Pipeline) and returning Fut
```
nihohit, line 537: same, please update

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+/// * `Fut` - Future returning Result<Option<T>, RedisError>
+///
+/// # Examples
+///
+/// ```rust,no_run
+/// use redis::{AsyncCommands, RedisResult, pipe};
+///
+/// async fn increment(con: redis::aio::ConnectionManager) -> RedisResult<isize> {
+///     let key = "my_counter";
+///     redis::aio::transaction_async(con, &[key], |mut con, mut pipe| async move {
```
nihohit, line 547: same, update

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+///             .query_async(&mut con)
+///             .await
+///     })
+///     .await
+/// }
+/// ```
+///
+/// # Notes
+///
+/// - The closure provided to this function may be executed multiple times if the
```
nihohit, line 563: same

nihohit (commented):
Excellent work, thanks for moving this forwards.

Besides the inline comments, this change is missing a "sad path" test, where a transaction fails an retries - preferably multiple transactions. I think you might need to avoid the built-in test infra and use a multithreaded runtime for that, in order to run transactions really concurrently.

redis/Cargo.toml
```diff
@@ -127,7 +127,7 @@ cache-aio = ["aio", "dep:lru"]
 tls = ["tls-native-tls"] # use "tls-native-tls" instead
 async-std-tls-comp = ["async-std-native-tls-comp"] # use "async-std-native-tls-comp" instead
 # Instead of specifying "aio", use either "tokio-comp" or "async-std-comp".
-aio = ["bytes", "dep:pin-project-lite", "dep:futures-util", "dep:tokio", "tokio/io-util", "dep:tokio-util", "tokio-util/codec", "combine/tokio"]
+aio = ["bytes", "dep:pin-project-lite", "dep:futures-util", "dep:tokio", "tokio/io-util", "dep:tokio-util", "tokio-util/codec", "combine/tokio", "dep:log"]
```
nihohit, line 130: see comment - please remove the added dependency.

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+/// - Returns the final result once the transaction succeeds
+///
+/// For more details on Redis transactions, see the [Redis documentation](https://redis.io/topics/transactions)
+pub async fn transaction_async<
+    Connection: ConnectionLike + Clone,
+    K: ToRedisArgs,
+    T: FromRedisValue,
+>(
+    mut connection: Connection,
+    keys: &[K],
```
nihohit, line 576: I wonder if this should take a configuration object, so that we could add later things like limited number of retries. WDYT?

alekece: I agree. It could be an excellent addition and future-proof way of tweaking the transaction.

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+        .filter(|cmd| cmd.data != b"UNWATCH")
+        .for_each(|cmd| {
+            transaction_pipeline.add_command(cmd);
+        });
+    transaction_pipeline.cmd("EXEC");
+
+    loop {
+        let response: Result<Option<Vec<_>>, RedisError> =
+            transaction_pipeline.query_async(&mut connection).await;
+        cmd("UNWATCH").exec_async(&mut connection).await?; //  // Always clean up the WATCH after the transaction attempt, regardless of success/failure
```
alekece, line 593: We may put the `UNWATCH` command directly at the very end of the pipeline ?

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+///
+/// For more details on Redis transactions, see the [Redis documentation](https://redis.io/topics/transactions)
+pub async fn transaction_async<
+    Connection: ConnectionLike + Clone,
+    K: ToRedisArgs,
+    T: FromRedisValue,
+>(
+    mut connection: Connection,
+    keys: &[K],
+    pipeline: &mut Pipeline,
```
alekece, line 577: We probably want to take a `Cmd` here instead of a not-used pipeline

redis/src/aio/connection.rs
```diff
@@ -508,3 +510,101 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+///
+/// - The closure provided to this function may be executed multiple times if the
+///   watched keys are modified by other clients between the WATCH and EXEC commands
+/// - Side effects in the closure should be idempotent since they may occur multiple times
+/// - The transaction is automatically rolled back if the closure returns an error
+/// - Returns the final result once the transaction succeeds
+///
+/// For more details on Redis transactions, see the [Redis documentation](https://redis.io/topics/transactions)
+pub async fn transaction_async<
+    Connection: ConnectionLike + Clone,
```
alekece, line 571: We can relax the type  here by removing the `Clone` constraint

HatemMn (commented):
Partial review, I still have to do the  "sad path" test I guess, if mandatory

I added a feature called "verbose" , this is discussable as we might as well simply not log

Finally I have a question, there is a non public function called `execute_transaction_async` in the `redis/src/pipeline.rs` file.  Can you please tell me what is the difference between this function and what we are currently developing, and why is it private ?

Thanks

redis/src/aio/connection.rs
nihohit: I prefer not to add an explicit feature to the crate - we have too many of them. I think that you can use `#[cfg(feature = "log")]` to check if the `log` crate is added as a feature.

redis/src/aio/connection.rs
nihohit: I don't think that the UNWATCH won't be executed. a `Pipeline` object is a collection of commands sent to the server. The server sees them just as if each command on the pipeline was sent separately, and early commands failure doesn't affect the next commands just because they're in a pipeline together. so executing command after a pipeline is equivalent to adding it to the pipeline - only, if the command is added to the pipeline, no other command can be interleaved, so its actually safer.

redis/src/aio/connection.rs
nihohit: `ignored_commands` are handled inside `Pipeline::query_async`, IIRC.

nihohit:
> Partial review, I still have to do the "sad path" test I guess, if mandatory

Well, the whole point of a transaction is the sad path. Without testing it, your test would've likely to have passed if you just ran the inputted pipeline, without any addition :)

> Finally I have a question, there is a non public function called execute_transaction_async in the redis/src/pipeline.rs file. Can you please tell me what is the difference between this function and what we are currently developing, and why is it private ?

That function wraps the pipeline with `MULTI/EXEC` pair, but doesn't handle `WATCH`. you need to add an API to received keys for watching.

redis/src/aio/connection.rs
HatemMn: I forgot to mention this has been added, mark solved if necessary please

redis/src/aio/connection.rs
HatemMn: okay so It's fine this way IG

HatemMn:
> > Partial review, I still have to do the "sad path" test I guess, if mandatory
> 
> Well, the whole point of a transaction is the sad path. Without testing it, your test would've likely to have passed if you just ran the inputted pipeline, without any addition :)

Can you please provide me of a similar (if any) test on the `redis-rs` codebase so that I can code a test that is similar to how the codebase typically works ?

nihohit:
I started writing the test, but then I realized that I don't understand the usecase you're trying to achieve here. 

The [redis documentation that explains WATCH](https://redis.io/docs/latest/develop/interact/transactions/) (scroll down to "Optimistic locking using check-and-set") gives an example where the user performs some actions *that return values to the user* between the WATCH and the transaction. If the only thing that the user wants is to send several commands that will be executed together, then `Pipeline::atomic` is sufficient for that. looks like WATCH only makes sense when you *don't* pipeline the commands.

So, I don't understand the usecase for this addition. Can you please explain it to me? Once we have the usecase, we'll have the test :)

HatemMn:
> I started writing the test, but then I realized that I don't understand the usecase you're trying to achieve here.
> 

Okay let's get back to the original message with which we started this : https://github.com/redis-rs/redis-rs/pull/904

Basically, the problem that some users of redis-rs face ( me & friends, for the least ) have trouble making updates whenever we migrate from a legacy code that uses the [transaction function](https://docs.rs/redis/latest/redis/fn.transaction.html) along with some [synchronous manager like this one](https://docs.rs/redis/latest/redis/struct.Connection.html) toward a codebase that will use a [ConnectionManager](https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html) instead.

**The main problem :** 
Clearly, we fail because [transaction](https://docs.rs/redis/latest/redis/fn.transaction.html) takes a [ConnectionLike](https://docs.rs/redis/latest/redis/trait.ConnectionLike.html) but [ConnectionManager](https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html) does not implement that.

As you said, the absence of the 'asyncTransaction' is compensated by the the presence of the `Pipeline:atomic` flow, but that's not clearly documented / stated - even tho it's written in some forums. A function that "wraps" all that boilerplate code would really be apreciated. Also that is why @alekece proposed the following PR : https://github.com/redis-rs/redis-rs/pull/1485

**The other problem:**

As you said previously, I cite : 
> The main issue I see is that because this is run on an async connection that could be used from other threads, there's nothing that promises you that external commands would be interleaved between the WATCH, the pipe and the UNWATCH, and thus potentially breaking the assumptions of this function. I recommend investigating changing F: Fn(C, Pipeline) -> Fut, to F: Fn(Pipeline) -> Pipeline - letting the user only populate the pipeline, and then inside the transaction_async function building a pipeline that begins with WATCH and MULTI

This kindof, indeed, changed the original purpose of all of this (which was having the exact same signature of `transaction` but as an async function). But If I understood your statement correctly, it's **impossible** to have such a function while still maintaining the safety garantees of redis-rs , right ? This is why, in my current proposal, I proposed a 'weaker' version of the boilerplate we originally wanted. It might not be what we want, but if we want a (theorically) unsafe function, then at least the version I proposed can be a helpful wrapper

nihohit:
Sorry, my question might've been unclear. I'm asking about a concrete workflow for tests - can you give me an example of an order of operations that could give a wrong result if it just used `Pipeline::atomic`, but would be safer or less error-prone using this function?

AFAIK, the WATCH function is usable only when you have a back-and-forth with the server before the transaction, and AFAIS this design makes it impossible. If this is so, it's my fault - I shouldn't have suggested a change without having a sample real-world case to understand the usecase behind this. I do apologize for that.

so, my question is - what is the real-world usecase that this implementation supports? If there is one, great, let's move forward with this, and write tests that simulate it. If not, let's find the use-case that doesn't have good coverage in the crate, and design a solution around it.

nihohit:
https://github.com/redis-rs/redis-rs/pull/1969 was merged

Issue #1481, opened 2025-01-09
make redis-rs sans-io


GlenDC:
There were proposals in the past to support aws-lc-rs as alternative to Ring, and I think that's a valid request and one that I also can help tackle at some point.

However as a first starting point i would like to separate the IO code (network layer etc) into either a separate module or a separate crate, I'm fine with both but prefer the latter. This to allow a redis-rs user to bring their own IO layer (network stack) and not end up with a binary with 2 or more network stacks just because they use redis-rs.

Even worse than the waste is that sometimes it's not even possible to have an additional stack, as for example Ring can compile-collide if you already have another incompatible stack. Etc...

I understand that this is rather niche because not many work on low level network stack stuff like me, but this change if done right should not remove any features for most users as they'll use the default stack provided by the redis-rs team, but it would unblock cases like mine.

What would be the first step once this work is approved? An outline from me on how I see this? Or I just straight up make a draft PR?

nihohit:
Hi Glen, thanks for the initiative!
This sounds like a rather large undertaking, and I'd really hate to waste your time here, so I think that the best course of action will be to start from getting rough goals and milestones, just so all of us understand the scope of this change, and making sure that this is a realistic project.

How well, do you know redis-rs? Would you like to schedule a video meeting so that I'll go through the library with you, or would you prefer to continue the conversation here?

GlenDC:
> This sounds like a rather large undertaking

Agreed, but it's either this or making my own redis library from scratch and I would hate that... Seems even more wasteful

I know redis itself pretty well and was also a past contributor to ledisdb (a Go sister project).

Redis-rs I don't know super well, but I think I'll be fine. I would need your guidance mostly here and in the PRs.

> so I think that the best course of action will be to start from getting rough goals and milestones, just so all of us understand the scope of this change, and making sure that this is a realistic project.

Great that's what I imagined you would want first. So my next step will be to provide you that here in my next comment somewhere in the next days.

nihohit:
Excellent!

I don't know if you know the `redis-protocol` crate which can handle a lot of the parsing of the redis messages. `redis-rs` is older than that crate, which means that ATM the two crates aren't compliant (I'm planning on integrating it into this repo), but if you'll decide that this is too large an undertaking, maybe it can help you developing your own client.

GlenDC:
Nice, maybe could be part of this work, would make it even more critical to define milestones and different steps correctly, but they are related.

[...]

PR #1514, opened 2025-02-04
refactor redis-rs to be sans-io backwards compat (sync only)

GlenDC:
In function of #1481,
for now as discussed the sync (non-aio) API's only, once we get this merged I can tackle the async version as well.

I think this PR covers the entire sync surface layer, it was a bit tricky to do this fully backwards compatible, which also means I had to make some design choices that I otherwise would not have done.

Do let me know if I missed something,
or need to do stuff different.
It's the first contribution to this codebase,
so I might also have done some things wrong
that are obvious to regular contributors. Do let me know.


nihohit:
Hi Glenn, thanks for contribution!

This is (understandably :) ) a very large change, and it would take me a while to review it. Apologies in advance. It would be helpful if you could provide some high-level overview of what you've done, to guide me through the review process.

GlenDC:
Take your time. I’m not in a hurry and busy with plenty of stuff. I’ll provide you with a summary somewhere in the next 24h.

nihohit:
I should thank you just for splitting `connection.rs` - it was on my plate for quite a while, and I never got to taking care of it.

The overall design looks good, and I've started looking at the code. I'll comment as I go.

nihohit:
oh, one last question, if you remember - which files contain actual additions, and which are just copied text from other files, or changes in imports?

GlenDC:
> which files contain actual additions, and which are just copied text from other files, or changes in imports?

No changes other than different import path for `use` statements:

- `redis/src/aio/async_std.rs`
- `redis/src/aio/mod.rs`
- `redis/src/aio/tokio.rs`
- `redis/src/cluster_client.rs`
- `redis/src/lib.rs`

No changes other than moved around:

- `redis/src/connection/addr.rs` new file, containing the `ConnectionAddr` code previously living inside of `src/connection.rs`
- `redis/src/connection/info.rs` new file, containing utility functions  and ConnectionInfo-like types all previously living inside of `src/connection.rs`
- `redis/src/connection/io/*.rs`: new files:
   - these contain the new `ConnectionDriver` trait that one can implement to support their own IO for sync Redis-rs
       - this trait is pretty much the only new code in this module
   - it also contains the `ActualConnection` enum and its logic previously found in `src/connection.rs`
   - it contains in sub files the constructors and data structures for the different variants of `ActualConnection` each in their own file and all of which was previously living in `src/connection.rs`
-`src/tls.rs` code has been moved into `io/tls_rustls.rs` for the rustls part and a parent mod for the enum.

Files with minor changes that are only additive in nature:

- `redis/src/client.rs` (extra methods to support connecting using your own IO
- `redis/src/cluster.rs` support for `Connect` impl for `GenericConnection<IO>`
- `redis/src/commands/mod.rs` support for `PubSubCommands` impl for `GenericConnection<IO>`

Changes that are hopefully backwards compatible:

- `redis/src/sentinel.rs`: add a generic with a default to `ActualConnection`
   - alternative that we can do instead is to make it also a wrapper around a generic version, but I think this should work as well? Up to you


No changes other than a comment:

- `redis/src/r2d2.rs`

GlenDC:
FYI this test fails, but given I am a bit new in this codebase it would be helpful if you can guide me in the right direction and perhaps give some of your own insights on why that might be @nihohit : 

```
running 1 test
test basic::test_connection_timeout_on_busy_server ... FAILED

failures:

failures:
    basic::test_connection_timeout_on_busy_server

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 74 filtered out; finished in 0.91s

──── TRY 3 STDERR:       redis::test_basic basic::test_connection_timeout_on_busy_server
thread 'basic::test_connection_timeout_on_busy_server' panicked at redis/tests/test_basic.rs:2178:9:
assertion failed: try_connect.is_err_and(|err| { err.is_timeout() })
stack backtrace:
```

nihohit:
@GlenDC sorry, took me a while to get back to you. I pulled the PR and `test_connection_timeout_on_busy_server` passes, so I can't fix the problem that I can't see. Does it fail locally for you?
Sadly, this condition doesn't log the source of the issue - did the connection attempt succeed, or did it fail, but with a different error?

Also, I see that there are some merge conflicts. I suggest that you don't rebase yet, since https://github.com/redis-rs/redis-rs/pull/1523 is likely to introduce some more conflicts.

nihohit:
My mistake, I needed to call the test with `REDISRS_SERVER_TYPE=tcp+tls RUST_BACKTRACE=1 cargo nextest run --locked -p redis --features=tls-native-tls test_connection_timeout_on_busy_server`. 
The test failed because the connection succeeded. 

I'm not sure why - I don't see any obvious change in the code, and when adding prints I see that the timeout seems to be passed correctly.

redis/src/client.rs
```diff
@@ -56,6 +60,14 @@ impl Client {
         connect(&self.connection_info, None)
     }
 
+    /// Same as [`Self::get_connection`] but with a custom IO Driver.
+    pub fn get_connection_with_io<IO: io::ConnectionDriver>(
+        &self,
+        driver: IO,
```
nihohit, line 66: nit: I prefer to use `driver: impl io::ConnectionDriver`

redis/src/client.rs
```diff
@@ -65,6 +77,15 @@ impl Client {
         connect(&self.connection_info, Some(timeout))
     }
 
+    /// Same as [`Self::get_connection_with_timeout`] but with a custom IO Driver.
+    pub fn get_connection_with_io_and_timeout<IO: io::ConnectionDriver>(
+        &self,
+        driver: IO,
```
nihohit, line 83: same

redis/src/connection/io/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+    fn send_bytes(&mut self, bytes: &[u8]) -> RedisResult<Value>;
+
+    /// Read a Redis value from the connection driver.
+    fn read_value(&mut self, parser: &mut Parser) -> RedisResult<Value>;
+
+    /// Reports if the connection driver is open.
+    fn is_open(&self) -> bool;
+
+    /// Closes the connection driver.
+    fn close_connection(&mut self);
```
nihohit, line 54: why is this needed? wouldn't it just be better to close on drop?

redis/src/r2d2.rs
```diff
@@ -3,6 +3,11 @@ use crate::sentinel::LockedSentinelClient;
 use crate::types::closed_connection_error;
 use crate::{ConnectionLike, RedisError};
 
+// NOTE: GenericConnection and generic friends
```
nihohit, line 6: generic friends? what does that mean?

redis/src/sentinel.rs
```diff
@@ -117,15 +117,17 @@ use crate::aio::MultiplexedConnection as AsyncConnection;
...
+    connection::{ActualConnection, ConnectionInfo},
+    types::RedisResult,
+    Client, Cmd, Connection, ConnectionDriver, ErrorKind, FromRedisValue, GenericConnection,
+    IntoConnectionInfo, RedisConnectionInfo, TlsMode, Value,
 };
 
 /// The Sentinel type, serves as a special purpose client which builds other clients on
 /// demand.
-pub struct Sentinel {
+pub struct Sentinel<IO = ActualConnection> {
```
nihohit, line 128: I'm not sure that this will work when adding sans-io support for aio work in sentinel - the users will need to declare 2 generic types, even if they care only about one, right?

nihohit (commented):
small comments. The work looks very impressive, thanks!

redis/src/sentinel.rs
GlenDC: I'm not sure i follow here. Also I realize after contributing to this crate that my redis-client usage was pretty simple and minimal.

Can you give some more context around `Sentinel` and also explain why this might be an issue for async IO? Is it because sentinel handles both the async api and the regular one? Would it perhaps make sense to split that up, to no longer do that if that is the case?

redis/src/r2d2.rs
GlenDC: Hmm probably a bad habit from mine to word it like that. Not even sure what I mean anymore. I think I meant to say that if we do the same thing for async that it would also not be supported here, but than again wouldn't be relevant anyway as I don't think this is ever used for that purpose?

redis/src/connection/io/mod.rs
GlenDC: Probably yes. I think I simply took it over from the original API and moved it into the trait.
Your comment is valid though. Want me to try to remove this and move it into drop logic?

redis/src/client.rs
GlenDC: Sure thing, will address it once I also understand better your other comments. This way I can handle it together with this one.

redis/src/client.rs
GlenDC: Same as https://github.com/redis-rs/redis-rs/pull/1514#discussion_r1957400543

redis/src/connection/io/mod.rs
nihohit: Yes, please. I prefer to make public traits minimal.

redis/src/r2d2.rs
nihohit: no, the r2d2 implementation AFAIK doesn't work with async connections, and it seems there's no public request for that :)

redis/src/sentinel.rs
nihohit: >  Is it because sentinel handles both the async api and the regular one?

Exactly. it contains both `async_connections_cache` and `connections_cache`. Adding the generic boundary to `connections_cache` implies that there will be a separate generic boundary for `async_connections_cache`, which will complicate their usage.
In any case, for backwards compatibility you'll have to use the same trick you used for `Connection` - Implement a `GenericSentinel<IO>` and then define `pub struct Sentinel(GenericSentinel<Connection>)`.

In any case, I think that you don't need to implement sentinel initially. I think It's a niche feature, and it can keep using `Connection` until there will be a request to implement it using sans-io, too.

nihohit: > Would it perhaps make sense to split that up, to no longer do that if that is the case?

Maybe, but again, I'd like to avoid backwards incompatibility.

redis/src/client.rs
GlenDC: It's not possible to use `impl` here though, as we also need to specify the same type as RedisResult<GenericConnection<IO>> which AFAIK you cannot make clear if you have 

```
pub fn get_connection_with_io_and_timeout(
        &self,
        driver: impl io::ConnectionDriver,
        timeout: Duration,
    ) -> RedisResult<GenericConnection<impl io::ConnectionDriver>>
```

^ the above works but you lose guarantee that the result type is the same as the input type

GlenDC: When I wrote my originalk reply 13 hours ago I hadn't read your `impl` part, I though you wanted to name it `Driver` instead of `IO`, which does make sense if you want that for some reason

redis/src/connection/io/mod.rs
GlenDC: It's a change from original code though as the original `Connection` used to have already 'close_connection' which now is changed to the below because of your request:

```rust
fn close_connection(&mut self) {
        // Notify the PushManager that the connection was lost
        self.send_disconnect();
        // NOTE: In original implementation this used to call `self.con.close_connection()`
        // now this happens in Drop
    }
```

Might be fine, but just saying that was the reason why I moved it also in the trait, to stay compatible with exact original behaviour

redis/src/r2d2.rs
GlenDC: Removed comment as this is no longer an issue, it can be requested if ever needed.

GlenDC:
> Also, I see that there are some merge conflicts. I suggest that you don't rebase yet, since https://github.com/redis-rs/redis-rs/pull/1523 is likely to introduce some more conflicts.

Is now merged already as well. Shall I rebase already or still wait for some other stuff?

nihohit:
You can rebase now, sorry for the hassle.

GlenDC:
> You can rebase now, sorry for the hassle.

Ok, should be good now 🤷

nihohit:
> When running the test separate it doesn't fail however.

Maybe because of the feature flags you use when running it separately? How do you run it separately?

nihohit:
In any case, the only advice I have is to go line-by-line in `Client::get_connection_with_timeout`, and print how long it takes, in order to find where the error appears.

Issue #1517, opened 2025-02-06
Client Side Cache with json.mget


kudlatyamroth:
When using JSON.MGET command with client side cache it get one more result then it is specified

i think it is because this line: https://github.com/redis-rs/redis-rs/blob/70da0c0ff82cacdc6c639b8530e2b6676f59d192/redis/src/caching/cache_manager.rs#L115

it doesn't consider path param from json command

nihohit:
Thanks for reporting this. 
@altanozlu, can you please take a look?

kudlatyamroth:
probably easiest fix will be by adding if statement like:
```
if command_name_str.starts_with("JSON") && cmd_len(cmd) - 2 == i {
    continue
}
```

nihohit:
you're welcome to open a PR with fix & test.

altanozlu:
Thanks for reporting it definitely requires a special case. I won't be able to provide a patch for a week. Using 'JSON.GET key path' as a singular command is needed when 'JSON.MGET key1 key2 path'  is provided. 
When I'm back I will provide patches for command coverage in tests.

kudlatyamroth:
ok PR opened #1520

PR #1520, opened 2025-02-07
Fix caching with the JSON module

kudlatyamroth:
fix problem when using json mget with client side cache returns one more result as None because it treats json command path param as key

fix issue: #1517


redis/tests/test_cache.rs
```diff
@@ -133,6 +134,79 @@ fn test_cache_mget(#[case] runtime: RuntimeType) {
...
+            let res2: Vec<Option<String>> = get_cmd("JSON.MGET", true)
+                .arg("key_1")
+                .arg("key_3")
+                .arg("key_2")
+                .arg(".")
+                .query_async(&mut con)
+                .await?;
+
+            assert_eq!(res2.len(), 3);
+            assert_eq!(res2, vec![Some(value_1.clone()), Some(value_2), None]);
```
altanozlu, line 189: Naming between key and value is mixed like key 3's vallue is named value_2

redis/src/caching/cache_manager.rs
```diff
@@ -111,25 +111,34 @@ impl CacheManager {
                     let mut commands = vec![];
                     let mut response: Vec<Value> = vec![];
                     let mut key_test_buffer: Vec<u8> = Vec::new();
+                    let arguments = cmd.args_iter().skip(1); // Skip the command name
+                    let last_key_index = arguments.clone().count() - 1; // Count arguments and adjust for zero-based index
```
altanozlu, line 115: Maybe you can use cmd len variable used above, instead of cloning?

redis/src/caching/cache_manager.rs
```diff
@@ -111,25 +111,34 @@ impl CacheManager {
...
                         if let crate::cmd::Arg::Simple(redis_key) = x {
                             key_test_buffer.clear();
                             key_test_buffer.extend_from_slice(single_command_name);
                             key_test_buffer.extend_from_slice(redis_key);
 
-                            if let Some(value) = self.get(redis_key, &key_test_buffer) {
-                                response.push(value);
-                                continue;
+                            // Check if the command is a JSON command and the key is the last one
+                            let is_json_last_key = command_name_str.starts_with("JSON") && (i == last_key_index);
```
altanozlu, line 124: command_name_str.starts_with("JSON")
Part can be calculated out of this loop

redis/src/caching/cache_manager.rs
```diff
@@ -111,25 +111,34 @@ impl CacheManager {
...
                     let mut key_test_buffer: Vec<u8> = Vec::new();
+                    let arguments = cmd.args_iter().skip(1); // Skip the command name
+                    let last_key_index = arguments.clone().count() - 1; // Count arguments and adjust for zero-based index
 
-                    for (i, x) in cmd.args_iter().skip(1).enumerate() {
+                    for (i, x) in arguments.enumerate() {
                         if let crate::cmd::Arg::Simple(redis_key) = x {
                             key_test_buffer.clear();
                             key_test_buffer.extend_from_slice(single_command_name);
                             key_test_buffer.extend_from_slice(redis_key);
```
altanozlu, line 121: If it's json.mget, extend key test buffer with path argument so it won't use cached value of another path of same key.

redis/src/caching/cache_manager.rs
```diff
@@ -111,25 +111,34 @@ impl CacheManager {
...
-                            commands.push((
-                                i,
-                                MultipleCachedCommandPart {
-                                    redis_key,
-                                    cmd_key: key_test_buffer.clone(),
-                                },
-                            ))
+
+                            if is_json_last_key {
+                                commands.push((i, command_part));
```
altanozlu, line 131: Is it needed to send path as a command?

altanozlu (changes requested):
Thanks for picking it up, please extend tests with different paths for same key.

redis/src/caching/cache_manager.rs
kudlatyamroth: cmd_len was my first approach, but cmd_len returns string lenght, not how many arguments was provided...
and we need to skip last argument :)

redis/src/caching/cache_manager.rs
kudlatyamroth: we need that when there are not cached keys, so we need to provide path to command

kudlatyamroth:
ok, i think i check every comment and make some adjustments

i also make some small refactor of code to split it into smaller chunks :)

redis/src/caching/cache_manager.rs
```diff
@@ -87,77 +87,168 @@ impl CacheManager {
...
+
+                match self.get(redis_key, &key_test_buffer) {
+                    Some(value) => response.push(value),
+                    None => {
+                        response.push(Value::Nil);
+                        commands.push((i, command_part));
+                    }
+                }
+            }
+        }
```
altanozlu, line 161: Just a wild idea of having tail_args in `CacheableCommand::Multiple` which would make it better IMHO
```suggestion
        let mut arguments: Vec<_> = cmd
            .args_iter()
            .skip(1)
            .filter_map(|arg| {
                if let crate::cmd::Arg::Simple(simple_arg) = arg {
                    Some(simple_arg)
                } else {
                    None
                }
            })
            .collect();
        let json_path_key = if is_json_command {
            let key = arguments.pop().unwrap();
            tail_args.push(key);
            Some(key)
        } else {
            None
        };
        let mut key_test_buffer: Vec<u8> = Vec::new();

        for (i, redis_key) in arguments.iter().enumerate() {
            self.prepare_key_buffer(
                &mut key_test_buffer,
                single_command_name,
                redis_key,
                json_path_key,
            );
            match self.get(redis_key, &key_test_buffer) {
                Some(value) => response.push(value),
                None => {
                    let command_part = MultipleCachedCommandPart {
                        redis_key,
                        cmd_key: key_test_buffer.clone(),
                    };

                    response.push(Value::Nil);
                    commands.push((i, command_part));
                }
            }
        }

```

redis/src/caching/cache_manager.rs
```diff
@@ -87,77 +87,168 @@ impl CacheManager {
...
+
+                self.prepare_key_buffer(
+                    &mut key_test_buffer,
+                    single_command_name,
+                    redis_key,
+                    last_key,
+                    is_json_command,
+                );
+
+                let command_part = MultipleCachedCommandPart {
```
altanozlu, line 141: Cloning the vector always worries me little bit :)

altanozlu (approved):
Great work for tests and breaking up the function!  It inspired me to improve current situation regarding json paths, until then this patch fixes the issue and tests passes.

redis/src/caching/cache_manager.rs
kudlatyamroth: i don't see where tail args are defined?

redis/src/caching/cache_manager.rs
kudlatyamroth: for me its raw unwrap, i know there should be value, but if somehow it will be None, we get error and don't know why...
for me its better expect or if we want to ignore this, we can use unwrap_or

redis/src/caching/cache_manager.rs
altanozlu: Sorry couldn't give the whole patch, diff is here: https://gist.github.com/altanozlu/07bb318bc5e806f578a409a0a44263d0

You don't have to do it, I was just playing with the idea.

redis/src/caching/cache_manager.rs
altanozlu: I agree unwrap counts should be reduced, also I was talking about ` cmd_key: key_test_buffer.clone(),` it's a pre-clone if 90% of commands are cached then it's just wasted clones. You can duplicate construction of MultipleCachedCommandPart to only when `commands.push((i, command_part));` is needed.
It's an optional thing and I can refactor this part with real perf benchmark later on.

kudlatyamroth: ok, changes pushed :)
changed part with unwrap with this:

```
        let json_path_key = is_json_command.then(|| {
            arguments.pop().map(|k| {
                tail_args.push(k);
                k
            })
        }).flatten();
```

kudlatyamroth:
fixed lint error

altanozlu (approved):

nihohit (approved):

Issue #1521, opened 2025-02-11
Docstrings are incorrect for `ttl` and `expire_time`


clbarnes:
```rust
    /// Get the time to live for a key in seconds.
    fn expire_time<K: ToRedisArgs>(key: K) {
        cmd("EXPIRETIME").arg(key)
        }
```

but the redis docs for EXPIRETIME say

> Returns the absolute Unix timestamp (since January 1, 1970) in seconds at which the given key will expire.

Also

```rust
    /// Get the expiration time of a key.
    fn ttl<K: ToRedisArgs>(key: K) {
        cmd("TTL").arg(key)
    }
```

where the redis docs for TTL say

> Returns the remaining time to live of a key that has a timeout.

We should document the TTL function as returning the time to live, and document the expire time function as returning the expiration time.

These errors are also present in the millisecond variants pttl and pexpire_time.

PR #1522, opened 2025-02-11
Improve docs around TTLs/ expiry times.

clbarnes:
See #1521


nihohit (approved):
Thanks for the correction!

Issue #1470, opened 2025-01-07
Move some feature flags into configuration


altanozlu:
There are feature flags like `tcp_nodelay, disable-client-setinfo, keep-alive` they could easily change behaviour of other libraries if they all depend on redis-rs and has different features.
Find features that should be in configuration of a connection and move them into configuration.

nihohit:
fixed in https://github.com/redis-rs/redis-rs/pull/1669

Issue #1147, opened 2024-04-10
Feature request: Add support for USER_TCP_TIMEOUT as a socket option.


barshaul:
We want to configure connections with the USER_TCP_TIMEOUT option. At the moment, we encounter situations where the server's host doesn't respond with an ACK to a transmitted message, leaving the connection hanging on send_recv. This issue was resolved by setting the `socket2` `set_tcp_user_timeout` function:

```rust
socket2.set_tcp_user_timeout(Duration::from_secs(5))?;
```

For deeper insights into why keep-alive isn't always sufficient, see this blog: [When TCP Sockets Refuse to Die](https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/).

To reproduce this issue, you can set an IP table rule on the server host to drop all received packats:
``` 
sudo iptables -I INPUT RuleNumber -i interfaceName -p tcp --dport port -j DROP
```

### Suggestion 1:
I propose using the existing "keep-alive" feature or introducing a new one named "socket-options", that would provide an interface to configure socket options. For example:

```rust
// connection.rs
#[cfg(any(feature = "socket-options", feature = "keep-alive"))]
#[derive(Default)]
pub struct SocketOptions {
    pub user_tcp_timeout: Option<Duration>,
    pub tcp_keep_alive: Option<socket2::TcpKeepalive>,
}
```

```rust
// tokio.rs
#[inline(always)]
async fn connect_tcp(
    addr: &SocketAddr,
    #[cfg(any(feature = "socket-options", feature = "keep-alive"))] socket_options: SocketOptions,
) -> io::Result<TcpStreamTokio> {
    let socket = TcpStreamTokio::connect(addr).await?;
    #[cfg(feature = "tcp_nodelay")]
    socket.set_nodelay(true)?;
    #[cfg(any(feature = "socket-options", feature = "keep-alive"))]
    {
        //For now rely on system defaults
        let keep_alive: socket2::TcpKeepalive = socket_options
            .tcp_keep_alive
            .unwrap_or(socket2::TcpKeepalive::new());

        //these are useless error that not going to happen
        let std_socket = socket.into_std()?;
        let socket2: socket2::Socket = std_socket.into();
        socket2.set_tcp_keepalive(&keep_alive)?;
        socket2.set_tcp_user_timeout(socket_options.user_tcp_timeout)?;
        TcpStreamTokio::from_std(socket2.into())
    }

    #[cfg(not(feature = "keep-alive"))]
    {
[...]

PR #1523, opened 2025-02-11
Allow users of async connections to set TCP settings.

nihohit:
https://github.com/redis-rs/redis-rs/issues/1470
https://github.com/redis-rs/redis-rs/issues/1147

ATM this change is only for async connections, in order to get early feedback quickly. This can technically be added to sync connections too.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "redis"
 version = "0.28.2"
-keywords = ["redis",  "valkey", "cluster", "sentinel", "pubsub"]
+keywords = ["redis", "valkey", "cluster", "sentinel", "pubsub"]
 description = "Redis driver for Rust."
 homepage = "https://github.com/redis-rs/redis-rs"
 repository = "https://github.com/redis-rs/redis-rs"
@@ -38,11 +38,19 @@ combine = { version = "4.6", default-features = false, features = ["std"] }
 
 # Only needed for AIO
 bytes = { version = "1", optional = true }
-futures-util = { version = "0.3.31", default-features = false, features = ["std", "sink"], optional = true }
+futures-util = { version = "0.3.31", default-features = false, features = [
+  "std",
+  "sink",
+], optional = true }
 pin-project-lite = { version = "0.2", optional = true }
 tokio-util = { version = "0.7", optional = true }
-tokio = { version = "1", features = ["rt", "net", "time", "sync"], optional = true }
-socket2 = { version = "0.5", default-features = false, optional = true }
+tokio = { version = "1", features = [
+  "rt",
+  "net",
+  "time",
+  "sync",
+], optional = true }
+socket2 = { version = "0.5", features = ["all"] }
 
 # Only needed for the connection manager
 arc-swap = { version = "1.7.1" }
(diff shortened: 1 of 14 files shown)


nihohit:
@altanozlu , thoughts?

redis/src/io/tcp.rs
```diff
@@ -0,0 +1,57 @@
...
+    user_timeout: Option<Duration>,
+}
+
+impl TcpSettings {
+    /// Sets the value of the `TCP_NODELAY` option on this socket.
+    pub fn set_nodelay(self, nodelay: bool) -> Self {
+        Self { nodelay, ..self }
+    }
+
+    /// Set parameters configuring TCP keepalive probes for this socket.
```
DoumanAsh, line 18: Please add comment default value is dependent on system kernel settings

altanozlu: I agree with @DoumanAsh, imho it's better to make None as default as documented [here](https://www.man7.org/linux/man-pages/man7/tcp.7.html) and [here](https://github.com/rust-lang/socket2/blob/0ae0c3a3e9e898677dc48eef88c74bf7fbd99a88/src/sys/unix.rs#L2579), having None(0) will keep default value of os at least in linux.

redis/src/io/tcp.rs
```diff
@@ -0,0 +1,57 @@
+use std::{io, net::TcpStream, time::Duration};
+
+/// Settings for a TCP stream.
+#[derive(Clone, Debug)]
+pub struct TcpSettings {
+    nodelay: bool,
+    keepalive: socket2::TcpKeepalive,
```
altanozlu, line 7: Now it will enable `SO_KEEPALIVE`  on socket as default, is this really something desired ? I'm just pointing out https://docs.rs/socket2/latest/src/socket2/socket.rs.html#2194-2197

nihohit: sorry, what do you mean to point out?

redis/src/io/tcp.rs
nihohit: changed behavior to not set this by default.

redis/src/io/tcp.rs
altanozlu: It will enable keep alive even without any configuration has been set, I believe it wasn't the case before.

altanozlu: I see it's already changed now

nihohit:
@altanozlu, @DoumanAsh, any more comments? Can I merge this?

altanozlu (approved):

DoumanAsh (approved):

redis/Cargo.toml
```diff
@@ -38,11 +38,19 @@ combine = { version = "4.6", default-features = false, features = ["std"] }
...
 tokio-util = { version = "0.7", optional = true }
-tokio = { version = "1", features = ["rt", "net", "time", "sync"], optional = true }
-socket2 = { version = "0.5", default-features = false, optional = true }
+tokio = { version = "1", features = [
+  "rt",
+  "net",
+  "time",
+  "sync",
+], optional = true }
+socket2 = { version = "0.5", features = ["all"] }
```
007gzs, line 54: I want to know the reason for changing from `socket2 = { version = "0.5", default-features = false, optional = true }` to `socket2 = { version = "0.5", features = ["all"] }`.

I use `use redis::{Cmd, ToRedisArgs, Value};` in wasm.

dependencies is `redis={version = "0", default-features = false}`.

When the version is 0.28, there is no dependency on socket2. However, after upgrading to 0.29, socket2 becomes a dependency

nihohit: Hi, @007gzs, if you feel that the dependency is unnecessary, you're welcome to open a PR removing it, and handle the code changes that requires.

PR #1526, opened 2025-02-14
Cluster connect hook

jorendorff:
I have a Redis 6 cluster. It unconditionally sends out node addresses as IPs. However, the nodes have certificates which only cover their hostnames, not their IPs. (I can't control the certificates.) So when I connect to the cluster with redis-rs, all non-initial connections fail with certificate errors.

Right now, I can connect by using `TlsMode::Insecure`, but I need a better solution.

The idea here is to add a hook that's called every time a ClusterClient connects to a node. It allows the client to rewrite the address and port used for the connection. This solves my problem: I can rewrite the IP addresses back to the hostnames that are in the certificates.

I'm sure this is not acceptable as-is (there aren't any tests and the connect hook API is not optimal), but consider it a proof of concept. Does the idea make sense?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -510,6 +510,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "connect-hook-test"
+version = "0.1.0"
+dependencies = [
+ "redis",
+]
+
+[[package]]
 name = "core-foundation"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


nihohit:
Hi, this design makes sense to me. It looks like it has a pretty small blast radius, and as long as the function is properly documented to make sure that it shouldn't be used lightly, I'm fine with it.
I think that the best way to test this would be using the [mock clusters](https://github.com/redis-rs/redis-rs/blob/ffaacccce240168f2b6f7ee4a38181b213790e4c/redis/tests/support/mock_cluster.rs#L289) - I'm not aware of another way to set up a cluster that sends "wrong" node names. The API isn't well documented or intuitive, but I think that if you'll review some of the existing tests, it should be fairly easy to use.

I would appreciate it if you could also implement the change for async cluster connections, but it's not a must-have.

jorendorff:
It looks like #1529 will work out better for everyone, so let's go with that. Closing this one.

PR #1529, opened 2025-02-14
Implement `danger_accept_invalid_hostnames` option.

### LINKED ISSUE #1500: More granular TLS configuration.
Would it be possible to make the TLS configuration more granular than `Secure` or `Insecure`? Specifically we are running into an issue where we may need to disable strict hostname checking, but I don't really want to disable all of the other things that `Insecure` disables (especially certificate validation). It seems like the `redis-cli` client doesn't even have strict host checking on by default, which made debugging rather confusing.

If this seems like a reasonable idea I am happy to working on a PR. Maybe by turning `Secure` and `Insecure` into `struct` variants and adding some fields?

jorendorff:
Fixes #1500.

This adds a new insecure connection mode between `TlsMode::Secure` and `TlsMode::Insecure`, called `danger_accept_invalid_hostnames` after the flag in `tls-native` that implements it. When this mode is enabled, certificate checking still happens, but the hostname in the certificate is not required to match the server we're trying to connect to.

**The new mode is not a new `TlsMode`** or a new field on `TlsMode::Insecure`. Either of those would be a better API, I think, but it'd be a breaking change. So this PR adds the new option as a `ClusterClientBuilder` method instead. When you use this, the `tls_mode` says `Secure` even though you've configured a hole in the TLS security. :-\

I don't actually need the rustls version, but I implemented the option for both `tls-rustls` and `tls-native-tls` to keep the features additive. That is, suppose we land this, and someone starts using it via the `tls-native-tls` feature. If they later add the `tls-rustls` feature, their project should keep building and working. (As implemented, it does keep building, but alas not working. They'll get a runtime error until they also add the `tls-rustls-insecure` feature. But I think that's already the case for the existing insecure mode. That should probably be fixed but it'll be a breaking change.)

This is an alternative to #1526.

### DIFF
--- a/redis/src/aio/tokio.rs
+++ b/redis/src/aio/tokio.rs
@@ -12,14 +12,14 @@ use tokio::{
     net::TcpStream as TcpStreamTokio,
 };
 
-#[cfg(all(feature = "tls-native-tls", not(feature = "tls-rustls")))]
+#[cfg(all(feature = "tokio-native-tls-comp", not(feature = "tokio-rustls-comp")))]
 use native_tls::TlsConnector;
 
-#[cfg(feature = "tls-rustls")]
+#[cfg(feature = "tokio-rustls-comp")]
 use crate::connection::create_rustls_config;
-#[cfg(feature = "tls-rustls")]
+#[cfg(feature = "tokio-rustls-comp")]
 use std::sync::Arc;
-#[cfg(feature = "tls-rustls")]
+#[cfg(feature = "tokio-rustls-comp")]
 use tokio_rustls::{client::TlsStream, TlsConnector};
 
 #[cfg(all(feature = "tokio-native-tls-comp", not(feature = "tokio-rustls-comp")))]
@@ -116,7 +116,7 @@ impl RedisRuntime for Tokio {
             .map(Tokio::Tcp)?)
     }
 
-    #[cfg(all(feature = "tls-native-tls", not(feature = "tls-rustls")))]
+    #[cfg(all(feature = "tokio-native-tls-comp", not(feature = "tokio-rustls-comp")))]
     async fn connect_tcp_tls(
         hostname: &str,
         socket_addr: SocketAddr,
@@ -144,7 +144,7 @@ impl RedisRuntime for Tokio {
             .map(|con| Tokio::TcpTls(Box::new(con)))?)
     }
 
-    #[cfg(feature = "tls-rustls")]
+    #[cfg(feature = "tokio-rustls-comp")]
     async fn connect_tcp_tls(
         hostname: &str,
         socket_addr: SocketAddr,
--- a/redis/src/aio/tokio.rs
+++ b/redis/src/aio/tokio.rs
@@ -25,10 +25,7 @@ use tokio_rustls::{client::TlsStream, TlsConnector};
 #[cfg(all(feature = "tokio-native-tls-comp", not(feature = "tokio-rustls-comp")))]
 use tokio_native_tls::TlsStream;
 
-#[cfg(feature = "tokio-rustls-comp")]
-use crate::tls::TlsConnParams;
-
-#[cfg(all(feature = "tokio-native-tls-comp", not(feature = "tls-rustls")))]
+#[cfg(any(feature = "tokio-rustls-comp", feature = "tokio-native-tls-comp"))]
 use crate::connection::TlsConnParams;
 
 #[cfg(unix)]
--- a/redis/src/aio/tokio.rs
+++ b/redis/src/aio/tokio.rs
@@ -121,7 +121,7 @@ impl RedisRuntime for Tokio {
         hostname: &str,
         socket_addr: SocketAddr,
         insecure: bool,
-        _: &Option<TlsConnParams>,
+        params: &Option<TlsConnParams>,
         tcp_settings: &crate::io::tcp::TcpSettings,
     ) -> RedisResult<Self> {
         let tls_connector: tokio_native_tls::TlsConnector = if insecure {
@@ -129,6 +129,10 @@ impl RedisRuntime for Tokio {
                 .danger_accept_invalid_certs(true)
                 .danger_accept_invalid_hostnames(true)
                 .use_sni(false)
+                .build()?
+        } else if let Some(params) = params {
+            TlsConnector::builder()
+                .danger_accept_invalid_hostnames(params.danger_accept_invalid_hostnames)
                 .build()?
         } else {
             TlsConnector::new()?
--- a/redis/src/cluster_client.rs
+++ b/redis/src/cluster_client.rs
@@ -39,7 +39,7 @@ struct BuilderParams {
     pub(crate) tcp_settings: TcpSettings,
 }
 
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub(crate) struct RetryParams {
     pub(crate) number_of_retries: u32,
     max_wait_time: u64,
--- a/redis/src/cluster_client.rs
+++ b/redis/src/cluster_client.rs
@@ -110,7 +110,13 @@ impl ClusterParams {
 
         #[cfg(any(feature = "tls-rustls-insecure", feature = "tls-native-tls"))]
         let tls_params = if value.danger_accept_invalid_hostnames {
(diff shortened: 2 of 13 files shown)


nihohit:
I think that this interface is better than https://github.com/redis-rs/redis-rs/pull/1526, because it presents a smaller change that requires less mental effort from the user. The problem is that the implementation is build around the assumption that passing `tls_params` when using native TLS is implicitly equivalent to using this unsafe feature. It is true ATM, but I wouldn't want the code to carry this implict assumption, which could be harmful down the line.

jorendorff:
> The problem is that the implementation is build around the assumption that passing `tls_params` when using native TLS is implicitly equivalent to using this unsafe feature.

If it says that, I didn't intend it. For example, when it says this:

```rust
        } else if let Some(params) = tls_params {
            TlsConnector::new()
                .danger_accept_invalid_hostnames(params.danger_accept_invalid_hostnames)
        } else {
```

the intent is to call `danger_accept_invalid_hostnames` unconditionally, but pass the boolean parameter from the TLSParams, which is still `false` by default.

jorendorff:
OK, let's close #1526 and proceed with this one. I pushed a few minor fixes, but the test I pushed is not working -- Redis server startup fails, I guess it doesn't like the way I'm messing with certificate generation? Will debug tomorrow.

jorendorff:
@nihohit I got further today; the server starts up fine. But I'm still having a lot of trouble getting test coverage for this feature.

#1534 is a nice first step.

My goal here is to get two tests passing:
- One showing that *without* the new danger mode, the client correctly rejects a certificate presented by the server that doesn't have the right `alt_names` for the IP we used to connect.
- And one showing that with the new danger mode enabled, it's allowed.

However, these tests both time out right now. I think Redis isn't starting up successfully, so I am probably generating a bad certificate somehow. Will continue debugging tomorrow.

Separately: test coverage for the `native-tls` side of things seems impossible for now. What do you think about that?

nihohit:
> However, these tests both time out right now

That sounds like bad TLS handshakes - either the cluster expects one but the client doesn't send it, or the client sends the handshake and it isn't answered by the server. I'm not sure what would happen if the client & server perform different handshakes - I expect an error to be returned, but I'm not sure it is.

nihohit:
> test coverage for the native-tls side of things seems impossible for now. What do you think about that?

I think that's regretable, but I don't have the time or the knowledge ATM to improve the testability or usability of native TLS. If you can do that, it will be very appreciated.

jorendorff:
@nihohit Thanks very much for your replies. It did turn out to be a bad TLS handshake, happening only in some test configurations, for several different reasons.

This is ready for review now! Sorry it's so long. If it's too much, the [first commit](https://github.com/redis-rs/redis-rs/pull/1529/commits/6718199a5d49037b7528b3f7277bbea7ceb521ae) could be split into its own PR. I could also split out the `default_reject_invalid_hostnames` tests and the supporting `certs_with_ip_alts` flag, since those should pass without the new feature.

jorendorff:
We've tested this in production and it fixes the problem for us.

jorendorff:
@nihohit Do we need to do anything else here? This is ready for review as far as we're concerned. :bow:

redis/src/aio/tokio.rs
```diff
@@ -124,7 +121,7 @@ impl RedisRuntime for Tokio {
         hostname: &str,
         socket_addr: SocketAddr,
         insecure: bool,
-        _: &Option<TlsConnParams>,
+        params: &Option<TlsConnParams>,
```
nihohit, line 124: can't comment on line 119 - shouldn't it also be replaced by `#[cfg(any(feature = "tokio-rustls-comp", feature = "tokio-native-tls-comp"))]`?

redis/src/cluster_client.rs
```diff
@@ -306,6 +313,26 @@ impl ClusterClientBuilder {
         self
     }
 
+    /// Configure hostname verification when connecting with TLS.
+    ///
+    /// If `insecure` is true, this **disables** hostname verification, while
+    /// leaving other aspects of certificate checking enabled (but that isn't
```
nihohit, line 319: "(but that isn't worth much: without hostname verification, TLS as a whole is not secure)" -> I'd prefer a phrasing that sounds more... "official", maybe?
"... checking enabled, this isn't secure enough for production-grade applications." maybe?

redis/src/connection.rs
```diff
@@ -40,13 +40,19 @@ use crate::PushInfo;
 use rustls_native_certs::load_native_certs;
 
 #[cfg(feature = "tls-rustls")]
-use crate::tls::TlsConnParams;
+use crate::tls::ClientTlsParams;
 
 // Non-exhaustive to prevent construction outside this crate
-#[cfg(not(feature = "tls-rustls"))]
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
nihohit, line 46: since this is about security, I'd prefer that the values here be always explicitly set - please remove the `Default` derivation.

redis/src/connection.rs
```diff
@@ -599,14 +705,18 @@ impl ActualConnection {
                 ref host,
```
nihohit, line 705: same for line 703 - shouldn't it be `#[cfg(any(feature = "tokio-rustls-comp", feature = "tokio-native-tls-comp"))]`?

redis/src/connection.rs
```diff
@@ -882,7 +990,33 @@ pub(crate) fn create_rustls_config(
                 })?
         } else {
             config_builder.with_no_client_auth()
-        }
+        };
+
+        // The cfg here matches the cfg on `danger_accept_invalid_hostnames`.
+        // This way adding `tls-rustls` on top of `tls-native-tls` does not
+        // remove the field or cause it to be silently ignored.
+        #[cfg(any(feature = "tls-rustls-insecure", feature = "tls-native-tls"))]
```
nihohit, line 998: since this is scoped under `#[cfg(feature = "tls-rustls")]` (line 954), the `feature = "tls-native-tls"` part should be dropped. This function is only for rustls, and you wouldn't want to accidentally enable it by setting `tls-native-tls`

nihohit (approved):
sorry it took me some time to get back to you. Some small comments, but overall this looks ready to be merged.

gorzell:
@nihohit what is your release cadence? It would be great if once this change is merged we could get a release with this and the timeout fix you made.

nihohit:
@gorzell I was planning on a release this tomorrow or on saturday. I don't mind waiting for this for a day or two.

redis/src/aio/tokio.rs
jorendorff: Right, all of these can use the `tokio-*-comp` feature names. Fixed now. I made the corresponding change in `async_std.rs` too.

There is a separate implementation for `rustls` builds, so the right cfg here is `#[cfg(all(feature = "tokio-native-tls-comp", not(feature = "tokio-rustls-comp")))]`.

redis/src/cluster_client.rs
jorendorff: I just deleted the part in parentheses. The warning that follows is clear and not too casual.

redis/src/connection.rs
jorendorff: I think in this case `tls-native-tls` is right since this isn't async code. The `tokio-*-comp` feature names aren't used in this file.

redis/src/connection.rs
jorendorff: :+1:

redis/src/connection.rs
jorendorff: Just to make totally sure you can merge this either way, I did two things

- change the existing code to make it clearer what's going on (now included in this PR)
- also make a separate commit that deletes the code (jorendorff/redis-rs#1)

I think you can merge that commit into this PR by doing:

```
git remote add jorendorff https://github.com/jorendorff/redis-rs.git
git fetch jorendorff accept-invalid-hostnames 
git checkout accept-invalid-hostnames
git fetch jorendorff simplify-invalid-hostnames
git cherry-pick jorendorff/simplify-invalid-hostnames
git push jorendorff accept-invalid-hostnames
```

nihohit:
Thanks for the contribution, and for cleaning up the config names in the async files!

PR #1534, opened 2025-02-20
Run most cluster tests in secure mode when rustls is enabled

jorendorff:
In trying to add a test for a new security mode in https://github.com/redis-rs/redis-rs/pull/1529, I found out no existing cluster tests run in `TlsMode::Secure` mode. It seems like a good idea to get some coverage for that case first.

All but 6 tests pass with certificate checking enabled, so this PR transitions tests to run in `Secure` mode by default when `rustls` is used, and explicitly opts out those 6 tests.

This is `tls-rustls` only for now. TLS support in the cluster tests works by generating a one-off root cert (to sign the server cert), and it appears `redis-rs` doesn't support adding a root cert to the client in `tls-native-tls` builds, so certificate checking would always fail.

### DIFF
--- a/redis-test/src/cluster.rs
+++ b/redis-test/src/cluster.rs
@@ -11,6 +11,7 @@ pub struct RedisClusterConfiguration {
     pub num_nodes: u16,
     pub num_replicas: u16,
     pub modules: Vec<Module>,
+    pub tls_insecure: bool,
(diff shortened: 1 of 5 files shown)


nihohit (approved):
Thanks for the contribution!

Issue #1540, opened 2025-02-26
SentinelClientBuilder docs Incorrect? Sentinel Connection Help


65001:
# Possible Docs Issue
In [SentinelClientBuilder:L1103](https://github.com/redis-rs/redis-rs/blob/1e4fc2990ee50fa285afa1012fe0bbde2e2349b1/redis/src/sentinel.rs#L1033) the build is rejected if the user uses  ConnectionAddr::Tcp .

However, the [docs](https://docs.rs/redis/0.29.0/redis/sentinel/index.html) list this as an example:
```rust
use redis::sentinel::SentinelClientBuilder;
use redis::{ConnectionAddr, TlsCertificates};
let nodes = vec![ConnectionAddr::Tcp(String::from("redis://127.0.0.1"), 6379), ConnectionAddr::Tcp(String::from("redis://127.0.0.1"), 6378), ConnectionAddr::Tcp(String::from("redis://127.0.0.1"), 6377)];

let mut builder = SentinelClientBuilder::new(nodes, String::from("master"), redis::sentinel::SentinelServerType::Master).unwrap();

builder = builder.set_client_to_sentinel_username(String::from("username1"));
builder = builder.set_client_to_sentinel_password(String::from("password1"));
builder = builder.set_client_to_sentinel_tls_mode(redis::TlsMode::Insecure);

builder = builder.set_client_to_redis_username(String::from("username2"));
builder = builder.set_client_to_redis_password(String::from("password2"));
builder = builder.set_client_to_redis_tls_mode(redis::TlsMode::Secure);


let client = builder.build().unwrap();
```

I think the documentation is incorrect, for the current version of the codebase.

# My Problem
I don't understand how to connect to a series of hosts say host-1-1. host-1-2, host-1-3 over TLS, when the Sentinel port is 456 and the Redis port is XYZ, and I have a different username/password for sentinel and redis.

nihohit:
Thanks for reporting these issues!
Would you like to open PRs to fix them? If not, I'll get to these fixes eventually.

65001:
Hey @nihohit I created a PR, that I think removes the requirement that a user has `ROLE` ACL to connect to a sentinel. 

I didn't touch the `SentinelBuilder`, to add the `min_other_sentinels`.

[...]

PR #1542, opened 2025-02-28
Hopefully fixed ACL issue faced by others

65001:
Will hopefully fix https://github.com/redis-rs/redis-rs/issues/1540

Allowed user to directly access client, from `SentinelClient`.


nihohit:
can you please add a test?

PR #1557, opened 2025-03-05
perf: Run reconnection attempts concurrently

Marwes:
I don't think there is any reason not to do this concurrently? Each connection will be to a different node and they will mostly just wait for IO which is nice to do concurrently.

Also refactored a bit to remove `ConnectionFuture` as it does not seem to serve a purpose anymore.

(Looking over the implementation to try and debug a connection problem and spotted it).

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -564,84 +564,76 @@ where
         let inner = self.inner.clone();
         async move {
             let mut write_guard = inner.conn_lock.write().await;
-            let mut connections = stream::iter(addrs)
-                .fold(
-                    mem::take(&mut write_guard.0),
-                    |mut connections, addr| async {
-                        let conn = Self::get_or_create_conn(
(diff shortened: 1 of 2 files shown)


nihohit (approved):

Marwes:
I have no idea how it would regress performance, every use of it the connection is already fully established and then that is wrapped into an immediately ready future. So all this does is remove indirection. It is possible that the shared connection could be used to eagerly start a reconnect attempt and only actually await on it once it is later accessed but that is not used right now.

As it is I think the `ConnectionFuture` is just an artifact of the earliest versions of the async cluster client being done without async await and storing the to-be-awaited connection in a map was just much easier  to implement https://github.com/redis-rs/redis-cluster-async/pull/6 .

nihohit:
As I said, very frustrating :)
I'll try to get the benchmarks working again and see if I can reproduce this. If I won't get to it, we'll just merge.

PR #1559, opened 2025-03-07
add -no-provider variant of the rustls features

kamulos:
This is an initial implementation of the [discussion in #1343](https://github.com/redis-rs/redis-rs/pull/1343#issuecomment-2706208818). Just to get a feeling if this is the right direction.

Relevant Points:

- All the features stay the same
- `ring` is still the default provider for `rustls`
- Like `reqwest` there is now `rustls-no-provider` variants of `rustls` features, that have all the same things enabled, but don't enable `ring` or any other provider
- The `rustls` features always are just the `rustls-no-provider` with `ring` as an addition
- All the checks in the code check against the `rustls-no-provider`  feature, because they automatically get enabled once the normal `rustls` features are enabled

Anyone using the `rustls-no-provider` features is responsible for enabling a provider as rustls feature themselves. Some might need to add a direct dependency, some might already have it.

An alternative would be to also reexport "rustls/aws_lc" in `redis-rs` so everyone could use that together with the `no-provider` feature.


nihohit:
Is it possible to add only a single feature? we check for every combination of features, and I'm sure that the flag-frenzy task is going to run significantly longer with this change.

kamulos:
I am not sure, seems difficult to me at the moment. There are quite a few `rustls` features already, that expect ring to be enabled. And I can't express via a single feature: "take away ring if it is enabled".

The following would work, but also would be quite breaking: all existing `rustls` features would stay the same, but `ring` would be removed. Then a single feature "rustls/ring" can be added to get ring back. But this would mean, that everyone upgrading needs to add this feature, unless they have it already set via another way in their workspace.

nihohit:
> that expect ring to be enabled

In what way they expect Ring to be enabled, besides setting the default provider?

nihohit:
@kamulos ping re: my question, and the failed flag-frenzy CI action.

nihohit:
@kamulos can you please check if https://github.com/redis-rs/redis-rs/pull/1632 solves this issue?

Issue #1561, opened 2025-03-11
Entered unreachable code error when retrieving Vec<(HashMap<String, String>, Vec<String>)>


mjcarson:
When trying to retrieve a Vec<(HashMap<String, String>, Vec<String>)> from a pipeline you get the following error:

```
thread 'main' panicked at /Users/mcarson/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/redis-0.29.1/src/types.rs:2596:1:
internal error: entered unreachable code
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9fb94b32df38073bf63d009df77ed10cb1c989d0/library/std/src/panicking.rs:695:5
   1: core::panicking::panic_fmt
             at /rustc/9fb94b32df38073bf63d009df77ed10cb1c989d0/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/9fb94b32df38073bf63d009df77ed10cb1c989d0/library/core/src/panicking.rs:145:5
   3: <(T11,T12) as redis::types::FromRedisValue>::from_owned_redis_values
             at /Users/mcarson/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/redis-0.29.1/src/types.rs:2558:33
   4: <alloc::vec::Vec<T> as redis::types::FromRedisValue>::from_owned_redis_value
             at /Users/mcarson/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/redis-0.29.1/src/types.rs:2250:44
   5: redis::types::from_owned_redis_value
             at /Users/mcarson/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/redis-0.29.1/src/types.rs:2675:5
   6: redis::pipeline::Pipeline::query_async::{{closure}}
             at /Users/mcarson/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/redis-0.29.1/src/pipeline.rs:207:9
   7: redis_tester::main::{{closure}}
             at ./main.rs:56:10
   8: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at /Users/mcarson/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.44.0/src/runtime/park.rs:284:60
   9: tokio::task::coop::with_budget
[...]

PR #1562, opened 2025-03-11
Fix unreachable error when parsing a nested tuple. 

nihohit:
The unreachable code assumed that a nested array has to be able to converted to the target tuple, and so panicked when when the array is larger. Instead, we should just fallback to the next piece of code, and allow the array to be split into multiple returns of the target tuple.

https://github.com/redis-rs/redis-rs/issues/1561


altanozlu:
Thanks for tackling the TODO

altanozlu (approved):

Issue #1563, opened 2025-03-13
discuss: Implement bb8 support in redis-rs


Xuanwo:
I'm starting this thread to discuss whether it's a good idea to implement `bb8` support in `redis-rs` directly, similar to how we do it for `r2d2`. I personally believe that will be good for the overall benefits around redis-rs and bb8.

- bb8 has native redis-rs support by [bb8-redis](https://crates.io/crates/bb8-redis)
- bb8 has an external redis cluster support by [bb8-redis-cluster](https://crates.io/crates/bb8-redis-cluster)
  - however, this support is based on [redis_cluster_async](https://crates.io/crates/redis_cluster_async) which has merged into `redis-rs` already.
  - `bb8-redis-cluster` itself is not updated 2 years ago

I'm willing to help implement bb8 support in redis-rs directly so our users can have a better experience when using Redis with bb8. I'm also interested in building a connection pool that can work with both Redis and a Redis Cluster setup.

What do you think? 

cc @nihohit and @djc

djc:
I'm happy to "donate" bb8-redis to the redis-rs org, whether you incorporate the code into the redis crate or otherwise.

Xuanwo:
> I'm happy to "donate" bb8-redis to the redis-rs org, whether you incorporate the code into the redis crate or otherwise.

Really appreciated! I prefer to have it in tree (like r2d2), but I'll let the maintainers of `redis-rs` decide. 

I volunteered to do all the work, includes:

- Move existsing `bb8-redis` code and implement support for redis cluster
- Add migrate docs for `bb8` project.

nihohit:
I'm willing to accept the code, assuming it is properly cordoned off from the rest of the crate - under a separate folder, with minimal changes outside of that folder.

Xuanwo:
Sure. I will work on this.

PR #1564, opened 2025-03-13
feat: Add bb8 support for async client

Xuanwo:
Part of https://github.com/redis-rs/redis-rs/issues/1563

---

My original plan was to use the bb8 code directly, but I decided to follow the same pattern as r2d2 instead.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -60,6 +60,9 @@ backon = { version = "1.4.0", optional = true, default-features = false }
 # Only needed for the r2d2 feature
 r2d2 = { version = "0.8.10", optional = true }
 
+# Only needed for the bb8 feature
+bb8 = { version = "0.9.0", optional = true }
+
 # Only needed for cluster
 crc16 = { version = "0.4", optional = true }
 rand = { version = "0.9", optional = true }
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -142,6 +142,8 @@ tcp_nodelay = []
 num-bigint = []
 disable-client-setinfo = []
 cache-aio = ["aio", "dep:lru"]
+r2d2 = ["dep:r2d2"]
+bb8 = ["dep:bb8"]
 
 # Deprecated features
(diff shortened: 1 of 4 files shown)


redis/Cargo.toml
```diff
@@ -60,6 +60,9 @@ backon = { version = "1.4.0", optional = true, default-features = false }
 # Only needed for the r2d2 feature
 r2d2 = { version = "0.8.10", optional = true }
 
+# Only needed for the bb8 feature
+bb8 = { version = "0.9.0", optional = true }
```
nihohit, line 64: please add this as an explicit feature, and add a rule in config/redis.toml that bb8 & r2d2 won't be built at the same time.

Xuanwo: > bb8 & r2d2 won't be built at the same time.

Hi, technically `bb8` be used alongside `r2d2` at the same time. Is there any concern about this?

nihohit: it's not a product consideration, just a heuristic to reduce test time. The `flag-frenzy` action checks an increasing number of combinations, and in order to reduce the time that step takes, I want to remove unlikely combinations from the checks. Our lint and test stages test with all feature enabled, so I'm doubtful that there's an interaction there that might be missed.

Dav1dde:
The [docs](https://docs.rs/redis/0.29.2/redis/#connection-pooling) currently recommend to not use a connection pool for async connections:

> For async connections, connection pooling isn’t necessary. The MultiplexedConnection is cloneable and can be used safely from multiple threads, so a single connection can be easily reused. For automatic reconnections consider using ConnectionManager with the connection-manager feature. Async cluster connections also don’t require pooling and are thread-safe and reusable.

@nihohit did this change?

nihohit:
@Dav1dde the docs recommend not using a connection pool for automatic reconnects, and this is still the recommendation. There are other scenarios where connection pooling might be relevant, such as concurrent blocking requests, or very high TPS workflows.

Issue #1565, opened 2025-03-15
PFCOUNT in cluster pipeline


stepanmracek:
Hi,
is there a specific reason why PFCOUNT is on the list of illegal cluster pipeline commands? As long as all keys are hashed to the same slot, there should be no problem. I removed PFCOUNT from the `fn is_illegal_cmd(cmd: &str)` in `src/cluster_pipeline.rs` and tried following script:
```rs
use redis::cluster::{ClusterClient, cluster_pipe};
use redis::cmd;

fn main() {
    let mut connection = ClusterClient::new([
        "redis://127.0.0.1:7001",
        "redis://127.0.0.1:7002",
        "redis://127.0.0.1:7003",
    ])
    .unwrap()
    .get_connection()
    .unwrap();

    cmd("FLUSHDB").exec(&mut connection).unwrap();

    let count: Vec<i32> = cluster_pipe()
        .cmd("PFCOUNT")
        .arg("{hll}-1")
        .arg("{hll}-2")
        .arg("{hll}-3")
        .query(&mut connection)
        .unwrap();
    println!("{count:?}");

    cluster_pipe()
        .cmd("PFADD")
        .arg("{hll}-1")
        .arg("foo")
        .cmd("PFADD")
        .arg("{hll}-2")
        .arg("foo")
        .cmd("PFADD")
        .arg("{hll}-2")
        .arg("bar")
        .cmd("PFADD")
        .arg("{hll}-3")
        .arg("bar")
        .exec(&mut connection)
        .unwrap();

    let count: Vec<i32> = cluster_pipe()
        .cmd("PFCOUNT")
        .arg("{hll}-1")
        .arg("{hll}-2")
        .arg("{hll}-3")
        .query(&mut connection)
        .unwrap();
    println!("{count:?}");
}
```

And the output was exactly what I expected:
```
[0]
[2]
```

nihohit:
>  As long as all keys are hashed to the same slot, there should be no problem

That's exactly the issue. While I wasn't around when ClusterPipeline was added, I assume that the forbidden commands were all commands that could accidentally be mapped to multiple slots. I see two ways of solving this - 
* either check that this assumption is correct ([using existing logic](https://github.com/redis-rs/redis-rs/blob/f40e6fd0fb79d51b69cdc36dd2a1753d07d5c4ec/redis/src/cluster_routing.rs#L359), preferably) during run time
[...]

PR #1566, opened 2025-03-15
Remove PFCOUNT from the list of illegal cluster pipeline commands (#1565)

stepanmracek:
Related to the issue https://github.com/redis-rs/redis-rs/issues/1565.

As long as all keys in PFCOUNT command are hashed to the same slot, there should be no problem using the command in the cluster pipeline.

### DIFF
--- a/redis/src/cluster_pipeline.rs
+++ b/redis/src/cluster_pipeline.rs
@@ -24,7 +24,7 @@ fn is_illegal_cmd(cmd: &str) -> bool {
         "KEYS" |
         "LASTSAVE" |
         "MGET" | "MOVE" | "MSET" | "MSETNX" |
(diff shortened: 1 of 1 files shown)


redis/src/cluster_routing.rs
```diff
@@ -312,6 +312,40 @@ where
...
+        slots.insert(get_slot(key));
+        key_index += 1;
+    }
+
+    if slots.len() != 1 {
+        return None;
+    }
+
+    Some(RoutingInfo::SingleNode(
+        SingleNodeRoutingInfo::SpecificNode(get_route(
```
stepanmracek, line 342: here is possible non-optimality where the slot of the first key is computed again inside get_route() :thinking:

nihohit:
Thanks for the contribution!

PR #1567, opened 2025-03-15
Add cache support to ConnectionManager.

nihohit:


altanozlu (changes requested):
When connection is lost, Connection Manager [should flush the local cache](https://redis.io/docs/latest/develop/reference/client-side-caching/#what-to-do-when-losing-connection-with-the-server), using [clear](https://docs.rs/lru/latest/lru/struct.LruCache.html#method.clear) on LRUs could work or just letting ConnectionManager creating new multiplexedconnection would be *better* in expense of CacheStatistics being resetted(which is doable).

nihohit:
why would recreating the cache be better?

altanozlu:
> why would recreating the cache be better?

By letting CacheManager to be re-created through Mpx you need less complexity/change in the code.

nihohit:
Fair enough, I agree that the implementation is much simpler. The question is what's the expectation for statistics from the connection manager.
Let's answer a more complex question - in a cluster connection, where some connections could be lost while other remain, what is the expectation for the statistics meaning.

altanozlu:
> Fair enough, I agree that the implementation is much simpler. The question is what's the expectation for statistics from the connection manager. Let's answer a more complex question - in a cluster connection, where some connections could be lost while others remain, what is the expectation for the statistics meaning? Do we hold statistics for only some connections? hold separate statistics for each connection? The simplest solution is not to provide statistics at all, but this should be an intentional design decision, not something we just do to simplify the code.

Hmm, when including cluster connections I feel like having a single shared `Statistics` for cluster/ConnectionManager or not having statistics at all is better. I still stand with having some information about how well caching works is a good thing.

altanozlu:
> > I still stand with having some information about how well caching works is a good thing.
> 
> I agree! So the question is how do we give _some_ information in a way which isn't confusing when using a connection that represents multiple connections. Statistics aren't the only problematic issue for cluster connections - you also let the user specify a limit to the cache size. Using a single shared cache is the simplest way to have this limit. So I think that having a shared cache manager makes sense - it allows us to share a cache and statistics with the user. We just need to understand how to (partially?) clear it when some nodes disconnect.

Shared cache manager sounds good and solves problems, it is possible to some epoch based GC. Think like Cluster has a number and with each connection/re-connection event it assigns incremented number to connection. When X connection has epoch 93 and request a key in CacheManager if that key exists but it's epoch is different than it will be invalidated.

nihohit:
I see the logic behind this, but you would also need to factor in topology changes. What happens when there's no disconnect, but some slots were moved between nodes, or the cluster scaled up, and a new node was introduced and some slots were moved to it?
The simple answer is just to invalidate everything then, but I believe that we can do better.

altanozlu:
> I see the logic behind this, but you would also need to factor in topology changes. What happens when there's no disconnect, but some slots were moved between nodes, or the cluster scaled up, and a new node was introduced and some slots were moved to it? The simple answer is just to invalidate everything then, but I believe that we can do better.

I think it also works fine with topology changes, correct me if I'm mistaken I'm not proficient at clustered redis.
there are A,B,C nodes,  they have mapped x,y,z and have 1,2,3 epochs respectively. If A disconnects and reconnects it will have epoch 4 every key node A requests will be invalidated during read. Let's assume node D has been added with epoch 5 and now it maps key z now old information related to key z will be also invalidated when node D requests that information. If it also holds truth with replications and failovers, it can be done, please correct me otherwise.

altanozlu (approved):

nihohit:
> it can be done, please correct me otherwise.

It can be done, I'm just not sure that it can be done in a simple manner. Would you like to try, or should I?

altanozlu:
> > it can be done, please correct me otherwise.
> 
> It can be done, I'm just not sure that it can be done in a simple manner. Would you like to try, or should I?

I will try it

nihohit:
> I will try it

Excellent, thanks. ping me if you want any help or inputs.

PR #1572, opened 2025-03-24
A spec compliant version, with less changed code.

65001:


65001:
Fixes #1540

redis/src/sentinel.rs
```diff
@@ -295,25 +292,80 @@ fn valid_addrs<'a>(
...
+fn old_check_role(connection_info: &ConnectionInfo) -> RedisResult<Role> {
+    let client = Client::open(connection_info.clone())?;
+    let mut conn = client.get_connection()?;
+    let role: RedisResult<Role> = crate::cmd("ROLE").query(&mut conn);
+    return role
+}
+
+fn check_role(conn: &mut Connection) -> RedisResult<Role> {
+    let role: RedisResult<Role> = crate::cmd("ROLE").query(conn);
+    role
```
nihohit, line 344: I think that `crate::cmd("ROLE").query(conn)` is sufficient for the full function, no need to set the variable

redis/src/sentinel.rs
```diff
@@ -295,25 +292,80 @@ fn valid_addrs<'a>(
...
+    let mut conn = client.get_connection()?;
+
+    //Once the client discovered the address of the master instance, it should attempt a connection with the master, and call the ROLE command in order to verify the role of the instance is actually a master.
+    let role = check_role(&mut conn);
+    if role.is_ok_and(|x| matches!(x, Role::Primary { .. })) {
+        return Ok(true);
+    }
+
+    //If the ROLE commands is not available (it was introduced in Redis 2.8.12), a client may resort to the INFO replication command parsing the role: field of the output.
+    let role = check_info_replication(&mut conn);
```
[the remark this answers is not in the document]
65001: When I tried:
```rust
    let results: RedisResult<Vec<crate::Value>> = crate::pipe()
        .cmd("ROLE")
        .cmd("INFO").arg("REPLICATION")
        .query::<Vec<crate::Value>>(&mut conn);
```
The pipeline will encounter a `ServerError::ExtensionError ` and only return:
```
 NOPERM: this user has no permissions to run the 'role' command or its subcommand,
```
it won't return the results of `INFO REPLICATION`. Is there another way of writing the pipeline so that I can do what you suggested?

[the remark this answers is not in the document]
nihohit: You'll need to use `req_packed_commands` instead, but it's a minor optimization so no need to rewrite everything for this.

nihohit:
you have a lot of unrequired formatting issues which cause CI to fail. please revert them.

nihohit:
also, when building with only sentinel feature, this error appears:
```
Checking: ["sentinel"]
redis/src/sentinel.rs:145:12: error[E0432]: unresolved import `crate::aio`: could not find `aio` in the crate root
redis/src/sentinel.rs:399:27: error[E0599]: no method named `get_multiplexed_async_connection` found for struct `Client` in the current scope: method not found in `Client`
redis/src/sentinel.rs:402:16: error[E0425]: cannot find function `async_check_role` in this scope: not found in this scope
redis/src/sentinel.rs:408:16: error[E0425]: cannot find function `async_check_info_replication` in this scope: help: a function with a similar name exists: `check_info_replication`
redis/src/sentinel.rs:472:85: error[E05[33](https://github.com/redis-rs/redis-rs/actions/runs/14093154067/job/39578395157?pr=1572#step:5:34)]: expected value, found struct variant `Role::Replica`: not a value

**Comment by 65001** [comment]
I ran:
```
cargo fmt --all -- --check
cargo clippy --all-features --all --tests --examples -- -D clippy::all -D warnings
```
see https://github.com/65001/redis-rs/actions/runs/14139751025 , and ran `flag-frenzy --package redis` on all the two package combinations locally.

File: redis/src/sentinel.rs [hunk 3]
```diff
@@ -133,19 +135,17 @@ use std::sync::Mutex;
 use std::{collections::HashMap, num::NonZeroUsize};
 
 #[cfg(feature = "aio")]
-use crate::aio::MultiplexedConnection as AsyncConnection;
-
+use crate::aio::MultiplexedConnection;
```
nihohit, line 138: why import this twice?

redis/src/sentinel.rs
```diff
@@ -345,14 +377,69 @@ fn find_valid_master(
...
+        return Ok(true);
+    }
+
+    //If the ROLE commands is not available (it was introduced in Redis 2.8.12), a client may resort to the INFO replication command parsing the role: field of the output.
+    let role = async_check_info_replication(&mut conn).await;
+    if role.is_ok_and(|x| x == "master") {
+        return Ok(true);
+    }
+
+    //TODO: Maybe there should be some kind of error message if both role checks fail due to ACL permissions?
```
nihohit, line 398: same
https://github.com/redis-rs/redis-rs/pull/1572/files#r2010961098

redis/src/sentinel.rs
```diff
@@ -345,14 +377,69 @@ fn find_valid_master(
...
+        return Ok(true);
+    }
+
+    //If the ROLE commands is not available (it was introduced in Redis 2.8.12), a client may resort to the INFO replication command parsing the role: field of the output.
+    let role = async_check_info_replication(&mut conn).await;
+    if role.is_ok_and(|x| x == "slave") {
+        return Ok(true);
+    }
+
+    //TODO: Maybe there should be some kind of error message if both role checks fail due to ACL permissions?
```
nihohit, line 421: same

redis/src/sentinel.rs
```diff
@@ -295,25 +295,57 @@ fn valid_addrs<'a>(
...
-        }
+fn determine_master_from_role_or_info_replication(
+    connection_info: &ConnectionInfo,
+) -> RedisResult<bool> {
+    let client = Client::open(connection_info.clone())?;
+    let mut conn = client.get_connection()?;
+
+    //Once the client discovered the address of the master instance, it should attempt a connection with the master, and call the ROLE command in order to verify the role of the instance is actually a master.
+    let role = check_role(&mut conn);
+    if role.is_ok_and(|x| matches!(x, Role::Primary { .. })) {
```
nihohit, line 306: the only error which is actually ok to continue here is unknown command (for redis 2), right? so maybe check if that's the error before continuing?

65001: Thats incorrect. The other error possible is NOPERM.

65001: In my env, they default to no role permissions but allow for the replication.

65001:
Hey @nihohit if this looks good, as a bug fix can you merge it in? I don't have much bandwidth this month due to a product launch at work. 

I can address the other changes you requested as feature enhancement, but I think the current PR fixes the bug I was encountering.

It would also allow me to stop using my own fork if this got up streamed.

nihohit:
Thanks for the contribution!

PR #1558, opened 2025-03-05
Add Valkey URL scheme

displexic:
Allows a URL like `valkey://127.0.0.1:6379`.

Addresses issue: https://github.com/redis-rs/redis-rs/issues/1555
Valkey supporting the scheme: https://github.com/valkey-io/valkey/pull/199

PR #1574, opened 2025-03-25
Fix valkey(s) url schemes not able to be converted to connection infos; Add valkey+unix url scheme

MarkusTieger:
This adds some changes, that (I assume) were forgotten in https://github.com/redis-rs/redis-rs/pull/1558.

- "valkey" and "valkeys" were not able to be converted to a connection info https://github.com/redis-rs/redis-rs/blob/8db06bf4e67c526a0604c8d3bc15ae7acf281a0e/redis/src/connection.rs#L463-L464
- "valkeys" did not get treated the same as "rediss" https://github.com/redis-rs/redis-rs/blob/8db06bf4e67c526a0604c8d3bc15ae7acf281a0e/redis/src/connection.rs#L363-L393
- "valkey+unix" protocol scheme was missing (equivalent to "redis+unix")

### DIFF
--- a/redis/src/connection.rs
+++ b/redis/src/connection.rs
@@ -75,7 +75,7 @@ fn connect_tcp_timeout(addr: &SocketAddr, timeout: Duration) -> io::Result<TcpSt
 pub fn parse_redis_url(input: &str) -> Option<url::Url> {
     match url::Url::parse(input) {
         Ok(result) => match result.scheme() {
-            "redis" | "rediss" | "valkey" | "valkeys" | "redis+unix" | "unix" => Some(result),
+            "redis" | "rediss" | "valkey" | "valkeys" | "redis+unix" | "valkey+unix" | "unix" => Some(result),
             _ => None,
         },
         Err(_) => None,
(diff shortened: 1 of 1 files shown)


redis/src/connection.rs
```diff
@@ -2220,6 +2220,7 @@ mod tests {
             ("valkey://127.0.0.1", true),
             ("valkey://[::1]", true),
             ("redis+unix:///run/redis.sock", true),
+            ("valkey+unix:///run/valkey.sock", true),
             ("unix:///run/redis.sock", true),
             ("http://127.0.0.1", false),
             ("tcp://127.0.0.1", false),
```
nihohit, line 2226: also, it looks like no other test in this file tests for valkey cases.

MarkusTieger: redis:// and valkey:// is literally the same in the code. The test where valkey is present, tests if the protocol schenes work, the other tests are unrelated for that. I don't think its necessary to have valkey test cases there

redis/src/connection.rs
```diff
@@ -2220,6 +2220,7 @@ mod tests {
             ("valkey://127.0.0.1", true),
```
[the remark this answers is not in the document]
MarkusTieger: Yeah, they probably should be in there. I will add them tomorrow

redis/src/connection.rs
nihohit: Well, the reason for this change is that we assumed that everything works in the code, but didn't test it. AFAIS `valkeys` still isn't tested anywhere, so any potential future change could break it again.

nihohit (approved):
Thanks for the contribution!

PR #1576, opened 2025-03-26
Add commands to flush database(s)

somechris:
We tend to run flushing commands in our tests, and it would be great to be able to do that directly on the connection, like

```
conn.flushdb().await?;
```


nihohit (approved):
Thanks for the contribution!

PR #1577, opened 2025-03-27
Support bb8 for cluster client

wiserfz:
Hi, this PR add support bb8 for `redis::cluster::ClusterClient` inspired by the #1564.

Thank you for these work, any feedback is appreciated.

### DIFF
--- a/redis/src/bb8.rs
+++ b/redis/src/bb8.rs
@@ -1,23 +1,47 @@
-use crate::{aio::MultiplexedConnection, ErrorKind};
-use crate::{Client, RedisError};
+use crate::aio::MultiplexedConnection;
+use crate::{Client, Cmd, ErrorKind, RedisError};
 
-impl bb8::ManageConnection for Client {
-    type Connection = MultiplexedConnection;
-    type Error = RedisError;
+#[cfg(feature = "cluster-async")]
+use crate::{cluster::ClusterClient, cluster_async::ClusterConnection};
 
-    async fn connect(&self) -> Result<Self::Connection, Self::Error> {
-        self.get_multiplexed_async_connection().await
-    }
+macro_rules! impl_bb8_manage_connection {
+    ($client:ty, $connectioin:ty, $get_conn:expr) => {
+        impl bb8::ManageConnection for $client {
+            type Connection = $connectioin;
+            type Error = RedisError;
 
-    async fn is_valid(&self, conn: &mut Self::Connection) -> Result<(), Self::Error> {
-        let pong: String = crate::cmd("PING").query_async(conn).await?;
-        match pong.as_str() {
-            "PONG" => Ok(()),
-            _ => Err((ErrorKind::ResponseError, "ping request").into()),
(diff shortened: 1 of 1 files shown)


redis/src/bb8.rs
```diff
@@ -1,23 +1,39 @@
...
+use crate::{cluster::ClusterClient, cluster_async::ClusterConnection};
+
+macro_rules! impl_bb8_manage_connection {
+    ($client:ty, $connectioin:ty) => {
+        impl bb8::ManageConnection for $client {
+            type Connection = $connectioin;
+            type Error = RedisError;
+
+            async fn connect(&self) -> Result<Self::Connection, Self::Error> {
+                self.get_multiplexed_async_connection().await
```
nihohit, line 14: why not use `get_async_connection`, since `get_multiplexed_async_connection` just calls that?

wiserfz: Because if use `get_async_connection` method, the implementation for `bb8::ManageConnection` can't use macro expression, must write separately, but the `is_valid` method and the `has_broken` method is same which is annoying, so I decided to write a `get_multiplexed_async_connection` method just for the trait implementation.

Maybe handle these in same way like the methods `get_async_connection` and `get_multiplexed_async_connection` of `redis::Client` or just write the implementation of `bb8::ManageConnection` separately, what do you think?

nihohit: I think you can just invert the call structure, and do `ClusterClient::get_async_connect(&self)`.

wiserfz: Sry, I don't understand, you mean remove the  `ClusterClient::get_multiplexed_async_connection` method? but how to write in the `bb8::ManageConnection::connect` method with macro expression?

nihohit:
```rs
macro_rules! impl_bb8_manage_connection {
    ($client:ty, $connectioin:ty, $get_conn:expr) => {
        impl bb8::ManageConnection for $client {
[..]
            async fn connect(&self) -> Result<Self::Connection, Self::Error> {
                $get_conn(self).await
            }
[..]
        }
    };
}

impl_bb8_manage_connection!(
    Client,
    MultiplexedConnection,
    Client::get_multiplexed_async_connection
);

#[cfg(feature = "cluster-async")]
impl_bb8_manage_connection!(
    ClusterClient,
    ClusterConnection,
    ClusterClient::get_async_connection
);

Comment (wiserfz) [reply redis/src/bb8.rs]: How stupid I am!, thank you for reply. I have fixed this and commit it.

File: redis/src/bb8.rs [hunk 2]
```diff
@@ -1,23 +1,47 @@
...
+);
+
+#[cfg(feature = "cluster-async")]
+impl_bb8_manage_connection!(
+    ClusterClient,
+    ClusterConnection,
+    ClusterClient::get_async_connection
+);
+
+// TODO: support bb8 for sentinel client which required
```
nihohit, line 45: is this solved by the addition of `$get_conn`?

PR #1579, opened 2025-03-28
Add `net` feature for compiling to embedded targets.

langyo:
I'm developing WASI modules that do not require TLS support. However, when compiling the code to WebAssembly (WASM), the build process fails due to dependencies introduced by the `socket2` crate. This appears to be caused by implicit TLS-related features in the dependency chain. To address this, I propose implementing a feature flag that explicitly disables TLS functionality in the library. Such configuration would align with the minimal requirements for Redis protocol parsing while maintaining WASI compatibility.

Here's the demo:

```rust
/*
[dependencies]
tokio = { version = "^1", features = ["macros", "rt"]}
redis = { version = "^0.29", default-features = false, features = [
  "script",
  "aio"
] }
*/
use redis::{Cmd, ConnectionLike, RedisResult, Value, parse_redis_value};

#[derive(Clone)]
pub struct MockConnection;

impl ConnectionLike for MockConnection {
    fn check_connection(&mut self) -> bool {
        true
    }

    fn is_open(&self) -> bool {
        true
    }

    fn req_command(&mut self, cmd: &Cmd) -> RedisResult<Value> {
        let result = parse_redis_value("+OK\r\n".as_bytes())?;
        Ok(result)
    }

    fn req_packed_command(&mut self, cmd: &[u8]) -> RedisResult<Value> {
        let result = parse_redis_value("+OK\r\n".as_bytes())?;
        Ok(result)
    }

    fn req_packed_commands(
        &mut self,
        cmd: &[u8],
        _offset: usize,
        count: usize,
    ) -> RedisResult<Vec<Value>> {
        let result = parse_redis_value("+OK\r\n".as_bytes())?;
        Ok(vec![result; count])
    }

    fn get_db(&self) -> i64 {
        0
    }
}

async fn run_single<C: ConnectionLike>(mut con: C) -> RedisResult<()> {
    interval.tick().await;
    println!("> GET a");
    let result: String = Cmd::get("a").query(&mut con)?;
    println!("< {result:?}");
}

#[tokio::main(flavor = "current_thread")]
async fn main() -> RedisResult<()> {
    run_single(MockConnection {}).await?;
    Ok(())
}
```

To build and run this demo by WASI target, try:

```shell
cargo build --target wasm32-wasip1
# Install wasmtime first: https://github.com/bytecodealliance/wasmtime
wasmtime ./target/wasm32-wasip1/debug/demo.wasm
```

### DIFF
--- a/redis/src/cluster_async/routing.rs
+++ b/redis/src/cluster_async/routing.rs
@@ -114,7 +114,7 @@ mod pipeline_routing_tests {
         let mut pipeline = crate::Pipeline::new();
 
         pipeline
-            .add_command(cmd("FLUSHALL")) // route to all masters
+            .flushall() // route to all masters
             .get("foo") // route to slot 12182
             .add_command(cmd("EVAL")); // route randomly
 
@@ -129,7 +129,7 @@ mod pipeline_routing_tests {
         let mut pipeline = crate::Pipeline::new();
 
         pipeline
-            .add_command(cmd("FLUSHALL")) // route to all masters
+            .flushall() // route to all masters
             .add_command(cmd("EVAL")); // route randomly
 
         assert_eq!(route_for_pipeline(&pipeline), Ok(None));
@@ -141,7 +141,7 @@ mod pipeline_routing_tests {
 
         pipeline
             .get("foo") // route to replica of slot 12182
-            .add_command(cmd("FLUSHALL")) // route to all masters
+            .flushall() // route to all masters
             .add_command(cmd("EVAL"))// route randomly
             .cmd("CONFIG").arg("GET").arg("timeout") // unkeyed command
             .set("foo", "bar"); // route to primary of slot 12182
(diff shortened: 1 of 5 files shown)


nihohit:
I want to get this in, but this requires some more thought & design. 
The first big issue with this change is that it breaks the library for users who don't use the default features. This is unacceptable. 
On a lower level, the amount of `#[cfg(net)]` through the code is rather distressing - it reduces readability and accessibility of the code. We need to add better mod scopes, so that not every new module has to be attributed like this.
I recommend looking into https://github.com/redis-rs/redis-rs/pull/1514 - I think that the sans-io system will fit WASI well, and while it is a larger change, it doesn't break existing users, only opens more usage scenarios.

nihohit:

nihohit:
https://github.com/redis-rs/redis-rs/pull/1818

PR #1607, opened 2025-04-13
Fix `make docs`

somechris:
This PR tries to fix the warnings of running `make docs`

The only missing issue is `redis/src/aio/connection.rs:99`

There, the `PubSub` link in the documentation

```
    /// Converts this [`Connection`] into [`PubSub`].
    #[deprecated(note = "aio::Connection is deprecated. Use [Client::get_async_pubsub] instead")]
    pub fn into_pubsub(self) -> PubSub<C> {
        PubSub::new(self)
    }
```

does not work. The warning I get is:

```
warning: public documentation for `into_pubsub` links to private item `PubSub`
  --> redis/src/aio/connection.rs:99:45
   |
99 |     /// Converts this [`Connection`] into [`PubSub`].
   |                                             ^^^^^^ this item is private
   |
   = note: this link will resolve properly if you pass `--document-private-items`
   = note: `#[warn(rustdoc::private_intra_doc_links)]` on by default
```

But in contrast to rustdoc's warning, PubSub there is not private. It's public.

The issue seems to be the wildcard import in `redis/src/aio/mod.rs` getting in the way of the `pub use` of `pubsub::PubSub` (`redis/src/aio/mod.rs:34`). Hence, the wildcard import cannot match the `PubSub` from `redis/src/aio/connection.rs:99`

Is there a simple way to make that link work?

(I put together a minimal example of the situation at https://github.com/somechris/rustdoc-wildcard-hiding-names )

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -666,7 +666,7 @@ impl ConnectionManager {
         Ok(())
     }
 
-    /// Gets [`CacheStatistics`] for current connection if caching is enabled.
+    /// Gets [`crate::caching::CacheStatistics`] for current connection if caching is enabled.
     #[cfg(feature = "cache-aio")]
     #[cfg_attr(docsrs, doc(cfg(feature = "cache-aio")))]
     pub fn get_cache_statistics(&self) -> Option<crate::caching::CacheStatistics> {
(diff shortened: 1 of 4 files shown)


nihohit (approved):
Thanks for the correction!
Hopefully we'll catch these in the future:
https://github.com/redis-rs/redis-rs/actions/runs/14461802375/job/40555670670?pr=1609

Issue #1610, opened 2025-04-15
How does set_push_sender handle errors?


rollo-b2c2:
```
        let mut ticker_redis =
            ConnectionManager::new_with_config(
                Client::open("redis://localhost")?,

                ConnectionManagerConfig::new()
                    .set_push_sender(|push_info| {
                        match push_info.kind {
                            redis::PushKind::PMessage | redis::PushKind::Message => {
                                let value: String = from_owned_redis_value(push_info.data.pop().unwrap())?;
                                let key: String = from_owned_redis_value(push_info.data.pop().unwrap())?;
							   let value: MyType = serde_json::from_str(&value).map_err(SendError)?;
							  sender.send(value).map_err(SendError)?;
                               Ok(())
                            }
                        };

                    })
                    .set_automatic_resubscription()
            ).await?;
```

Been digging through the code and I can't really see how errors are handled in set_push_sender.  I appreciate my toy example abuses it a little, but point still stands.  I can't actually see anything in the redis-rs codebase that handles the error.  In fact it kinda worries me that errors will subtly break things, which I think is a bug.

If there's an error in set_push_sender, I think it will terminate the reconnection task *silently*.

https://github.com/redis-rs/redis-rs/blob/cf1e15693e07cc397e4155f2b37e45db6053703a/redis/src/aio/connection_manager.rs#L536

```
    async fn check_for_disconnect_pushes(
        receiver: oneshot::Receiver<(
            ConnectionManager,
            UnboundedReceiver<PushInfo>,
            Arc<dyn AsyncPushSender>,
        )>,
    ) {
        let Ok((this, mut internal_receiver, external_sender)) = receiver.await else {
            return;
        };
        while let Some(push_info) = internal_receiver.recv().await {
            if push_info.kind == PushKind::Disconnection {
                this.reconnect(this.0.connection.load());
            }
            if external_sender.send(push_info).is_err() {
                return;
            }
        }
    }
```

nihohit:
Thanks for pointing out the issue! you're right that there's a connection here between 2 things that maybe should be separate. I need to think about what's the right thing to do here.

> If there's an error in set_push_sender, I think it will terminate the reconnection task silently.

By "reconnection task", do you mean the `check_for_disconnect_pushes` loop, or an actual reconnect attempt? Because reconnects are spawned, and ATM can't be aborted.

> Might I suggest that the _task_handle memember of the ConnectionManager struct impl future so people can await the background reconnection task.

[...]

PR #1612, opened 2025-04-16
ConnectionManager: Reconnect when RESP3 is set.

nihohit:
This changes the behavior of the connection manager, so that if it was configured with RESP3 but without a push sender, or with a push sender after it returned an error, it would still handle disconnect messages from the underlying connection and reconnect without a user request.

https://github.com/redis-rs/redis-rs/issues/1610

### DIFF
--- a/redis/tests/test_async.rs
+++ b/redis/tests/test_async.rs
@@ -1797,20 +1797,17 @@ mod basic_async {
     #[cfg(feature = "connection-manager")]
     #[rstest]
     #[cfg_attr(feature = "tokio-comp", case::tokio(RuntimeType::Tokio))]
-    #[case::async_std(RuntimeType::AsyncStd)]
-    fn manager_should_reconnect_without_actions_if_push_sender_is_set(
-        #[case] runtime: RuntimeType,
-    ) {
+    #[cfg_attr(feature = "async-std-comp", case::async_std(RuntimeType::AsyncStd))]
+    #[cfg_attr(feature = "smol-comp", case::smol(RuntimeType::Smol))]
+    fn manager_should_reconnect_without_actions_if_resp3_is_set(#[case] runtime: RuntimeType) {
         let ctx = TestContext::new();
         if ctx.protocol == ProtocolVersion::RESP2 {
             return;
         }
-        let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
 
         let max_delay_between_attempts = 2;
         let config = redis::aio::ConnectionManagerConfig::new()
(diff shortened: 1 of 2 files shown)


redis/src/aio/mod.rs
```diff
@@ -225,7 +243,7 @@ impl<T, Func: Fn(PushInfo) -> Result<(), T> + Send + Sync + 'static> AsyncPushSe
     fn send(&self, info: PushInfo) -> Result<(), SendError> {
         match self(info) {
             Ok(_) => Ok(()),
-            Err(_) => Err(SendError),
+            Err(_) => Err(SendError::new(true)),
```
rollo-b2c2, line 246: Hm, this isn't necessarily permanent.  Let's say your sender was like the lock example but instead it was doing opportunistic try_lock updates.

I suppose the I could just implement my own AsyncPushSender then.

rollo-b2c2: It's ever so slightly annoying that the Func wasn't initially defined as.

```
impl<T, Func: Fn(PushInfo) -> Result<(), SendError> + Send + Sync + 'static> AsyncPushSender for Func {
    fn send(&self, info: PushInfo) -> Result<(), SendError> {
        self(info)
    }
}
```

nihohit: You're right. Now I wonder whether the default should be true or false.

rollo-b2c2:
[`1d5264c` (#1614)](https://github.com/redis-rs/redis-rs/pull/1614/commits/1d5264ce25ece27b61e95de429d01190ebc18027)

Hey please take a look at my suggestion when you get a chance.

nihohit:
@rollo-b2c2 I think that until we figure out what's the right interface to report the sender closing, we should get this merged, since there's value in this fix even without the report. WDYT?

redis/tests/test_async.rs
```diff
@@ -1766,14 +1766,54 @@ mod basic_async {
...
+    #[cfg_attr(feature = "async-std-comp", case::async_std(RuntimeType::AsyncStd))]
+    #[cfg_attr(feature = "smol-comp", case::smol(RuntimeType::Smol))]
+    fn manager_should_reconnect_without_actions_if_push_sender_is_set_even_after_sender_returns_error(
         #[case] runtime: RuntimeType,
     ) {
         let ctx = TestContext::new();
         if ctx.protocol == ProtocolVersion::RESP2 {
             return;
         }
+        println!("running");
```
cheban96, line 1816: it's necessary?

PR #1613, opened 2025-04-16
**Breaking change** Remove deprecated aio::Connection.

nihohit:
This means adding an implementation for `Monitor` that isn't based on `aio::Connection`.


altanozlu (approved):
At first glance everything looks good, I didn't dig into it deeply.

PR #1614, opened 2025-04-16
ConnectionManager: Reconnect when RESP3 is set.

rollo-b2c2:
@nihohit just doing it against main for your visibility.

This changes the behavior of the connection manager, so that if it was configured with RESP3 but without a push sender, or with a push sender after it returned an error, it would still handle disconnect messages from the underlying connection and reconnect without a user request.

#1610This changes the behavior of the connection manager, so that if it was configured with RESP3 but without a push sender, or with a push sender after it returned an error, it would still handle disconnect messages from the underlying connection and reconnect without a user request.

#1610

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -564,9 +564,7 @@ impl ConnectionManager {
             if let Some(sender) = external_sender.as_ref() {
                 if let Err(err) = sender.send(push_info) {
                     if err.is_permanent() {
-                        external_sender.take();
-                    } else {
-                        return Err(err);
+                        return Err(external_sender.take());
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


nihohit:
What's the benefit of this suggestion? I try to avoid adding additional APIs and promises to the user.

rollo-b2c2:
Imagine if the push_sender has a non-recoverable error.

How do you know that connection manager is now in a bad state?



```
let (send, recv) = tokio::sync::mpsc::UnboundedSender::new();

async fn listen_to_updates(sender: tokio::sync::mpsc::UnboundedSender) {
	
	let conn = client.get_connection_manager_with_config().set_push_sender(send);
	conn.subscribe("channel-that-sends-once-a-second").await;
	
	// I want this to keep running and reconnecting forever until it has an unrecoverable error
	// From this scope there's no way to tell.  
    // So unless I add a oneshot I have to use

    futures::future::pending().await

    // and hope for the best
	
    // This is really a poor substitute for the internal task_handle.  Ideally I could just do this
    conn._task_handle.await;
}
```

nihohit:
Wouldn't it be simpler to just set some flag?

nihohit:
I see, but this requires a bit more thought. The same issue could happen with the multiplexed connection, cluster connection and even the sync connection. I'd like an interface that will be correct for all 3.

PR #1617, opened 2025-04-23
Document variadic tuple implementations for FromRedisValue and ToRedisValue

Kriskras99:
This is done using the rustdoc internal feature `fake_variadic`, which requires a nightly compiler. Therefore this is gated behind the `docsrs` feature which is set by docs.rs.

This can be tested by running `RUSTDOCFLAGS="--cfg docsrs" cargo doc`.

Based on implementations in [Bevy](https://github.com/bevyengine/bevy/pull/14703) and [Serde](https://docs.rs/serde/latest/src/serde/ser/impls.rs.html#416).

I tried to avoid duplicating the trait implementations for the tuple with a single item, but the `fake_variadic` attribute needs to be on the implementation of the tuple with the single item, and can't be on the tuple with 12 items. Alternative would've been reworking how the macros work completely.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1820,8 +1820,8 @@ impl_to_redis_args_for_map!(
 
 macro_rules! to_redis_args_for_tuple {
     () => ();
-    ($($name:ident,)+) => (
-        #[doc(hidden)]
+    ($(#[$meta:meta],)*$($name:ident,)+) => (
+        $(#[$meta])*
         impl<$($name: ToRedisArgs),*> ToRedisArgs for ($($name,)*) {
             // we have local variables named T1 as dummies and those
             // variables are unused.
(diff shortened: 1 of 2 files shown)


redis/src/types.rs
```diff
@@ -2410,6 +2427,110 @@ impl FromRedisValue for () {
 
 macro_rules! from_redis_value_for_tuple {
     () => ();
+    ($name:ident, ) => (
```
nihohit, line 2430: Improving the documentation at the cost of increasing code complexity isn't really a tradeoff I'm willing to make.

Kriskras99: If I rework the macros to be less complex, would that be acceptable or are the cfg_attrs also unwantend?:
```rust
#![cfg_attr(docsrs, feature(doc_cfg, rustdoc_internals))]
#![cfg_attr(docsrs, expect(internal_features)]
```

nihohit:
> If I rework the macros to be less complex, would that be acceptable

Sure, but I don't want to lead you on - the macro is complex for a good reason. I'm not sure that there's a way to simplify it enough to make the repetition not problematic.

> are the cfg_attrs also unwantend?:

I'm not a rustdoc expert, but AFAIS these feature flags only enable other features, right? They don't change anything if you just add them?

Kriskras99:
Yes, that's right.

I've force-pushed a new implementation which doesn't duplicate the complicated macro.
With `--cfg docsrs` enabled, it looks like this:
Screenshot From 2025-04-26 20-21-32
And this is without:
Screenshot From 2025-04-26 20-21-55

(It's also possible to just hide the tuple implementation, but that's up to you)

nihohit:
That looks great! Please fix the CI and we'll merge this.

Kriskras99:
Should be fixed now

nihohit (approved):

nihohit:
Thanks! I appreciate the effort :)

PR #1618, opened 2025-04-24
Add an optional bytes::BufMut function to RedisWrite

Kriskras99:
I'm dealing with some legacy stuff where I need to serialize some data as ProtoBuf and send it via Redis. Currently this requires creating an extra buffer where I serialize the data, and then copy that data to the writer. This PR allows serializing directly into the writer.

This function is gated behind the "bytes" feature, which is automatically enabled with any of the async features.

A default implementation is provided, so that this is not a breaking change for upstream.

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -353,6 +353,48 @@ impl RedisWrite for Cmd {
 
         CmdBufferedArgGuard(self)
     }
+
+    #[cfg(feature = "bytes")]
+    fn bufmut_for_next_arg(&mut self, capacity: usize) -> impl bytes::BufMut + '_ {
+        self.data.reserve(capacity);
+        struct CmdBufferedArgGuard<'a>(&'a mut Cmd);
+        impl Drop for CmdBufferedArgGuard<'_> {
+            fn drop(&mut self) {
+                self.0.args.push(Arg::Simple(self.0.data.len()));
+            }
+        }
+        unsafe impl bytes::BufMut for CmdBufferedArgGuard<'_> {
+            fn remaining_mut(&self) -> usize {
+                self.0.data.remaining_mut()
+            }
+
+            unsafe fn advance_mut(&mut self, cnt: usize) {
+                self.0.data.advance_mut(cnt);
+            }
+
+            fn chunk_mut(&mut self) -> &mut bytes::buf::UninitSlice {
+                self.0.data.chunk_mut()
+            }
+
+            // Vec specializes these methods, so we do too
+            fn put<T: bytes::buf::Buf>(&mut self, src: T)
+            where
+                Self: Sized,
+            {
+                self.0.data.put(src);
+            }
+
+            fn put_slice(&mut self, src: &[u8]) {
+                self.0.data.put_slice(src);
+            }
+
+            fn put_bytes(&mut self, val: u8, cnt: usize) {
+                self.0.data.put_bytes(val, cnt);
+            }
+        }
+
+        CmdBufferedArgGuard(self)
+    }
 }
 
 impl Default for Cmd {
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -354,6 +354,13 @@ impl RedisWrite for Cmd {
         CmdBufferedArgGuard(self)
     }
 
+    fn reserve_space_for_args(&mut self, additional: &[usize]) {
+        // Reserve space for the sum of the data size of all the arguments
+        self.data.reserve(additional.iter().sum());
+        // Reserve space for the argument markers
+        self.args.reserve(additional.len());
+    }
+
     #[cfg(feature = "bytes")]
     fn bufmut_for_next_arg(&mut self, capacity: usize) -> impl bytes::BufMut + '_ {
         self.data.reserve(capacity);
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -354,11 +354,15 @@ impl RedisWrite for Cmd {
         CmdBufferedArgGuard(self)
     }
 
-    fn reserve_space_for_args(&mut self, additional: &[usize]) {
-        // Reserve space for the sum of the data size of all the arguments
-        self.data.reserve(additional.iter().sum());
-        // Reserve space for the argument markers
-        self.args.reserve(additional.len());
+    fn reserve_space_for_args(&mut self, additional: impl Iterator<Item = usize>) {
+        let mut capacity = 0;
+        let mut args = 0;
+        for add in additional {
+            capacity += add;
+            args += 1;
+        }
+        self.data.reserve(capacity);
+        self.args.reserve(args);
     }
 
     #[cfg(feature = "bytes")]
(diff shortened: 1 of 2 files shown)


redis/src/cmd.rs
```diff
@@ -353,6 +353,48 @@ impl RedisWrite for Cmd {
 
         CmdBufferedArgGuard(self)
     }
+
+    #[cfg(feature = "bytes")]
+    fn bufmut_for_next_arg(&mut self, capacity: usize) -> impl bytes::BufMut + '_ {
+        self.data.reserve(capacity);
```
nihohit, line 359: The underlying vector might have excess capacity, so this should calculate exactly how much to reserve.

redis/src/types.rs
```diff
@@ -1351,6 +1351,68 @@ pub trait RedisWrite {
...
     fn writer_for_next_arg(&mut self) -> impl std::io::Write + '_;
+
+    #[cfg(feature = "bytes")]
+    /// Appends an empty argument to the command, and returns a
+    /// [`bytes::BufMut`] instance that can write to it.
+    ///
+    /// `capacity` should be equal or greater to the amount of bytes
+    /// expected, as some implementations might not be able to resize
+    /// the returned buffer.
+    fn bufmut_for_next_arg(&mut self, capacity: usize) -> impl bytes::BufMut + '_ {
```
nihohit, line 1362: maybe this should return a concrete type / an extending trait to BufMut, that allows writing multiple args? That way the user doesn't have to pay the cost of reallocation for each arg.

redis/src/cmd.rs
```diff
@@ -353,6 +353,48 @@ impl RedisWrite for Cmd {
...
+        struct CmdBufferedArgGuard<'a>(&'a mut Cmd);
+        impl Drop for CmdBufferedArgGuard<'_> {
+            fn drop(&mut self) {
+                self.0.args.push(Arg::Simple(self.0.data.len()));
```
nihohit, line 363: what happens if the guard is dropped without adding any argument?

redis/src/types.rs
```diff
@@ -1351,6 +1351,68 @@ pub trait RedisWrite {
     /// Appends an empty argument to the command, and returns a
     /// [`std::io::Write`] instance that can write to it.
     fn writer_for_next_arg(&mut self) -> impl std::io::Write + '_;
+
+    #[cfg(feature = "bytes")]
+    /// Appends an empty argument to the command, and returns a
```
nihohit, line 1356: This would be the final part of the PR, but the documentation should include the safety guarantees / assumptions of this function - for example, what happens if the user writes more than a single argument, or doesn't write any data at all.

nihohit (commented):
I think that this is a good change, but we need to understand whether this is the best API, and whether there are any footguns hidden here,

redis/src/cmd.rs
Kriskras99: > After calling `reserve`, capacity wil be greater than or equal to `self.len() + additional`. Does nothing if capacity is already sufficient.

So extra complexity on our side is not needed.

redis/src/types.rs
Kriskras99: Maybe something like:
```rust
fn as_bufmut_writer(&mut self) -> impl RedisBufMutWriter + '_;

pub trait RedisBufMutWriter {
    // Reserve at least `additional` capacity.
    fn reserve(&mut self, additional: usize);
    // Get a `bytes::BufMut` to write an arg to.
    //
    // `size` is the expected size of the arg, if `size` is less than
    // the amount of bytes written the implementation is allowed to panic.
    fn write_arg(&mut self, size: usize) -> impl bytes::BufMut + '_;
}
```
Or we can add the `reserve` function to the RedisWriter implementation (with a default implementation that does nothing).

redis/src/cmd.rs
Kriskras99: It works the same way as `writer_for_next_arg`, see https://github.com/redis-rs/redis-rs/pull/905#discussion_r1947894825

redis/src/types.rs
Kriskras99: When adding that I should also add those same safety guarantees to `writer_for_next_arg`

redis/src/cmd.rs
nihohit: And that's just from February. I'm never going to learn 🤦 
Yes, please also add a clarification to all relevant docs.

redis/src/types.rs
nihohit: How `RedisWriter::reserve` be implemented for `Vec<Vec<u8>>`?

Kriskras99: It can't and will be a no-op in that case, but that's fine as it's an optional optimisation. For `Vec<Vec<u8>>` you would need a `reserve_args(n: usize)` function or something similar.

Kriskras99: I would document `reserve` like this:
```rust
/// Tell the writer to expect at least `additional` more bytes
///
/// These additional bytes can be part of one or more arguments. The implementer is allowed to
/// ignore this information if it's internal command buffer is not continuous, like [`Vec<Vec<u8>>`]
fn reserve(&mut self, additional: usize);
```
And maybe rename `reserve` to something more fitting.

redis/src/types.rs
Kriskras99: I've added some documentation to `bufmut_for_next_arg` and `writer_for_next_arg` about possible issues.

redis/src/types.rs
Kriskras99:
```rust
    /// Reserve space for `additional` arguments in the command
    ///
    /// `additional` is a list of the byte sizes of the arguments.
    ///
    /// It's strongly recommended that implementors implement this function,
    /// as the default implementation is a no-op.
    fn reserve_space_for_args(&mut self, additional: &[usize])
```
I came up with this function for the RedisWriter trait. This allows both `Cmd` and `Vec<Vec<u8>>` to utilize the information. I've again provided a default implementation (a no-op), so that this isn't a breaking change for anyone.

Kriskras99:
I've added some tests, however there's not really a way to test `reserve_space_for_args` as there's no way to observe the capacity of `Cmd`.

redis/src/types.rs
```diff
@@ -1350,7 +1350,93 @@ pub trait RedisWrite {
...
+    /// as an argument.
     fn writer_for_next_arg(&mut self) -> impl std::io::Write + '_;
+
+    /// Reserve space for `additional` arguments in the command
+    ///
+    /// `additional` is a list of the byte sizes of the arguments.
+    ///
+    /// It's strongly recommended that implementors implement this function,
+    /// as the default implementation is a no-op.
+    fn reserve_space_for_args(&mut self, additional: &[usize]) {
```
nihohit, line 1367: please provide an example for when this can be used.

nihohit (approved):

redis/src/types.rs
Kriskras99: I've added an example, and am thinking to change the function signature of `reserve_space_for_args` to:
```rust
fn reserve_space_for_args(&mut self, additional: impl Itererator<Item = usize>);
```

nihohit: yes, that makes sense to me.

Kriskras99: Done

Kriskras99:
Thanks!

PR #1622, opened 2025-04-30
Make `reserve_space_for_args` take a `IntoIterator`

Kriskras99:
With `IntoIterator` a Vec or array can be provided without the user needing to call `.into_iter()`. As this trait also has a default implementation for all traits (a no-op that returns the same iterator) this also works with iterators.

Also simplifies the example in the documentation and fixed a typo.


nihohit (approved):

Kriskras99:
That was quick, thanks!

PR #1632, opened 2025-05-01
Remove the `ring` dependency from `rustls`.

nihohit:


kamulos:
Thanks! Did a quick test with my service: it seems to work perfectly and `rustls` is now using `aws-lc` for me  :+1:

avifenesh:
@nihohit while AWS one is better, and for ring has fewer options, the size of the aws is much, much heavier.  What consideration were taken for the decision? I'm in the same question currently. rustls decided to maintain the ring crate if this is the reason.

PR #1634, opened 2025-05-02
Implement Error::source for the RedisError struct

StefanPalashev:
# Implement the `source` method of the `error::Error` _trait_ for the `RedisError` struct.

### Description
`Error::source()` is the recommended method for accessing the error chain, and it should be used instead of `cause()` in modern Rust code.

### Testing
Add unit tests that cover cases with and without an underlying source error.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -766,6 +766,13 @@ impl error::Error for RedisError {
             _ => None,
         }
     }
+
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        match self.repr {
+            ErrorRepr::IoError(ref err) => Some(err),
+            _ => None,
+        }
+    }
 }
 
 impl fmt::Display for RedisError {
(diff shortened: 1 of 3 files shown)


redis/src/types.rs
```diff
@@ -749,6 +749,12 @@ impl From<(ErrorKind, &'static str, String)> for RedisError {
     }
 }
 
+impl From<(String, Option<String>)> for RedisError {
```
nihohit, line 752: this is just for testing right? I prefer to keep the `make_extension_error` function, to make this explicit.

nihohit (approved):
Only one comment. Thanks for the contribution!

redis/src/types.rs
StefanPalashev: Yes, this is done only for testing purposes.

The `make_extension_error` function has to be exposed in order for it to be usable in the tests.

Additionally, all other variants of the `ErrorRepr` enum are already covered by `impl From<...> for RedisError` functions, so this addition simply completes the set.

nihohit: Right, so I prefer that the `make_extension_error` function be exposed than to add this, because I want it to be explicit that this creates an extension error.

nihohit: or even simpler, just add these tests to `types.rs` instead of the tests suite.

nihohit (approved):

nihohit:
Thanks for the contribution!

PR #1636, opened 2025-05-08
Allow for buffer reuse with `Cmd`

Kriskras99:
This adds a `clear()` function, which clears the buffers but keeps any settings. It also makes the `write_packed_command()` function public, so the command can be serialized to an existing buffer.

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -473,6 +473,35 @@ impl Cmd {
         (self.args.capacity(), self.data.capacity())
     }
 
+    /// Clears the command, resetting it completely.
+    ///
+    /// This is equivalent to [`Cmd::new`], except the buffer capacity is kept.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// # use redis::{Client, Cmd};
+    /// # let client = Client::open("redis://127.0.0.1/").unwrap();
+    /// # let mut con = client.get_connection().expect("Failed to connect to Redis");
+    /// let mut cmd = Cmd::new();
+    /// cmd.arg("SET").arg("foo").arg("42");
+    /// cmd.query::<()>(&mut con).expect("Query failed");
+    /// cmd.clear();
+    /// // This reuses the allocations of the previous command
+    /// cmd.arg("SET").arg("bar").arg("42");
+    /// cmd.query::<()>(&mut con).expect("Query failed");
+    /// ```
+    pub fn clear(&mut self) {
+        self.data.clear();
+        self.args.clear();
+        self.cursor = None;
+        self.no_response = false;
+        #[cfg(feature = "cache-aio")]
+        {
+            self.cache = None;
+        }
+    }
+
     /// Appends an argument to the command.  The argument passed must
     /// be a type that implements `ToRedisArgs`.  Most primitive types as
     /// well as vectors of primitive types implement it.
@@ -516,6 +545,10 @@ impl Cmd {
     }
 
     /// Returns the packed command as a byte vector.
+    ///
+    /// This is a wrapper around [`write_packed_command`] that creates a [`Vec`] to write to.
+    ///
+    /// [`write_packed_command`]: Self::write_packed_command
     #[inline]
     pub fn get_packed_command(&self) -> Vec<u8> {
         let mut cmd = Vec::new();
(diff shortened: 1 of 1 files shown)


nihohit (commented):
please add tests, demonstrating that this is safe to use even after `cursor` and `no_response` have been set.

redis/src/cmd.rs
```diff
@@ -473,6 +473,34 @@ impl Cmd {
...
+    /// // This reuses the allocations of the previous command
+    /// cmd.arg("SET").arg("bar").arg("42");
+    /// cmd.query::<()>(&mut con).expect("Query failed");
+    /// ```
+    ///
+    /// [`set_cache_config`]: Self::set_cache_config
+    /// [`set_no_response`]: Self::set_no_response
+    pub fn clear(&mut self) {
+        self.data.clear();
+        self.args.clear();
```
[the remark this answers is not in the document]
Kriskras99: Done

redis/src/cmd.rs
```diff
@@ -473,6 +473,35 @@ impl Cmd {
...
+    /// ```
+    pub fn clear(&mut self) {
+        self.data.clear();
+        self.args.clear();
+        self.cursor = None;
+        self.no_response = false;
+        #[cfg(feature = "cache-aio")]
+        {
+            self.cache = None;
+        }
```
altanozlu, line 502: It's cleaner without brackets
```suggestion
        #[cfg(feature = "cache-aio")]
        self.cache = None;
```

redis/src/cmd.rs
```diff
@@ -800,6 +841,43 @@ mod tests {
...
+        // Everything should be reset, but the capacity should still be there
+        assert!(cmd.data.is_empty());
+        assert!(cmd.data.capacity() > 0);
+        assert!(cmd.args.is_empty());
+        assert!(cmd.args.capacity() > 0);
+        assert_eq!(cmd.cursor, None);
+        assert!(!cmd.no_response);
+        assert!(cmd.cache.is_none());
+    }
+
```
altanozlu, line 880: I think there is no need to add 2 tests just for cache-aio difference.

```suggestion
    #[test]
    fn test_cmd_clean() {
        let mut cmd = Cmd::new();
        cmd.arg("key").arg("value");
        cmd.set_no_response(true);
        cmd.cursor_arg(24);
        #[cfg(feature = "cache-aio")]
        cmd.set_cache_config(crate::CommandCacheConfig::default());
        cmd.clear();

        // Everything should be reset, but the capacity should still be there
        assert!(cmd.data.is_empty());
        assert!(cmd.data.capacity() > 0);
        assert!(cmd.args.is_empty());
        assert!(cmd.args.capacity() > 0);
        assert_eq!(cmd.cursor, None);
        assert!(!cmd.no_response);
        #[cfg(feature = "cache-aio")]
        assert!(cmd.cache.is_none());
    }
```

redis/src/cmd.rs
Kriskras99: Doesn't work:
```
    Checking redis v0.30.0 (/home/kriskras99/Source/redis-rs/redis)
error[E0658]: attributes on expressions are experimental
   --> redis/src/cmd.rs:499:9
    |
499 |         #[cfg(feature = "cache-aio")]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #15701 <https://github.com/rust-lang/rust/issues/15701> for more information
    = help: add `#![feature(stmt_expr_attributes)]` to the crate attributes to enable
    = note: this compiler was built on 2025-05-10; consider upgrading it if it is out of date

For more information about this error, try `rustc --explain E0658`.
error: could not compile `redis` (lib) due to 1 previous error
```

redis/src/cmd.rs
Kriskras99: Same compiler error as other suggestion

Kriskras99: So I could put both expressions in brackets, and put the cfg on that (like I did in `clear`). Or do it in two separate tests. I don't have a preference.

altanozlu (approved):

redis/src/cmd.rs
altanozlu: In that case I also don't have strong opinions, LGTM

Kriskras99:
@nihohit I've added tests to assert that `Cmd::clear() == Cmd::new()` and improved the documentation. Anything else?

nihohit (approved):
Thanks for the contribution!

PR #1638, opened 2025-05-12
feat: extend sentinel api to be a bit more flexible

hugobpx:
The current `redis::sentinel::SentinelClient` api is limited to `get_connection` and `async_get_connection`. In order to establish connections to the correct (master/replica) Redis instance, it will first attempt to find the master/replica and create a `redis::Client` pointing at it. 

This PR makes `get_client` and `async_get_client` publicly available so users have the ability of getting such `redis::Client` and use it directly without going through sentinel facade.

Additionally, I'm adding a method that allows a user to obtain a `redis::Client` that points to one of the provided Sentinel instances. Having such a client allows, for instance, to leverage `pubsub` to subscribe to Sentinel events.

### DIFF
--- a/redis/src/sentinel.rs
+++ b/redis/src/sentinel.rs
@@ -1058,7 +1058,7 @@ impl SentinelClient {
         })
     }
 
-    /// Returns a `redis::Client` considering the server type.
+    /// Returns a [`Client`] considering the server type.
     pub fn get_client(&mut self) -> RedisResult<Client> {
         match self.server_type {
             SentinelServerType::Master => self
@@ -1070,13 +1070,13 @@ impl SentinelClient {
         }
     }
 
-    /// Returns a [`redis::Client`] connected to **the first Sentinel node that
+    /// Returns a [`Client`] connected to **the first Sentinel node that
     /// responds successfully to `ROLE` with the expected value**.
     ///
     /// The function walks through the list supplied in
     /// `Sentinel::sentinels_connection_info`, trying each address in order:
     ///
-    /// 1. `Client::open` attempts to build a client for that Sentinel.
+    /// 1. [Client::open] attempts to build a client for that Sentinel.
     /// 2. We immediately issue `ROLE` to confirm the node is reachable and is a Setinel node.
     /// 3. The **first** node that passes both checks, the client is returned.
     ///
@@ -1133,7 +1133,7 @@ impl SentinelClient {
 #[cfg(feature = "aio")]
 #[cfg_attr(docsrs, doc(cfg(feature = "aio")))]
 impl SentinelClient {
-    /// Returns a `redis::Client` considering the server type.
+    /// Returns a [`Client`] considering the server type.
     pub async fn async_get_client(&mut self) -> RedisResult<Client> {
         match self.server_type {
             SentinelServerType::Master => {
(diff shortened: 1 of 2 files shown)


nihohit:
Hi, thanks for the contribution!

This is a good change, but it requires both tests, and that the documentation make it clear that the client only points to the right node at the time of the function execution - the rust node layout could change later, and a node that was previously a replica might now be a primary, or vice versa.

hugobpx:
Hi @nihohit, apologies for the late reply.

I'm unsure what tests should cover as existing tests should already be covering the `get_client` and `async_get_client`. Regarding `get_sentinel_client`, I can add a very simple test just to add coverage but all its doing is taking the first sentinel node present in `sentinel.sentinels_connection_info`. At that point the config value should already be a proper type and gone through existing validation. Additionally, sentinel nodes don't have the concept of master/replica, that's why I'm just taking the first one defined and creating a `redis::Client`.

nihohit:
> I can add a very simple test just to add coverage but all its doing is taking the first sentinel node present in sentinel.sentinels_connection_info

That's fine, but please add to the test a check that the returned node is actually a sentinel node. But the more important part of the comment was the documentation - it should be full, and explain when would the user want to use these functions, and warn against the fact that the role of the node that the client points to might change.

hugobpx:
hey @nihohit, finally got around to finish this. Changed things a bit:

- Instead of just blindly returning a `redis::Client` to the first sentinel node configured, I'm now traversing the list of provided sentinel nodes to find the first that says its a sentinel node
- Added docs that should help understand what the methods are for
- Added async and sync versions
- Included tests

redis/src/sentinel.rs
```diff
@@ -1098,6 +1149,57 @@ impl SentinelClient {
         }
     }
 
+    /// Returns a [`redis::Client`] connected to **the first Sentinel node that
+    /// responds successfully to `ROLE` with the expected value**.
+    ///
+    /// The function walks through the list supplied in
+    /// `Sentinel::sentinels_connection_info`, trying each address in order:
```
nihohit, line 1156: this is the async implementation - it can try all servers concurrently, and return according to the first successful answer. no need to iterate in order.
I think that `select_ok` is the algorithm to use here.

redis/tests/test_sentinel.rs
```diff
@@ -295,6 +295,91 @@ fn test_sentinel_client() {
...
+    );
+
+    assert_eq!(
+        first_configured_sentinel.addr,
+        sentinel_client_2.get_connection_info().addr
+    );
+}
+
+#[test]
+fn test_sentinel_client_async_io_error() {
```
nihohit: not async - test_sentinel_client_io_error

what is being tested here, exactly?

redis/tests/test_sentinel.rs
```diff
@@ -295,6 +295,91 @@ fn test_sentinel_client() {
     }
 }
 
+#[test]
+fn test_sentinel_client_async() {
```
nihohit: not async

redis/tests/test_sentinel.rs
hugobpx: this branch [here](https://github.com/redis-rs/redis-rs/pull/1638/files#diff-6e43330d9f65fae075a9bb65ccb71db69f54fe2da7a94512ec00c7f423bf1058R1113). So that the user is aware of the error when reaching out to the sentinel servers. In this case, an io error because the address doesn't resolve.

redis/src/sentinel.rs
hugobpx: Yeah contemplated a concurrent approach but gets tricky because I opted to reuse existing pieces and `async_try_single_sentinel` takes in a `&mut Option<AsyncConnection>` to reuse existing connections.

nihohit:
ok, let's fix the lints and get this merged

nihohit:
LGTM. let's fix those lints and merge this.

nihohit:
Thanks for the contribution!

PR #1642, opened 2025-05-25
Add Client Side Caching for async Cluster Connections

altanozlu:
This is quite a small change because most of the logic is already implemented in other modules used by ClusterConnection.

I wanted to consolidate cache tests nicely across mpx,cm and cluster connections but so far failed to do so.

With that reason patch adds two new tests.

Also I have tested slot change on Redis 8, when there is a slot change redis will send invalidation but couldn't write a simple test for it.

### DIFF
--- a/redis/src/cluster_async/mod.rs
+++ b/redis/src/cluster_async/mod.rs
@@ -119,6 +119,8 @@ use crate::{
     RedisFuture, RedisResult, ToRedisArgs, Value,
 };
 
+#[cfg(feature = "cache-aio")]
+use crate::caching::{CacheManager, CacheStatistics};
 use crate::ProtocolVersion;
 use futures_sink::Sink;
 use futures_util::{
@@ -137,6 +139,8 @@ struct ClientSideState {
     _task_handle: HandleContainer,
     response_timeout: Option<Duration>,
     runtime: Runtime,
+    #[cfg(feature = "cache-aio")]
+    cache_manager: Option<CacheManager>,
 }
 
 /// This represents an async Redis Cluster connection.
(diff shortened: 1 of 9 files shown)


redis/src/cluster_async/mod.rs
```diff
@@ -1332,6 +1348,10 @@ where
     if let Some(resolver) = dns_resolver {
         config = config.set_dns_resolver_internal(resolver.clone());
     }
+    #[cfg(feature = "cache-aio")]
+    if let Some(cache_manager) = cache_manager {
+        config = config.set_cache_manager(cache_manager.clone_and_increase_epoch());
```
nihohit, line 1353: don't we need to also increase epoch after a topology refresh? what would happen if slots were migrated between nodes without a disconnect?

altanozlu: > Also I have tested slot change on Redis 8, when there is a slot change redis will send invalidation but couldn't write a simple test for it.

I don't think there is a need for it since old slot owner will send invalidation.

nihohit:
also need to fix the flag-frenzy build.

redis/src/cluster_async/mod.rs
altanozlu: Added a case of slot change with and without migrating the key, it's also exposed caching of ServerError's and I have also changed that.

altanozlu: Redis 6.2.18 doesn't send invalidation when migrate isn't used. I would like to document it and leave 6.2.X like this, what do you think ?

nihohit: > I would like to document it and leave 6.2.X like this, what do you think ?

Yup, sounds great.

nihohit (approved):

PR #1648, opened 2025-05-30
Fix AsyncIter implementation of stream.

nihohit:
When the `safe_iterators` feature isn't implemented, the `Stream` implementation of AsyncIter should return T, not `RedisResult<T>`.


nihohit:
@threema-donat, anything I might've missed?

PR #1658, opened 2025-06-04
Minimize the lifetime of mutex lock guards

svix-jplatte:
We've been investigating some tokio poll latency spikes that then lead to timeouts in some redis accesses. One of the theories going around is that it might be bug in this library. I have audited `Mutex` and `RwLock` uses and found some lock guards that are being kept alive a little longer than necessary, so here's a PR to fix that (though it seems very unlikely that any of these changes are related to our issue - the bug hunt continues).

### DIFF
--- a/redis/src/caching/sharded_lru.rs
+++ b/redis/src/caching/sharded_lru.rs
@@ -94,6 +94,7 @@ impl ShardedLRU {
                 }
             }
         }
+        drop(lru_cache);
         self.statistics.increase_miss(1);
         None
     }
@@ -138,7 +139,9 @@ impl ShardedLRU {
     }
 
     pub(crate) fn invalidate(&self, cache_key: &Vec<u8>) {
-        if let Some(cache_holder) = self.get_shard(cache_key).pop(cache_key) {
+        // Use a separate binding for this to release the lock guard before calling increase_invalidate.
+        let maybe_cache_holder = self.get_shard(cache_key).pop(cache_key);
+        if let Some(cache_holder) = maybe_cache_holder {
             self.statistics
                 .increase_invalidate(cache_holder.value_list.len());
         }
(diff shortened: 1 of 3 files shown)


redis/src/caching/sharded_lru.rs
```diff
@@ -138,7 +139,9 @@ impl ShardedLRU {
     }
 
     pub(crate) fn invalidate(&self, cache_key: &Vec<u8>) {
-        if let Some(cache_holder) = self.get_shard(cache_key).pop(cache_key) {
+        // Use a separate binding for this to release the lock guard before calling increase_invalidate.
+        let maybe_cache_holder = self.get_shard(cache_key).pop(cache_key);
+        if let Some(cache_holder) = maybe_cache_holder {
```
altanozlu, line 144: I believe this will be fixed by itself when edition becomes 2024 ?
ref: https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-if-let-scope.html

svix-jplatte: I'm pretty sure it doesn't. That page talks about things being dropped before `else`, not before entering the condition body. The latter would be very annoying in a lot of cases.

nihohit: is there a measurable difference? I tried finding it in the rust playground, but failed.
I'm asking out of curiosity,

svix-jplatte: I haven't actually done any benchmarking. I made these changes because they seemed like unambiguous wins, regardless of how small.

nihohit: I'm not sure that the variable split are an unambiguous wins, but I also don't think that they're harmful, so I'm fine with this :)

nihohit:
thanks for the contribution!

PR #1663, opened 2025-06-06
feat: Make transaction status of pipelines public

somechris:
When trying to mock Redis connections, the `ConnectionLike`'s `req_packed_commands` expects a different response shape on `transactional` and non-`transactional` pipelines.

(Vec with single entry of Redis array of results, resp. Vec of results)

In order to hide this difference from users, it would be helpful to check if a pipeline is in transaction mode or not.

To allow this, we make the accessor function public.

### DIFF
--- a/redis/src/pipeline.rs
+++ b/redis/src/pipeline.rs
@@ -72,8 +72,9 @@ impl Pipeline {
         self
     }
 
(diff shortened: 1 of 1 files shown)


nihohit:
Thanks for the contribution!

PR #1664, opened 2025-06-07
Pipeline: Add comments explaining offset and count

somechris:
While working on a mocking `ConnectionLike` it at first was not straight forward to me why we use the offset/count that we do.

After the understanding what `atomic` does, it's clear.

But adding some explanation will hopefully help the next person who needs to understand them.


nihohit (approved):
Thanks for the contribution. documentation is always appreciated :)

PR #1669, opened 2025-06-13
Move TCP features to settings.

nihohit:
Additionally, hide internal structure of the connection settings, to allow for non-backwards-compatible changes.


nihohit:
@altanozlu WDYT?

redis/src/connection.rs
```diff
@@ -231,23 +236,108 @@ impl fmt::Display for ConnectionAddr {
...
+
+    /// Sets the TCP settings for the connection.
+    pub fn set_tcp_settings(mut self, tcp_settings: TcpSettings) -> Self {
+        self.tcp_settings = tcp_settings;
+        self
+    }
+
+    /// Sets the username for the connection's ACL
+    pub fn set_username(mut self, username: impl Into<String>) -> Self {
+        self.redis.username = Some(username.into());
```
altanozlu: Since these ones are just coy of RedisConnectionInfo, isn't better to call it directly ?

altanozlu (approved):
I didn't find a chance to go throughly (probably won't find any in time), thanks for removing feature backed connection settings.

Issue #1668, opened 2025-06-13
`num-bigint` is not optional


tbrezot:
The `num-bigint` crate is [not marked as optional](https://github.com/redis-rs/redis-rs/blob/c8df73ddf2e424b4c1f464b4a842ff4321121e9c/redis/Cargo.toml#L101), even though is seems to be (there is a feature with its name and the comment line 98 makes it rather explicit).

This has impact in the build-time of all client crates (building the `num-bigint` from scratch takes about 8 seconds on my machine).

nihohit:
It was optional in the past, but now it isn't since RESP3 added [a big int value](https://github.com/redis-rs/redis-rs/blob/c8df73ddf2e424b4c1f464b4a842ff4321121e9c/redis/src/types.rs#L297). Do you have a better representation for a big integer?

tbrezot:
No, I don't know of any better representation. A string could be returned instead, but this cripples ergonomics. Wrapping all occurrences behind the feature as you propose in #1670 could be okay. Otherwise just fixing the doc and removing the feature to acknowledge it is indeed mandatory would be enough imho.

nihohit:
fixed in branch 1.0.x.

PR #1670, opened 2025-06-13
Make the num-bigint crate optional again.

nihohit:
https://github.com/redis-rs/redis-rs/issues/1668

### DIFF
--- a/redis-test/Cargo.toml
+++ b/redis-test/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.75"
 bench = false
 
 [dependencies]
-redis = { version = "1.0.0-alpha", path = "../redis" }
+redis = { version = "1.0.0-alpha", path = "../redis", default-features = false }
 bytes = { version = "1", optional = true }
 futures = { version = "0.3", optional = true }
 tempfile = "=3.20.0"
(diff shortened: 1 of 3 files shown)


redis/src/parser.rs
```diff
@@ -623,12 +629,15 @@ mod tests {
...
-                BigInt::parse_bytes(b"3492890328409238509324850943850943825024385", 10).unwrap()
-            )
+        #[cfg(feature = "num-bigint")]
+        let expected = Value::BigNumber(
+            num_bigint::BigInt::parse_bytes(b"3492890328409238509324850943850943825024385", 10)
+                .unwrap(),
         );
+        #[cfg(not(feature = "num-bigint"))]
+        let expected =
+            Value::BigNumber(b"(3492890328409238509324850943850943825024385\r\n".to_vec());
```
altanozlu:
```suggestion
            Value::BigNumber("3492890328409238509324850943850943825024385".to_string());
```
If num-bigint is going to be optional, I would expect getting the number as a string from the library, instead of getting bytes including CRLF and (.

nihohit: good catch! this revealed the fact that we don't actually test for no default features, and this flow wasn't at all tested!

PR #1673, opened 2025-06-16
Add support for vector set commands

StefanPalashev:
# Add Support for Redis Vector Set Commands

This Pull Request implements the full suite of Redis Vector Set commands as defined in the [Redis documentation](https://redis.io/docs/latest/develop/data-types/vector-sets/), along with unit tests to verify correctness and expected behavior.

The following commands have been implemented:
- ➕ `VADD` – Add a new element to a vector set.
- 🧮 `VCARD` – Return the number of elements in a vector set.
- 📏 `VDIM` – Return the dimensionality of a vector set.
- 🧬 `VEMB` – Return the approximate vector associated with a given element in the vector set.
- 🔍 `VGETATTR` – Retrieve attribute values associated with an element of a vector set.
- ℹ️ `VINFO` – Return information about a vector set.
- 🔗 `VLINKS` – Return the neighbors of a specified element in a vector set.
- 🎲 `VRANDMEMBER` – Return one or more random elements from a vector set.
- ➖ `VREM` – Remove an element from a vector set.
- 🛠️ `VSETATTR` – Associate a JSON object with an element in a vector set.
- 🧠 `VSIM` – Return elements similar to a given vector or element. 

Command aliases:
- ❌ `VDELATTR` - A utility command that can be used to remove the associated attributes of an element in a vector set. It is an alias for `VSETATTR` with an empty string.

## 🧪 Tests

Each command is covered by unit tests that verify:

- Correct output for valid inputs
- Expected behavior for edge cases (e.g., non-existent keys or elements)
- Consistency with the Redis vector set specification

The tests serve both as validation and usage examples for future contributors.

---


.github/workflows/rust.yml
```diff
@@ -54,6 +54,12 @@ jobs:
               db-name: redis,
               db-version: 8.0-rc1
             },
+            {
+              rust: stable,
```
nihohit: please replace the 8.0-rc1 config, instead of adding a new row to the matrix.

redis/Cargo.toml
```diff
@@ -91,9 +91,9 @@ rustls-native-certs = { version = "0.8", optional = true }
 tokio-rustls = { version = "0.26", optional = true, default-features = false }
 futures-rustls = { version = "0.26", optional = true, default-features = false }
 
-# Only needed for RedisJSON Support
-serde = { version = "1.0.219", optional = true }
-serde_json = { version = "1.0.140", optional = true }
+# Needed for RedisJSON Support and vector sets
+serde = { version = "1.0.219" }
```
nihohit: please keep those optional. Making these non-optional significantly increases compilation times.

StefanPalashev:
> I still need to go over most of the PR, but moving serde to a non-optional dependency is a no-go. This will have to be made into an optional feature - preferably, the same feature as the JSON module, since I'd like to cut down on the number of features in the crate. I suggest targeting the 1.0.x branch, and renaming the JSON feature to account for all extra functionality.

Thank you for the feedback!

Would it make sense to keep the JSON feature separate so it can still be enabled on its own, without pulling in the other extras? I’m happy to consolidate if that’s the preferred direction — just wanted to check in case keeping them separate would offer more flexibility for users.

The target branch will be updated once the other proposed changes are implemented.

nihohit:
It makes sense to introduce this as a separate feature, but this crate uses flag-frenzy to check for potentially problematic feature combinations, which means that introducing new features makes the CI run slower, so I try to avoid that.
I assume you're planning to introduce more Redis 8 features - do they contain more dependencies? how many more of them will require serde or serde_json?

StefanPalashev:
> It makes sense to introduce this as a separate feature, but this crate uses flag-frenzy to check for potentially problematic feature combinations, which means that introducing new features makes the CI run slower, so I try to avoid that. I assume you're planning to introduce more Redis 8 features - do they contain more dependencies? how many more of them will require serde or serde_json?

The plan is to implement new Redis 8 features, but based on my understanding, none of the planned additions are expected to introduce dependencies on `serde` or `serde_json`.

StefanPalashev:
@nihohit The vector sets feature is now optional. However, if this isn’t quite the approach you had in mind, please feel free to let me know.

redis/src/types.rs
StefanPalashev, line 1: All of the newly introduced types (`VectorQuantization`, `EmbeddingInput<'a>`, `VectorAddInput<'a>`, `VectorSimilaritySearchInput<'a>`) and their implementations (`impl<'a> ToRedisArgs`) need to be annotated with `#cfg(feature = "vector-sets")]`.

redis/src/lib.rs
StefanPalashev, line 1: All of the new vector sets related types should be conditionally exported.

```
#[cfg(feature = "vector-sets")]
#[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
pub use crate::types::{
    EmbeddingInput, VectorAddInput, VectorQuantization, VectorSimilaritySearchInput,
};
```

.github/workflows/rust.yml
StefanPalashev: Thank you for pointing it out. The change has been applied.

redis/Cargo.toml
```diff
@@ -91,7 +91,7 @@ rustls-native-certs = { version = "0.8", optional = true }
 tokio-rustls = { version = "0.26", optional = true, default-features = false }
 futures-rustls = { version = "0.26", optional = true, default-features = false }
 
-# Only needed for RedisJSON Support
+# Needed for RedisJSON Support
```
nihohit: Needed for modules support

redis/Cargo.toml
```diff
@@ -118,6 +118,7 @@ default = ["acl", "streams", "geospatial", "script", "keep-alive", "num-bigint"]
 acl = []
 geospatial = []
 json = ["dep:serde", "serde/derive", "dep:serde_json"]
+vector-sets = ["json"]
```
nihohit: why not `"dep:serde", "serde/derive", "dep:serde_json"`?

redis/src/commands/mod.rs
```diff
@@ -1257,6 +1263,157 @@ implement_commands! {
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vadd<K: ToRedisArgs, E: ToRedisArgs>(key: K, input: VectorAddInput<'a>, element: E) -> (bool) {
+        cmd("VADD").arg(key).arg(input).arg(element)
+    }
+
+    /// Add a new element into the vector set specified by key with optional parameters for fine-tuning the insertion process.
+    /// [Redis Docs](https://redis.io/commands/VADD)
+    #[cfg(feature = "vector-sets")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vadd_extended<K: ToRedisArgs, E: ToRedisArgs>(key: K, input: VectorAddInput<'a>, element: E, options: &'a VAddOptions) -> (bool) {
```
nihohit: please be consistent with the naming scheme - we use `_options`, not `_extended`.

redis/src/commands/mod.rs
```diff
@@ -1257,6 +1263,157 @@ implement_commands! {
...
+        cmd("VEMB").arg(key).arg(element)
+    }
+
+    /// Return the raw internal representation of the approximate vector associated with a given element in the vector set.
+    /// Vector sets normalize and may quantize vectors on insertion.
+    /// VEMB reverses this process to approximate the original vector by de-normalizing and de-quantizing it.
+    /// [Redis Docs](https://redis.io/commands/VEMB)
+    #[cfg(feature = "vector-sets")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vemb_raw<K: ToRedisArgs, E: ToRedisArgs>(key: K, element: E) -> Generic {
```
nihohit: it's preferable to use an option type here, in case that more values will be added to the command in the future.

redis/src/commands/mod.rs
```diff
@@ -1257,6 +1263,157 @@ implement_commands! {
...
+    }
+
+    /// Associate a JSON object with an element in a vector set.
+    /// Use this command to store attributes that can be used in filtered similarity searches with VSIM.
+    /// [Redis Docs](https://redis.io/commands/VSETATTR)
+    #[cfg(feature = "vector-sets")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vsetattr<K: ToRedisArgs, E: ToRedisArgs, J: Serialize>(key: K, element: E, json_object: &'a J) -> (bool) {
+        let attributes_json = match serde_json::to_value(json_object) {
+            Ok(serde_json::Value::String(s)) if s.is_empty() => "".to_string(),
```
nihohit, line 1333: what happens on the `Err` case? why not fail then?
Also, why bother with the `if s.is_empty()` check? what happens if it is removed?

redis/src/commands/mod.rs
```diff
@@ -1257,6 +1263,157 @@ implement_commands! {
...
+    fn vgetattr<K: ToRedisArgs, E: ToRedisArgs>(key: K, element: E) -> (Option<String>) {
+        cmd("VGETATTR").arg(key).arg(element)
+    }
+
+    /// Return metadata and internal details about a vector set, including
+    /// size, dimensions, quantization type, and graph structure.
+    /// [Redis Docs](https://redis.io/commands/VINFO)
+    #[cfg(feature = "vector-sets")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vinfo<K: ToRedisArgs>(key: K) -> Generic {
```
nihohit: doesn't this return a `HashMap<String, Value>`?

redis/src/commands/mod.rs
```diff
@@ -1257,6 +1263,157 @@ implement_commands! {
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vlinks_with_scores<K: ToRedisArgs, E: ToRedisArgs>(key: K, element: E) -> Generic {
+        cmd("VLINKS").arg(key).arg(element).arg("WITHSCORES")
+    }
+
+    /// Return one random elements from a vector set.
+    /// [Redis Docs](https://redis.io/commands/VRANDMEMBER)
+    #[cfg(feature = "vector-sets")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vrandmember<K: ToRedisArgs>(key: K) -> (Option<Vec<String>>) {
```
nihohit: Why return a vec? and why have both this and vrandmember_multiple?

redis/src/commands/mod.rs
```diff
@@ -1257,6 +1263,157 @@ implement_commands! {
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vsim<K: ToRedisArgs>(key: K, input: VectorSimilaritySearchInput<'a>) -> Generic {
+        cmd("VSIM").arg(key).arg(input)
+    }
+
+    /// Performs a vector similarity search with optional parameters for customization.
+    /// [Redis Docs](https://redis.io/commands/VSIM)
+    #[cfg(feature = "vector-sets")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+    fn vsim_extended<K: ToRedisArgs>(key: K, input: VectorSimilaritySearchInput<'a>, options: &'a VSimOptions) -> Generic {
```
nihohit: again, `_options`.

redis/src/types.rs
```diff
@@ -77,6 +77,21 @@ pub enum FieldExistenceCheck {
     FXX,
 }
 
+/// Quantization options for vector storage
+#[cfg(feature = "vector-sets")]
+#[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+#[derive(Clone, Copy)]
+pub enum VectorQuantization {
```
nihohit: why are these here and not in `commands/mod.rs`?

redis/src/types.rs
```diff
@@ -2909,6 +2924,126 @@ impl ToRedisArgs for ExpireOption {
...
 
+/// Input data formats that can be used to generate vector embeddings:
+///
+/// - 32-bit floats
+/// - 64-bit floats
+/// - Strings (e.g., numbers as strings)
+#[cfg(feature = "vector-sets")]
+#[cfg_attr(docsrs, doc(cfg(feature = "vector-sets")))]
+#[derive(Clone)]
+pub enum EmbeddingInput<'a> {
```
nihohit: same for all types below

nihohit (commented):
commented on the new functions. Still need to review types and tests.

redis/Cargo.toml
StefanPalashev: There is no need to duplicate the dependencies already included in the `json` feature, because `vector-sets` simply relies on them as of now. Furthermore, if the approach to JSON processing changes in the future, it's likely that `vector-sets` will need to adapt to it as well.

redis/src/commands/mod.rs
StefanPalashev: This aligns with the established convention of using a separate function definition to add a simple parameter such as `"RAW"`. The pattern can be observed in the names of functions that use `"WITHSCORES"` as a parameter. As of now this is the only variable to the VEMB command and this influenced the decision not to wrap it in an option type. With that being said, I don't mind changing it if you think it is necessary.

redis/Cargo.toml
nihohit: Yes, but this means that users who add the vector-sets feature also implicitly add the json feature, even if they're not interested in it. Sharing dependencies isn't a reason to force features on the users.

redis/src/commands/mod.rs
nihohit: please do.

redis/src/commands/mod.rs
StefanPalashev: Without this check empty strings will be serialized as `\"\"`, which is not the intended behavior. An empty string should remove the attributes. Without it an empty JSON object will be set as an attribute.

If the matching fails completely `unwrap()` will panic at runtime. The fail case when the passed string is not empty ("") simply tries to convert the passed in object to string.

redis/src/commands/mod.rs
StefanPalashev: Thank you for pointing this out.

redis/src/commands/mod.rs
StefanPalashev: It does. It will be wrapped with an Option for the cases when there is no such key.
This would also significantly shorten the code in one of the test cases.

redis/src/types.rs
StefanPalashev: Because they are `enums` and there were similar scenarios.
Should they be moved to `commands/mod.rs`?

StefanPalashev: Are types like `VAddOptions` and `VSimOptions` supposed to be in `commands/mod.rs`?
There should be a proper way to group things.

redis/src/commands/mod.rs
StefanPalashev: The reason for having two separate functions is to follow the same pattern as the standard set APIs (`srandmember` and `srandmember_multiple`).
The functions were both returning vectors as the documentation states that the response will be an `array reply`.

redis/src/commands/mod.rs
StefanPalashev: Done

nihohit:
please move all added types either to commands/mod.rs, or to a separate module - maybe commands/vector_sets.rs?

nihohit:
Thanks!

PR #1681, opened 2025-06-18
Fix connection leak when connection_manager is dropped

magec:
Whenever a `ConnectionManager` is dropped, its connection
is not finished until the runtime is shut down. This is because the
tokio::task responsible for listening to disconnection events from redis
keeps running (and holding a reference to it).

This change adds a cancellation token and uses it in the `drop()` implementation
(which is also added).


nihohit (approved):
Excellent fix, thanks!

magec:
Lol, tests don't seem to agree. Will check what's wrong.

nihohit:
instead of the added dependency, please just use `futures::select!`, which is already in use in the crate.

magec:
The problem is that futures is a dev-dependency. Will add it and use it instead of tokio

nihohit:
futures-util isn't. doesn't it contain this macro?

magec:
I had to use `fuse()` as read here https://users.rust-lang.org/t/why-doesnt-tokio-select-require-fusedfuture/46975/6

Not a `futures` expert, maybe I'm missing something? I just want to signal the task so it stops running.

redis/Cargo.toml
```diff
@@ -50,6 +50,7 @@ tokio = { version = "1", features = [
   "net",
   "time",
   "sync",
+  "macros",
```
nihohit: please remove

nihohit:
I'm fine with the added code. It isn't pretty, but it looks correct.

nihohit:
The test failure is because the cancellation token is triggered on `Drop`, but because `ConnectionManager` is cloneable, clones can be dropped while some are still active, and cancel the operation.
Cancellation should be triggered only when the last clone is dropped.

magec:
It does not have to do with the cancel logic. It's the `select!` If I refactor the code to be like :

```rust
    async fn check_for_disconnect_pushes(
        receiver: oneshot::Receiver<(
            ConnectionManager,
            UnboundedReceiver<PushInfo>,
            OptionalPushSender,
        )>,
        _cancel_token: CancellationToken,
    ) {
        let Ok((this, mut internal_receiver, external_sender)) = receiver.await else {
            return;
        };

        let recv_future = internal_receiver.recv().fuse();
        pin_mut!(recv_future);

        futures_util::select! {
            maybe_push_info = recv_future => {
                if let Some(push_info) = maybe_push_info {
                    if push_info.kind == PushKind::Disconnection {
                        this.reconnect(this.0.connection.load());
                    }

                    if let Some(sender) = external_sender.as_ref() {
                        let _ = sender.send(push_info);
                    }
                }
            },
        }
    }

```
and without the drop trait implemented, it also fails.

magec:
Weird it does not get triggered when I implement it in the Internals struct, which is not clonable. Whenever the last Arc is dropped, the Internals should be also dropped, but it does not seem to be the case, because the connection is not shut down.

magec:
Lol, now I get it, it does not get triggered because the task is holding the reference.

magec:
I don't like the implementation, tbh, because if it gets cloned at the same time a reference is dropped, then you end up with a valid manager with a stopped push task. That said, this event is very unlikely, because it only happens when you are dropping the manager itself.

Still, leaving the responsibility of the connection close to the runtime does not seem to be a good thing to do either.

magec:
This whole thing came up because we are trying to connect to a Valkey instance running in Google cloud using authentication. When you do that, google leverages IAM and opaque tokens which are refreshed every hour. 

I need to use multiplexed connections because I think is the most performant way of connecting, and also, would like to have the reconnection logic for obvious reasons. The way I implemented it was by using an `ArcSwap` holding a manager and changing it every hour. Without this fix, I saw that I was leaking connections.

I am saying this to make an example of a use case where you want to create and drop managers, because I understand that the typical way of using it would be haivng just one for the entire life of the process.

nihohit:
Ok, let's try to tackle this problem in another way - the problem is that `check_for_disconnect_pushes` holds a copy of `ConnectionManager`, which leaks the connection. Can we just solve the problem by sending a weak pointer instead of a strong pointer to the manager? Essentially we want a non-owning copy to live there, and that would solve the problem without cancellation tokens or `select`s, right?

nihohit:
Here, I think that will solve the issue - https://github.com/nihohit/redis-rs/commit/30d44e36b264005ea59ede848040acf2096aaa67

magec:
You were faster, I was now changing `reconnect` to accept Weak<Internals> :D

nihohit:
Feel free to pull the change - this is your investigation and you've done 90% of the work, you should have the credit on the PR.

magec:
Just commited it.

nihohit:
NNNNOOOOOOO!

nihohit:
line 509 should be 
`let connection_config = internals.connection_config.clone();`
🤦

magec:
Yes, changed it already

nihohit:
Thanks for your contribution!

PR #1682, opened 2025-06-19
Test Suite - Added 512MB GET SET Test

65001:
> By default, a single Redis string can be a maximum of 512 MB.

https://redis.io/docs/latest/develop/data-types/strings/

These tests ensure that both async and sync connections can support strings of the maximum size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ implements the `FromRedisValue` trait.
 ```rust
 use redis::TypedCommands;
 
-fn fetch_an_integer() -> redis::RedisResult<isize> {
+fn fetch_an_integer() -> Option<isize> {
 	// connect to redis
 	let client = redis::Client::open("redis://127.0.0.1/")?;
 	let mut con = client.get_connection()?;


nihohit:
Hi @65001 would you like to fix the tests?

65001:
Hey @nihohit ,

I'm looking at [build (1.75.0, redis, redis, 7.2.4) logs](https://github.com/redis-rs/redis-rs/actions/runs/15760912898/job/44427183993?pr=1682). I see:
```
PASS [  12.261s] redis::test_async basic_async::test_async_getset::case_1_tokio
PASS [  14.466s] redis::test_async basic_async::test_async_getset::case_2_async_std
PASS [  13.918s] redis::test_async basic_async::test_async_getset::case_3_smol
PASS [  14.095s] redis::test_basic basic::test_getset_large
 ```
I'm seeing it fail under `Testing Connection Type TCP with all features and Rustls support`. 

Is this just a case of the timeout not being large enough or the test runner potentially running out of memory? I can add a [Per-test settings](https://nexte.st/docs/configuration/per-test-overrides) to increase the timeout if you think it's time out related.

If it's memory related, I don't know how to fix it, other than suggest a [large test runner](https://docs.github.com/en/actions/concepts/runners/about-larger-runners).

The other solution might be to try and ensure the string generation is optimal, if the above two things aren't acceptable.

Would appreciate your help in figuring out what's acceptable for the project long term.

nihohit:
I don't know what caused it. I don't is the timeout, since I doubt that a
large message takes longer than setting up a clister with replicas. I guess
we should first try it with a larger runner, or just print the amount of
available memory.

nihohit:
@65001 if you're still interested in this, the issue seems to be only with the `smol` runtime when using TLS. If you run this only on tokio the test should pass.

65001:
@nihohit - Is an acceptable solution to just drop smol for this unit test? And add a link to this conversation documenting why?

If so I'll make the change. If it's more extensive than that, I'll have to pass till December or so, when classes for my masters + full time work gets less busy.

nihohit:
It is acceptable, I just hope that it works - there's nothing that ensures that the issue isn't also reproducible in Tokio.
Maybe the right solution is just to increase the connection and request timeouts in the connection configuration.

PR #1687, opened 2025-06-24
Add AsRef<Client> to connection manager

rollo-b2c2:
Just useful to know what client is being passed.

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -293,6 +293,12 @@ macro_rules! reconnect_if_io_error {
     };
 }
 
+impl AsRef<Client> for ConnectionManager {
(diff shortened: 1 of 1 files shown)


nihohit:
Can you please explain when this is useful? I prefer not to add to the public API unless there's clear value.

rollo-b2c2:
If you pass in a ConnectionManager anywhere, and you have an error, there's basically no way to determine what redis you've actually hit.

You can't even debug the ConnectionManager.

nihohit:
For that I'd prefer to implement fmt:: Debug instead of expanding a
functional api.

nihohit:
closed in favor of https://github.com/redis-rs/redis-rs/pull/1829

PR #1694, opened 2025-06-26
Propagate connection error when failing to connect to clustered redis

Nathan-Fenner:
When attempting to connect to a clustered Redis instance, if there is some kind of network misconfiguration, the only error message you get back is `It failed to check startup nodes.` which doesn't help to diagnose the connection problem.

This PR formats and combines the individual reasons for failing to connect to each initial node in the cluster, and concatenates them into the "details" of the error message.

### DIFF
--- a/redis/src/cluster.rs
+++ b/redis/src/cluster.rs
@@ -401,22 +401,54 @@ where
     /// `BrokenPipe` error.
     fn create_initial_connections(&self) -> RedisResult<()> {
         let mut connections = HashMap::with_capacity(self.initial_nodes.len());
+        let mut failed_connections = Vec::new();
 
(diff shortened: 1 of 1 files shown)


Nathan-Fenner:
I'm happy to fix the lints - but none of them refer to any of the code that I touched in `cluster.rs` - I think that the main branch has bad code on it somehow? Mostly concerned that I'm going to introduce conflicts from trying to fix them on this branch.

nihohit:
@Nathan-Fenner, you're right - looks like clippy introduced new lints recently.  I merged fixes, feel free to rebase over main.

nihohit:
Thanks for the contribution!

Issue #1693, opened 2025-06-26
Suggestion: Implement Into<String> for ValueType


skyleo:
Implementing `Into<String>` for ValueType would enable one to use it in scan_options for example:

```rust
con.scan_options::<String>(ScanOptions::default().with_type(ValueType::ZSet));
```

(I'd much prefer to use that instead of a magic string literal)

Although it'd be nicer to have this more type-safe at compile-time already. (Only accepting ValueType)

nihohit:
Sure, feel free to contribute either suggestion - the second will probably need to go to the 1.0.x branch, since it breaks API.

skyleo:
> Sure, feel free to contribute either suggestion - the second will probably need to go to the 1.0.x branch, since it breaks API.

Oki-dokey. I contributed the first suggestion. Second suggestion is a bit too spooky for me as an outsider.

skyleo:
Closing as the PR for this has been merged.

PR #1695, opened 2025-06-27
Implement conversions between ValueType and String and use it in its FromRedisValue implementation

skyleo:
This PR does:
- implement `From<String>` and `From<&String>` for ValueType
- implement `From<ValueType>` for String
- Replace code that would be duplicated due to introduction of `From<String>` and `From<&String>` with the use of `Ok(s.into())`

This allows one to pass a ValueType directly such as in:

```rust
con.scan_options::<String>(ScanOptions::default().with_type(ValueType::ZSet));
```

Related issue: https://github.com/redis-rs/redis-rs/issues/1693

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -2954,33 +2954,45 @@ pub enum ValueType {
     Unknown(String),
 }
 
+impl<T: AsRef<str>> From<T> for ValueType {
+    fn from(s: T) -> Self {
+        match s.as_ref() {
+            "string" => ValueType::String,
+            "list" => ValueType::List,
+            "set" => ValueType::Set,
+            "zset" => ValueType::ZSet,
+            "hash" => ValueType::Hash,
+            "stream" => ValueType::Stream,
+            s => ValueType::Unknown(s.to_string()),
+        }
+    }
+}
+
+impl From<ValueType> for String {
+    fn from(v: ValueType) -> Self {
+        match v {
+            ValueType::String => "string".to_string(),
+            ValueType::List => "list".to_string(),
+            ValueType::Set => "set".to_string(),
+            ValueType::ZSet => "zset".to_string(),
+            ValueType::Hash => "hash".to_string(),
+            ValueType::Stream => "stream".to_string(),
+            ValueType::Unknown(s) => s,
+        }
+    }
+}
+
 impl FromRedisValue for ValueType {
     fn from_redis_value(v: &Value) -> RedisResult<Self> {
         match v {
-            Value::SimpleString(s) => match s.as_str() {
-                "string" => Ok(ValueType::String),
-                "list" => Ok(ValueType::List),
-                "set" => Ok(ValueType::Set),
-                "zset" => Ok(ValueType::ZSet),
-                "hash" => Ok(ValueType::Hash),
-                "stream" => Ok(ValueType::Stream),
-                _ => Ok(ValueType::Unknown(s.clone())),
-            },
+            Value::SimpleString(s) => Ok(s.into()),
             _ => invalid_type_error!(v, "Value type should be a simple string"),
         }
     }
 
     fn from_owned_redis_value(v: Value) -> RedisResult<Self> {
         match v {
-            Value::SimpleString(s) => match s.as_str() {
-                "string" => Ok(ValueType::String),
-                "list" => Ok(ValueType::List),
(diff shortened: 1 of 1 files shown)


skyleo:
After this force-push due to me having forgotten to implement `From<ValueType>` in my initial proposal,
this PR can be reviewed now.

Thank you ~

redis/src/types.rs
```diff
@@ -2954,33 +2954,59 @@ pub enum ValueType {
...
+            "set" => ValueType::Set,
+            "zset" => ValueType::ZSet,
+            "hash" => ValueType::Hash,
+            "stream" => ValueType::Stream,
+            _ => ValueType::Unknown(s),
+        }
+    }
+}
+
+impl From<&String> for ValueType {
```
nihohit: from `&str` will be broader, and I think that the compiler knows to replace &String with &str.

redis/src/types.rs
```diff
@@ -2954,33 +2954,59 @@ pub enum ValueType {
...
+            "set" => ValueType::Set,
+            "zset" => ValueType::ZSet,
+            "hash" => ValueType::Hash,
+            "stream" => ValueType::Stream,
+            _ => ValueType::Unknown(s.clone()),
+        }
+    }
+}
+
+impl From<ValueType> for String {
```
nihohit, line 2971: replace `String` with `&'static str` - it doesn't need to assume allocations that way.

nihohit:
also, please rebase over the latest main, for lint fixes.

redis/src/types.rs
skyleo: This isn't trivial to do, due to ValueType::Unknown holding a String which it owns, thus I can't return the slice of it in this trait's function.

Suggestions or just keep as is?

redis/src/types.rs
skyleo: Unfortunately type coercion will not work that well when used on stuff like `From<&str>`, `Option<&str>` etc.

So this:
```rust
    fn from_redis_value(v: &Value) -> RedisResult<Self> {
        match v {
            Value::SimpleString(s) => Ok(s.into()),
            _ => invalid_type_error!(v, "Value type should be a simple string"),
        }
    }
```

Will not work (the `s.into()` part of it).

One alternative is to do it as such:
```rust
impl<T: AsRef<str>> From<T> for ValueType {
    fn from(s: T) -> Self {
        match s.as_ref() {
            "string" => ValueType::String,
            "list" => ValueType::List,
            "set" => ValueType::Set,
            "zset" => ValueType::ZSet,
            "hash" => ValueType::Hash,
            "stream" => ValueType::Stream,
            s => ValueType::Unknown(s.to_string()),
        }
    }
}
```

And in that case the type coercion will work just fine in the first example I gave.
But we have a collision with
```rust
impl From<String> for ValueType {
    fn from(s: String) -> Self {
        match s.as_str() {
            "string" => ValueType::String,
            "list" => ValueType::List,
            "set" => ValueType::Set,
            "zset" => ValueType::ZSet,
            "hash" => ValueType::Hash,
            "stream" => ValueType::Stream,
            _ => ValueType::Unknown(s),
        }
    }
}
```

which was introduced in this PR. Of course this can just be removed and the `T: AsRef<Str>` approach then just works fine.
But there may be an allocation instead of a move even when one passes an owned String.

Are you fine with that trade-off?

I mean tbf ValueType::Unknown shouldn't really happen, and when it does happen, someone's code is very wrong, so I personally think this trade-off is fine.

Inspiration taken from:
- https://stackoverflow.com/questions/34974732/why-does-optionstring-not-coerce-to-optionstr
- https://users.rust-lang.org/t/converting-between-str-and-string/53595/1

skyleo:
Pushed the fixup commit for the first suggested change with my proposal for now.

Waiting on feedback in your both suggestions that I replied.

Thank you.

redis/src/types.rs
nihohit: keep as-is. if someone will really want it, we can implement `TryFrom<&ValueType> for &str` later.

skyleo: Okay, will rebase and squash the fixup into the only commit this PR will have.

skyleo:
All feedback as discussed has been applied and the fixup(s) have been squashed / rebased into the only commit this PR will have now.

nihohit:
Thanks for the contribution!

PR #1697, opened 2025-06-29
Ensure Sentinel#master_for propagates connection errors

### LINKED ISSUE #1240: Master with given name not found in sentinel- MasterNameNotFoundBySentinel
Dears,

I'm trying to use redis-rs to connect Redis Sentinel.
But response with below error, I can connect using the same connection information with Java library.
Please help on this if I'm using the redis-rs in a wrong way.

**Error Message:**
called `Result::unwrap()` on an `Err` value: Master with given name not found in sentinel- MasterNameNotFoundBySentinel

**My Code:**

```
let nodes = vec!["redis://:Avaya123@10.133.105.69:32002/","redis://:Avaya123@10.133.105.69:32002/","redis://:Avaya123@10.133.105.69:32002/"];
    let mut master = SentinelClient::build(nodes.clone(), String::from("redis1"),None, SentinelServerType::Master).unwrap();
    let connection_options = AsyncConnectionConfig::new().with_response_timeout(std::time::Duration::from_secs(60))
        .with_connection_timeout(std::time::Duration::from_secs(60));
    master.get_async_connection_with_config(&connection_options).await.unwrap(); -- Here is the line which pop up the error.
```

Thanks & Regards

bittrance:
`master_for` will first ask sentinel for the master and then connect to that node and verify that it considers itself master. This is likely to be the first connection to a node, so any misconfiguration or permission error will surface here. Those errors are currently eaten and will just result in `MasterNameNotFoundBySentinel` which is quite confusing if the actual errors is that the password used to connect is wrong. This PR makes sure to propagate those errors to the library user.

The PR further takes care to surface the error from the `ROLE` command unless there is reason to assume it was met with "unknown command". There is only a small number of patch revisions between the introduction of Sentinel v2 and the `ROLE` command, so this fallback is quite esoteric. Surfacing the first errors avoids confusing the caller with messages about `NOPERM` on `INFO` when in fact the developer should focus on why we could not call `ROLE`.

This PR does not address a similar problem with finding replicas ("slaves") as there seems to be variations between blocking and async versions of the code that I do not understand. If we want to standardize the behavior there, it should probably be a separate PR?

Fixes #1240.

Ping @65001 .


nihohit:
Thanks for the contribution! I think that `async_determine_slave_from_role_or_info_replication` also needs this fix.

redis/src/sentinel.rs
```diff
@@ -302,19 +302,24 @@ fn determine_master_from_role_or_info_replication(
...
 
-    //TODO: Maybe there should be some kind of error message if both role checks fail due to ACL permissions?
-    Ok(false)
+    let role_err = role_result.unwrap_err();
+    // unknown commands are expressed as response errors
+    if !matches!(role_err.kind(), ErrorKind::ResponseError) {
+        return Err(role_err);
+    }
+
+    Err(fallback_role_result.unwrap_err())
```
nihohit: the end of the function is repeated twice (or three times, with `async_determine_slave_from_role_or_info_replication`). Can you please move it to a function, for DRY reasons?

bittrance: Sure, but why is there no blocking version of that method?

bittrance: Fixed.

nihohit: > Sure, but why is there no blocking version of that method?

I assume because nobody contributed it 🤷

PR #1706, opened 2025-07-01
Errors2

nihohit:

PR #1718, opened 2025-07-10
Improve pipeline errors

nihohit:
This change is rebased over https://github.com/redis-rs/redis-rs/pull/1706.
This change adds an error variant for pipeline errors, that lists a all the failed commands - command index and error.
For users that want the full response, they can replace the query type with `Vec<Value>` or `Vec<RedisResult<T>>`, in order to get the responses for both the successful and failed commands.

fixes https://github.com/redis-rs/redis-rs/issues/746

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -285,24 +285,6 @@
 //! # Ok(()) }
 //! ```
 //!
-//! If the user wants to handle server errors while also receiving values from successful requests,
-//! the user can define the target type of a pipeline query as `Vec<RedisResult<T>>` or `Vec<Value>`. This vector
-//! will contain only non-ignored values, if all requests completed successfully. If some failed, this
-//! vector will contain all server responses, including the ones that were ignored.
-//!
-//! ```rust,no_run
-//! # fn do_something() -> redis::RedisResult<()> {
-//! # let client = redis::Client::open("redis://127.0.0.1/").unwrap();
-//! # let mut con = client.get_connection().unwrap();
-//! let results : Vec<redis::RedisResult<String>> = redis::pipe() //results: Vec<redis::Value> would also work
-//!         .set("x", "").ignore()
-//!         .hset("x", "field", "field_value").ignore()
-//!         .get("x")
-//!         .query(&mut con)?; // This will result with a vec![Ok("OK"), Err(WRONGTYPE), OK("x-value")]
-//!
-//! # Ok(()) }
-//! ```
-//!
 //! If you want the pipeline to be wrapped in a `MULTI`/`EXEC` block you can
 //! easily do that by switching the pipeline into `atomic` mode.  From the
 //! caller's point of view nothing changes, the pipeline itself will take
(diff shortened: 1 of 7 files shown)


redis/src/lib.rs
```diff
@@ -285,6 +285,24 @@
...
+//!
+//! ```rust,no_run
+//! # fn do_something() -> redis::RedisResult<()> {
+//! # let client = redis::Client::open("redis://127.0.0.1/").unwrap();
+//! # let mut con = client.get_connection().unwrap();
+//! let results : Vec<RedisResult<String>> = redis::pipe() //results: Vec<redis::Value> would also work
+//!         .set("x", "").ignore()
+//!         .hset("x", "field", "field_value").ignore()
+//!         .get("x")
+//!         .query(&mut con)?; // This will result with a vec![Ok("OK"), Err(WRONGTYPE), OK("x-value")]
```
altanozlu: For me this one feels wrong, probably you thinked about this but, before `Vec<RedisResult<T>>` you could get an error for failure of the whole pipe, now you will get everything even successful command results you have ignored before. 
Can you explain more about the approach you have took ?

altanozlu (commented):
Hats off for `T::can_collect_errors()`. 
I have a disagreement with the approach for ignore

redis/src/lib.rs
nihohit: It's a fair question. I think that some users might want both the successes and the failures when they happen, and I didn't want to add another function to opt into this behavior, I thought that using a type that represents an error is implicitly a way of showing you're interested in this.
Now, if you're interested in errors, we can't skip an error that would've previously failed the request - that would hide the error from the user. So we can't skip ignored failed responses.

If you think this behavior is too confusing, we can just remove it.

As a side note, I thought [you were in favor of this](https://github.com/redis-rs/redis-rs/issues/1043#issuecomment-3053383125) :)

altanozlu: > As a side note, I thought https://github.com/redis-rs/redis-rs/issues/1043#issuecomment-3053383125 :)

I favor `Vec<RedisResult<T>>` but having different Vec size on ignore mixed pipeline if it's `RedisResult<T>` or just `T` doesn't look good.


> I thought that using a type that represents an error is implicitly a way of showing you're interested in this.

Yes but user is also explicitly saying ignore this command, maybe having `ignore` and `Vec<RedisResult<T>>` options together is deceiving and shouldn't be allowed. Because if user processes pipe's results, they need to compare size of Vec then act upon it.

nihohit: Fair enough. I'll revert this change. Thanks for correcting me :)

nihohit: done

altanozlu (commented):
when it's a type error it will only return first error like below, I would assume it will give three error two for x being string and one for nil value.
```
Pipeline failure: [(Index 1, error: \"WRONGTYPE\": Operation against a key holding the wrong kind of value)
```

example code:
```
        pipe.set("x", "x-value").ignore().get("x").get("x").get("d");
        let res: RedisResult<Vec<i32>> = pipe.query(&mut con);
```

note: multiple errors works fine when it's a served sided error.

nihohit:
Sorry, I didn't understand the comment. Why would this command trigger multiple errors? Is it because you expect the pipeline to try and deserialize the received non-error values, even if it encountered server errors?

altanozlu:
> Sorry, I didn't understand the comment. Why would this command trigger multiple errors? Is it because you expect the pipeline to try and deserialize the received non-error values, even if it encountered server errors?

I expected to receive all errors in the pipeline failure. For example, when there is a server-side error due to HSET to a scalar value, multiple errors can be listed at once. However, when there is a client-side error because `T` cannot be generated from the `Value`, only one error will be listed in the pipeline failure, even though there could be multiple.

It's not a huge deal but wanted to point out.

nihohit:
It's true - essentially we're passing only client side or server side errors, not both.
I guess I can change that, but I think that might be make the code rather complex. One thing that I do need is to ensure that if there are multiple parsing issues, they're all passed to the user.

nihohit:
> One thing that I do need is to ensure that if there are multiple parsing issues, they're all passed to the user.

I don't think this can be done in a simple manner either :(
Since the parsing is recursive, this means that we need some mechanism to propagate internal errors and their indices to the external parsing error.

altanozlu (approved):
It can get merged as is, maybe you can also add an example about this ?

nihohit:
I don't think an example is needed, since the final change was just making a more explicit error message.

PR #1728, opened 2025-07-15
Made `get_sentinel_replicas` & `async_get_sentinel_replicas` pub

siddharthteli12:


nihohit:
Thanks!
It might be unclear to the user what does the `Vec<HashMap<String, String>>` return type actually mean. Looks like the functions should return host name + port, or maybe `ConnectionAddr`?

siddharthteli12:
Lol, I didn’t think this through. I think we should create a new method — maybe `replicas_for` — that returns a `RedisResult<Vec<Client>>`, to be consistent with `master_for`?

siddharthteli12:
@nihohit , Can you review it now

nihohit:
`Client` isn't a connection, it's just a baseline object that is used to create connections, so the function names & docs need to be corrected.

redis/tests/test_sentinel.rs
```diff
@@ -103,6 +103,82 @@ fn assert_connect_to_known_replicas(
...
 
+#[test]
+fn test_get_all_replicas_success() {
+    let number_of_replicas = 3;
+    let master_name = "master1";
+    let mut context = TestSentinelContext::new(2, number_of_replicas, 3);
+    let sentinel = context.sentinel_mut();
+    let replicas = sentinel.get_all_replicas(master_name, None).unwrap();
+
+    assert!(replicas.len() == number_of_replicas as usize);
```
nihohit: use assert_eq instead of assert

redis/tests/test_sentinel.rs
```diff
@@ -103,6 +103,82 @@ fn assert_connect_to_known_replicas(
...
+}
+
+#[test]
+fn test_get_all_replicas_invalid_master_name() {
+    let mut cluster = TestSentinelContext::new(2, 2, 3);
+    let sentinel = cluster.sentinel_mut();
+
+    let result = sentinel.get_all_replicas("invalid_master_name", None);
+    assert!(result.is_err());
```
nihohit: use assert_eq and assert on the error kind

redis/tests/test_sentinel.rs
```diff
@@ -103,6 +103,82 @@ fn assert_connect_to_known_replicas(
...
+        .arg("set")
+        .arg(master_name)
+        .arg("down-after-milliseconds")
+        .arg("1000")
+        .query::<()>(&mut conn)
+        .unwrap();
+
+    context.cluster.servers[1].stop();
+
+    std::thread::sleep(std::time::Duration::from_secs(3));
```
nihohit: please reduce the time, I prefer to avoid slow tests when possible

redis/tests/test_sentinel.rs
```diff
@@ -1062,6 +1138,123 @@ pub mod async_tests {
...
+                redis::cmd("SENTINEL")
+                    .arg("set")
+                    .arg(master_name)
+                    .arg("down-after-milliseconds")
+                    .arg("1000")
+                    .query_async::<()>(&mut conn)
+                    .await?;
+
+                context.cluster.servers[1].stop();
+                std::thread::sleep(std::time::Duration::from_secs(4));
```
nihohit: please reduce the wait time

redis/tests/test_sentinel.rs
siddharthteli12: Reduced time to 800ms is that fine?

siddharthteli12:
> `Client` isn't a connection, it's just a baseline object that is used to create connections, so the function names & docs need to be corrected.

Makes sense. Updated the method name to `get_replica_clients` & doc comment too. Let me know if it's fine?

redis/src/sentinel.rs
```diff
@@ -261,6 +261,22 @@ fn random_replica_index(max: NonZeroUsize) -> usize {
     rand::rng().random_range(0..max.into())
 }
 
+fn try_connect_to_all_replica(
```
nihohit: still not a connection :)

nihohit (approved):

redis/tests/test_sentinel.rs
nihohit: if it works in the 10s of ms it would be best.

redis/src/sentinel.rs
siddharthteli12: Changed this to `build_replica_clients`

siddharthteli12:
Test cases are failing with `NoValidReplicasFoundBySentinel` probably getting empty `addresses` for `build_replica_clients` method. Will look into it.

siddharthteli12:
Fixed test cases & also checked clippy logs. Should be fine now.

nihohit:
Thanks for the contribution!

PR #1731, opened 2025-07-16
Make `Script`'s `load_cmd` public

somechris:
Having the `load_cmd` public allows to re-use it when loading scripts in pipelines.

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -2700,11 +2700,12 @@ implement_commands! {
 
     // script commands
 
-    /// Adds a prepared script command to the pipeline.
+    /// Invoke a prepared script.
     ///
-    /// Note: unlike a call to [`invoke`](crate::ScriptInvocation::invoke), if the script isn't loaded during the pipeline operation,
-    /// it will not automatically be loaded and retried. The script can be loaded using the
-    /// [`load`](crate::ScriptInvocation::load) operation.
+    /// Note: Unlike[`ScriptInvocation::invoke`](crate::ScriptInvocation::invoke), this function
+    /// does _not_ automatically load the script. If the invoked script did not get loaded beforehand, you
+    /// need to manually load it (e.g.: using [`ScriptInvocation::load`](crate::ScriptInvocation::load))
+    /// or this command will fail.
     #[cfg_attr(feature = "script", doc = r##"
 
 # Examples:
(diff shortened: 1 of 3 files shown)


nihohit:
please add a test, and an example to the documentation demonstrating how to use this function.

somechris:
Sorry for the delay.

I've switched to a different approach now. because the motivation for making `load_cmd` public was to be able to load scripts in a `Pipeline`.

But since `Pipeline` already has `invoke_script`, it might be generally more useful to have a `load_script` function for `Pipeline`s instead of making `load_cmd` public and having to do things manually.


> please add a test, and an example to the documentation demonstrating how to use this function.

Done for the public function. (The other code only got moved around.)



(The [failing linter job](https://github.com/redis-rs/redis-rs/actions/runs/16804997078/job/47595063961?pr=1731) seems unrelated. See #1748)

redis/src/commands/mod.rs
```diff
@@ -2700,11 +2700,44 @@ implement_commands! {
...
+    fn load_script<>(script: &'a crate::Script) -> Generic {
+        &mut script.load_cmd()
+    }
+
     /// Adds a prepared script command to the pipeline.
     ///
     /// Note: unlike a call to [`invoke`](crate::ScriptInvocation::invoke), if the script isn't loaded during the pipeline operation,
     /// it will not automatically be loaded and retried. The script can be loaded using the
-    /// [`load`](crate::ScriptInvocation::load) operation.
+    /// [`load_script`](Self::load_script) operation on the pipeline or out of band using the
```
nihohit: the documentation shouldn't reference a pipeline, since these commands are generated for multiple types, not only pipelines.

redis/src/commands/mod.rs
```diff
@@ -2700,11 +2700,44 @@ implement_commands! {
 
     // script commands
 
+    /// Adds a command to the pipeline to load a script.
```
nihohit: same - please don't reference a pipeline.

redis/src/commands/mod.rs
somechris: True. Thanks. Done.

I've also adapted the (since 2024) existing `script_invoke` reference to pipeline accordingly.

redis/src/commands/mod.rs
somechris: Done :+1:

Issue #1730, opened 2025-07-16
the trait `FromRedisValue` is not implemented for `!`


GustavAlbrecht:
Hi i'm using a `redis::aio::MultiplexedConnection` and `redis::AsyncCommands` like this `redis.set("key", &value).await?; `.

I'm getting the following error: error[E0277]: the trait bound `!: FromRedisValue` is not satisfied.

I can fix this by declaring the return type as `()` via turbofish syntax  however it's inconvenient and seems unnecessary.

Am i doing something wrong or any fix in sight?

rustc 1.88.0 (6b00bc388 2025-06-23)
redis = { version = "0.32.4", features = ["tokio-native-tls-comp"] }

nihohit:
Import AsyncTypedCommands instead of AsyncCommands. That trait has concrete
typed outputs, so the compiler doesn't need to attempt to derive the output
type.

PR #1732, opened 2025-07-17
Enhance visibility of typed traits in docs.

nihohit:
We want more users to be aware of the typed traits, in order to save issues like https://github.com/redis-rs/redis-rs/issues/1730.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,24 @@ Documentation on the library can be found at
 To open a connection you need to create a client and then to fetch a
 connection from it.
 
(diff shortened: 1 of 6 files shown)


nihohit:
@altanozlu WDYT?

altanozlu:
It's good, it can be also be enhanced more by having more examples using TypeCommands Instead of Commands, and having one or two example with `Commands` for sake of visibility.

nihohit:
Thanks. I'll merge this, and look into more examples later.

PR #1743, opened 2025-08-01
Support Redis 8.2 stream command extensions and new deletion commands

StefanPalashev:
# Add support for new stream commands and extensions introduced in Redis 8.2

## Overview
This Pull Request introduces new Redis Streams commands (`XDELEX` and `XACKDEL`) that provide enhanced control over message deletion and acknowledgment operations, specifically targeting **Redis 8.2+** features. The implementation adds support for fine-grained deletion policies that control how stream entries are handled with respect to consumer group Pending Entries Lists (PELs). It also extends the `XADD` and `XTRIM` commands with the same capabilities.

#### Syntax
`XDELEX key [KEEPREF | DELREF | ACKED] IDS numids id [id ...]`

`XDELEX` accepts one of the following options: `KEEPREF` (default), `DELREF` or `ACKED`.
- Using the `KEEPREF` deletion policy deletes entries from the stream, but preserves existing references to those entries in all consumer groups' PELs. This behavior is similar to that of `XDEL`.
- Using the `XDELEX` deletion policy deletes entries from the stream and also removes all references to those entries from all consumer groups' PELs, effectively cleaning up all traces of the entries.
- Using the `ACKED` deletion policy deletes only those entries that have been read and acknowledged by **all** consumer groups. If an entry has not yet been read or acknowledged by some group, it will not be deleted.

### `XACKDEL` - combines the functionality of `XACK` and `XDEL` in Redis Streams. 
It acknowledges the specified entry IDs in the given consumer group and conditionally deletes the corresponding entries from the stream.

#### Syntax
`XACKDEL key group [KEEPREF | DELREF | ACKED] IDS numids id [id ...]`

`XACKDEL` accepts one of the following options: `KEEPREF` (default), `DELREF` or `ACKED`.
- `KEEPREF` acknowledges the entries in the specified consumer group and deletes them from the stream, but preserves any existing references to those entries in all consumer groups' PELs.
- `DELREF` acknowledges the entries in the specified consumer group, deletes them from the stream, and also removes all references to those entries from all consumer groups' PELs - effectively cleaning up all traces. If an entry ID is no longer in the stream but still has dangling references, `XDELEX` with `DELREF` would still remove those references.
- `ACKED` acknowledges the entries in the specified consumer group and deletes only those entries that have been read and acknowledged by **all** consumer groups. If an entry has not yet been read or acknowledged by any consumer group, it will not be deleted.

## Altered commands
Since `XADD` and `XTRIM` can delete entries, they were both extended with the `[KEEPREF | DELREF | ACKED]` deletion policies as well.
`KEEPREF` - 
Removes entries from the stream according to the specified strategy (`MAXLEN` or `MINID`), regardless of whether they are still referenced by any consumer groups. However, it **preserves** existing references to those entries in all consumer groups’ PELs.
This is the **default behavior** of these commands, so it introduces **no breaking change**.
`DELREF` –
Removes entries from the stream based on the specified strategy (`MAXLEN` or `MINID`), and also removes **all** references to those entries from all consumer groups’ PELs — effectively cleaning up all traces of the entries.
`ACKED` –
Only removes entries that have been **read and acknowledged by all consumer groups**. If an entry has not been read or acknowledged by some group, it will not be deleted. When `MAXLEN` is specified, **no more than** `MAXLEN` entries will be deleted.

#### Syntax
`XTRIM key <MAXLEN | MINID> [= | ~] threshold [LIMIT count] [KEEPREF | DELREF | ACKED]`

#### Syntax
`XADD key [NOMKSTREAM] [<MAXLEN | MINID> [= | ~] threshold [LIMIT count]] [KEEPREF | DELREF | ACKED] <* | id> field value [field value ...]`

---

## 🧪 Tests

- Each of the deletion policies in the `XADD` and `XTRIM` commands has its own individual test.
- The `XDELEX` and `XACKDEL` commands each have separate tests that showcase and verify various use cases.

---

### 📚 References

- [Pull Request](https://github.com/redis/redis/pull/14130) that introduces the commands in Redis

### DIFF
--- a/redis/tests/support/mod.rs
+++ b/redis/tests/support/mod.rs
@@ -442,8 +442,6 @@ pub fn is_version(expected_major_minor: (u16, u16), version: Version) -> bool {
 }
 
 // Redis version constants for version-gated tests
-pub const REDIS_VERSION_CE_8_0_RC1: Version = (7, 9, 240);
-#[cfg(feature = "vector-sets")]
 pub const REDIS_VERSION_CE_8_0: Version = (8, 0, 0);
 pub const REDIS_VERSION_CE_8_2: Version = (8, 1, 240);
 
--- a/redis/tests/support/mod.rs
+++ b/redis/tests/support/mod.rs
@@ -441,6 +441,30 @@ pub fn is_version(expected_major_minor: (u16, u16), version: Version) -> bool {
         || (expected_major_minor.0 == version.0 && expected_major_minor.1 <= version.1)
 }
 
+// Redis version constants for version-gated tests
+pub const REDIS_VERSION_CE_8_0_RC1: Version = (7, 9, 240);
+#[cfg(feature = "vector-sets")]
+pub const REDIS_VERSION_CE_8_0: Version = (8, 0, 0);
+pub const REDIS_VERSION_CE_8_2: Version = (8, 1, 240);
+
+/// Macro to run tests only if the Redis version meets the minimum requirement.
+/// If the version is insufficient, the test is skipped with a message.
+#[macro_export]
+macro_rules! run_test_if_version_supported {
+    ($minimum_required_version:expr) => {{
+        let ctx = $crate::support::TestContext::new();
+        let redis_version = ctx.get_version();
+
+        if redis_version < *$minimum_required_version {
+            eprintln!("Skipping the test because the current version of Redis {:?} doesn't match the minimum required version {:?}.",
+            redis_version, $minimum_required_version);
+            return;
+        }
+
+        ctx
+    }};
+}
+
 #[cfg(feature = "tls-rustls")]
 fn load_certs_from_file(tls_file_paths: &TlsFilePaths) -> TlsCertificates {
     let ca_file = File::open(&tls_file_paths.ca_crt).expect("Cannot open CA cert file");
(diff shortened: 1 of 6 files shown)


redis/tests/support/mod.rs
```diff
@@ -440,6 +440,30 @@ pub fn is_version(expected_major_minor: (u16, u16), version: Version) -> bool {
         || (expected_major_minor.0 == version.0 && expected_major_minor.1 <= version.1)
 }
 
+// Redis version constants for version-gated tests
+pub const REDIS_VERSION_CE_8_0_RC1: Version = (7, 9, 240);
```
nihohit: let's remove this and use `REDIS_VERSION_CE_8_0` where this is used, since the RC isn't used in CI.

StefanPalashev: Thank you for the review!
Does this have to be done in the current Pull Request or should a new one be opened? This serves as a minimal server version that guarantees support for some of the hash field expiration commands, which are irrelevant to the current Pull Request.

nihohit: does this break any behavior and requires more than just the deletion? if not, in this PR to save the effort.

nihohit:
Thanks for the contribution :)

PR #1748, opened 2025-08-07
test: Fix clippy warnings about `ref instead of clone` in stream tests

somechris:
Clippy complained that `std::slice::from_ref` should get used instead of cloning when calling `xclaim_options`. So we switch.

An example of a failed CI run is at:

  https://github.com/redis-rs/redis-rs/actions/runs/16804997078/job/47595063961?pr=1731


nihohit (approved):
nice, thanks for the contribution!

PR #1750, opened 2025-08-07
Clarify difference between RESP2 and RESP3 pubsub.

nihohit:
Clarify the docs on the `PubSub` structs that they're meant to be used with RESP2, not RESP3, and made the top-level docs clearer on how to use RESP3 pubsub, also by adding sync connection pubsub methods.


nihohit:
@altanozlu

redis/tests/test_basic.rs
```diff
@@ -3688,23 +3688,19 @@ mod basic {
...
-            // `set_no_response` is used because in RESP3
-            // SUBSCRIPE/PSUBSCRIBE and UNSUBSCRIBE/PUNSUBSCRIBE commands doesn't return any reply only push messages
-            redis::cmd("SUBSCRIBE")
-                .arg("foo")
-                .set_no_response(true)
-                .exec(&mut pubsub_con)
-                .unwrap();
-        }
+        // `set_no_response` is used because in RESP3
+        // SUBSCRIPE/PSUBSCRIBE and UNSUBSCRIBE/PUNSUBSCRIBE commands doesn't return any reply only push messages
```
altanozlu: no need for these comments

redis/src/connection.rs
```diff
@@ -1695,6 +1697,50 @@ impl Connection {
         }
         result
     }
+
+    /// Subscribes to a new channel(s).
+    ///
+    /// This only works if the connection was configured with [ProtocolVersion::RESP3] and [Self::set_push_sender].
+    pub fn subscribe_resp3<T: ToRedisArgs>(&mut self, channel: T) -> RedisResult<()> {
```
altanozlu, line 1704: Why not just `subscribe` ? Let's push users for RESP3!

redis/src/lib.rs
```diff
@@ -245,6 +245,55 @@
...
+//! # #[cfg(feature = "aio")]
+//! # {
+//! # use futures::prelude::*;
+//! # use redis::AsyncCommands;
+//!
+//! # async fn func() -> redis::RedisResult<()> {
+//! let client = redis::Client::open("redis://127.0.0.1/?protocol=resp3").unwrap();
+//! let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+//! let config = redis::AsyncConnectionConfig::new().set_push_sender(tx);
+//! let mut con = client.get_multiplexed_async_connection_with_config(&config).await?;
```
altanozlu, line 264: What do you think about making example with `ConnectionManager` + `resubscribe_automatically` ?

altanozlu (changes requested):
example `async-pub-sub` also should reflect this, having an example with also normal commands would be amazing.
Maybe something like below ?
```
use redis::{AsyncTypedCommands, Msg};

#[tokio::main]
    let client = redis::Client::open("redis://127.0.0.1/?resp3=true").unwrap();
    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
    let config = redis::aio::ConnectionManagerConfig::new().set_push_sender(tx);
    let mut cm = client.get_connection_manager_with_config(config).await?;

    cm.publish("wavephone", "banana").await?;

    let pubsub_msg: Msg = Msg::from_push_info(rx.recv().await.unwrap()).unwrap();
    let message: String = pubsub_msg.get_payload().unwrap();
    assert_eq!(&message, "banana");
    cm.set(format!("received-message:{message}"), 1).await?;

    Ok(())
}

```

Also I think There should be TryInto<Msg> for PushInfo for easier conversion.

redis/src/connection.rs
nihohit: Because of https://github.com/redis-rs/redis-rs/blob/1cca609df52516a06a539379a61af31330825876/redis/src/commands/mod.rs#L2833

which I wasn't aware of at all. The names are already taken, and I don't want to break existing users :(

nihohit:
> Also I think There should be TryInto for PushInfo for easier conversion.

Please open a PR :)

redis/src/lib.rs
nihohit: adjusted the examples you provided accordingly.

redis/tests/test_basic.rs
```diff
@@ -3733,14 +3736,45 @@ mod basic {
...
+        pubsub_con.unsubscribe_resp3("foo").unwrap();
+        pubsub_con.punsubscribe_resp3("bar*").unwrap();
+        pubsub_con.ping().unwrap();
+
+        assert_eq!(con.publish("foo", 23), Ok(0));
+        assert_eq!(con.publish("barvaz", 42), Ok(0));
+
+        // We have received verification from Redis that it's subscribed to channel.
```
altanozlu:
```suggestion
        // We have received verification from Redis that it's unsubscribed from the channel.
```

redis/tests/test_basic.rs
```diff
@@ -3688,23 +3688,18 @@ mod basic {
...
-        {
-            // `set_no_response` is used because in RESP3
-            // SUBSCRIPE/PSUBSCRIBE and UNSUBSCRIBE/PUNSUBSCRIBE commands doesn't return any reply only push messages
-            redis::cmd("SUBSCRIBE")
-                .arg("foo")
-                .set_no_response(true)
-                .exec(&mut pubsub_con)
-                .unwrap();
-        }
+        // `set_no_response` is used because in RESP3
```
altanozlu:
```suggestion
```

redis/tests/test_basic.rs
```diff
@@ -3733,14 +3736,45 @@ mod basic {
...
+                    Value::BulkString("bar*".as_bytes().to_vec()),
+                    Value::BulkString("barvaz".as_bytes().to_vec()),
                     Value::BulkString("23".as_bytes().to_vec())
                 ]
             ),
             (kind, data)
         );
+
+        // `set_no_response` is used because in RESP3
+        // SUBSCRIPE/PSUBSCRIBE and UNSUBSCRIBE/PUNSUBSCRIBE commands doesn't return any reply only push messages
```
altanozlu:
```suggestion
```

altanozlu (approved):
LGTM

nihohit:
Thanks for the help @altanozlu!

Issue #1798, opened 2025-09-26
XAUTOCLAIM error when PEL has invalid entries for Redis 6


BatmanAoD:
The behavior for `XAUTOCLAIM` [changed in Redis v7](https://github.com/redis/redis/pull/10227). Prior to v7, any entries that have been deleted from the stream but still exist in the PEL were returned as `(nil)`. This causes a deserialization error with the built-in `StreamAutoClaimReply` type:

```
Response was of incompatible type - TypeError: array([nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil]) (response was \"Incorrect type\")
```

Since the error occurs during deserialization, there's not much that can be done by a user to handle the issue. There are a couple things the client could do instead:

* Ignore the invalid values, but add a field to the `StreamAutoClaimReply` type indicating that there were ignored invalid PEL entries. This could simply be a `usize`, since, as far as I can tell, the `xautoclaim` response doesn't actually tell you anything else about the entries, including their IDs.
* Still fail the command, but return an error that can be programmatically checked to see if the problem is that there were invalid PEL entries.

With either of these approaches, the code changes could presumably be behind a feature flag, since they wouldn't apply to anyone using Redis v7.

Alternatively, if the crate isn't intended to support Redis v6, that should be noted in the README and/or docs somewhere.

nihohit:
Thanks for reporting this.
The crate should support redis 6, but this scenario isn't tested, apparently. If you can write me a test case that reproduces this, I can try writing a fix.

BatmanAoD:
I'm not familiar with how testing this crate works, but I'll take a look!

BatmanAoD:
[...]

PR #1800, opened 2025-09-27
introduces a test that fails with redis v6

BatmanAoD:
This is not a fix for #1798, but should demonstrate the issue when run against Redis v6.

I'm not entirely clear on whether there's currently a way to have different test behavior depending on the Redis engine version.


nihohit:
Fixed in https://github.com/redis-rs/redis-rs/pull/1822

PR #1822, opened 2025-10-04
Fixed deserialization of autoclaim messages

nihohit:
https://github.com/redis-rs/redis-rs/issues/1798

### DIFF
--- a/redis/src/commands/streams.rs
+++ b/redis/src/commands/streams.rs
@@ -456,6 +456,10 @@ pub struct StreamAutoClaimReply {
     pub claimed: Vec<StreamId>,
     /// The list of stream ids that were removed due to no longer being in the stream
     pub deleted_ids: Vec<String>,
+    /// If set, this means that the reply contained invalid nil entries, that were skipped during parsing.
+    ///
+    /// This should only happen when using Redis 6, see <https://github.com/redis-rs/redis-rs/issues/1798>
+    pub invalid_entries: bool,
 }
 
 /// Reply type used with [`xread`] or [`xread_options`] commands.
@@ -673,7 +677,7 @@ pub struct StreamKey {
 }
 
 /// Represents a stream `id` and its field/values as a `HashMap`
-#[derive(Default, Debug, Clone)]
+#[derive(Default, Debug, Clone, PartialEq)]
 pub struct StreamId {
     /// The stream `id` (entry ID) of this particular message.
     pub id: String,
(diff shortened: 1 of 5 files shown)


redis/src/commands/streams.rs
```diff
@@ -456,6 +456,10 @@ pub struct StreamAutoClaimReply {
     pub claimed: Vec<StreamId>,
     /// The list of stream ids that were removed due to no longer being in the stream
     pub deleted_ids: Vec<String>,
+    /// If set, this means that the reply contained invalid nil entries, that were skipped during parsing.
+    ///
+    /// This should only happen when using Redis 6, see <https://github.com/redis-rs/redis-rs/issues/1798>
+    pub invalid_entries: bool,
```
BatmanAoD, line 462: Since this is only relevant to Redis 6, would it make sense to put the field behind a config flag?

It may be useful to return a count of IDs rather than a bool; but since the user doesn't have the IDs either way, that may not be terribly useful.

Also, if this field is enabled, then the `deleted_ids` field probably shouldn't be enabled, since it's impossible for them both to be set (either you're using Redis 6 and `xclaim` can contain stale values, or you're using 7+ and stale entries are automatically deleted and their IDs are returned in `deleted_ids`).

redis/tests/test_streams.rs
```diff
@@ -1955,3 +1940,125 @@ fn test_xrevrange() {
...
+            1,
+            claim.id.clone(),
+            StreamAutoClaimOptions::default().count(4).with_justid(),
+        )
+        .unwrap();
+    // we expect to receive just IDs, so we remove the maps
+    std::mem::take(&mut claimed_entries[0].map);
+    std::mem::take(&mut claimed_entries[1].map);
+
+    if ctx.get_version().0 > 6 {
```
BatmanAoD, line 2039: I didn't take a close look at how the test context works; is the context using `REDIS_RS_REDIS_JSON_PATH`? 
I wasn't able to get that working for Redis 6 the other day.

BatmanAoD (commented):
I'm not very familiar with the codebase, but this looks right.

redis/tests/test_streams.rs
nihohit: It uses `REDIS_RS_REDIS_JSON_PATH` only when the JSON module is specified. When no module is specified you don't need to build the JSON module.

redis/src/commands/streams.rs
nihohit: > Since this is only relevant to Redis 6, would it make sense to put the field behind a config flag?

We use a library that checks all of the combination of feature flags, so we try to avoid adding those, as it slows down CI. This doesn't seem critical, so I think I won't add the flag.

> It may be useful to return a count of IDs rather than a bool; but since the user doesn't have the IDs either way, that may not be terribly useful.

I don't think it serves a purpose - we don't even have the IDs. 

> Also, if this field is enabled, then the deleted_ids field probably shouldn't be enabled, since it's impossible for them both to be set

Do you think this should be a single field enum? That would make sense, but its backwards incompatible, and I'm not sure that the gain here is worth the break.

BatmanAoD: Nope, not worth the break IMO.

redis/tests/test_streams.rs
BatmanAoD: So what redis engine is used by the tests?

nihohit: The [CI matrix is running](https://github.com/redis-rs/redis-rs/blob/3d2a103f89d59256fd76c0ccb38779692a1412a6/.github/workflows/rust.yml#L24) Redis 6.2, 7.2, 8.0, 8.2 & valkey 7.2 and 8.0.

PR #1825, opened 2025-10-09
Add more ToRedisArgs & FromRedisValue impls

QuarticCat:
The `S` generic arg is missing in some `ToRedisArgs` impls so we cannot use customized hasher in `HashMap` while we can do this for `HashSet`. A similar situation happens in `FromRedisValue` as well.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -2086,18 +2086,18 @@ macro_rules! from_redis_value_for_tuple {
                     return Ok(vec![]);
                 }
 
-                if items.iter().all(|item|item.is_collection_of_len(n)) {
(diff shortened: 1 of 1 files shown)


QuarticCat:
Lint CI should be fixed now.

QuarticCat:
The nightly lint error seems to be a problem of clippy: https://github.com/rust-lang/rust-clippy/issues/15852

And it is unrelated to code modified in this PR.

QuarticCat:
I managed to work around those nightly Clippy warnings anyway.

nihohit:
Thanks for the contribution!

PR #1826, opened 2025-10-11
docs: add bb8 feature to Optional Features section

tottoto:
Adds the bb8 feature to the Optional Features section.


nihohit:
Thanks for the contribution!

PR #1831, opened 2025-10-15
Add token-based authentication

StefanPalashev:
# 🔐 Add Token-Based Authentication with Azure Entra ID Support for Redis Connections

## 📋 Summary

This Pull Request introduces comprehensive token-based authentication support, with first-class integration for Azure Entra ID (formerly Azure Active Directory) for Redis connections, enabling secure authentication for Azure-hosted Redis instances and Redis Enterprise with Azure integration.

### 🔑 **Multiple Authentication Flows**
- **DeveloperToolsCredential** - Authenticates through developer tools such as the Azure CLI.
It tries the following credential types, in this order, stopping when one provides a token:
    * `AzureCliCredential`
    * `AzureDeveloperCliCredential`
- **Service Principal** - `Client secret` and `certificate-based` authentication
- **Managed Identity** - Both `system-assigned` and `user-assigned`
- **Custom Credentials** - Support for any `TokenCredential` implementation

### **Token Management**
- **Automatic Token Refresh & Streaming of Credentials** - Seamlessly handle token expiration and stream updated credentials to prevent connection failures due to token expiration
- **Configurable Refresh Policies** - Customizable refresh thresholds and retry behavior

### 🛡️ **Input Validation**

- **Empty Scope Prevention** - Rejects empty or invalid scope configurations
- **Scope Validation** - Ensures that scopes are not empty. 
The validation can be extended to follow Azure's `https://resource/.default` format.

### **Async-First Design**
- Full async/await support for non-blocking operations
- Seamless integration with multiplexed connections

### **Streaming Credentials Pattern**
The implementation uses a streaming approach where:
1. The credentials provider (`EntraIdCredentialsProvider`) runs a background task for token refresh
2. Connections subscribe to credential updates via async streams
3. Automatic re-authentication occurs when new tokens are available
4. Token refresh occurs without interrupting the connection

### **Error Handling**
The library provides detailed error messages for troubleshooting and properly handles authentication and token refresh failures. These messages are emitted via the `log` crate, allowing observation of the library’s runtime behavior through configured loggers.
The background token refresh service will automatically retry failed token refreshes according to the retry configuration, logging information about each attempt at appropriate levels.
Once the maximum number of attempts is exceeded, the service will stop retrying and the underlying error will be propagated to the subscribers, which are `multiplexed connections`, rendering them unusable.

### **Feature Flags**
- `token-based-authentication`: General traits and types for token-based authentication support
- `entra-id`: Azure Entra ID specific implementation includes the main dependencies - `azure_identity` and `azure_core`

### **New Modules**
```
redis/
├── src/
│   ├── auth.rs    # Core authentication traits and types
│   ├── auth_management.rs    # Token refresh configuration and utilities
│   └── entra_id.rs    # Azure Entra ID credentials provider implementation
└── tests/
    ├── test_acl.rs    # Integration tests using ACL and a mocked streaming credentials provider
    └── test_auth.rs    # Integration tests
    └── test_credentials_provider_failures.rs    # Tests for connection-layer behavior when streaming credentials providers fail.
```

### **Enhanced Modules**
```
redis/
└── src/
    ├──client.rs    # Attach a credentials provider to a client via the `with_credentials_provider` function
    ├──connection.rs    # Credentials provider in the `RedisConnectionInfo` struct
    ├──lib.rs    # New public APIs
    └── aio/
       └── multiplexed_connection.rs    # Streaming credentials subscription and re-authentication
```

### **Basic Usage**
```rust
use redis::{Client, EntraIdCredentialsProvider, RetryConfig};

// Using DeveloperToolsCredential
let mut provider = EntraIdCredentialsProvider::new_developer_tools()?;
provider.start(RetryConfig::default());

let client = Client::open_with_credentials_provider("redis://your-redis.azure.com:6380", provider).unwrap();

let mut connection = client.get_multiplexed_async_connection().await?;
```

### **Service Principal Authentication**
```rust
use redis::{Client, EntraIdCredentialsProvider, RetryConfig};

// Using ClientSecretCredential
let mut provider = EntraIdCredentialsProvider::new_client_secret(
    "tenant-id".to_string(),
    "client-id".to_string(), 
    "client-secret".to_string(),
)?;
provider.start(RetryConfig::default());

let client = Client::open_with_credentials_provider("redis://your-redis.azure.com:6380", provider).unwrap();

let mut connection = client.get_multiplexed_async_connection().await?;
```

### **Managed Identities**
```rust
use redis::{Client, EntraIdCredentialsProvider, RetryConfig};

let mut provider = EntraIdCredentialsProvider::new_system_assigned_managed_identity()?;
// or
/* 
User assigned managed identities can be identified by one of the values in the following enumeration:
pub enum UserAssignedId {
    /// The client ID of a user-assigned identity
    ClientId(String),
    /// The object or principal ID of a user-assigned identity
    ObjectId(String),
    /// The Azure resource ID of a user-assigned identity
    ResourceId(String),
}
*/

let provider = EntraIdCredentialsProvider::new_user_assigned_managed_identity_with_scopes(
    vec!["your-scope".to_string()],
    Some(ManagedIdentityCredentialOptions {
        // Specify the user-assigned identity using one of:
        user_assigned_id: Some(UserAssignedId::ClientId("your-client-id".to_string())),
        // or: user_assigned_id: Some(UserAssignedId::ObjectId("your-object-id".to_string())),
        // or: user_assigned_id: Some(UserAssignedId::ResourceId("your-resource-id".to_string())),
        ..Default::default()
    }),
)?;
provider.start(RetryConfig::default());

let client = Client::open_with_credentials_provider("redis://your-redis.azure.com:6380", provider).unwrap();

let mut connection = client.get_multiplexed_async_connection().await?;
```

### **Comprehensive Test Suite**
- Unit tests for token refresh configuration
- Unit tests for scopes validation
- JWT parsing and validation tests
- Mock credential provider tests simulating various scenarios:
  - Successful authentication
  - Authentication failures
  - Retry mechanisms
  - Multiple token refreshes over time
  - Multiple subscribers
  - Provider cleanup
- Mocked streaming credentials provider tests with ACL, validating automatic re-authentications and connection stability
- Tests for connection-layer behavior when streaming credentials providers fail

### **Mock Infrastructure**
- `MockStreamingCredentialsProvider` and `MockTokenCredential` for testing without real Azure dependencies
- Configurable response patterns (success, failure, alternating)
- Call counting and verification

### **Integration Tests**
The integration tests are designed to ensure correct interaction with real Azure services and identity endpoints.

The tests rely on standard Azure environment variables to authenticate against Azure Entra ID:
* `AZURE_TENANT_ID`
* `AZURE_CLIENT_ID`
* `AZURE_CLIENT_SECRET`
* `AZURE_CLIENT_CERTIFICATE_PATH`

### **New Dependencies**
```toml
[dependencies]
azure_identity = { version = "0.31", optional = true, features = ["client_certificate"] }
azure_core = { version = "0.31", optional = true }
base64 = { version = "0.22.1", optional = true }

[features]
token-based-authentication = ["dep:futures-util", "dep:log", "dep:backon"]
entra-id = ["dep:azure_identity", "dep:azure_core", "dep:base64", "token-based-authentication", "tokio-comp"]
```

### **Feature Gating**
All Entra ID functionality is properly feature-gated with `#[cfg(feature = "entra-id")]` to avoid dependency bloat for users who don't need Azure authentication.

### **Cargo.toml**
```toml
[dependencies]
redis = { version = "1.0", features = ["entra-id", "tokio-comp"] }
```

## 🛡️ **Security Considerations**

- **Token Security** - Tokens are handled securely
- **Scope Validation** - Prevents misconfiguration that could lead to authentication failures
- **Error Handling** - Azure-specific errors are properly mapped to Redis errors.

### **Added Documentation**
- `redis/docs/token_authentication.md` - Detailed usage guide with examples and integration patterns for different deployment scenarios
- `README.md` - Updated with token-based authentication examples

## **Backward Compatibility**

- **Fully backward compatible** - No changes to existing APIs
- **Optional feature** - Only affects builds that explicitly enable `entra-id`
- **No breaking changes** - All existing authentication methods continue to work

## 🔍 **Code Quality**

- **Zero Code Duplication** - Common patterns extracted into reusable functions
- **Error Handling** - All error paths properly handled with descriptive messages
- **Memory Efficiency** - Minimal allocations and efficient token caching

## 🚦 **Testing Instructions**
The integration tests are ignored by default, as they are designed to be executed only against live environments.
They can be run manually using the following command:

```bash
cargo test --test test_auth --features entra-id -- --ignored
```
To run a specific test, the following command can be used:

```bash
cargo test --test test_auth --features entra-id <test_name> -- --ignored
```

## 🎉 **Benefits**

- **🔐 Enterprise Security** - Leverage Azure's enterprise-grade authentication
- **🔄 Zero-Downtime Authentication** - Automatic token refresh prevents connection drops
- **📈 Scalability** - Supports high-throughput applications with efficient token management
- **🛠️ Developer Experience** - Simple, intuitive API with examples
- **🏢 Enterprise Ready** - Supports all Azure authentication patterns used in production

## **Contact / Support**
For any future questions, feedback, or discussion related to this feature, members of the open source community can reach out at [ecosystems@redis.com](mailto:ecosystems@redis.com)


redis/Cargo.toml
```diff
@@ -144,6 +151,8 @@ num-bigint = ["dep:num-bigint"]
 cache-aio = ["aio", "dep:lru"]
 r2d2 = ["dep:r2d2"]
 bb8 = ["dep:bb8"]
+token-based-authentication = ["aio"]
```
nihohit: why is this enabled only with aio?

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,217 @@
...
+- `expiration_refresh_ratio`: Fraction of token lifetime before refresh (0.0-1.0)
+- `retry_config`: Configuration for retry behavior on failures
+
+### RetryConfig
+
+- `max_attempts`: Maximum number of retry attempts
+- `initial_delay`: Initial delay before first retry
+- `max_delay`: Maximum delay between retries
+- `backoff_multiplier`: Exponential backoff multiplier
+- `jitter_percentage`: Random jitter percentage (0.0-1.0)
```
nihohit: clarify what these retries mean

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Authentication Flows
+
+### DefaultAzureCredential (Recommended for Development)
+
+The `DefaultAzureCredential` tries multiple credential sources in order:
+
+```rust
+let provider = EntraIdCredentialsProvider::new_default()?;
+```
```
nihohit: all of the code documentation should also (or only) appear in code, where it will be tested by rust doc tests

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,217 @@
...
+};
+
+provider.start(config);
+```
+
+## Error Handling
+
+The library provides comprehensive error handling for authentication and token refresh failures.
+The background token refresh service will automatically retry failed token refreshes according to the retry configuration.
+Once the maximum number of attempts is reached, the service will stop retrying and the underlying error will be propagated to the subscribers.
```
nihohit: how & when will it propagate?

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,217 @@
...
+### 1. Use Appropriate Credential Types
+
+- **Development**: `DefaultAzureCredential`
+- **Production Services**: Service Principal with certificate
+- **Azure-hosted Apps**: Managed Identity
+
+### 2. Configure Appropriate Refresh Ratios
+
+- **High-frequency apps**: 0.5-0.7 (refresh early)
+- **Low-frequency apps**: 0.8-0.9 (refresh later)
```
nihohit: what's a high/low-frequency app?

redis/src/aio/multiplexed_connection.rs
```diff
@@ -487,6 +489,13 @@ pub struct MultiplexedConnection {
     _task_handle: Option<SharedHandleContainer>,
     #[cfg(feature = "cache-aio")]
     pub(crate) cache_manager: Option<CacheManager>,
+    /// Connection info for re-authentication purposes
+    #[allow(dead_code)]
```
nihohit: remove, or set the appropriate `cfg`

redis/tests/test_auth.rs
```diff
@@ -0,0 +1,397 @@
+#![cfg(feature = "token-based-authentication")]
```
nihohit: these tests are all unit tests, and should be in the relevant files, not with external integration tests.

what you're missing is actual tests that check that connections get new credentials, and send these new credentials to the server.
You should create a test credentials provider, that doesn't rely on entra, and then either check that it refreshes the credentials with a mock server, or by spawning a task that periodically* changes the ACL password, and seeing that the connection is still usable.

*periodically here means every couple of milliseconds, because we don't want a test that runs for minutes or more.

redis/docs/token_authentication.md
StefanPalashev: In this context, an app refers to any application or service registered in Microsoft Entra ID that requests tokens to access protected resources.

Normally, high-frequency apps are those that access protected resources often or run continuously, so they should refresh tokens earlier to avoid expiration issues. Whereas, low-frequency apps access protected resources less often, so their token refreshes can be delayed to minimize unnecessary authentication requests.

StefanPalashev: Based on the authentication flow, the returned tokens have different lifetimes. Access tokens obtained via **service principals** (using **client credentials** or **certificates**) typically expire after one hour.

As stated in the [guidelines by Microsoft:](
https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview-for-developers?tabs=dotnet#guidelines-if-youre-handling-tokens-directly)

> "Managed identity tokens are valid for 24 hours, there's no benefit in requesting new tokens regularly, as a cached one will be returned from the token issuing endpoint."

> When you acquire a token, you can set your token cache to expire 5 minutes before the expires_on (or equivalent property) that will be returned when the token is generated.

The `azure_identity` crate implements a cache internally, which keeps the latest token up to 5 minutes before its expiration for most of the authentication flows.

Based on these fixed lifetimes, the commit with message "`EntraIdCredentialsProvider and documentation refactoring`" simplifies the `EntraIdCredentialsProvider` by relying only on the `RetryConfig` struct, however it leaves the `TokenRefreshConfig` available for future use.

redis/Cargo.toml
StefanPalashev: Thanks for pointing this out!

The `token-based-authentication` used to be the only flag, before `entra-id` was introduced. It's supposed to bring in the `traits` needed for such authentication. With the latest changes, the only dependency for it is the `futures:util`.

config/redis.toml
```diff
@@ -99,6 +99,28 @@ require = ["smol-rustls-comp"]
...
+[[rule]]
+when = "entra-id"
+require = ["tokio-comp"]
+forbid = ["smol-rustls-comp", "OR", "smol-native-tls-comp"]
+
+# Entra ID with any Rustls-based TLS backend requires `tokio-rustls-comp`
+[[rule]]
+when = ["entra-id", "AND", ["tls-rustls", "OR", "tls-rustls-webpki-roots"]]
+require = ["tokio-rustls-comp"]
+forbid = ["NOT", "tokio-rustls-comp"]
```
nihohit: this can be removed

config/redis.toml
```diff
@@ -99,6 +99,28 @@ require = ["smol-rustls-comp"]
...
+
+# Entra ID with any native-tls backend requires `tokio-native-tls-comp`
+[[rule]]
+when = ["entra-id", "AND", "tls-native-tls"]
+require = ["tokio-native-tls-comp"]
+forbid = ["NOT", "tokio-native-tls-comp"]
```
nihohit: same

redis/src/auth_management.rs
```diff
@@ -125,3 +125,74 @@ pub mod credentials_management_utils {
         Err("OID claim not found".to_string())
     }
 }
+
+#[cfg(all(feature = "token-based-authentication", test))]
+mod auth_management_tests {
```
nihohit, line 143: since the tests are here, then things that shouldn't be `pub` can be just `pub(crate)` or even private in the file above. please ensure everything is minimally public.

redis/src/entra_id.rs
```diff
@@ -587,3 +587,330 @@ mod tests {
         assert_eq!(provider.scopes, custom_scopes);
     }
 }
+
+#[cfg(all(feature = "entra-id", test))]
```
nihohit, line 763: note to self: didn't go over these yet.

redis/tests/test_auth.rs
```diff
@@ -1 +1,110 @@
...
+            .unwrap_or_else(|_| panic!("The `REDIS_URL` environment variable is not set."))
+    }
+
+    fn get_env_var(var_name: &str) -> String {
+        std::env::var(var_name)
+            .unwrap_or_else(|_| panic!("The `{var_name}` environment variable is not set."))
+    }
+
+    async fn test_redis_connection(provider: EntraIdCredentialsProvider, test_key: &str) {
+        let mut provider = provider;
```
nihohit: why redeclare the argument?

redis/tests/test_auth.rs
```diff
@@ -1 +1,110 @@
+#[cfg(feature = "entra-id")]
```
nihohit, line 8: ~~note to self - review this~~
done

redis/tests/test_auth.rs
```diff
@@ -1 +1,110 @@
...
+mod entra_id_tests {
```
nihohit, line 9: we still need (here or in another file) tests the take an implementation of the credentials provider that can run in our CI - probably a mocked credentials provider, that can refresh credentials in a sub-second rate.
Just to see that the connection reacts correctly - sends AUTH, etc.
(as mentioned [here](https://github.com/redis-rs/redis-rs/pull/1831#discussion_r2438514015)already)

redis/src/auth_management.rs
StefanPalashev: Will do, thanks!

redis/tests/test_auth.rs
StefanPalashev: Thank you for noting it, the provider needs to be marked as mutable, hence the redeclaration. 
It's now passed as mutable instead.

redis/docs/token_authentication.md
StefanPalashev: All of the examples provided here are either fragments or pseudo-code samples of what can be seen in the tests within `test_auth.rs` and `entra_id.rs`. The goal was to have simple fragments in one place for quick reference.

Should they be removed from this file?

redis/tests/test_auth.rs
StefanPalashev: [a6229a7a](https://github.com/redis-rs/redis-rs/pull/1831/commits/a6229a7aa2321d5b7a938de31120c8dde57bd305) Introduces the tests.

redis/tests/test_auth.rs
```diff
@@ -1 +1,109 @@
...
+            .arg(test_key)
+            .query_async(&mut con)
+            .await
+            .unwrap();
+
+        assert_eq!(result, Some("42".to_string()));
+    }
+
+    #[tokio::test]
+    #[ignore]
```
nihohit: please add documentation explaining why all of these tests are ignored.
or better yet, add at the top of the file something like `#![cfg(false)]` and document there why the whole file is ignored - that will allow removing the per-test ignores.

redis/docs/token_authentication.md
nihohit: no need to remove them from here, I just want also documentation to appear in the docs.rs representation.
Most users are unlikely to find this file, unless something will point them towards this.

redis/src/entra_id.rs
```diff
@@ -0,0 +1,922 @@
...
+//! ```rust,no_run
+//! use redis::{Client, EntraIdCredentialsProvider, RetryConfig};
+//!
+//! # async fn example() -> redis::RedisResult<()> {
+//! // Create the credentials provider using the DeveloperToolsCredential
+//! let mut provider = EntraIdCredentialsProvider::new_developer_tools()?;
+//! provider.start(RetryConfig::default());
+//!
+//! // Create Redis client with credentials provider
+//! let client = Client::open("redis://your-redis-instance.com:6380")?
```
nihohit: why is the client opened and only afterwards the credentials provider is added?
shouldn't the credentials provider be a part of the `ConnectionInfo`, just like authentication is?
and does it make sense to ahve authentication and the credentials provider both enabled?

redis/tests/test_auth.rs
StefanPalashev: These tests are normally ignored because they **should only be run against existing Azure environments**.
Furthermore, some of them depend on **managed identities** (both `system` and `user-assigned`), **which are only available on Azure-hosted machines that expose the Instance Metadata Service (IMDS)**.

There is an existing Redis infrastructure that provisions such environments and hosts, and runs these tests on them.
Because of this, they cannot be annotated with `#![cfg(false)]`, as this would prevent them from being compiled.
An alternative approach would be to create a dedicated feature flag and annotate the tests with it.

redis/src/entra_id.rs
StefanPalashev: By design, the credentials provider is intended to be attached to existing clients, allowing multiple clients to share the same provider and obtain the same credentials. Additionally, some of the tests (the ones located in `test_acl.rs`), rely on the `TestContext` infrastructure, which currently can utilize the identity provider by it being attached to the client, as there is no way to create the client with an identity provider. While modifications might be possible, they would likely require significant effort with minimal benefit.

The credentials provider is part of the `RedisConnectionInfo`, which is part of the `ConnectionInfo`. It makes more sense to be placed in there, since the username and password field are placed within it and the implementation targets Redis connections.

Having both authentication and credentials provider enabled might make sense in scenarios where some of the connections would require specific credentials, while others would rely on credentials that get updated over time.

redis/tests/test_auth.rs
StefanPalashev: The file was altered with documentation that explains that.

redis/src/entra_id.rs
StefanPalashev: The discussion above revealed a gap in the implementation that prevented the provider from being shared across clients. This gap is now addressed, and new tests have been added to demonstrate how a credentials provider can be shared between clients and connections.

nihohit: > By design, the credentials provider is intended to be attached to existing clients, allowing multiple clients to share the same provider and obtain the same credentials.

Why allowing multiple clients to share the same provider means that the provider has to be attached to existing clients? Just clone or Arc it, and pass it to the client when you open it.

> The credentials provider is part of the RedisConnectionInfo, which is part of the ConnectionInfo.

Right, but `ConnectionInfo` is passed to the client it's being created, so why mutate the value afterwards? Why not add the provider to the `ConnectionInfo` before creating the client?

StefanPalashev: Commit [c1f092ca](https://github.com/redis-rs/redis-rs/pull/1831/commits/c1f092cacadb034a3eae54202347122c4614a694) adds a new constructor that allows the provider to be passed when creating a client.

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,225 @@
+# Token-Based Authentication with Azure Entra ID
```
nihohit: instead of adding this as a separate document in a new folder, let's try to integrate this into the library's rustdoc, in a way that will use our ability to check that the code actually compiles (we just had issues with stale code in the README). Please move this to the `entra_id` module, and check the generated rust docs to see that it looks like you expect the doc to look, and that each named type is linked to the type's documentation.

redis/docs/token_authentication.md
nihohit: @StefanPalashev this comment still wasn't answered.

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,225 @@
...
+
+### 1. Use Appropriate Credential Types
+
+- **Development**: `DeveloperToolsCredential`
+- **Production Services**: Service Principal with certificate
+- **Azure-hosted Apps**: Managed Identity
+
+### 2. Handle Token Expiration
+
+- Use background refresh services for long-running applications
```
nihohit: what does this mean? what do users have to do in order to do this?

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,225 @@
...
+- Implement proper error handling for authentication failures
```
nihohit: same

redis/docs/token_authentication.md
```diff
@@ -0,0 +1,225 @@
...
+
+### RetryConfig
+
+Configuration for handling failed token refresh attempts.
+
+- `max_attempts`: Maximum number of retry attempts for token refresh.
+- `initial_delay`: Initial delay before attempting to refresh the token after a failure. Subsequent retries use exponential backoff based on this value.
+- `max_delay`: Upper bound for retry delays to prevent excessively long waits. Delays will never exceed this value, even with exponential backoff.
+- `backoff_multiplier`: Growth factor for exponential backoff (typically 2.0 for doubling). Each retry delay is multiplied by this value: `delay = delay * backoff_multiplier`.
+- `jitter_percentage`: Random variation added to delays as a fraction of the calculated delay (0.0 to 1.0). For example, 0.5 means up to ±50% variation to prevent synchronized retries.
```
nihohit: once you integrate this into the rustdoc, there won't be a need to re-document each type, just link to the type's documentation.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -557,9 +566,32 @@ impl MultiplexedConnection {
             })
             .transpose()?;
 
+        #[cfg(feature = "token-based-authentication")]
+        let mut connection_info = connection_info.clone();
+        #[cfg(not(feature = "token-based-authentication"))]
+        let connection_info = connection_info.clone();
+
+        #[cfg(feature = "token-based-authentication")]
+        if let Some(ref cp) = connection_info.credentials_provider {
```
nihohit: nit: cp -> credentials_provider

redis/src/aio/multiplexed_connection.rs
```diff
@@ -587,6 +633,36 @@ impl MultiplexedConnection {
...
+                            } else {
+                                println!("Re-authenticated async connection");
+                            }
+                        }
+                        Err(err) => {
+                            eprintln!("Credential stream error for async connection: {err}");
+                        }
+                    }
+                }
+                println!("Re-authentication stream ended");
```
nihohit: please remove these prints. We should consider how to pass failures back to the user, but the console isn't the right place for it.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -596,6 +672,12 @@ impl MultiplexedConnection {
         self._task_handle = Some(SharedHandleContainer::new(handle));
     }
 
+    /// This function sets the handle for the credentials subscription task when a credentials provider is used.
```
nihohit: why do we need this function? can't this be inlined in line 663?

redis/src/aio/multiplexed_connection.rs
```diff
@@ -587,6 +633,36 @@ impl MultiplexedConnection {
             cache_manager: cache_manager_opt,
         };
 
+        // Set up streaming credentials subscription if provider is available
+        #[cfg(feature = "token-based-authentication")]
+        if let Some(streaming_provider) = connection_info.credentials_provider {
+            let mut inner_connection = con.clone();
```
nihohit, line 666: I don't think that this task should hold a full `MultiplexedConnection` - this might lead down the line to circular pointers, and stuff not getting released. Can this be reduced to only holding `Pipeline` and the added fields necessary for the credentials usage (which will be removed from the multiplexed connection)?

redis/src/auth.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// The password for authentication
+    pub password: String,
+}
+
+/// Trait for providing authentication credentials
+///
+/// This trait allows different authentication mechanisms to be plugged into
+/// the Redis client, such as static passwords, token-based authentication,
+/// or dynamic credential providers like Azure Entra ID.
+pub trait CredentialsProvider: Send + Sync {
```
nihohit: I don't see where this, or StaticCredentialsProvider, are used.

redis/src/auth.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Get the current authentication credentials
+    ///
+    /// This function should return valid credentials that can be used for
+    /// Redis authentication. If the credentials are expired or invalid,
+    /// the implementation should refresh them before returning.
+    fn get_credentials(&self) -> RedisResult<BasicAuth>;
+}
+
+/// Async version of the credentials provider trait
+pub trait AsyncCredentialsProvider: Send + Sync {
```
nihohit: is this used anywhere?

redis/src/auth.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/// Trait for providing credentials in a streaming fashion
+///
+/// This allows connections to subscribe to credential updates and automatically
+/// re-authenticate when tokens are refreshed, preventing connection failures
+/// due to token expiration.
+pub trait StreamingCredentialsProvider: Send + Sync {
+    /// Get an independent stream of credentials.
+    fn subscribe(&self) -> Pin<Box<dyn Stream<Item = RedisResult<BasicAuth>> + Send + 'static>>;
```
nihohit, line 49: shouldn't this take as input the connection info? how would we handle a cluster configuration with different credentials for each node?

I don't know if such a thing exists, but it seems wrong to assume that it doesn't.

redis/src/auth_management.rs
```diff
@@ -0,0 +1,200 @@
+//! This module contains utilities for managing token-based authentication
+use std::time::{Duration, SystemTime};
+
+/// Configuration for token refresh behavior
+#[derive(Debug, Clone)]
+pub struct TokenRefreshConfig {
```
nihohit, line 7: please add `#[non_exhaustive]`.

redis/src/auth_management.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Fraction of token lifetime after which refresh should be triggered (0.0 to 1.0).
+    /// For example, 0.8 means refresh when 80% of the token's lifetime has elapsed.
+    pub expiration_refresh_ratio: f64,
```
nihohit, line 10: should we allow for jitter around this, to prevent thundering herd authentication failures?

redis/src/auth_management.rs
```diff
@@ -0,0 +1,200 @@
...
+        Self {
+            expiration_refresh_ratio: 0.8,
+            retry_config: RetryConfig::default(),
+        }
+    }
+}
+
+/// Configuration for handling failed token refresh attempts
+#[derive(Debug, Clone)]
+pub struct RetryConfig {
```
nihohit, line 41: add `non_exhaustive`

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+        Self::validate_scopes(&scopes)?;
+        let credential_provider =
+            ClientSecretCredential::new(&tenant_id, client_id, client_secret.into(), options)
+                .map_err(Self::convert_error)?;
+        Ok(Self {
+            credential_provider: Arc::new(
+                std::sync::Arc::try_unwrap(credential_provider).map_err(|_| {
+                    RedisError::from((
+                        ErrorKind::AuthenticationFailed,
+                        "Failed to unwrap credential",
```
nihohit: what does this error mean to the user? how can they handle it?

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+            //     )));
+            // }
+
+            // if !scope.ends_with("/.default") {
+            //     return Err(RedisError::from((
+            //         ErrorKind::InvalidClientConfig,
+            //         "Invalid scope: must end with '/.default'",
+            //         format!("Scope: '{scope}'"),
+            //     )));
+            // }
```
nihohit: why is this commented out?

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+/// If the PKCS12 archive is password-protected, provide the password via `password`.
+#[derive(Debug, Clone)]
+pub struct ClientCertificate {
+    /// Base64-encoded PKCS12 certificate data
+    pub base64_pkcs12: String,
+    /// The certificate's password if any
+    pub password: Option<String>,
+}
+
+type Subscriptions = Vec<Arc<Sender<RedisResult<BasicAuth>>>>;
```
nihohit: I don't think that the `Arc` here is necessary, the sender is already cheaply clonable.

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+                    Ok(Self::convert_credentials(username.to_owned(), access_token))
+                }
+                Err(error) => Err(Self::convert_error_ref(error)),
+            };
+
+            sender.send(response)
+        }))
+        .await;
+
+        subscribers_arc
```
nihohit: why isn't this done on the initial lock, and before sending the messages?

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+        retry_config: RetryConfig,
+        compute_sleep_duration_on_success: F,
+    ) where
+        F: Fn(&AccessToken) -> std::time::Duration + Send + Sync + 'static,
+    {
+        // Prevent multiple calls to start
+        if self
+            .background_handle
+            .lock()
+            .expect("mutex poisoned")
```
nihohit: this doesn't give more information than just an `unwrap`.

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+
+    /// Start the background refresh service
+    fn start_refresh_service<F>(
+        &mut self,
+        retry_config: RetryConfig,
+        compute_sleep_duration_on_success: F,
+    ) where
+        F: Fn(&AccessToken) -> std::time::Duration + Send + Sync + 'static,
+    {
+        // Prevent multiple calls to start
```
nihohit, line 383: how can multiple calls to start happen? shouldn't that be an error?

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+            subscribers: Default::default(),
+            current_credentials: Default::default(),
+        })
+    }
+
+    /// Start the background refresh service
+    pub fn start(&mut self, retry_config: RetryConfig) {
+        self.start_refresh_service(retry_config, |access_token| {
+            let remaining = access_token.expires_on - OffsetDateTime::now_utc();
+            let remaining_duration = match remaining.try_into() {
```
nihohit, line 640: what's the type conversion here? are you converting duration to duration?

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+                Ok(duration) => duration,
+                Err(_) => std::time::Duration::from_secs(0),
+            };
+            remaining_duration
+                .checked_sub(std::time::Duration::from_secs(240))
```
nihohit: why is this a constant value?

redis/src/entra_id.rs
```diff
@@ -0,0 +1,951 @@
...
+            background_handle: Default::default(),
+            subscribers: Default::default(),
+            current_credentials: Default::default(),
+        })
+    }
+
+    /// Start the background refresh service
+    pub fn start(&mut self, retry_config: RetryConfig) {
+        self.start_refresh_service(retry_config, |access_token| {
+            let remaining = access_token.expires_on - OffsetDateTime::now_utc();
```
nihohit, line 639: why do we need the `time` crate and UTC time? isn't everything using local machine time?

redis/docs/token_authentication.md
StefanPalashev: They are going to be propagated via the stream.

When the maximum number of allowed retry attempts is reached, the function `Self::notify_subscribers(&subscribers_arc, &username, token_response).await;` will be called. At this point, the `token_response` parameter (`token_response: Result<AccessToken, azure_core::Error>`) will contain the Error variant.

nihohit: please clarify that in the docs, from the user's, not the implementers', perspective.

redis/src/auth.rs
StefanPalashev: These traits are kept for potential future use.

Implementing any of them would allow a user to fetch credentials on-demand from external sources. This could be useful if there's a need for a single, one-time token.

The static implementation serves as a wrapper around `BasicAuth`, which consists of a simple username and password.

redis/src/auth_management.rs
StefanPalashev: This can be achieved by using the field inside `retry_config` - `jitter_percentage`. A previous implementation included jittering; however, it was removed because of constraints imposed by the tokens’ lifetimes.

> Based on the authentication flow, the returned tokens have different lifetimes. Access tokens obtained via **service principals** (using **client credentials** or **certificates**) typically expire after one hour.
> 
> As stated in the [guidelines by Microsoft:](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview-for-developers?tabs=dotnet#guidelines-if-youre-handling-tokens-directly)
> 
> > "Managed identity tokens are valid for 24 hours, there's no benefit in requesting new tokens regularly, as a cached one will be returned from the token issuing endpoint."
> 
> > When you acquire a token, you can set your token cache to expire 5 minutes before the expires_on (or equivalent property) that will be returned when the token is generated.
> 
> The `azure_identity` crate implements a cache internally, which keeps the latest token up to 5 minutes before its expiration for most of the authentication flows.
> 
> Based on these fixed lifetimes, commit "`EntraIdCredentialsProvider and documentation refactoring`" simplifies the `EntraIdCredentialsProvider` by relying only on the `RetryConfig` struct, however it leaves the `TokenRefreshConfig` available for future use.

As mentioned above (link to the quoted comment) - https://github.com/redis-rs/redis-rs/pull/1831#discussion_r2463850065), due to the implementation details, `TokenRefreshConfig` remains reserved for future use.

redis/src/aio/multiplexed_connection.rs
StefanPalashev: Thank you, there was no need for an extra function and the call is now inlined.

redis/docs/token_authentication.md
StefanPalashev: The examples and most of the relevant information in the file has been moved to `entra_id` as part of its documentation.
This file is now deleted.

redis/src/aio/multiplexed_connection.rs
StefanPalashev: Fixed

StefanPalashev:
Thanks for the review! I’m working through the comments now and will follow up once everything is addressed.

redis/src/auth.rs
nihohit: but these traits aren't used ATM, so implementing them now won't give the user anything, right?
let's remove them, and re-introduce them once there's actual usage.

redis/src/entra_id.rs
StefanPalashev: As stated before, in the documentation - https://github.com/redis-rs/redis-rs/pull/1831#discussion_r2463850065:

> The azure_identity crate implements a cache internally, which keeps the latest token up to 5 minutes before its expiration for most of the authentication flows.

The caching inside the `azure_identity` crate uses the following code to determine whether a token should be refreshed:
`fn should_refresh(token: &AccessToken) -> bool {
    token.expires_on <= OffsetDateTime::now_utc() + Duration::seconds(300)
}`

Because of this, the lifetime approach, which used `TokenRefreshConfig` was abandoned and the fixed value of 4 minutes was used in here.

redis/src/entra_id.rs
StefanPalashev: The access token provided by the `azure_core` crate uses types from the `time` crate, with expiration timestamps expressed in UTC.

redis/src/entra_id.rs
StefanPalashev: The conversion is from `time::Duration` to `std::time::Duration`.

redis/src/entra_id.rs
StefanPalashev: `start` is expected to be called exactly once. Additional calls indicate a programming error and should be ignored.

redis/src/entra_id.rs
StefanPalashev: The code has been replaced with an `unwrap()`.

redis/src/entra_id.rs
StefanPalashev: Thanks for the suggestion!
The code has been modified to reflect it. As a result, the necessary cleanup is performed upfront, ensuring that only active subscribers are being notified about the messages.

redis/src/entra_id.rs
StefanPalashev: Correct, it must have been a part of an older version of the code. It is now removed.

redis/src/entra_id.rs
StefanPalashev: A new test that verifies the behavior has been added.

redis/docs/token_authentication.md
StefanPalashev: These recommendations are no longer applicable, as the `StreamingCredentialsProvider` handles all of these cases internally. They were only relevant when alternative credentials providers could be used, which is no longer possible since the other traits have been removed.

As a result, users do not need to take any action—starting the credentials provider automatically covers these scenarios.

redis/src/auth.rs
StefanPalashev: As far as I know, different credentials will not be used for each node when Entra ID is used, however I'll have to double check this again.
Edit: That's the case - there won't be different tokens at least for Entra ID. All of the nodes will use the token provided by the credentials provider.

The cluster support is planned as a separate Pull Request.

As for the handling part, we'll have to alter the `ClusterConnection` a bit. There seems to be a `Vec<ConnectionInfo>` inside of it, which could potentially hold info about the provider.

redis/src/entra_id.rs
StefanPalashev: I’m not sure this is needed, so I’ll remove it for now and add it back if necessary.

redis/src/entra_id.rs
StefanPalashev: This shouldn't be possible. If this is to happen it would mean that the credential provider returned by the `azure_identity` has unexpected shared ownership.

redis/src/aio/multiplexed_connection.rs
StefanPalashev: The challenge in here is that this happens asynchronously and most probably **after** `new_with_config()` has already returned.

We'll need to decide how to pass these failures to the user. There are a few variants:
 - Terminate the connection
 - Create an error channel using the multiplexed connection’s push sender. This would allow re-authentication errors to be sent through the channel, enabling users to optionally listen for them and react accordingly.
 - Panic (although we should probably avoid this)

redis/src/aio/multiplexed_connection.rs
StefanPalashev: I think the comment above about propagating failures back to the user needs to be resolved first, as it appears to block this change.

StefanPalashev:
@nihohit 
I’ve addressed the feedback from the previous review. When you have a moment, I’d appreciate another look and any guidance you may have on the remaining items (especially the ones targeting `redis/src/aio/multiplexed_connection.rs`).

redis/src/entra_id.rs
nihohit: I don't think that means that we need a direct dependency on the crate. AFAIK we only need direct dependencies if we use types / functions from. if we only call functions from return values from `azure_core`, then I don't think the dependency is necessary.

nihohit: Essentially, let's try using `into` :)

redis/src/entra_id.rs
nihohit: then please put it under a named, commented constant.

redis/src/aio/multiplexed_connection.rs
nihohit: terminate the connection.

redis/src/entra_id.rs
nihohit: the put this in the returned error - some text that will clarify that this is an unexpected situation, and should be raised.

redis/src/aio/multiplexed_connection.rs
nihohit: which comment, [this](https://github.com/redis-rs/redis-rs/pull/1831#discussion_r2662293913)? Is it blocking because we need to decide on a propagation mechanism?

redis/src/aio/multiplexed_connection.rs
```diff
@@ -585,7 +585,7 @@ impl MultiplexedConnection {
                     return Err(err);
                 }
                 None => {
-                    println!("Credential stream closed");
+                    unreachable!("Credential stream closed");
```
nihohit: if there's no possible way to output that this happens without crashing the user program, it's better to keep this codepath empty.
Just do
```
None => {
  // This path is taken when the credential stream is closed. ATM we do nothing - 
  //  the user will notice this issue if / when the connection will stop authenticating.
}

File: redis/src/entra_id.rs [hunk 41]
```diff
@@ -245,7 +413,10 @@ impl EntraIdCredentialsProvider {
                             retry_config.backoff_multiplier,
                             retry_config.max_delay,
                         );
-                        println!("An error occurred while refreshing the token. Attempt {attempt}. Sleeping for {:?}", error_delay);
+                        println!(
```
nihohit: again, remove printlns from the code. If you want, instead use the `log` dependency in the feature.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -558,9 +567,32 @@ impl MultiplexedConnection {
...
+        #[cfg(feature = "token-based-authentication")]
+        if let Some(ref credentials_provider) = connection_info.credentials_provider {
+            // Retrieve the initial credentials from the provider and apply them to the connection info
+            match credentials_provider.subscribe().next().await {
+                Some(Ok(credentials)) => {
+                    connection_info.username = Some(ArcStr::from(credentials.username));
+                    connection_info.password = Some(ArcStr::from(credentials.password));
+                }
+                Some(Err(err)) => {
+                    eprintln!("Error while receiving credentials from stream: {err}");
```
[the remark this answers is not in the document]
nihohit: use the `log` crate

redis/src/aio/multiplexed_connection.rs
```diff
@@ -588,6 +634,37 @@ impl MultiplexedConnection {
...
+                            } else {
+                                println!("Re-authenticated async connection");
+                            }
+                        }
+                        Err(err) => {
+                            eprintln!("Credential stream error for async connection: {err}");
+                        }
+                    }
+                }
+                println!("Re-authentication stream ended");
```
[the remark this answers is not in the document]
nihohit: use the `log` crate

redis/src/entra_id.rs
```diff
@@ -0,0 +1,1164 @@
...
+                            retry_config.max_delay,
+                        );
+                        println!(
+                            "An error occurred while refreshing the token. Attempt {attempt}. Sleeping for {:?}",
+                            error_delay
+                        );
+                        tokio::time::sleep(error_delay).await;
+                        continue;
+                    }
+                    println!("Max attempts reached. Stopping token refresh.");
```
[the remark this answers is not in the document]
nihohit: use the `log` crate

redis/src/entra_id.rs
```diff
@@ -0,0 +1,1164 @@
...
+        self.start_refresh_service(retry_config, |access_token| {
+            let remaining = access_token.expires_on - OffsetDateTime::now_utc();
+            let remaining_duration = match remaining.try_into() {
+                Ok(duration) => duration,
+                Err(_) => std::time::Duration::from_secs(0),
+            };
+            remaining_duration
+                .checked_sub(std::time::Duration::from_secs(240))
+                .unwrap_or_else(|| {
+                    eprintln!("Token expires soon; refreshing immediately");
```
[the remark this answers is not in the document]
nihohit: use the `log` crate

nihohit:
@StefanPalashev I marked as resolved the Copilot comments that I found irrelevant.

redis/tests/test_auth.rs
```diff
@@ -0,0 +1,138 @@
...
+    async fn test_service_principal_client_secret() {
```
nihohit, line 87: I think that we're missing a long-running test, that checks for actual identity updates over time.

redis/src/aio/multiplexed_connection.rs
StefanPalashev: The `None` arm is now handled properly.

redis/src/entra_id.rs
StefanPalashev: The `println!` calls were replaced with logging via the `log` crate.

redis/src/entra_id.rs
```diff
@@ -0,0 +1,1164 @@
...
+        let credential_provider =
+            DeveloperToolsCredential::new(options).map_err(Self::convert_error)?;
+        Ok(Self {
+            credential_provider: Arc::new(
+                std::sync::Arc::try_unwrap(credential_provider).map_err(|_| {
+                    RedisError::from((
+                        ErrorKind::AuthenticationFailed,
+                        "Failed to unwrap credential",
+                    ))
+                })?,
```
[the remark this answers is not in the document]
StefanPalashev: The code was extracted into a helper function.

redis/src/entra_id.rs
StefanPalashev: This code was duplicated across multiple constructors. It is now a separate function with a more descriptive message.

redis/src/entra_id.rs
StefanPalashev: The value has been extracted to a constant.

redis/src/entra_id.rs
StefanPalashev: The `azure_core` crate re-exports types from the `time` crate, so a direct dependency on it is no longer necessary.

redis/src/entra_id.rs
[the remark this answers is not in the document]
StefanPalashev: The `log` crate has been introduced as a dependency for the feature.

redis/src/entra_id.rs
```diff
@@ -0,0 +1,1164 @@
...
+    /// Start the background refresh service
+    pub fn start(&mut self, retry_config: RetryConfig) {
+        self.start_refresh_service(retry_config, |access_token| {
+            let remaining = access_token.expires_on - OffsetDateTime::now_utc();
+            let remaining_duration = match remaining.try_into() {
+                Ok(duration) => duration,
+                Err(_) => std::time::Duration::from_secs(0),
+            };
+            remaining_duration
+                .checked_sub(std::time::Duration::from_secs(240))
```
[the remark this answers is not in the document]
StefanPalashev: The requested change has been applied.

redis/src/entra_id.rs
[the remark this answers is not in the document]
StefanPalashev: The `log` crate has been introduced as a dependency for the feature.

redis/src/aio/multiplexed_connection.rs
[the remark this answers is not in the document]
StefanPalashev: The `log` crate has been introduced as a dependency for the feature.

redis/Cargo.toml
```diff
@@ -50,6 +50,7 @@ tokio = { version = "1", features = [
   "net",
   "time",
   "sync",
+  "macros"
```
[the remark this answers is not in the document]
StefanPalashev: The requested changes have been applied.

README.md
```diff
@@ -163,6 +163,36 @@ To enable insecure mode, append `#insecure` at the end of the URL:
...
+```toml
+redis = { version = "0.32.7", features = ["entra-id", "tokio-comp"] }
+```
+
+### Supported Authentication Flows
+
+- **DeveloperToolsCredential**: Authenticates through developer tools such as the Azure CLI.
+It tries the following credential types, in this order, stopping when one provides a token:
+* [`AzureCliCredential`]
+* [`AzureDeveloperCliCredential`]
```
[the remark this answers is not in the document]
StefanPalashev: This suggestion was ignored.

redis/src/aio/multiplexed_connection.rs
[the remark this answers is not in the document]
StefanPalashev: The `log` crate has been introduced as a dependency for the feature.

StefanPalashev:
As discussed above (in https://github.com/redis-rs/redis-rs/pull/1831#discussion_r2684022129), the connection has to be terminated in the event of re-authentication errors or errors caused by the credentials provider. However, the latest changes introduce a mechanism that does not terminate the connection, but instead leaves it in a practically unusable state by causing all subsequent commands to fail with a re-authentication error.

Is the correct way to terminate the connection to drop the Pipeline’s `mpsc::Sender` so that the channel closes and the driver task is aborted?

@nihohit, could you please provide guidance on how to approach this, as the current solution is probably not sufficient enough? Thanks in advance!

redis/tests/test_auth.rs
StefanPalashev: `test_mock_multiple_tokens_over_time` in `entra_id.rs` and `test_token_rotation_with_mock_streaming_credentials_provider` in `test_acl.rs` are doing something similar.

Do you think they are enough? There's the option to add an integration test against a real `EntraID` deployment, but it would run for nearly an hour due to the long token expiration time.

README.md
```diff
@@ -163,6 +163,36 @@ To enable insecure mode, append `#insecure` at the end of the URL:
...
 ```
 
+## Token-Based Authentication with Azure Entra ID
+
+Redis-rs supports token-based authentication using Azure Entra ID, providing secure, dynamic authentication for Redis connections. This feature is particularly useful for Azure-hosted applications and enterprise environments.
+
+To enable Entra ID authentication, add the `entra-id` feature to your Cargo.toml:
+
+```toml
+redis = { version = "1.0.3", features = ["entra-id", "tokio-comp"] }
```
nihohit:
```suggestion
redis = { version = "1.0", features = ["entra-id", "tokio-comp"] }
```

nihohit:
> Is the correct way to terminate the connection to drop the Pipeline’s mpsc::Sender so that the channel closes and the driver task is aborted?

That will only work if all the clones of the connection did this, and I don't think that this is enforceable. I think that if the authentication stream terminates and an error is logged, then the rest of the failed requests will be clear. I guess we'll just have to rely on the error logs.

config/redis.toml
```diff
@@ -99,6 +99,16 @@ require = ["smol-rustls-comp"]
 when = ["tls-rustls-webpki-roots", "tokio-comp"]
 require = ["tokio-rustls-comp"]
 
+# ------------------------------------------------------------------------
+# Entra ID integration (requires Tokio-based async + TLS compatibility)
+# ------------------------------------------------------------------------
+
+# Entra ID is Tokio-only; forbid Smol-based TLS runtimes
```
nihohit, line 106: I couldn't find this in the docs. why is this?

.github/workflows/rust.yml
```diff
@@ -289,7 +289,15 @@ jobs:
...
+          vcpkg install openssl:x64-windows
+          echo "OPENSSL_DIR=C:\vcpkg\packages\openssl_x64-windows" >> $env:GITHUB_ENV
+          echo "OPENSSL_INCLUDE_DIR=C:\vcpkg\packages\openssl_x64-windows\include" >> $env:GITHUB_ENV
+          echo "OPENSSL_LIB_DIR=C:\vcpkg\packages\openssl_x64-windows\lib" >> $env:GITHUB_ENV
+
       - name: lint
         uses: ./.github/actions/lint-and-check
         with:
-          rust-toolchain: stable    
+          rust-toolchain: stable
```
nihohit: revert

nihohit:
@StefanPalashev, sorry, pushed another commit with the fix to flag-frenzy. once you pull that and answer https://github.com/redis-rs/redis-rs/pull/1831#discussion_r2777883459, I think that we'll be done :)

StefanPalashev:
> > Is the correct way to terminate the connection to drop the Pipeline’s mpsc::Sender so that the channel closes and the driver task is aborted?
> 
> That will only work if all the clones of the connection did this, and I don't think that this is enforceable. I think that if the authentication stream terminates and an error is logged, then the rest of the failed requests will be clear. I guess we'll just have to rely on the error logs.

This should mean that we can invoke `drop` on the sender when `re_authentication_failed` has been set, eventually all of the clones will get their senders closed if they attempt to send commands.

nihohit:
> This should mean that we can invoke drop on the sender when re_authentication_failed has been set, eventually all of the clones will get their senders closed if they attempt to send commands.

Is this an answer to my last comment?
I was asking why the change is only for tokio - I couldn't find any reference that says entra is only supported by tokio.

StefanPalashev:
> > This should mean that we can invoke drop on the sender when re_authentication_failed has been set, eventually all of the clones will get their senders closed if they attempt to send commands.
> 
> Is this an answer to my last comment? I was asking why the change is only for tokio - I couldn't find any reference that says entra is only supported by tokio.

No, about that, initially there was an observation of some of the code in the `azure_crates`, which steered into `tokio`. Internally it uses `reqwest` as its HTTP client in some of the calls as well, which also uses `tokio` under the hood.

nihohit:
Thanks for the contribution! I'll create a release soon.

Issue #1832, opened 2025-10-16
native-tls TlsCertificates custom


hepengju:
when use rustls, redis-rs support use custom certificates
when use native-tls, redis-rs have no way to custom Identity & Certificate.

Maybe It will be nice if add a method like this.

**client.get_connection(tlsConnector: TlsConnector)**

(Remark: Sometimes Redis use ssl with: X509v1, 
But rustls giveup support this old version. https://github.com/rustls/rustls/issues/2364. 
So we have to use native-tls at this time)

nihohit:
Feel free to contribute a fix for this.

PR #1833, opened 2025-10-16
native-tls custom tls_connector (#1832)

hepengju:


redis/src/connection.rs
```diff
@@ -53,6 +52,9 @@ pub struct TlsConnParams {
     pub(crate) root_cert_store: Option<RootCertStore>,
     #[cfg(any(feature = "tls-rustls-insecure", feature = "tls-native-tls"))]
     pub(crate) danger_accept_invalid_hostnames: bool,
+
+    #[cfg(feature = "tls-native-tls")]
+    pub(crate) tls_connector: Option<TlsConnector>,
```
nihohit, line 57: this is never set. how can a user use this change?

hepengju: 3Q for review. I will add more code and test it for this feature later.

PR #1840, opened 2025-10-22
Add docs notes as suggested in #1836

### LINKED ISSUE #1836: Non-obvious single-result pipeline behavior
I'm using pipes for a few things with sorted sets and I stumbled across a non-obvious error.

My code looks roughly like this:

```rs
let conn = ...;

let mut pipe = redis::pipe();
pipe.zrembyscore(&key, 0, 10).ignore();
pipe.zrangebyscore(&key, 0, -1);

let (resp,): (Vec<u128>,) = pipe.query_async(&mut conn).await?;
```

My issue is with the last line of that snippet. Before getting to that conclusion, I was using

```rs
let resp: Vec<u128> = pipe.query_async(&mut conn).await?;
```

This seemed obvious to me since the first command of the pipe is ignored, so then only one result should allow me to just type it without a tuple. This is not like that.

The issue is not only that it does not work like that (which is pretty counterintuitive IMO), but also that the error does not point to that being the issue.

```
Response was of incompatible type - TypeError: "Response type not convertible to numeric." (response was array([]))
```

The thing here is that since my `resp` variable is typed as `Vec<u128>`, then it seemed obvious to me that `array([])` was an empty `Vec<u128>` and for some reason since it's empty it couldn't parse numbers inside it (that's why the "Response type not convertible to numeric."). That's not the case, `array([])` is the root array of responses from the pipeline, and the `Vec<u128>` is going to be a nested array within the root array.

This is neither obvious nor documented, so my suggestion here is to support doing `let resp: Vec<u128> = pipe.query_async(&mut conn).await?;` and then assuming that non-tuple types are single-response pipelines.

I'm willing to make a PR for this to either document the behavior or add support for non-tuple pipeline results if I can be guided a bit into where to look at.

Nekidev:
Fixes #1836

I added two notes about single-result pipeline responses needing to be wrapped in a collection for better clarity.

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -378,7 +378,7 @@
 //! # let mut con = client.get_connection().unwrap();
 //! let (k1,): (i32,) = redis::pipe()
 //!     .cmd("SET").arg("key_1").arg(42).ignore()
-//!     .cmd("GET").arg("key_1").query(&mut conn).unwrap();
+//!     .cmd("GET").arg("key_1").query(&mut con).unwrap();
 //! ```
 //!
 //! # Transactions


redis/src/lib.rs
```diff
@@ -370,6 +370,17 @@
 //! # Ok(()) }
 //! ```
 //!
+//! NOTE: Pipelines return a collection of results, even when there's only a sinle response.
```
nihohit: sinle -> single

redis/src/pipeline.rs
```diff
@@ -110,6 +110,17 @@ impl Pipeline {
     /// NOTE: A Pipeline object may be reused after `query()` with all the commands as were inserted
     ///       to them. In order to clear a Pipeline object with minimal memory released/allocated,
     ///       it is necessary to call the `clear()` before inserting new commands.
+    ///
+    /// NOTE: This method returns a collection of results, even when there's only a sinle response.
```
nihohit: same

nihohit (commented):
thanks!

nihohit:
Thanks for the contribution!

PR #1859, opened 2025-11-06
Add getters to `aio::ConnectionManagerConfig`, `ConnectionInfo`, `RedisConnectionInfo` and `io::tcp::TcpSettings` types

LazyMechanic:
This PR introduces getter methods in `aio::ConnectionManagerConfig`, `ConnectionInfo`, `RedisConnectionInfo` and `io::tcp::TcpSettings` types to improve library usability and configurability without exposing internal fields directly.

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -94,12 +94,12 @@ impl ConnectionManagerConfig {
         Self::default()
     }
 
-    /// Returns a minimum delay between connection attempts.
+    /// Returns the minimum delay between connection attempts.
     pub fn min_delay(&self) -> Duration {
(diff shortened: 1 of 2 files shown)


redis/src/aio/connection_manager.rs
```diff
@@ -94,6 +94,51 @@ impl ConnectionManagerConfig {
         Self::default()
     }
 
+    /// Returns a minimum delay between connection attempts.
```
nihohit: here and below - `the` instead of `a`

LazyMechanic: Fixed

nihohit:
Thanks for the contribution!

PR #1862, opened 2025-11-10
Support Compare and set [CAS] and Compare and delete [CAD]

StefanPalashev:
# Support Compare and set and Compare and delete
## Summary

Redis Release 8.4 introduces a straightforward, single-command approach for implementing single-key optimistic concurrency control with `compare-and-set` and `compare-and-delete` operations.

---
# Changes introduced with this Pull Request:

## Extend the `SET` Command with a new parameter - Value comparison

**Syntax:**
**SET** _key_ _value_ [NX | XX | :new: **IFEQ match-value** | **IFNE match-value** | **IFDEQ match-digest** | **IFDNE match-digest**] [GET] [EX seconds | PX milliseconds | EXAT unix-time-seconds | PXAT unix-time-milliseconds | KEEPTTL]

### Value comparisons

**IFEQ** `match-value` - Set the key’s value and expiration **only if** its current value is **equal to** `match-value`.  
  If the key doesn’t exist, it **won’t be created**.
**IFNE** `match-value` - Set the key’s value and expiration **only if** its current value is **not equal to** `match-value`.  
  If the key doesn’t exist, it **will be created**.
**IFDEQ** `match-digest` - Set the key’s value and expiration **only if** the **digest of its current value** is **equal to** `match-digest`.
  If the key doesn’t exist, it **won’t be created**.
**IFDNE** `match-digest` - Set the key’s value and expiration **only if** the **digest of its current value** is **not equal to** `match-digest`.
  If the key doesn’t exist, it **will be created**.

---

## New `DELEX` Command
**Description:**
Conditionally deletes a key, which holds a string value, if it exists and the provided value comparison condition is held.

**Syntax:**
**DELEX** _key_ [IFEQ match-value | IFNE match-value | IFDEQ match-digest | IFDNE match-digest]

### Value comparisons

**IFEQ** `match-value` - Delete the key **only if** its value is **equal to** `match-value`.  
**IFNE** `match-value` - Delete the key **only if** its value is **not equal to** `match-value`.  
**IFDEQ** `match-digest` - Delete the key **only if** the **digest of its value** is **equal to** `match-digest`.  
**IFDNE** `match-digest` - Delete the key **only if** the **digest of its value** is **not equal to** `match-digest`.  

### Reply

* **Error** - If the key exists but holds a non-string value.
* **Integer** - `0` - Key not deleted (does not exist or condition was false), `1` - Key successfully deleted.

---

## New `DIGEST` Command
**Description:**
Returns the hash digest of the value stored in the key, as a hex string.
NOTE: For the digest, Redis will use XXH3

**Syntax:**
**DIGEST** _key_

### Reply

- **Nil** - Key does not exist.  
- **Error** - Key exists but holds a non-string value.  
- **Bulk string** - The **hex string** representation of the **XXH3 digest** of the key’s value .

---

## **Dependencies**

Introduced the `xxhash-rust` dependency to enable hash calculation. This allows developers to compute fast XXH3 digests of key values for conditional operations such as **IFDEQ** and **IFDNE**.

### **New Dependencies**
```
xxhash-rust = { version = "0.8", features = ["xxh3"] }
```

## **Testing**:
- Tests have been added for each `ValueComparison` used by the **SET** command, along with dedicated tests for the **DELEX** and **DIGESTS** commands.
- Additional tests were also added to verify the behavior of the digest utility.

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -378,7 +378,7 @@ implement_commands! {
     /// IFDEQ `match-digest` - Delete the key only if the digest of its value is equal to `match-digest`
     /// IFDNE `match-digest` - Delete the key only if the digest of its value is not equal to `match-digest`
     /// [Redis Docs](https://redis.io/commands/DELEX)
-    fn del_ex<K: ToRedisArgs>(key: K, value_comparison: ValueComparison) -> (usize) {
+    fn del_ex<K: ToSingleRedisArg>(key: K, value_comparison: ValueComparison) -> (usize) {
         cmd("DELEX").arg(key).arg(value_comparison)
     }
 
@@ -3343,7 +3343,7 @@ impl<Db: ToString> ToSingleRedisArg for CopyOptions<Db> {}
 /// ) -> RedisResult<Vec<usize>> {
 ///     let opts = SetOptions::default()
 ///         .conditional_set(ExistenceCheck::NX)
-///         .value_comparison(ValueComparison::IFEQ("old_value".to_string()))
+///         .value_comparison(ValueComparison::ifeq("old_value"))
 ///         .get(true)
 ///         .with_expiration(SetExpiry::EX(60));
 ///     con.set_options(key, value, opts)
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -141,6 +141,17 @@ pub enum Value {
 }
 
 /// Helper enum that is used to define comparisons between values and their digests
+///
+/// # Example
+/// ```rust
+/// use redis::ValueComparison;
+///
+/// // Create comparisons using constructor methods
+/// let eq_comparison = ValueComparison::ifeq("my_value");
+/// let ne_comparison = ValueComparison::ifne("other_value");
+/// let deq_comparison = ValueComparison::ifdeq("digest_hash");
+/// let dne_comparison = ValueComparison::ifdne("other_digest");
+/// ```
 #[derive(Clone, Debug)]
 #[non_exhaustive]
 pub enum ValueComparison {
@@ -152,6 +163,57 @@ pub enum ValueComparison {
     IFDEQ(String),
     /// Value's digest is not equal
     IFDNE(String),
+}
+
+impl ValueComparison {
+    /// Create a new IFEQ (if equal) comparison
+    ///
+    /// Performs the operation only if the key's current value is equal to the provided value.
+    ///
+    /// For SET: Sets the key only if its current value matches. Non-existent keys are not created.
+    /// For DEL_EX: Deletes the key only if its current value matches. Non-existent keys are ignored.
+    pub fn ifeq(value: impl ToSingleRedisArg) -> Self {
+        ValueComparison::IFEQ(Self::arg_to_string(value))
+    }
+
+    /// Create a new IFNE (if not equal) comparison
+    ///
+    /// Performs the operation only if the key's current value is not equal to the provided value.
+    ///
+    /// For SET: Sets the key only if its current value doesn't match. Non-existent keys are created.
+    /// For DEL_EX: Deletes the key only if its current value doesn't match. Non-existent keys are ignored.
+    pub fn ifne(value: impl ToSingleRedisArg) -> Self {
+        ValueComparison::IFNE(Self::arg_to_string(value))
+    }
+
+    /// Create a new IFDEQ (if digest equal) comparison
+    ///
+    /// Performs the operation only if the digest of the key's current value is equal to the provided digest.
+    ///
+    /// For SET: Sets the key only if its current value's digest matches. Non-existent keys are not created.
+    /// For DEL_EX: Deletes the key only if its current value's digest matches. Non-existent keys are ignored.
+    ///
+    /// Use [`calculate_value_digest`] to compute the digest of a value.
+    pub fn ifdeq(digest: impl ToSingleRedisArg) -> Self {
+        ValueComparison::IFDEQ(Self::arg_to_string(digest))
+    }
+
+    /// Create a new IFDNE (if digest not equal) comparison
+    ///
+    /// Performs the operation only if the digest of the key's current value is not equal to the provided digest.
+    ///
+    /// For SET: Sets the key only if its current value's digest doesn't match. Non-existent keys are created.
+    /// For DEL_EX: Deletes the key only if its current value's digest doesn't match. Non-existent keys are ignored.
+    ///
+    /// Use [`calculate_value_digest`] to compute the digest of a value.
+    pub fn ifdne(digest: impl ToSingleRedisArg) -> Self {
+        ValueComparison::IFDNE(Self::arg_to_string(digest))
+    }
+
+    fn arg_to_string(value: impl ToSingleRedisArg) -> String {
+        let args = value.to_redis_args();
+        String::from_utf8_lossy(&args[0]).into_owned()
+    }
 }
 
 impl ToRedisArgs for ValueComparison {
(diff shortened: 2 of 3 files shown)


redis/src/commands/mod.rs
```diff
@@ -3317,22 +3335,32 @@ impl<Db: ToString> ToSingleRedisArg for CopyOptions<Db> {}
...
     conditional_set: Option<ExistenceCheck>,
+    /// IFEQ `match-value` - Set the key's value and expiration only if its current value is equal to `match-value`.
+    /// If the key doesn't exist, it won't be created.
+    /// IFNE `match-value` - Set the key's value and expiration only if its current value is not equal to `match-value`.
+    /// If the key doesn't exist, it will be created.
+    /// IFDEQ `match-digest` - Set the key's value and expiration only if the digest of its current value is equal to `match-digest`.
+    /// If the key doesn't exist, it won't be created.
+    /// IFDNE `match-digest` - Set the key's value and expiration only if the digest of its current value is not equal to `match-digest`.
+    /// If the key doesn't exist, it will be created.
+    value_comparison: Option<ValueComparison>,
```
nihohit, line 3363: I don't see these in the [SET](https://redis.io/docs/latest/commands/set/) docs

redis/src/commands/mod.rs
```diff
@@ -3317,22 +3335,32 @@ impl<Db: ToString> ToSingleRedisArg for CopyOptions<Db> {}
...
 ///         .get(true)
 ///         .with_expiration(SetExpiry::EX(60));
 ///     con.set_options(key, value, opts)
 /// }
 /// ```
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Default)]
 pub struct SetOptions {
     conditional_set: Option<ExistenceCheck>,
+    /// IFEQ `match-value` - Set the key's value and expiration only if its current value is equal to `match-value`.
```
nihohit, line 3355: since these are new, please add to the docs from which version they're available, and phrase it in a way that clarifies that it's only on Redis.

redis/src/commands/mod.rs
```diff
@@ -371,6 +372,23 @@ implement_commands! {
         cmd("DEL").arg(key)
     }
 
+    /// Conditionally removes the specified key. A key is ignored if it does not exist.
```
nihohit, line 375: since these are new, please add to the docs from which version they're available, and phrase it in a way that clarifies that it's only on Redis.

redis/src/commands/mod.rs
```diff
@@ -371,6 +372,23 @@ implement_commands! {
...
+    /// IFEQ `match-value` - Delete the key only if its value is equal to `match-value`
+    /// IFNE `match-value` - Delete the key only if its value is not equal to `match-value`
+    /// IFDEQ `match-digest` - Delete the key only if the digest of its value is equal to `match-digest`
+    /// IFDNE `match-digest` - Delete the key only if the digest of its value is not equal to `match-digest`
+    /// [Redis Docs](https://redis.io/commands/DELEX)
+    fn del_ex<K: ToRedisArgs>(key: K, value_comparison: ValueComparison) -> (usize) {
+        cmd("DELEX").arg(key).arg(value_comparison)
```
nihohit, line 382: same, I couldn't find this in the docs

nihohit (commented):
it's hard for me to review code for an undocumented API. is there somewhere where I can read the commands' documentation?

StefanPalashev:
Marking the Pull Request as draft, as the 8.4 release has been postponed.

StefanPalashev:
The 8.4 Release is now official.
Links to the relevant documentation:
https://redis.io/docs/latest/commands/set/
https://redis.io/docs/latest/commands/delex/
https://redis.io/docs/latest/commands/digest/

redis/src/types.rs
```diff
@@ -140,6 +140,46 @@ pub enum Value {
...
 }
 
+/// Helper enum that is used to define comparisons between values and their digests
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum ValueComparison {
+    /// Value is equal
+    IFEQ(String),
+    /// Value is not equal
+    IFNE(String),
```
nihohit, line 161: why are these strings and not `ToRedisArgs`? Doesn't that move the responsibility of serializing the values to the user?

redis/src/types.rs
```diff
@@ -2273,6 +2313,44 @@ pub fn from_redis_value<T: FromRedisValue>(v: Value) -> Result<T, ParsingError>
...
+/// use redis::{calculate_value_digest, ValueComparison, SetOptions};
+///
+/// let value = "my_value";
+/// let digest = calculate_value_digest(value);
+///
+/// // Use the digest in a value comparison
+/// let opts = SetOptions::default()
+///     .value_comparison(ValueComparison::IFDEQ(digest));
+/// ```
+pub fn calculate_value_digest<T: ToRedisArgs>(value: T) -> String {
```
nihohit, line 2393: these are just helper functions to make usage of the feature easier for the users, right?
What's the rationale for adding these?
I'm wondering because AFAIS they require the addition of a new dependency, and I'd like to know that's actually needed.

StefanPalashev: > these are just helper functions to make usage of the feature easier for the users, right?
Yes, you are right about this. 

The rationale is that this became an explicit design requirement for the server, which states that clients must also implement a method for calculating a string’s digest.

They can be placed behind a feature flag for users who often need digest comparisons and would benefit from local digest calculation. The server can also compute the digest via the `digest` command, though this requires an additional network call.

redis/src/types.rs
StefanPalashev: It does. Changing the `enum` to accept `ToRedisArgs` requires changes in the `SetOptions` struct as well, which in turn requires changes in the places where it is used. Is this the correct approach as these seem to be breaking changes?

nihohit: I guess it can take `Box<dyn ToRedisArgs>`, which wouldn't propagate the types outside.

redis/src/types.rs
nihohit: no, it's a pretty small dependency, so IMO this is acceptable.

redis/src/types.rs
nihohit: oh, this reminds me - the enum should be `#[nonexhaustive]`
maybe with a constructor that takes `impl ToRedisArgs`, so that if I change the internal implementation, it won't be a major breakage.

redis/src/types.rs
StefanPalashev: Thanks

redis/src/types.rs
StefanPalashev: The second approach was taken (Implement a constructor that takes `impl ToRedisArgs`). It seems that `ToSingleRedisArg` fits better, so I've decided to pick it instead.

redis/src/commands/mod.rs
```diff
@@ -3343,13 +3343,14 @@ impl<Db: ToString> ToSingleRedisArg for CopyOptions<Db> {}
...
 ///         .conditional_set(ExistenceCheck::NX)
-///         .value_comparison(ValueComparison::IFEQ("old_value".to_string()))
+///         .value_comparison(ValueComparison::ifeq("old_value"))
 ///         .get(true)
 ///         .with_expiration(SetExpiry::EX(60));
 ///     con.set_options(key, value, opts)
 /// }
 /// ```
 #[derive(Clone, Default)]
+#[non_exhaustive]
```
nihohit: unnecessary, since the struct's fields are private.

nihohit (commented):
LGTM. fix last comment.

nihohit:
Thanks for the contribution!

PR #1863, opened 2025-11-13
feat: Allow the async cluster to be created without waiting for connections

Marwes:
We have a use case where we don't want to prevent our service from either starting or operating if there are problems with our redis cluster. Once everything is running the current cluster connection already handles re-connections however creating the connection during startup would, with the current API, force us to connect to the cluster once (which could cause problems if there are connection issues just as we spin up a new service).

By deferring the initial connection step to done with `reconnect_to_initial_nodes` we can just let the connection run in the background.

### DIFF
--- a/redis/src/cluster_handling/async_connection/mod.rs
+++ b/redis/src/cluster_handling/async_connection/mod.rs
@@ -167,49 +167,38 @@ where
         initial_nodes: &[ConnectionInfo],
         cluster_params: ClusterParams,
     ) -> RedisResult<ClusterConnection<C>> {
-        let protocol = cluster_params.protocol.unwrap_or_default();
-        let response_timeout = cluster_params.response_timeout;
-        #[cfg(feature = "cache-aio")]
-        let cache_manager = cluster_params.cache_manager.clone();
-        let runtime = Runtime::locate();
-        let inner = ClusterConnInner::new(initial_nodes, cluster_params);
-        let inner = inner.wait_for_initial_connection().await?;
-
-        let (sender, mut receiver) = mpsc::channel::<Message<_>>(100);
-        let stream = async move {
-            let _ = stream::poll_fn(move |cx| receiver.poll_recv(cx))
-                .map(Ok)
-                .forward(inner)
-                .await;
-        };
-        let _task_handle = HandleContainer::new(runtime.spawn(stream));
-
-        Ok(ClusterConnection {
-            sender,
-            state: Arc::new(ClientSideState {
-                protocol,
-                _task_handle,
-                response_timeout,
-                runtime,
-                #[cfg(feature = "cache-aio")]
-                cache_manager,
-            }),
-        })
+        let (connection, connect_receiver) = Self::new_inner(initial_nodes, cluster_params);
+        dbg!(connect_receiver.await).map_err(|_| {
+            RedisError::from((ErrorKind::Io, "Cluster connection task were dropped"))
+        })??;
+        Ok(connection)
     }
 
     pub(crate) fn new_pending(
         initial_nodes: &[ConnectionInfo],
         cluster_params: ClusterParams,
     ) -> ClusterConnection<C> {
+        let (connection, _connect_receiver) = Self::new_inner(initial_nodes, cluster_params);
+        connection
+    }
+
+    pub(crate) fn new_inner(
+        initial_nodes: &[ConnectionInfo],
+        cluster_params: ClusterParams,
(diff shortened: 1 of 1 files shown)


redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -172,30 +170,62 @@ where
...
-                        #[cfg(feature = "cache-aio")]
-                        cache_manager,
-                    }),
-                }
-            })
+    pub(crate) fn new_pending(
+        initial_nodes: &[ConnectionInfo],
+        cluster_params: ClusterParams,
+    ) -> ClusterConnection<C> {
+        let protocol = cluster_params.protocol.unwrap_or_default();
```
nihohit, line 189: since this is almost completely a repetition of `new`, I'd prefer it if the logic was consolidated into a single function, with arguments that describe the difference in behavior.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -995,119 +1105,43 @@ impl<C> ClusterConnInner<C>
...
+            )),
             cluster_params,
             pending_requests: Mutex::new(Vec::new()),
             initial_nodes: initial_nodes.to_vec(),
             subscription_tracker,
         });
-        let connection = ClusterConnInner {
-            inner: Core(inner),
+        let core = Core(inner);
+        ClusterConnInner {
```
nihohit: A lot of code is moved, with git history removed, just because you can't call `reconnect_to_initial_nodes` without creating `ClusterConnInner`, but this PR can be much smaller by just creating the inner representation and then mutating its state:
```rs
let core = Core(inner);
let mut inner = ClusterConnInner {
    inner: core.clone(),
    in_flight_requests: Default::default(),
    refresh_error: None,
    state: ConnectionState::PollComplete,
};
inner.state = ConnectionState::Recover(RecoverFuture::Reconnect(Box::pin(
    inner.reconnect_to_initial_nodes(),
)));
inner
```

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -995,119 +1105,43 @@ impl<C> ClusterConnInner<C>
...
+            inner: core.clone(),
```
nihohit, line 1040: nit: why clone?

redis/src/cluster_handling/client.rs
```diff
@@ -571,6 +571,24 @@ impl ClusterClient {
         .await
     }
 
+    /// Creates new connections to Redis Cluster nodes with a custom config and returns a
+    /// [`cluster_async::ClusterConnection`]. The connections to the cluster are done in the
+    /// background.
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if there is a failure while creating connections or slots.
```
nihohit: This doesn't return result, an error cannot be returned here.
Instead, you should document clearly that using this means that the user won't know if the cluster is unreachable, or that the addresses were false.

redis/src/cluster_handling/async_connection/mod.rs
Marwes: The two function differ in that one is async and the other (used for `get_pending_`) isn't , I don't see a way to combine them into one (without forcing `get_pending_` to be async for no reason).

redis/src/cluster_handling/async_connection/mod.rs
Marwes: Because it is used in `reconnect_to_initial_nodes` 3 lines below.

redis/src/cluster_handling/async_connection/mod.rs
Marwes: It wasn't to let me call the function in line with the constructor, them main reson to refactor were to replace the `-> impl Future` functions with plain `async fn` variants which I think simplifies the code a bit, but I reverted that part.

redis/src/cluster_handling/client.rs
Marwes: Fixed

redis/src/cluster_handling/async_connection/mod.rs
nihohit: I think that these functions are just complex enough that it's likely that future changes will only be applied in one implementation and not the other, and I'd really like to avoid that.

To your point, you could call a unified sync function that runs a generic function in the middle, and in that generic function you could spawn a task, and send its handle on a oneshot channel for the async `new` function.
It's a bit convoluted, but IMO it's worth the cost.

Marwes: Done

PR #1865, opened 2025-11-18
feat: Add ignore_errors mode to Pipeline for partial results

kudlatyamroth:
This PR introduces a new ignore_errors mode to the Pipeline struct. This feature allows users to execute a pipeline and retrieve results for all commands, even if some of them fail, disabling the default fail-fast behavior.

Currently, redis-rs pipelines operate in a fail-fast mode. If a single command within the pipeline returns an error from the server, the entire operation returns an error, and successful results from other commands are lost.

There are scenarios (e.g., json.get with path) where it is desirable to process successful responses and handle errors individually. This change allows users to opt-in to such behavior.

### DIFF
--- a/redis/tests/test_async.rs
+++ b/redis/tests/test_async.rs
@@ -4,19 +4,22 @@ mod support;
 mod basic_async {
     use std::{collections::HashMap, time::Duration};
 
+    use crate::support::*;
     use futures::{prelude::*, StreamExt};
     use futures_time::{future::FutureExt, task::sleep};
+    #[cfg(feature = "json")]
+    use redis::JsonAsyncCommands;
     use redis::{
         aio::ConnectionLike, cmd, pipe, AsyncCommands, ErrorKind, IntoConnectionInfo, ParsingError,
         ProtocolVersion, PushKind, RedisConnectionInfo, RedisError, RedisResult, ScanOptions,
         ServerErrorKind, Value,
     };
+    #[cfg(feature = "json")]
+    use redis_test::server::Module;
     use redis_test::server::{redis_settings, use_protocol};
     use rstest::rstest;
     use test_macros::async_test;
     use tokio::sync::mpsc::error::TryRecvError;
-
-    use crate::support::*;
 
     #[rstest::rstest]
     #[cfg_attr(feature = "tokio-comp", case::tokio(RuntimeType::Tokio))]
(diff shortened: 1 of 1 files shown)


kudlatyamroth:
atomic only adds multi / exec, but results are interpreted like every command would be run separately, so it will work :)
redis alone, when i run:
```
mutli
json.get some_key .
json.get some_key .some_path
json.get some_key .
exec
```
return results like:
```
1) some_value
2) Err path not exist
3) some_value
```

nihohit:
IIRC That's not how MULTI/EXEC works. Please add a test demonstrating how it works.

kudlatyamroth:
Ok, test added

when we have transaction enabled, ignore errors not make sense without json module.
for example when we don't have enabled json module, something like this:
```
set x 34
multi
get x
json.get x .
get y
get x
exec
```
will fail transaction with error about not existent method, but when we have enabled json module, it will run whole transaction with response:
```
1) 34
2) Err wrong type
3) (nil)
4) 34
```

and when we don't have atomic operation and json disabled, ignore_errors will return response, but in second place will be Not Existent method 

Because of this, i think about changing `ignore_errors` to `ignore_response_errors`

~~@nihohit can you check errors here on test? locally it runs without problem, but locally i tested only on redis 8~~
nvm. i had to change test name

redis/src/pipeline.rs
```diff
@@ -71,6 +73,41 @@ impl Pipeline {
         self
     }
 
+    /// Configures the pipeline to return partial results even if some commands fail.
+    ///
+    /// By default (`false`), if any command within the pipeline returns an error from the Redis server,
+    /// the entire `query()` method immediately returns that error (fail-fast).
+    ///
+    /// When set to `true`, the pipeline will execute all commands and return the results.
```
nihohit: `When set,` (because the user won't see the undelying boolean)

redis/src/pipeline.rs
```diff
@@ -71,6 +73,41 @@ impl Pipeline {
         self
     }
 
+    /// Configures the pipeline to return partial results even if some commands fail.
+    ///
+    /// By default (`false`), if any command within the pipeline returns an error from the Redis server,
```
nihohit: `By default,` (because the user won't see the undelying boolean)

redis/src/pipeline.rs
```diff
@@ -71,6 +73,41 @@ impl Pipeline {
...
+    /// the entire `query()` method immediately returns that error (fail-fast).
```
nihohit: this phrasing implies that the next commands won't be sent, which isn't correct. Maybe 
`the query will only return the errors, and not the results of the successful calls`?

redis/src/pipeline.rs
```diff
@@ -71,6 +73,41 @@ impl Pipeline {
...
+    ///
+    /// When set to `true`, the pipeline will execute all commands and return the results.
+    /// If a specific command failed, its error is preserved in the response.
+    ///
+    /// **Important:** When enabling this, you should adjust your return type to handle individual errors.
+    /// Instead of `Vec<T>`, you should likely use `Vec<RedisResult<T>>`.
```
nihohit: also, clarify that the type shouldn't be a tuple, since the returned collection will receive errors from calls that are marked with `ignore`, and so the user might receive more values than expected.

kudlatyamroth: but, when we return response, we filter ignored results, even if it is error:
```rust
                if self.ignore_errors {
                    return Ok(from_redis_value(Value::Array(
                        self.filter_ignored_results(response),
                    ))?);
                }
```

i updated test, to show that :)

redis/src/pipeline.rs
kudlatyamroth: changed

redis/src/pipeline.rs
kudlatyamroth: changed, and added more specification

redis/src/pipeline.rs
kudlatyamroth: changed

nihohit:
Thanks for the contribution!

kudlatyamroth:
thank you for merging and guidance :)

PR #1871, opened 2025-11-28
Add support for the MSETEX command

StefanPalashev:
# Add support for the MSETEX command
## Summary
Redis 8.4 introduces the [MSETEX](https://redis.io/docs/latest/commands/msetex/) command. It is an extension to the [MSETNX](https://redis.io/docs/latest/commands/msetnx/) command, which allows multiple string keys to be set with an optional shared expiration in a single atomic operation.

## Command Specifications
**Description**:
Set multiple string keys with an optional shared expiration in a single atomic operation.

**Syntax**:
```
MSETEX numkeys key value [key value ...] 
[NX | XX] [EX seconds |  PX milliseconds | EXAT unix-time-seconds |  PXAT unix-time-milliseconds | KEEPTTL]
```
**Options**:

- Existence options:
**NX** -- Only set the keys and their expiration time if **none** of them already exist.
**XX** -- Only set the keys and their expiration time if **all** of them already exist.

- Expiration options:
**EX** _seconds_ -- Set expiration time in seconds.
**PX** _milliseconds_ -- Set expiration time in milliseconds.
**EXAT** _timestamp-seconds_ -- Set absolute Unix expiration time (seconds).
**PXAT** _timestamp-milliseconds_ -- Set absolute Unix expiration time (milliseconds).
**KEEPTTL** -- Retain the time to live associated with the key(s).

**Reply**:
Integer reply:
`0` – No keys were set.
`1` – All keys were successfully set.

## Testing
Unit tests have been added to verify the correct behavior of the command.

## Additional Changes:
Implemented the `ToRedisArgs` trait for some of the existing `enums` to reduce code duplication.

## References:
Official documentation of the command - [MSETEX](https://redis.io/docs/latest/commands/msetex/)

Relevant Pull Requests from the Redis server:
[Use fixed position keys parameter for MSETEX command](https://github.com/redis/redis/pull/14470)
[Add MSETEX command](https://github.com/redis/redis/pull/14434)

### DIFF
--- a/redis/src/cluster_handling/sync_connection/pipeline.rs
+++ b/redis/src/cluster_handling/sync_connection/pipeline.rs
@@ -22,7 +22,7 @@ fn is_illegal_cmd(cmd: &str) -> bool {
         "INFO" |
         "KEYS" |
         "LASTSAVE" |
-        "MGET" | "MOVE" | "MSET" | "MSETNX" |
+        "MGET" | "MOVE" | "MSET" | "MSETNX" | "MSETEX" |
         "PING" | "PUBLISH" |
         "RANDOMKEY" | "RENAME" | "RENAMENX" | "RPOPLPUSH" |
         "SAVE" | "SCAN" |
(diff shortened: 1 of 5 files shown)


StefanPalashev:
@nihohit, Do you think all common trait implementations (for example, `ToRedisArgs`) should be centralized in `redis/src/types.rs` rather than distributed across it and `redis/src/commands/mod.rs`?

nihohit:
IMO trait implementation for common /external types should be in types.rs, while internal types should be implemented in their relevant module. Personally, I think that commands/mod.rs is full enough, and there's room for more commands modules.

nihohit:
Thanks!

PR #1896, opened 2025-12-11
Add to the readme that this is a client

yesbox:
Adds to the readme that this library is a client for Redis/Valkey servers, which was not clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,12 @@
 [![crates.io](https://img.shields.io/crates/v/redis.svg)](https://crates.io/crates/redis)
 [![Chat](https://img.shields.io/discord/976380008299917365?logo=discord)](https://discord.gg/WHKcJK9AKP)
 
-Redis-rs is a high level Rust library for Redis, Valkey and any other RESP 
-(Redis Serialization Protocol) compliant DB. It provides convenient access
-to all Redis functionality through a very flexible but low-level API. It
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,12 +4,12 @@
 [![crates.io](https://img.shields.io/crates/v/redis.svg)](https://crates.io/crates/redis)
 [![Chat](https://img.shields.io/discord/976380008299917365?logo=discord)](https://discord.gg/WHKcJK9AKP)
 
-Redis-rs is a high level Rust library for Redis, Valkey and any other RESP 
-(Redis Serialization Protocol) compliant DB. It provides convenient access
-to all Redis functionality through a very flexible but low-level API. It
-uses a customizable type conversion trait so that any operation can return
-results in just the type you are expecting. This makes for a very pleasant
-development experience.
+Redis-rs is a client Rust library for Redis, Valkey and any other RESP (Redis
```
nihohit: `a high-level client`

(the high level terminology is used again later, and I want to stay consistent)

yesbox: OK. I restructured a bit, the sentence felt it a bit heavy otherwise when only adding. The other reason I removed/replaced "high-level" initially is that claim is followed up by describing "a flexible but low-level API". It's both high and low in the first paragraph, and I thought it unclear if this is intentional.

nihohit:
Thanks for the contribution!

PR #1898, opened 2025-12-19
MockRedisConnection with consumation assert

meierfra-ergon:
Hi
we use redis-test in our unit-tests, and found, that if not all commands, defined in a `MockRedisConnection`, are not fully consumed, there is no (easy) way to assert that (with public functions).

So I would suggest a public function to check if all commands have been consumed, to assert on. In my example `is_empty()`.

Even more comfortable, but a bit breaking, would be to assert automatically in the drop() method of  the MockRedisConnection itself. We also could make it configurable to be less breaking.

What do you think? It`s only a suggestion to start a discussion.

### DIFF
--- a/redis-test/src/lib.rs
+++ b/redis-test/src/lib.rs
@@ -177,8 +177,14 @@ impl MockRedisConnection {
     {
         MockRedisConnection {
             commands: Arc::new(Mutex::new(VecDeque::from_iter(commands))),
-            assert_is_empty_on_drop: true,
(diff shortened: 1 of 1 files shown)


nihohit:
Hi, this sounds reasonable, but let's make the default `false`, in order not to break existing users.

nihohit:
Thanks for the contribution!

nihohit:
released in the latest redis-test version.

Issue #1727, opened 2025-07-15
ConnectionManager with CLIENT SETNAME


felipou:
I was using CLIENT SETNAME right after getting a connection (and handling reconnection manually). I'm now switching to use the ConnectionManager (great feature BTW), but there is no way to issue commands automatically after a connection.

I was thinking about contributing a PR with one of the following features:

- Add client name to ConnectionManagerConfig and call CLIENT SETNAME after reconnection
- Add initialization pipeline to ConnectionManager to be called after reconnection
- Add client name to AsyncConnectionConfig and call CLIENT SETNAME on connection initialization (along with CLIENT SETINFO)
- Add initialization pipeline to AsyncConnectionConfig to be called on connection initialization (somewhere near CLIENT SETINFO)

Would any of these be desirable? Should I go ahead with implementing something like this?

nihohit:
> Add initialization pipeline to ConnectionManager to be called after reconnection
> Add initialization pipeline to AsyncConnectionConfig to be called on connection initialization

These sound like good ideas, thanks!

hepengju:
Has this feature been implemented? The corresponding configuration was not found in the latest version redis-1.0.2. 
(Note: I am developing a Redis desktop management client: [RedisME](https://www.hepengju.com), currently using a synchronous approach. When the connection fails, manual refresh of the connection is required, which does not provide a good user experience. Therefore, I plan to switch to an asynchronous connection: it can avoid locking and support automatic reconnection. During reconnection, I want to set the client name and modify the db to the one previously selected by the user.)

nihohit:
this hasn't been implemented. you're welcome to contribute such the change.

nihohit:
another implementation option is to add those to the [RedisConnectionInfo](https://github.com/redis-rs/redis-rs/blob/f46e4ead1dc383a256bdb55a1a8f7a65eeec75c5/redis/src/connection.rs#L286), and then the info will be available for each client type to implement.

hepengju:
Implementing this feature is a bit challenging for me, but the unmerged code above gives me some inspiration.
A simple implementation would be to add initialization pipeline commands to the ConnectionManager. I will try to contribute this feature.

somechris:
@hepengju Are you still working on this?

(Judging by the comments on the PR and as it's a month old, I assume this is free to grab?)

hepengju:
[...]

PR #1922, opened 2026-01-14
add init_pipeline feature to ConnectionManager

hepengju:
#1727

### DIFF
--- a/redis/src/client.rs
+++ b/redis/src/client.rs
@@ -4,15 +4,12 @@ use std::time::Duration;
 use crate::aio::{AsyncPushSender, DefaultAsyncDNSResolver};
 #[cfg(feature = "aio")]
 use crate::io::AsyncDNSResolver;
-use crate::{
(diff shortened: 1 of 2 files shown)


nihohit (commented):
it's missing tests.

redis/src/client.rs
```diff
@@ -259,6 +258,12 @@ impl AsyncConnectionConfig {
         self.set_push_sender_internal(std::sync::Arc::new(sender))
     }
 
+    /// Sets pipeline to be executed on connection creation (when use connection_manager)
+    pub fn set_init_pipeline(mut self, pipeline: Option<Pipeline>) -> Self {
```
nihohit, line 262: why is this set here? I don't see the multiplexed connection using it.

hepengju: Just considering saving init_pipeline for reuse during reconnection, perhaps it's more appropriate to place it inside struct Internals. Note: Not very proficient in Rust, I can't run the unit tests for this async manager, I'll think about how to optimize and improve it in future.

Issue #1933, opened 2026-01-28
Feature Request: Make MultiplexedConnection buffer size configurable


paz-raon:
## Summary                                                                                                                                                                                                     
                                                                                                                                                                                                                 
  Make the internal `BUFFER_SIZE` (currently hardcoded to 50) configurable via `AsyncConnectionConfig`.                                                                                                          
                                                                                                                                                                                                                 
  ## Problem                                                                                                                                                                                                     
                                                                                                                                                                                                                 
  `MultiplexedConnection` uses a hardcoded buffer size of 50 for its internal mpsc channel:                                                                                                                      
                                                                                                                                                                                                                 
  ```rust                                                                                                                                                                                                        
  // multiplexed_connection.rs                                                                                                                                                                                   
  const BUFFER_SIZE: usize = 50;                                                                                                                                                                                 
  let (sender, mut receiver) = mpsc::channel(BUFFER_SIZE);                                                                                                                                                       
```
  In high-throughput scenarios with many concurrent tokio::spawn tasks, this buffer fills up quickly and causes backpressure.                                                                                    
                                                                                                                                                                                                                 
##  Benchmark                                                                                                                                                                                                      
                                                                                                                                                                                                                 
  Tested with concurrent MGET operations:                                                                                                                                                                        
```
  ┌──────────────────────────────────────────┬──────────────┐                                                                                                                                                    
  │              Configuration               │ Success Rate │                                                                                                                                                    
  ├──────────────────────────────────────────┼──────────────┤                                                                                                                                                    
  │ Single connection (buffer=50)            │ ~91%         │                                                                                                                                                    
  ├──────────────────────────────────────────┼──────────────┤                                                                                                                                                    
  │ 8-connection pool (effective buffer=400) │ ~98%         │                                                                                                                                                    
  └──────────────────────────────────────────┴──────────────┘                                                                                                                                                    
```
  Using a pool improves throughput, but it defeats the purpose of MultiplexedConnection being "cheaply cloneable" and pooling-unnecessary.                                                                       
                                                                                                                                                                                                                 
##  Proposed Solution                                                                                                                                                                                              
                                                                                                                                                                                                                 
  Add set_buffer_size() to AsyncConnectionConfig:                                                                                                                                                                
```
  let config = AsyncConnectionConfig::new().set_buffer_size(200);
```
  Default remains 50 for backward compatibility.                                                                                                                                                                 
                                                                                                                                                                                                                 
[...]

PR #1935, opened 2026-01-28
Make MultiplexedConnection buffer size configurable #1933

paz-raon:
## Summary

Add `set_pipeline_buffer_size()` to `AsyncConnectionConfig` and `ConnectionManagerConfig`.

The internal pipeline buffer size was hardcoded to 50. This allows users to customize it.

## Usage

```rust
let config = ConnectionManagerConfig::new()
    .set_pipeline_buffer_size(100);
```

Fully backward compatible - default remains 50.

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -44,6 +44,7 @@ pub struct ConnectionManagerConfig {
     resubscribe_automatically: bool,
     #[cfg(feature = "cache-aio")]
     pub(crate) cache_config: Option<crate::caching::CacheConfig>,
+    pipeline_buffer_size: Option<usize>,
 }
 
 impl std::fmt::Debug for ConnectionManagerConfig {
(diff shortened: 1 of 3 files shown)


redis/src/aio/connection_manager.rs
```diff
@@ -224,6 +227,12 @@ impl ConnectionManagerConfig {
             ..self
         }
     }
+
+    /// Sets the buffer size for the internal pipeline channel.
```
nihohit, line 231: please explain what this means, so users could make an informed decision without reading source code.

redis/tests/test_async.rs
```diff
@@ -1900,4 +1900,54 @@ mod basic_async {
 
         Ok(())
     }
+
+    #[cfg(feature = "tokio-comp")]
```
nihohit: please remove these tests - they don't test anything meaningful. If you want to to write integration tests for this, you need to check that the backpressure changes according to this setting.

redis/src/aio/connection_manager.rs
paz-raon: 👌

redis/tests/test_async.rs
paz-raon: 👌

nihohit:
Thanks for your contribution!

Issue #1950, opened 2026-02-06
Sentinel connection with non-zero DB breaks due to attempt to SELECT


palfrey:
Run the following code:
```rust
use redis::sentinel::{SentinelClient, SentinelServerType};

fn do_redis_code(url: &str) -> redis::RedisResult<()> {
    let mut client = SentinelClient::build(vec![url], "master".into(), None, SentinelServerType::Master)?;
    client.get_connection()?;
    Ok(())
}

fn main() {
    let url = "redis://127.0.0.1:26379/1";
    do_redis_code(url).unwrap();
}
```
gets you
```
thread 'main' (751047) panicked at redis/examples/sentinel.rs:11:24:
called `Result::unwrap()` on an `Err` value: Redis server refused to switch database - Server(ResponseError): unknown command 'SELECT', with args beginning with: '1' 
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I've worked around this in my code by explicitly setting the DB to zero on connection, and then doing a manual `SELECT` directly to the actual redis node afterwards.

<details>
<summary>docker compose config I've used</summary>

```yaml
services:
  redis:
    image: redis:8.4-alpine3.22
    ports:
      - 6379:6379
    command: redis-server --loglevel debug

  sentinel:
    image: redis:8.4-alpine3.22
    depends_on:
      - redis
    ports:
      - 26379:26379
    command: >
      sh -c 'echo "sentinel resolve-hostnames yes" > /etc/sentinel.conf &&
            echo "sentinel monitor master redis 6379 2" >> /etc/sentinel.conf &&
            echo "sentinel down-after-milliseconds master 1000" >> /etc/sentinel.conf &&
            echo "sentinel failover-timeout master 5000" >> /etc/sentinel.conf &&
            echo "sentinel parallel-syncs master 1" >> /etc/sentinel.conf &&
            redis-server /etc/sentinel.conf --sentinel'
```
</details>

nihohit:
What exactly are you reporting here?
You've constructed a URL that asks for DB 1, sent it to a node that doesn't support switching DBs, and received an error. What's wrong with this flow?

palfrey:
[...]

PR #1952, opened 2026-02-08
Warn if someone tries to set DB on a sentinel connection

palfrey:
https://github.com/redis-rs/redis-rs/issues/1950 was primarily caused by my confusion about the interface, with DB-using URLs being parsed fine by `SentinelClient`, but them later causing issues. This PR spits out a warning about that.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -289,7 +289,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
+      - name: Install OpenSSL
+        shell: powershell
+        run: |
+          vcpkg install openssl:x64-windows
+          echo "OPENSSL_DIR=C:\vcpkg\packages\openssl_x64-windows" >> $env:GITHUB_ENV
+          echo "OPENSSL_INCLUDE_DIR=C:\vcpkg\packages\openssl_x64-windows\include" >> $env:GITHUB_ENV
+          echo "OPENSSL_LIB_DIR=C:\vcpkg\packages\openssl_x64-windows\lib" >> $env:GITHUB_ENV
+
       - name: lint
         uses: ./.github/actions/lint-and-check
         with:
-          rust-toolchain: stable    
+          rust-toolchain: stable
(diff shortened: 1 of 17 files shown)


nihohit:
please also fix the docs in the sentinel file - [this ](https://github.com/redis-rs/redis-rs/blob/92d7d3456fef0deab0ebee425474337d7034f6cb/redis/src/sentinel.rs#L55)looks like the same issue you had.

palfrey:
> please also fix the docs in the sentinel file - [this ](https://github.com/redis-rs/redis-rs/blob/92d7d3456fef0deab0ebee425474337d7034f6cb/redis/src/sentinel.rs#L55)looks like the same issue you had.

I've clarified very slightly the docs there. I think it possibly needs a clearer warning there about the DB case, but not quite sure where to fit in in the existing structure.

redis/src/sentinel.rs
```diff
@@ -700,6 +702,16 @@ impl Sentinel {
             .map(|p| p.into_connection_info())
             .collect::<RedisResult<Vec<ConnectionInfo>>>()?;
 
+        #[cfg(feature = "log")]
```
nihohit: I don't think that this is necessary, since an error will be returned by the DB. If anything, it's better to return a client error immediately than to log a warning.

palfrey: This PR does both. So, yeah the client error is good, but I'd regard it as confusing. This log's explicit purpose is clarifying _why_ the client error has occurred to make things easier to figure out when fixing stuff.

nihohit: at the cost of another feature, which increases the complexity of finding problematic feature combinations.
If you're sure that logging is needed, I'd prefer that the sentinel feature take a dependency on `log`, and that relevant logs be added throughout the code. As this change happens, I doubt that users will know to enable the new feature, only to receive one new warning.

palfrey: I've folded the log feature into sentinel.

PR #1956, opened 2026-02-13
Prepare a version

nihohit:


nihohit:
@StefanPalashev, I found that the token based change breaks semver by removing the UnwindSafe auto-trait from `RedisConnectionInfo`, so I'm reverting that change. I think that's the simplest way to fix this is to add `+ UnwindSafe` to all of the dynamic trait declarations, but AFAIS this also requires a change in the entra-id crate.

StefanPalashev:
@nihohit It seems that the `StreamingCredentialsProvider` has to be marked with `Send + Sync + UnwindSafe + RefUnwindSafe`.

The implementations of these traits could look like:
``` rust
impl std::panic::UnwindSafe for EntraIdCredentialsProvider {}
impl std::panic::RefUnwindSafe for EntraIdCredentialsProvider {}
```

The manual implementation is needed because of the `credential_provider` inside, which is of type `Arc<dyn TokenCredential + Send + Sync>.`

Then this test can be used to verify that the type is indeed compliant:
``` rust
  #[test]
    fn test_redis_connection_info_is_unwind_safe() {
        use std::panic::{RefUnwindSafe, UnwindSafe};

        const fn assert_unwind_safe<T: UnwindSafe + RefUnwindSafe>() {}

        assert_unwind_safe::<RedisConnectionInfo>();
    }
```

nihohit:
> The manual implementation is needed because of the credential_provider inside, which is of type Arc<dyn TokenCredential + Send + Sync>.

The question is why this is so. If the internal provider is actually unwind unsafe, I want to understand why its container isn't unsafe, not just arbitrarilly mark it as safe. If it is unwind safe and just not marked as such, maybe you can upstream this change to the entra crate.

StefanPalashev:
### The `TokenCredential` trait cannot be marked with these traits (`UnwindSafe` + `RefUnwindSafe`).

It is used internally (in `azure_core`) in a manner similar to how the `EntraIdCredentialsProvider` uses it (`Arc<dyn TokenCredential>`).

These are some key factors that led to that conclusion:

The `azure_core` crate, where the `TokenCredential` trait resides, uses the `async-lock` crate as a dependency.
`async_lock::RwLock` is **NOT** `RefUnwindSafe` as it contains an `UnsafeCell`.

Here's more information on it: https://doc.rust-lang.org/core/panic/trait.RefUnwindSafe.html

This affects the `TokenCache`, which uses a `RwLock<HashMap<...>>` and almost all of the credentials types use the `TokenCache`. In addition, most of those types also use either `dyn Policy` or `dyn Executor` traits that are both _marked_ as:
`Send + Sync + fmt::Debug`

nihohit:
I see from the docs that while it isn't RefUnwindSafe, [it is UnwindSafe](https://docs.rs/async-lock/latest/async_lock/struct.RwLock.html#impl-UnwindSafe-for-RwLock%3CT%3E), and the breakage was about `UnwindSafe`, not about `RefUnwindSafe`, so this change might be sufficient.

StefanPalashev:
Isn't the breakage about both?
```
Failed in:
  type RedisConnectionInfo is no longer UnwindSafe, in /home/runner/work/redis-rs/redis-rs/redis/src/connection.rs:288
  type RedisConnectionInfo is no longer RefUnwindSafe, in /home/runner/work/redis-rs/redis-rs/redis/src/connection.rs:288
```

nihohit:
You're right, my mistake. In this case, we'll need to find some other ergonomic way to pass the credentials provider to the connection, which doesn't break backwards compatibility. Maybe as another parameter to the connection creation process.

Issue #1958, opened 2026-02-17
[Feature Request] Zone-Aware Read Affinity for Cluster Mode


danielamiao:
### Description
Currently, cluster routing is topology-agnostic, which leads to unpredictable cross-zone data transfer costs and increased tail latency in multi-zone (availability zone) cloud deployments like Amazon ElastiCache or GCP Memorystore. We propose an "Affinity Tag" system that allows the client to prioritize reads from nodes that share a common metadata identifier (e.g., Affinity Tag) with the client.

### Proposed Implementation
To remain cloud-agnostic, the library should not implement cloud-specific APIs. Instead, we suggest extending ClusterConfig to accept a user-provided mapping:

1. Metadata Injection: Users provide a HashMap<String, String> mapping Node Hostnames to Affinity Tags (e.g., node-001.cache. -> us-west-2a).
2. Client Identity: The client is initialized with its own local_affinity_tag.
3. Hostname Priority: The discovery logic must prioritize the hostname field from CLUSTER SLOTS over IP addresses to ensure the mapping remains resilient during node replacements or IP churn.
4. Routing Logic: When `read_from_replicas` is enabled, the client should attempt to match its local_affinity_tag against the discovered replicas' tags. If a match is found, that connection is prioritized; otherwise, it falls back to default behavior.

### Use Cases

1. Cost Optimization: Eliminating the cross-zone data transfer charges on cloud providers.
2. Latency Reduction: Avoiding the tail latency penalty incurred by crossing zone boundaries.

This is an initial proposal to get feedback, if the feature is reasonable, we are happy to contribute to the design and implementation.

nihohit:
AZ awareness is a great feature, and I'm sure lots of people would love to have it. Thanks for suggesting this :)
Question - would this be preferable to latency tracking, and prefering replicas based on some latency metric? That would account both for different zones and load. 

> Metadata Injection

I'm not sure that this mechanism is robust. What happens when the cluster scales up, and new nodes are missing from the map? Are there guarantees that nodes will remain in the same AZ after failovers or restarts?

> The discovery logic must prioritize the hostname field from CLUSTER SLOTS over IP addresses to ensure the mapping remains resilient during node replacements or IP churn.

Can you explain this?

> the client should attempt to match its local_affinity_tag against the discovered replicas' tags.

This assumes that all AZs that aren't the same AZ are equidistant, which I believe isn't true.

[...]

PR #1961, opened 2026-02-18
Expose control over cluster read routing strategy

PDXKimani:
Addresses #1958.

Allows users to control how reads will be routed in cluster connections. This is done via two new traits:
- `ReadRoutingStrategy`: These strategies are instantiated per-connection and have three methods to implement:
  - `choose_any` is responsible for deciding which node a read should be routed to when it is valid to choose a primary or any replica.
  - `choose_replica` is responsible for deciding which node a read should be routed to when it is only valid to choose a replica. (This is currently only invoked when routing is manually set to ReplicaRequired).
  - `on_topology_changed` is invoked when the connection discovers a new SlotMap (e.g. in response to a MOVED). This allows strategies to keep track of the cluster topology.
- `ReadRoutingStrategyFactory`: This is set at the ClusterClient level and is responsible for instantiating `ReadRoutingStrategy`s when connections are created via `create_strategy`. A blanket impl is provided for any `ReadRoutingStrategy` which implements Clone, for convenience.

The preexisting  `read_from_replicas` configuration now installs a simple `RandomReplica` strategy to the client, which preserves the original behavior (choosing a random replica for reads, or falling back to the primary if no replicas are available.)

An example has been added showing how latency-aware routing could be achieved with this system.

I've also fixed an adjacent extant bug where `READONLY` would not be set on replica connections unless `read_from_replicas` was true (which would now check if a strategy was provided). It is possible due to ReplicaRequired that reads may be sent to replicas even if that is not the case, so we should just always attempt to set READONLY.

### DIFF
--- a/redis/tests/support/mock_cluster.rs
+++ b/redis/tests/support/mock_cluster.rs
@@ -105,8 +105,12 @@ pub fn contains_slice(xs: &[u8], ys: &[u8]) -> bool {
     false
 }
 
+pub fn is_connection_check(cmd: &[u8]) -> bool {
+    contains_slice(cmd, b"READONLY") || contains_slice(cmd, b"PING")
+}
+
 pub fn respond_startup(name: &str, cmd: &[u8]) -> Result<(), RedisResult<Value>> {
-    if contains_slice(cmd, b"PING") {
+    if is_connection_check(cmd) {
         Err(Ok(Value::SimpleString("OK".into())))
     } else if contains_slice(cmd, b"CLUSTER") && contains_slice(cmd, b"SLOTS") {
         Err(Ok(Value::Array(vec![Value::Array(vec![
@@ -117,8 +121,6 @@ pub fn respond_startup(name: &str, cmd: &[u8]) -> Result<(), RedisResult<Value>>
                 Value::Int(6379),
             ]),
         ])])))
-    } else if contains_slice(cmd, b"READONLY") {
-        Err(Ok(Value::SimpleString("OK".into())))
     } else {
         Ok(())
     }
@@ -171,7 +173,7 @@ pub fn respond_startup_with_replica_using_config(
             slot_range: (8192..16383),
         },
     ]);
-    if contains_slice(cmd, b"PING") {
+    if is_connection_check(cmd) {
         Err(Ok(Value::SimpleString("OK".into())))
     } else if contains_slice(cmd, b"CLUSTER") && contains_slice(cmd, b"SLOTS") {
         let slots = slots_config
@@ -199,8 +201,6 @@ pub fn respond_startup_with_replica_using_config(
             })
             .collect();
         Err(Ok(Value::Array(slots)))
-    } else if contains_slice(cmd, b"READONLY") {
-        Err(Ok(Value::SimpleString("OK".into())))
     } else {
         Ok(())
     }
--- a/redis/tests/test_cluster.rs
+++ b/redis/tests/test_cluster.rs
@@ -363,7 +363,7 @@ mod cluster {
             "test_cluster_can_connect_to_server_that_sends_cluster_slots_with_null_host_name";
 
         let MockEnv { mut connection, .. } = MockEnv::new(name, move |cmd: &[u8], _| {
-            if contains_slice(cmd, b"PING") {
+            if is_connection_check(cmd) {
                 Err(Ok(Value::SimpleString("OK".into())))
             } else if contains_slice(cmd, b"CLUSTER") && contains_slice(cmd, b"SLOTS") {
                 Err(Ok(Value::Array(vec![Value::Array(vec![
@@ -387,7 +387,7 @@ mod cluster {
         let name = "test_cluster_can_connect_to_server_that_sends_cluster_slots_with_partial_nodes_with_unknown_host_name";
 
         let MockEnv { mut connection, .. } = MockEnv::new(name, move |cmd: &[u8], _| {
-            if contains_slice(cmd, b"PING") {
+            if is_connection_check(cmd) {
                 Err(Ok(Value::SimpleString("OK".into())))
             } else if contains_slice(cmd, b"CLUSTER") && contains_slice(cmd, b"SLOTS") {
                 Err(Ok(Value::Array(vec![
@@ -492,7 +492,7 @@ mod cluster {
             }
             started.store(true, atomic::Ordering::SeqCst);
 
-            if contains_slice(cmd, b"PING") {
+            if is_connection_check(cmd) {
                 return Err(Ok(Value::SimpleString("OK".into())));
             }
 
@@ -596,7 +596,7 @@ mod cluster {
             }
             started.store(true, atomic::Ordering::SeqCst);
 
-            if contains_slice(cmd, b"PING") {
+            if is_connection_check(cmd) {
                 return Err(Ok(Value::SimpleString("OK".into())));
             }
 
--- a/redis/tests/test_cluster_async.rs
+++ b/redis/tests/test_cluster_async.rs
@@ -689,7 +689,7 @@ mod cluster_async {
             async_connection: mut connection,
             ..
         } = MockEnv::new(name, move |cmd: &[u8], _| {
-            if contains_slice(cmd, b"PING") {
+            if is_connection_check(cmd) {
                 Err(Ok(Value::SimpleString("OK".into())))
             } else if contains_slice(cmd, b"CLUSTER") && contains_slice(cmd, b"SLOTS") {
                 Err(Ok(Value::Array(vec![Value::Array(vec![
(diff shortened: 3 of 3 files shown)


nihohit:
do you intend to provide the other implementations in a later PR? I'd like to be sure that the design works for complex strategies. For example, a latency based implementation might require usage of the cluster connection in order to measure latency.

PDXKimani:
> do you intend to provide the other implementations in a later PR? I'd like to be sure that the design works for complex strategies. For example, a latency based implementation might require usage of the cluster connection in order to measure latency.

@nihohit I'd implemented an az-aware thing for elasticache locally that has a few too many dependencies to commit to this crate. I'm looking at adding a latency-aware test case. 

Giving access to the actual client/connection is somewhat nontrivial, because you end up with circular references - the client/connection depend on the strategy, so the strategy probably shouldn't depend on the client/connection. For strategies sourcing info externally, this shouldn't matter, I think. For a latency-aware strategy, I think it might be necessary for the strategy to make more direct connections using the known node addresses, rather than a circular dependency through the cluster client/connection.

Thinking about additional hooks, I'm looking at adding a "topology updated" hook for this trait to inform the strategy of discovered changes to cluster toplogy.

I've also been thinking that the strategy instance should probably be per-connection rather than per-client - the client is really just a container for configuration, as far as I can tell. Each connection is maintaining its own slot map, so it might make sense for each connection to maintain its own routing strategy as well. The strategy could choose to share information across connections via shared memory (e.g. behind an Arc), of course.

redis/src/cluster_handling/sync_connection/mod.rs
```diff
@@ -518,7 +515,7 @@ where
         let info = get_connection_info(node, &self.cluster_params)?;
 
         let mut conn = C::connect(info, Some(self.cluster_params.connection_timeout))?;
-        if self.cluster_params.read_from_replicas {
+        if self.cluster_params.read_routing_strategy.is_some() {
             // If READONLY is sent to primary nodes, it will have no effect
             cmd("READONLY").exec(&mut conn)?;
         }
```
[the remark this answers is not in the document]
PDXKimani: I think this is an extant bug with `ReplicaRequired`, but I can fix it while I'm here.

nihohit:
> Giving access to the actual client/connection is somewhat nontrivial, because you end up with circular references

You can use weak pointers to overcome that. Just clone + upgrade before querying for latencies.

PDXKimani:
> > Giving access to the actual client/connection is somewhat nontrivial, because you end up with circular references
> 
> You can use weak pointers to overcome that. Just clone + upgrade before querying for latencies.

I think I'm more concerned about the logical circular dependency on strategy. If you try to write a strategy that uses a cluster connection which depends on the strategy for routing, it's probably pretty easy to unintentionally write a strategy that deadlocks/never returns due to logical dependency on itself. You could give out connections with the strategy overridden to None, I suppose... but I'm unsure what the direct utility of that connection really is?

As a strategy implementer you're provided with knowledge of the topology of the cluster as it's discovered now, which you can use as I do in the example latency-aware routing to directly communicate with nodes via single-node connections. You could use that approach to interrogate the cluster nodes about anything you think would be useful for routing, I think?

The main patterns I envision people using this for are:
- Simple routing policies like "first replica" or " random replica" or "round robin", that are entirely internal logic.
- Latency/measurement-based routing like the included example that use direct node connections to monitor the cluster.
- External-information routing populated from cloud provider APIs or static configurations or otherwise.
- Perhaps something with slot-based routing, which I've tried to account for.

danielamiao:
@nihohit any feedback on @PDXKimani 's latest response above?

nihohit:
@danielamiao @PDXKimani Sorry I wasn't available last week. This is a fairly large PR, and I'll try to start reviewing it soon.
Off the top of my head, I don't like that the latency aware routing strategy is a part of the examples, and practically untested. We can consider adding another crate to the repo with strategies, but for something like this, I'd like to think also about how to benchmark it - it contains implicit performance assumptions, and I'd like to be able to know that they're actually true, and that the strategy is actually useful, but I'm not sure how to do that, since all of the testing is done locally.

nihohit:
> I've also fixed an adjacent extant bug where READONLY would not be set on replica connections unless read_from_replicas was true (which would now check if a strategy was provided). It is possible due to ReplicaRequired that reads may be sent to replicas even if that is not the case, so we should just always attempt to set READONLY.

Let's fix this in a separate PR. It's a small fix, and it will probably be merged quickly.

redis/src/cluster_handling/slot_map.rs
```diff
@@ -122,30 +136,57 @@ impl SlotMap {
...
 
-    pub(crate) fn slot_addr(&self, slot_addr: &SlotAddr, read_from_replica: bool) -> &ArcStr {
+    pub(crate) fn slot_addr(
+        &self,
+        slot: u16,
+        slot_range_start: u16,
+        slot_range_end: u16,
+        slot_addr: &SlotAddr,
+        strategy: Option<&dyn ReadRoutingStrategy>,
+    ) -> &ArcStr {
```
nihohit, line 161: should we return here ArcStr and not NodeAddress? Or should we avoid the usage of `ArcStr` throughout the cluster codebase?

redis/src/cluster_handling/routing.rs
```diff
@@ -957,6 +959,308 @@ impl Slot {
...
+/// (e.g. `127.0.0.1`), an IPv6 address (e.g. `dead::cafe:beef`), or a hostname
+/// (e.g. `redis-node-1.example.com`). Use [`address`](NodeAddress::address) to get
+/// the full string, or [`host`](NodeAddress::host) and [`port`](NodeAddress::port)
+/// for the individual components.
+///
+/// This type is used in [`AnyNodeRoutingInfo`] and [`ReadRoutingStrategy`] to identify
+/// cluster nodes.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct NodeAddress {
+    pub(crate) addr: ArcStr,
```
nihohit, line 974: why save the address as a string, instead of saving host & port separate, and constantly splitting them?

redis/src/cluster_handling/routing.rs
```diff
@@ -957,6 +959,308 @@ impl Slot {
...
+/// This struct is passed to [`ReadRoutingStrategy::choose_any`] when the command can be
+/// served by either the primary or a replica. The strategy may return a reference to
+/// `primary` or any entry in `replicas`.
+pub struct AnyNodeRoutingInfo<'a> {
+    /// The hash slot number (0–16383) that the command targets.
+    pub slot: u16,
+    /// The first slot owned by this node group (inclusive).
+    pub slot_range_start: u16,
+    /// The last slot owned by this node group (inclusive).
+    pub slot_range_end: u16,
```
nihohit, line 1078: why would the stragey require these?

redis/src/cluster_handling/routing.rs
```diff
@@ -957,6 +959,308 @@ impl Slot {
...
+        self.inner.iter()
+    }
+}
+
+/// Routing information for a read command that may be sent to any node (primary or replica).
+///
+/// This struct is passed to [`ReadRoutingStrategy::choose_any`] when the command can be
+/// served by either the primary or a replica. The strategy may return a reference to
+/// `primary` or any entry in `replicas`.
+pub struct AnyNodeRoutingInfo<'a> {
```
nihohit, line 1072: I think that if you'll read the code outside of the context of this PR, it won't be immediately clear why do we need another routing abstraction, on top of `SingleNodeRoutingInfo`. This implies that maybe the strategy types need to be a part of a separate module/file, to clarify that they're not about external control of request routing, but only relevant to internal routing strategies.

redis/src/cluster_handling/routing.rs
```diff
@@ -957,6 +959,308 @@ impl Slot {
...
+    /// The returned reference must point to either `info.primary` or one of the entries
+    /// in `info.replicas`, if present.
+    fn choose_any<'a>(&self, info: &AnyNodeRoutingInfo<'a>) -> &'a NodeAddress;
+
+    /// Choose which replica to route a read command to, when a replica is required.
+    ///
+    /// The returned reference must point to one of the entries in `info.replicas`.
+    /// Replicas are guaranteed non-empty. If no replicas are available, the primary will be used,
+    /// and this method will not be called.
+    fn choose_replica<'a>(&self, info: &ReplicaRoutingInfo<'a>) -> &'a NodeAddress;
```
nihohit, line 1183: would a single function with a 
```rs
enum Requirement { // bikeshed the name :)
  Any,
  Replica
}
```
argument be betetr? I'm trying to imagine future extensions to the trait, and I wonder what's the best way to future proof it, and make implementing it better.

redis/src/cluster_handling/routing.rs
```diff
@@ -957,6 +959,308 @@ impl Slot {
...
+///
+/// For [`choose_any`](ReadRoutingStrategy::choose_any), a random replica is selected,
+/// falling back to the primary if no replicas are available.
+/// For [`choose_replica`](ReadRoutingStrategy::choose_replica), a random replica is selected.
+///
+/// This is the strategy installed by [`ClusterClientBuilder::read_from_replicas`].
+///
+/// [`ClusterClientBuilder::read_from_replicas`]: crate::cluster::ClusterClientBuilder::read_from_replicas
+#[derive(Debug, Clone, Default)]
+pub struct RandomReplica;
```
nihohit, line 1249: nit: RandomReplicaStrategy

PDXKimani:
Closing this PR in favor of #1985, which implements this functionality on top of the already-merged base PRs #1981 and #1982

PR #2015, opened 2026-03-24
Redis Search Support, Part I: Fundamentals

StefanPalashev:
# 🔍 Redis Search Support, Part I: Fundamentals

## Overview

This Pull Request implements a complete, type-safe API for the Redis [**FT.CREATE**](https://redis.io/docs/latest/commands/ft.create/) command, enabling users to create **_Redis Search_** indexes programmatically with compile-time safety and ergonomic builder patterns. It is the first in a series of changes aimed at introducing support for the entire module.

## Motivation

**_Redis Search_** is a powerful full-text search and vector similarity search engine for Redis. However, constructing **FT.CREATE** commands manually is error-prone due to:
- Complex syntax with many optional parameters
- Field-specific modifiers with different constraints
- Mutually exclusive options that can cause runtime errors
- Vector fields requiring precise attribute counting and ordering

This implementation provides a type-safe, idiomatic Rust API that prevents common mistakes at compile time and follows Rust's "make invalid states unrepresentable" principle.

## File Structure

```
redis/src/commands/
├── mod.rs                   # Registers the search module
├── search.rs                # Module re-exports for public API
└── query_engine/create/
    ├── command.rs           # FtCreateCommand builder
    ├── types.rs             # All schema types and builders
    └── tests.rs             # Unit tests

redis/tests/query_engine/
└── test_ft_create.rs        # Integration tests
```

All of the upcoming commands will follow the same directory structure:
- A separate folder inside the `query_engine` folder with files for:
  - the command's builder
  - the types needed for the command's implementation
  - unit tests

### Type Composition Hierarchy

The types follow a layered composition pattern that maximizes code reuse:

```
┌─────────────────────────────────────────────────────────────────────────┐
│                           FieldDefinition                               │
│  (Top-level enum that wraps all field types for schema insertion)       │
└─────────────────────────────────────────────────────────────────────────┘
                                 │
         ┌───────────────────────┼───────────────────────────┐
         │                       │                           │
         ▼                       ▼                           ▼
┌─────────────────┐     ┌─────────────────┐     ┌─────────────────────────┐
│ Simple Fields   │     │ GeoShape Field  │     │ Vector Fields           │
│  TEXT           │     │                 │     │  FLAT                   │
│  TAG            │     │                 │     │  HNSW                   │
│  NUMERIC        │     │                 │     │  VAMANA                 │
│  GEO            │     │                 │     │                         │
└────────┬────────┘     └────────┬────────┘     └────────────┬────────────┘
         │                       │                           │
         ▼                       ▼                           ▼
┌─────────────────┐     ┌─────────────────┐     ┌─────────────────────────┐
│SchemaCommonField│     │SchemaCommonField│     │   SchemaVectorField     │
│  sortable       │     │  only no_index! │     │     algorithm           |
│  no_index       │     │                 │     │     dim                 │
│                 │     │                 │     │     vector_type         │
│                 │     │                 │     │     distance_metric     │
└────────┬────────┘     └────────┬────────┘     └────────────┬────────────┘
         │                       │                           │
         └───────────────────────┴───────────────────────────┘
                                 │
                                 ▼
                        ┌─────────────────┐
                        │ BaseSchemaField │
                        │  field_type     │
                        │  alias          │
                        │  index_missing  │
                        └─────────────────┘
```

#### `BaseSchemaField` (Foundation)
Serves as a base for all field types.

```rust
pub struct BaseSchemaField {
    field_type: FieldType, // The Redis field type (TEXT, TAG, NUMERIC, GEO, VECTOR, GEOSHAPE)
    alias: Option<String>, // Optional AS clause for field aliasing
    index_missing: bool, // INDEXMISSING modifier
}
```

#### `SchemaCommonField` (Extended Base)
Extends `BaseSchemaField` with the `sortable` and `no_index` modifiers.

```rust
pub struct SchemaCommonField {
    base: BaseSchemaField,
    sortable: Option<Sortable>, // SORTABLE or SORTABLE UNF modifier
    no_index: bool, // NOINDEX modifier
}
```
Used by: **TEXT**, **TAG**, **NUMERIC**, **GEO** and **GEOSHAPE**.
Note: **GeoShape** fields don't support **SORTABLE**. Because of that the **SchemaGeoShapeField** does not provide an interface to **SORTABLE**.

#### `SchemaVectorField` (Vector Base)
Specialized base for vector fields containing the core vector parameters (_type_, _dimension_, _distance metric_)

```rust
pub struct SchemaVectorField {
    base: BaseSchemaField,
    algorithm: VectorAlgorithm
    vector_type: VectorType,
    dim: u32,
    distance_metric: DistanceMetric,
}
```

The implementation provides support for all available vector algorithms.
- **FLAT**: Brute-force exact search. Best for small datasets (< 1M vectors) where perfect accuracy is required.
- **HNSW**: Hierarchical Navigable Small World graph-based approximate search. Best for large datasets (> 1M vectors) where search performance and scalability are more important than perfect accuracy.
- **SVS-VAMANA**: Intel's Scalable Vector Search with graph-based approximate search and compression support.
Best when high performance with reduced memory usage is required, especially on Intel hardware.
More information at: https://intel.github.io/ScalableVectorSearch/intro.html
---

### Auxiliary Type - **FtCreateCommand**
A command builder that orchestrates index creation using the _Typestate pattern_ to enforce correctness at **compile time**.
It is a vital part of the tests suite, which allows command inspection.

Type States:
- `FtCreateCommand<WithoutSchema>` - Initial state, `into_cmd()` not available
- `FtCreateCommand<WithSchema>` - After calling `schema()`, `into_cmd()` available

Example modeled after one of the examples on the website:
```rust
    /*
    Create an index that stores the title, publication date, and categories of blog post hashes whose keys start with blog:post: (for example, blog:post:1).
    FT.CREATE idx ON HASH PREFIX 1 blog:post: SCHEMA title TEXT SORTABLE published_at NUMERIC SORTABLE category TAG SORTABLE
    */
    let schema = schema! {
        "title" =>  SchemaTextField::new().sortable(Sortable::Yes),
        "published_at" => SchemaNumericField::new().sortable(Sortable::Yes),
        "category" => SchemaTagField::new().sortable(Sortable::Yes),
    };
    let options = CreateOptions::new()
        .on(IndexDataType::Hash)
        .prefix("blog:post:");
        
    let ft_create = FtCreateCommand::new("idx").options(options).schema(schema);
```
In the example above, the `FtCreateCommand` would yield the same plain-text command as commented.

This command can either be executed by converting to a regular `redis::Cmd` by calling `into_cmd()` or it can be converted to a String representation (similar to how it is done in the tests) via the `into_args()` function.

**Key features:**
- Enforces non-empty schema at **compile time** via _Typestate pattern_
- `schema()` transitions **WithoutSchema** → **WithSchema**
- Converts to `redis::Cmd` via `into_cmd()` (only available in **WithSchema** state)
- Test helper `into_args()` for command inspection (only available in **WithSchema** state)

### Field Type Structs

| Type | Unique Modifiers | Common Modifiers |
|------|---------------------|-----------------------|
| `SchemaTextField`  | weight, no_stem, phonetic, with_suffix_trie, index_empty | alias, indexmissing,<br> sortable, no_index |
| `SchemaTagField`  | separator, case_sensitive, with_suffix_trie, index_empty | alias, indexmissing,<br> sortable, no_index |
| `SchemaNumericField`  | N/A | alias, indexmissing,<br> sortable, no_index |
| `SchemaGeoField`  | N/A | alias, indexmissing,<br> sortable, no_index |
| `SchemaGeoShapeField` | coord_system (SPHERICAL/FLAT) | alias, indexmissing,<br> no_index |

---

#### `VectorField` Enum
Wraps all of the Vector algorithms along with their specific options.

```rust
pub enum VectorField {
    Flat(SchemaVectorField, FlatVectorOptions),
    Hnsw(SchemaVectorField, HnswVectorOptions),
    Vamana(SchemaVectorField, VamanaVectorOptions),
}
```

Each variant contains:
1. `SchemaVectorField` - Common vector configuration
2. Algorithm-specific options struct

#### Algorithm-Specific Options

| Algorithm | Options Struct | Parameters | Common Modifiers |
|------------|------------------|-------------|-----------------------|
| FLAT | `FlatVectorOptions` | `block_size` | alias, indexmissing |
| HNSW | `HnswVectorOptions` | `m`, `ef_construction`, `ef_runtime`, `epsilon` | alias, indexmissing |
| VAMANA | `VamanaVectorOptions` | `compression`, `construction_window_size`, `graph_max_degree`, `search_window_size`, `epsilon`, `training_threshold`, `reduce` | alias, indexmissing |

Note:  The table lists all of the possible modifiers for the sake of completeness. However, common modifiers are applied at the `VectorField` level and are therefore not exposed through algorithm-specific option structs.

#### Vector Field Builders

Each algorithm has a dedicated builder.

```rust
// FLAT builder
VectorField::flat(VectorType::Float32, 128, DistanceMetric::Cosine)
    .block_size(1000)
    .build()

// HNSW builder
VectorField::hnsw(VectorType::Float32, 128, DistanceMetric::Cosine)
    .m(16)
    .ef_construction(200)
    .build()

// VAMANA builder (note: uses VamanaVectorType for type safety)
VectorField::vamana(VamanaVectorType::Float32, 128, DistanceMetric::Cosine)
    .compression(CompressionType::LVQ8)
    .graph_max_degree(64)
    .build()
```

---

### FieldDefinition and From Traits

`FieldDefinition` is the sum type that wraps all field types:

```rust
pub enum FieldDefinition {
    Text(SchemaTextField),
    Tag(SchemaTagField),
    Numeric(SchemaNumericField),
    Geo(SchemaGeoField),
    GeoShape(SchemaGeoShapeField),
    Vector(VectorField),
}
```

Each field type implements `From<T> for FieldDefinition`, which enables ergonomic schema construction with `.into()`.

---
### RediSearchSchema and schema! Macro
The `RediSearchSchema` is a **mandatory**, **non-empty** parameter for the **FT.CREATE** command. It defines the set of fields to be indexed and uses the _Typestate pattern_ to enforce schema validity at **compile time**.

#### How It Works

1. **Marker Types**: Zero-sized types (`Empty`, `NonEmpty`) that exist only at the type level.
They carry no runtime data and are erased during compilation, resulting in zero overhead.

2. **Generic State Parameter**: `RediSearchSchema<State>` is generic over its state.
The default is `Empty`, representing a newly created schema.

3. **State Transitions**: Methods consume `self` and return a new type:
- `RediSearchSchema<Empty>::insert()` → `RediSearchSchema<NonEmpty>`
- `RediSearchSchema<NonEmpty>::insert()` → `RediSearchSchema<NonEmpty>`

4. **Restricted APIs**: `ToRedisArgs` is only implemented for `NonEmpty`,
so empty schemas cannot be serialized or passed to commands.

#### Benefits

- **Compile-time guarantees**: Invalid states are caught by the compiler
- **Zero runtime cost**: No Option checks, no panics, no assertions
- **Self-documenting**: The type signature shows what's required

```rust
#[must_use = "Schema has no effect unless passed to a command"]
#[derive(Debug, Clone)]
pub struct RediSearchSchema<State = Empty> {
    fields: Vec<(String, FieldDefinition)>,
    _state: PhantomData<State>,
}
```

The `schema!` macro offers a concise syntax for defining schemas and guarantees at **compile time** that at least one field is specified. Empty schemas are not allowed, and invoking the macro with no fields (`schema! {}`) will result in a compile-time error.
Screenshot 2026-01-27 130448
This also holds true for manually created schemas:
Screenshot 2026-01-27 131055

#### Example use of the macro:
```rust
let schema = schema! {
    "title" => SchemaTextField::new().weight(2.0),
    "price" => SchemaNumericField::new().sortable(Sortable::Sortable),
    "embedding" => VectorField::hnsw(VectorType::Float32, 128, DistanceMetric::Cosine).m(16).build(),
};
```
---

## Serialization: **ToRedisArgs**

All types implement `ToRedisArgs` for Redis protocol serialization. Most types are serialized using a generic approach, with a few exceptions (notably **GeoShape** and **Vector** fields). In general, the implementation follows this pattern:

```rust
impl ToRedisArgs for Schema< FieldType(Text, Tag, Numeric, Geo, Vector, GeoShape) >Field {
    fn write_redis_args<W>(&self, out: &mut W) {
        self.common.base.write_redis_args(out);  // Writes: <TYPE> [AS alias] [INDEXMISSING]

        // Write: Field type specific modifiers

        self.common.write_redis_args(out);  // Writes: [SORTABLE] and/or [NOINDEX] if applicable.
    }
}
```

**GeoShapes** use a different serialization format, where the coordinate system must immediately follow the field type. This requirement prevents reuse of the common base implementation. The following code shows how a **GeoShape** is serialized:

```rust
        if let Some(alias) = &self.common.base.alias {
            out.write_arg(b"AS");
            alias.write_redis_args(out);
        }

        self.common.base.field_type.write_redis_args(out);
        self.coord_system.write_redis_args(out);

        if self.common.base.index_missing {
            out.write_arg(b"INDEXMISSING");
        }

        self.common.write_redis_args(out);
```

### Vector Field Attribute Counting

Vector fields require a count of attributes in the Redis protocol:

```
VECTOR <algorithm> <attributes_count> TYPE <type> DIM <dim> DISTANCE_METRIC <metric> [options...]
```
The implementation calculates this dynamically via the algorithm-specific options' `ToRedisArgs` implementation of `num_of_args()`:

```rust
impl ToRedisArgs for VectorField {
    fn write_redis_args<W>(&self, out: &mut W) {
        // Write field type and algorithm
        base.write_redis_args(out);

        // Calculate and write attribute count
        let attributes_count = match self {
            VectorField::Flat(base, opts) => base.num_of_args() + opts.num_of_args(),
            VectorField::Hnsw(base, opts) => base.num_of_args() + opts.num_of_args(),
            VectorField::Vamana(base, opts) => base.num_of_args() + opts.num_of_args(),
        };
        attributes_count.write_redis_args(out);

        // Write attributes...
    }
}
```

---

## VAMANA Parameter Details
https://redis.io/docs/latest/develop/ai/search-and-query/vectors/#svs-vamana-index

### Conditional Parameters
The builder adjusts values to conform to the constraints defined in the documentation. A warning is issued whenever an adjustment occurs.

```rust
    /// Number of vectors needed to learn compression parameters.
    /// Applicable only when used with COMPRESSION. Increase if recall is low.
    /// Note: setting this too high may slow down search. If a value is provided, it must be less than MAX_TRAINING_THRESHOLD (100 * DEFAULT_BLOCK_SIZE, where DEFAULT_BLOCK_SIZE is 1024).
    /// The default is 10 * DEFAULT_BLOCK_SIZE.
    pub fn training_threshold(mut self, training_threshold: u32) -> Self {
        if self.compression.is_some() {
            let clamped = std::cmp::min(training_threshold, MAX_TRAINING_THRESHOLD);
            if clamped != training_threshold {
                warn!(
                    "training_threshold exceeded the maximum allowed value; clamped from {} to {}.",
                    training_threshold, clamped
                );
            }
            self.training_threshold = Some(clamped);
        } else {
            warn!("training_threshold ignored: applies only when compression is enabled.");
        }
        self
    }

    /// The dimension used when using LeanVec4x8 or LeanVec8x8 compression for dimensionality reduction.
    /// If a value is provided, it should be less than DIM. Lowering it can speed up search and reduce memory use.
    /// The default is DIM / 2.
    pub fn reduce(mut self, reduce: u32) -> Self {
        if self
            .compression
            .is_some_and(|c| matches!(c, CompressionType::LeanVec4x8 | CompressionType::LeanVec8x8))
        {
            let max_reduce = self.base.dim.saturating_sub(1).max(1);
            let clamped = std::cmp::min(reduce, max_reduce).max(1);
            if clamped != reduce {
                warn!(
                    "reduce value {} out of valid range 1..={}; clamped to {}.",
                    reduce, max_reduce, clamped
                );
            }
            self.reduce = Some(clamped);
        } else {
            warn!("reduce ignored: applies only to LeanVec4x8 and LeanVec8x8 compression types.");
        }
        self
    }
```

---
### Unit Tests (`tests.rs`)
76 unit tests covering:
- Individual field types with all modifiers
- Vector fields with all algorithms and parameters
- `CreateOptions` with all combinations
- Edge cases (empty schema, zero dimensions, clamping of VAMANA vector parameters and etc.)
- Command serialization correctness

### Integration Tests (`test_ft_create.rs`)
Comprehensive integration tests against real Redis instances:

**Test Structure:**
1. **Individual modifiers** - Each parameter tested in isolation
2. **Progressive combinations** - Cumulative modifier application
3. **All combinations** - Full parameter matrix testing
4. **Mutually exclusive validation** - Error cases

**Coverage:**
- TEXT fields: 8 modifiers + 2 mutually exclusive (`indexmissing` and `no_index`)
Text

- TAG fields: 7 modifiers + 2 mutually exclusive (`indexmissing` and `no_index`)
Tag

- NUMERIC fields: 3 modifiers + 2 mutually exclusive (`indexmissing` and `no_index`)
Numeric

- GEO fields: 3 modifiers + 2 mutually exclusive (`indexmissing` and `no_index`)
Geo

- GEOSHAPE fields: 2 coordinate systems + 3 modifiers (alias + the two mutually exclusive `indexmissing` and `no_index`)
GeoShape

The following images show only a subset of the generated indices for brevity:
- FLAT vectors: 6 vector types × 3 distance metrics × 1 builder param × 2 field modifiers
Flat

- HNSW vectors: 6 vector types × 3 distance metrics × 4 builder params × 2 field modifiers
HNSW

- VAMANA vectors: 2 vector types × 3 distance metrics × 6 compressions × 6-7 builder params × 2 field modifiers
Vamana

Note: The statement above lists either 6 or 7 builder parameters, because reduce depends on the compression and is only supported by `LeanVec4x8` and `LeanVec8x8`. The image demonstrates that the reduce parameter is omitted when the compression mode is not supported.

### Test metrics
**Total test indices created:** **1166** unique index configurations
**Total time elapsed**: **< 1** sec

**Vector specific indices** + debug information (prints of the serialized command as seen in the images above):
| Algorithm | Number of indices | Time elapsed |
|------------|----------------------|----------------|
|FLAT | 144 | 0.16s |
|HNSW | 252 | 0.30s |
|VAMANA | 672 | 0.83s |

## Breaking Changes
None - this is a new feature addition.

## Other changes
The test version guard macro (`run_test_if_version_supported`) has been extended with a pattern that accepts modules, which allows the tests to be adapted to the bundled modules.

## Feature Gating
All Redis Search functionalities will be feature-gated with `#[cfg(feature = "search")]`.
```
[features]
search = ["dep:serde", "serde/derive", "dep:serde_json", "dep:log"]
```

## Future Work

**FT.SEARCH**, **FT.AGGREGATE** and **FT.DROPINDEX** command implementations.

## Documentation and relevant information:
https://redis.io/docs/latest/commands/ft.create/
https://redis.io/docs/latest/develop/ai/search-and-query/vectors/
https://redis.io/docs/latest/develop/ai/search-and-query/vectors/svs-compression/

### Additional resources:
https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/sorting/
https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/phonetic_matching/
https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/stemming/
https://intel.github.io/ScalableVectorSearch/intro.html


StefanPalashev:
Important notes:
The more observant may notice that the Pull Request is named "**Redis Search Support**", whereas the branch and the first commit use the term **Redis Query Engine**.

This is because the **Redis Query Engine**, also known as **RediSearch**, is being renamed to **Redis Search**.
This will be reflected in the upcoming Pull Requests.

.github/workflows/rust.yml
```diff
@@ -132,9 +138,53 @@ jobs:
             exit 1
           fi
 
-      - name: Run module-specific tests
+      - name: Verify RediSearch module
```
nihohit: this mostly repeats the JSON module logic, right? can it be moved to a reusable action?

redis/src/commands/query_engine/create/command.rs
```diff
@@ -0,0 +1,131 @@
...
+///
+/// let cmd = FtCreateCommand::new("my_index")
+///     .options(CreateOptions::new().on(IndexDataType::Hash))
+///     .schema(schema! { "title" => SchemaTextField::new() })
+///     .into_cmd();
+/// ```
+pub struct FtCreateCommand<State = WithoutSchema> {
+    index: String,
+    options: CreateOptions,
+    schema: Option<RediSearchSchema<NonEmpty>>,
```
nihohit: instead of using an option, you can set the type of the field to be the schema. Then you won't need the `WithSchema` marker -  the State type will be RediSearchSchema<_>

nihohit:
I'm missing references to cluster support, and I also don't see cluster routing rules. I assume that this means that the command will always be routed to a single node, using `index` as the key for determining slot. This is probably not the expected behavior, but I couldn't find a reference to how this command should be forwarded in cluster nodes.

StefanPalashev:
Clients should **NOT** broadcast **FT.CREATE** and other DDL command to shards, this should be handled by the **DMC** proxy.

.github/workflows/rust.yml
StefanPalashev: The shared logic has been extracted into a reusable function.

nihohit:
What's the DMC proxy, where is it used, and how?

uglide:
@nihohit I think Stefan confused the DMC Proxy with the Search coordinator. So, starting with Redis 8.0, the client library sends FT.CREATE as a keyless command, and let the Search module relay it automatically to other Redis Cluster shards.

nihohit:
@uglide so there's no correct node to send these messages to?
is there documentation explaining this?

uglide:
@nihohit It's documented here https://redis.io/docs/latest/develop/reference/command-tips/#request_policy

Since the index name is not a key, FT.CREATE and similar commands are "keyless" and the following rule applies:

> This tip can help clients determine the shards to send the command in clustering mode. The default behavior a client should implement for commands without the request_policy tip is as follows:
> 1. **The command doesn't accept key name arguments: the client can execute the command on an arbitrary shard.**

In the past, many client libraries just took the shortcut there and treated the index name as a key to simplify the implementation, but since Redis 8.0, it was fixed, and now we have fixed it in other client libraries too.

nihohit:
I need the time to read the docs, but @StefanPalashev sounds like we need to add this to the cluster routing table.

StefanPalashev:
> I need the time to read the docs, but @StefanPalashev sounds like we need to add this to the cluster routing table.

Will do, thanks!

redis/tests/query_engine/test_ft_create.rs
```diff
@@ -0,0 +1,902 @@
...
+static GEOSHAPE_FIELD_NAME: &str = "area";
+
+#[test]
+fn test_ft_create_with_an_empty_index_name() {
+    let ctx = run_test_if_version_supported!(&REDIS_VERSION_CE_8_0, &[Module::Search]);
+    let mut con = ctx.connection();
+    let schema = schema! {
+        TEXT_FIELD_NAME => SchemaTextField::new()
+    };
+    assert_eq!(
```
nihohit, line 44: what happens here on the second call? will it fail?

redis/tests/query_engine/test_ft_create.rs
```diff
@@ -0,0 +1,902 @@
...
+
+static TEXT_FIELD_NAME: &str = "title";
+static NUMERIC_FIELD_NAME: &str = "price";
+static TAG_FIELD_NAME: &str = "condition";
+static GEO_FIELD_NAME: &str = "location";
+static VECTOR_FIELD_NAME: &str = "embedding";
+static GEOSHAPE_FIELD_NAME: &str = "area";
+
+#[test]
+fn test_ft_create_with_an_empty_index_name() {
```
nihohit, line 32: should these tests also run against a cluster connection?

redis/tests/query_engine/test_ft_create.rs
```diff
@@ -0,0 +1,902 @@
...
+        let schema = schema! {
+            TEXT_FIELD_NAME =>  modifier(combined_schema_text_field.clone())
+        };
+        assert_eq!(
+            con.ft_create(&combined_index_name, &CreateOptions::new(), &schema),
+            Ok("OK".to_string())
+        );
+    }
+    // Test that mutually exclusive modifiers are mutually exclusive indeed
+    assert!(
```
nihohit: use assert_eq on the concrete returned error

redis/tests/query_engine/test_ft_create.rs
```diff
@@ -0,0 +1,902 @@
...
+        let schema = schema! {
+            TAG_FIELD_NAME =>  modifier(combined_schema_tag_field.clone())
+        };
+        assert_eq!(
+            con.ft_create(&combined_index_name, &CreateOptions::new(), &schema),
+            Ok("OK".to_string())
+        );
+    }
+    // Test that mutually exclusive modifiers are mutually exclusive indeed
+    assert!(
```
nihohit: the repeated structure of these tests should probably be encoded in some function.

nihohit (commented):
some comments on the tests

redis/src/commands/query_engine/create/command.rs
StefanPalashev: Thank you for the suggestion. The code has been updated accordingly. Please let me know whether this is what you had in mind.

redis/tests/query_engine/test_ft_create.rs
StefanPalashev: Yes, similarly to the test below. The code has been updated to explicitly make a second call and assert that it fails.

redis/tests/query_engine/test_ft_create.rs
StefanPalashev: Introduced a new auxiliary function (`assert_no_index_and_index_missing_exclusivity_for_field`) that does that.

redis/tests/query_engine/test_ft_create.rs
StefanPalashev: I agree that the tests share a similar overall structure. However, there are also a number of subtle differences between them that would make extracting a common helper function somewhat complex and potentially reduce readability.

redis/tests/query_engine/test_ft_create.rs
StefanPalashev: When a call to `con.ft_create(...)` is made on a `ClusterConnection`, the client looks up `FT.CREATE` in its routing table and resolves it to `RouteBy::Random`. In practice, this means the command is sent to exactly one primary selected from the current cluster topology. There is no fan-out, slot-based routing or aggregation at the client level. It is just a single RESP request to a single node, and the client returns that node’s response.

On the server side, the Search module handles the command in two phases:

1) Local creation
The receiving primary creates the index in its local memory.

2) Internal propagation to peer primaries
Before replying to the client, the module opens internal TCP connections to the other primaries and sends `_FT.CREATE` (note the underscore) commands to them. Those commands register the already-built index locally without triggering further propagation, so the process is strictly one-hop.

As a result, after a single successful `FT.CREATE`, all primaries already contain the index and subsequent `FT.CREATE` calls deterministically fail with `SEARCH_INDEX_EXISTS`, regardless of which primary receives the request.

Because of that behavior, the cluster case effectively mirrors the standalone semantics, so I'm not sure adding additional tests against a cluster connection would provide much value.

redis/tests/query_engine/test_ft_create.rs
StefanPalashev: It seems the error formats vary across different versions of the search module, which makes this testing approach unreliable. For now, keeping the assertion as-is is probably the most reasonable option.

@nihohit, what do you think?

redis/tests/query_engine/test_ft_create.rs
nihohit: > When a call to con.ft_create(...) is made on a ClusterConnection, the client looks up FT.CREATE in its routing table and resolves it to RouteBy::Random. In practice, this means the command is sent to exactly one primary selected from the current cluster topology. 

That sounds like an assumption that should be tested.

redis/tests/query_engine/test_ft_create.rs
nihohit: if we can't assert on this, we shouldn't. shame about the changed errors.

StefanPalashev:
@somechris, Hi, do you mind taking a look at the changes in the following files:
- .config/nextest.toml
- ‎.github/workflows/rust.yml
- ‎redis-test/src/server.rs
and the **Makefile**?

Should we incorporate a similar check for the bloom module in the `Verify Redis modules` step in `‎.github/workflows/rust.yml`?

.config/nextest.toml
```diff
@@ -12,16 +12,19 @@ slow-timeout = { period = "20s", terminate-after = 2 }
...
+default-filter = 'not (test(cluster) | binary(test_sentinel) | test(tls) | test(test_module) | binary(test_ft_create))'
 
 [profile.module_bloom]
 default-filter = 'binary(test_module_bloom*)'
 
 [profile.module_json]
 default-filter = 'test(test_module_json)'
+
+[profile.module_search]
+default-filter = 'binary(test_ft_create)'
```
somechris, line 30: I think this should work.

And I see that you have more things planned around search.

But what about (at least for now) renaming `query_engine/test_ft_create.rs` to something that fits the current layout (drop the extra dir, and use `test_module_$MODULE`)?

Maybe simply `test_module_search.rs`?

That name is a good fit for what this PR does, and it makes things simpler.
Not only here, but it'll allow you to drop the `path` in `redis/Cargo.toml`.

Then the test would fit in way better with the other integration tests.

If `test_module_search` every grew too big, we could split into `test_module_search_some_feature.rs` and `test_module_search_some_other_feature.rs` and still could reap the benefits of the current layout.

WDYT?

redis-test/src/server.rs
```diff
@@ -216,6 +217,13 @@ impl RedisServer {
 
                     redis_cmd.arg("--loadmodule").arg(path);
                 }
+                Module::Search => {
```
somechris, line 220: nit: This arm is essentially doing the same as the `Module::Bloom` part above.

But `bloom`s formulation is less nested and hence simpler in many metrics.

So I would re-use that formulation for `search`, which also makes things more locally consistent.

.github/workflows/rust.yml
```diff
@@ -110,46 +117,79 @@ jobs:
           rm ./Cargo.toml; mv ./Cargo.toml.actual ./Cargo.toml
           rm -rf ./__ci/redis-json
 
-      - name: Verify RedisJSON module
-        if: matrix.config.json-module-version != ''
+      - name: Verify Redis modules
+        if: matrix.config.json-module-version != '' || matrix.config.search-module-version != ''
```
somechris: I understand why you updated that part; and that's good.

And I think that checking that the module is functional was for sure great for debugging when adding initial support for modules in CI.

However, that part is working reliably right now.
I do not recall that part ever failing.

And if the module would not be functional, all tests using that module would fail ... that'd be more than a good enough signal that the module is broken :-)

However, such lengthy shell jobs make our CI code hard to read as fewer and fewer people can read shell scripts.

So, I see little need for that whole job right now.

Psssshhh! ... /me glances conspiratively left and right ... don't tell nihohit :-) ... but in #2170, I plan to remove that part for good.

So I'd not bother updating it and leave that part as it was.

That spares me time to review it. And it spares you time to think about questions like:

* Why are different ports used for different modules?
* Why did trailing `|| exit 1`s get added?
* ...

redis/src/commands/query_engine/create/tests.rs
```diff
@@ -0,0 +1,1198 @@
...
+ToRedisArgs::write_redis_args() serializes to Redis protocol
+      │
+      └── Vector fields: writes algorithm, attribute count, then attributes
+      └── GeoShape fields: writes field type and coordinate system, then modifiers
+      └── Other fields: writes field type, then modifiers
+      ▼
+FtCreateCommand collects all args and offers conversion to redis::Cmd
+      └── FtCreateCommand<NoSchema> ──>[options()]──>schema()──> FtCreateCommand<RediSearchSchema<NonEmpty>>
+      └── FtCreateCommand<RediSearchSchema<NonEmpty>> ──>[options()]──>into_cmd()──> Cmd
+*/
```
somechris, line 31: Does this comment really belong to `tests.rs`?

redis/src/commands/search.rs
```diff
@@ -0,0 +1,9 @@
+//! RediSearch module - provides search and indexing functionality.
+#[path = "query_engine/create/types.rs"]
+pub mod create_types;
```
somechris, line 3: I'd expect that search functionality lives somewhere around

```
redis/src/commands/search...
```

... like this file :+1:

But I find it very confusing that it needlessly shoehorns in files from a path that has no `search` in it, but uses `query_engine`.

I'd suggest to align files/dirs and to use `#[path = ...]` only when necessary. It breaks reader's expectations. :+1:

redis/src/commands/search.rs
```diff
@@ -0,0 +1,9 @@
...
+
+#[path = "query_engine/create/command.rs"]
+pub mod create;
```
somechris, line 6: .... and please make module and file names align. To find the code for the `create` crate in a file called `command.rs` is surprising 😀

redis/src/commands/query_engine/create/types.rs
```diff
@@ -0,0 +1,1919 @@
...
+        W: ?Sized + RedisWrite,
+    {
+        match self {
+            FieldDefinition::Text(tf) => tf.write_redis_args(out),
+            FieldDefinition::Numeric(nf) => nf.write_redis_args(out),
+            FieldDefinition::Geo(gf) => gf.write_redis_args(out),
+            FieldDefinition::Tag(tf) => tf.write_redis_args(out),
+            FieldDefinition::Vector(v) => v.write_redis_args(out),
+            FieldDefinition::GeoShape(gs) => gs.write_redis_args(out),
+            FieldDefinition::JustType(t) => t.write_redis_args(out),
```
somechris, line 1755: I for sure understand too little about this PR ... but is `JustType` really meant to be here?

redis/src/commands/query_engine/create/types.rs
```diff
@@ -0,0 +1,1919 @@
...
+/// let schema = schema! {
+///     "title" => SchemaTextField::new(),
+///     "price" => SchemaNumericField::new()
+/// };
+///
+/// // Using the builder pattern
+/// let schema = RediSearchSchema::new()
+///     .insert("title", SchemaTextField::new())
+///     .insert("price", SchemaNumericField::new());
+/// ```
```
somechris, line 1833: I don't fully get the value proposition for the `schema` macro that you explain here.

Is it to make the typestate pattern less annoying and still be able to assure that there is at least one field?

If so, couldn't we get rid of the typestate and the macro by requiring a field definition in schema builder's `new`?

Something like (ignoring proper type names, names for fields, ...)

```
pub struct SchemaBuilder {
    fields: Vec<Field>
}

impl SchemaBuilder {
    pub fn new(field: Field) -> Self {
        Self {
            fields: vec![field]
        }
    }

    pub fn field(self, field: Field) -> Self {
        self.fields.push(field);
        self
    }
}
```

redis/src/commands/query_engine/create/types.rs
```diff
@@ -0,0 +1,1919 @@
...
+#[macro_export]
+macro_rules! schema {
+    // Requires at least one field - empty invocation won't match
+    ($first_key:expr => $first_value:expr $(, $key:expr => $value:expr)* $(,)?) => {{
+        $crate::search::RediSearchSchema::new()
+            .insert($first_key, $first_value)
+            $(
+                .insert($key, $value)
+            )*
+    }};
```
somechris, line 1918: I think you can simplify that a lot by using `+` repetition instead of `*`.

somechris:
> Should we incorporate a similar check for the bloom module in the `Verify Redis modules` step in `‎.github/workflows/rust.yml`?

I'd hope that "no". I'd even remove the check for the search module :grin: ... see [my comments there](https://github.com/redis-rs/redis-rs/pull/2015/changes#r3660899705).

redis/tests/query_engine/test_ft_create.rs
StefanPalashev: The `Add integration tests for clustered deployments` commit adds tests that verify the behavior in clustered deployments.

redis/src/commands/query_engine/create/types.rs
StefanPalashev: Yes, it is. This allows users to create simple queries such as:
`FT.CREATE idx SCHEMA title TEXT price NUMERIC category TAG`

The code for the schema and the command would look like this:
```
let schema = RediSearchSchema::new()
    .insert("title", FieldType::Text)
    .insert("price", FieldType::Numeric)
    .insert("category", FieldType::Tag);

let ft_create = FtCreateCommand::new("idx").schema(schema);
```
The inserts will use `JustType(FieldType::Text)`,` JustType(FieldType::Numeric)` and `JustType(FieldType::Tag)`.

redis-test/src/server.rs
StefanPalashev: Thanks, applied.

.github/workflows/rust.yml
StefanPalashev: Thank you for taking the time to review that part and for the work that you are about to put in.
The fragment has been restored to its previous state.

redis/src/commands/query_engine/create/types.rs
somechris: Ah! That's meant to be syntactic sugar?

Then you could just `into` the `FieldType`s into `FieldDefintion` differently.

If you make that:

```rust
impl From<FieldType> for FieldDefinition {
    fn from(field_type: FieldType) -> Self {
        match field_type {
            FieldType::Text => { FieldDefinition::Text(SchemaTextField::new()) },
            ...
        }
    }
}
```

Then you can get rid of `JustType` and we don't have 2 paths for `TEXT`, 2 paths for `TAG`, 2 paths for `NUMERIC`, ... but only 1 path for each, once they are `FieldDefinition`s.

redis/src/commands/query_engine/create/command.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Set the schema for the command.
+    ///
+    /// The schema must be non-empty (contain at least one field).
+    /// This is enforced at compile time by the type system.
+    ///
+    /// This transitions the builder from `FtCreateCommand<NoSchema>` to
+    /// `FtCreateCommand<RediSearchSchema<NonEmpty>>`, making `into_cmd()` available.
+    pub fn schema(
+        self,
+        schema: RediSearchSchema<NonEmpty>,
```
somechris, line 82: Are we sold on the name `RediSearchSchema`?

To allow re-using the code for Valkey's search implementation, I'd prefer a more neutral name;

What about simply `SearchSchema` or some other more re-usable name?

StefanPalashev: > Important notes: The more observant may notice that the Pull Request is named "**Redis Search Support**", whereas the branch and the first commit use the term **Redis Query Engine**.
> 
> This is because the **Redis Query Engine**, also known as **RediSearch**, is being renamed to **Redis Search**. This will be reflected in the upcoming Pull Requests.

The `Schema` is also getting renamed, in fact to exactly your suggested name, but this is part of the 2nd pull request of the series.

PR #2018, opened 2026-03-26
fix(cluster): clamp retry backoff upper bound after min-wait floor

cuiweixie:
## Summary

Cluster retry jitter used `.min(max_wait).max(min_wait + 1)`, which can exceed `max_retry_wait` when `min_retry_wait + 1 > max_retry_wait`, contradicting [`retry_wait_formula`](https://github.com/redis-rs/redis-rs/blob/main/redis/src/cluster_handling/client.rs) (documented as `rand(min_wait_retry .. min(max_retry_wait, factor * exponent_base^retry))`).

## Changes

- Extract `cluster_retry_wait_upper_ms` with `.max(min_wait + 1).min(max_wait)` so the cap applies after the floor.
- Add regression test `cluster_retry_wait_upper_caps_max_when_min_plus_one_exceeds_max` (fails with the previous clamp order, passes with the fix).

## Verification

```text
# buggy: assert left 101, right 50
```

### DIFF
--- a/redis/src/cluster_handling/client.rs
+++ b/redis/src/cluster_handling/client.rs
@@ -97,8 +97,16 @@ impl Default for RetryParams {
 }
 
 /// Exclusive upper bound (ms) for [`RetryParams::wait_time_for_retry`] jitter; see [`ClusterClientBuilder::retry_wait_formula`].
+///
+/// The result is always strictly greater than `min_wait` so the half-open jitter range is never empty
+/// (e.g. when `max_retry_wait == min_retry_wait`, or when the clamp would otherwise equal `min_wait`).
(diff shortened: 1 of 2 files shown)


nihohit (approved):
Thanks for the fix!

nihohit:
@cuiweixie please fix the linting issue

nihohit:
the test is failing due to a valid issue - what happens when the min & max are equal?
please fix the implementation.

nihohit:
Wouldn't it make sense to just avoid the jitter if the min & max are equal?

nihohit:
@cuiweixie there's an open question, and the lints need fixing.

nihohit:
closed in favor of #2158

PR #2030, opened 2026-04-01
ValueType: Add missing possible types

somechris:
Adding missing types is backwards incompatible. In order to keep the
number of such backward incompatible changes low, as many types as
possible get added in one go:

* `None` - for keys without value (was `Unknown("none")` before).
* `VectorSet` - for bundled vectorset module.
* `BloomFilter`, `CuckooFilter`, `CountMin`, `TDigest`, `TopK` from
  Redis' `bloom` module.
* `TimeSeries` from Redis' `timeseries` module.
* `Trie` from Redis' `search` module

Redis' `search` also defines `scdtype00`, and `ft_invidx0` which are
ignored as they do not occur in the keyspace and `ft_invidx`,
`numericdx`, and `ft_tagidx` which are omitted as they are legacy
types.

ValKey's `json` module uses the same type as Redis', so it is covered
as well.

ValKey's `search` module's indices live outside the keyspace. So they
do not need coverage.

None of the added types are currently covered by tests, as they rely
on modules to create suitable values.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -2578,6 +2578,8 @@ pub enum ValueType {
     TopK,
     /// A time series. [Redis Docs](https://redis.io/docs/latest/develop/data-types/timeseries/)
     TimeSeries,
+    /// A Trie. [Redis Docs](https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/autocomplete/)
+    Trie,
     /// Any other value type not explicitly defined in [Redis Docs](https://redis.io/docs/latest/commands/type/)
     Unknown(String),
 }
@@ -2601,6 +2603,8 @@ impl<T: AsRef<str>> From<T> for ValueType {
             "MBbloomCF" => ValueType::CuckooFilter,
             "TDIS-TYPE" => ValueType::TDigest,
             "TopK-TYPE" => ValueType::TopK,
+            // Search module
+            "trietype0" => ValueType::Trie,
             // Timeseries module
             "TSDB-TYPE" => ValueType::TimeSeries,
             // Fallback
@@ -2628,6 +2632,8 @@ impl From<ValueType> for String {
             ValueType::TDigest => "TDIS-TYPE".to_string(),
             ValueType::TopK => "TopK-TYPE".to_string(),
             ValueType::CountMin => "CMSk-TYPE".to_string(),
+            // Search module
+            ValueType::Trie => "trietype0".to_string(),
             // Timeseries module
             ValueType::TimeSeries => "TSDB-TYPE".to_string(),
             // Fallback
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -2548,6 +2548,8 @@ pub(crate) type SyncPushSender = std::sync::mpsc::Sender<PushInfo>;
 #[derive(Debug, Clone, PartialEq)]
 #[non_exhaustive]
 pub enum ValueType {
+    /// Key does not have a value
+    None,
     /// Generally returned by anything that returns a single element. [Redis Docs](https://redis.io/docs/latest/develop/data-types/strings/)
     String,
     /// A list of String values. [Redis Docs](https://redis.io/docs/latest/develop/data-types/lists/)
@@ -2569,6 +2571,7 @@ pub enum ValueType {
 impl<T: AsRef<str>> From<T> for ValueType {
     fn from(s: T) -> Self {
         match s.as_ref() {
+            "none" => ValueType::None,
             "string" => ValueType::String,
             "list" => ValueType::List,
             "set" => ValueType::Set,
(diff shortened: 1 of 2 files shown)


redis/src/types.rs
```diff
@@ -2560,22 +2562,52 @@ pub enum ValueType {
     Hash,
     /// A Redis Stream. [Redis Docs](https://redis.io/docs/latest/develop/data-types/stream)
     Stream,
+    /// A vector set. [Redis Docs](https://redis.io/docs/latest/develop/data-types/vector-sets/)
+    VectorSet,
     /// A RedisJSON value. [Redis Docs](https://redis.io/docs/latest/develop/data-types/json/)
     JSON,
+    /// A Bloom filter. [Redis Docs](https://redis.io/docs/latest/develop/data-types/probabilistic/bloom-filter/)
+    BloomFilter,
```
somechris: This `ValueType::BloomFilter` is a bit problematic.

Both Redis' `bloom` module and ValKey's `bloom` module come with Bloom filter implementations. They are incompatible and use different strings for their type. Redis' `bloom` module uses `MBbloom--` while ValKey's uses `bloomfltr`.

Hence, when using a plain `ValueType::BloomFilter` variant as this PR does, the `Into<String>` has to decide whether it maps to Redis' or ValKey's variant. It cannot do both.

A simple alternative would be to have two `ValueType` Bloom filer variants. E.g.: `ValueType::RedisBloomFilter` for Redis' Bloom filters and `ValueType::ValKeyBloomFilter` for ValKey. But that's not so great from a usability perspective. And it's inconsistent with other module provided types (as `ValueType::JSON`) which lack a Redis/ValKey prefix.

Is there a good, future-proof way around this issue?

(I don't see a worry-free solution)

Or are there some general `redis-rs` guidelines what to do if implementations differ (e.g.: [ValKey's `HSETEX`](https://valkey.io/commands/hsetex/) allows `NX`/`XX`, [Redis' `HSETEX`](https://redis.io/docs/latest/commands/hsetex/) does not)?

nihohit: I don't see an alternative to RedisBloomFilter / ValkeyBloomFilter (although maybe name them BloomFilterRedis / BloomFilterValkey, for ease of use?) - these are 2 distinct values.
Maybe 
```rs
enum ServerType {
  Redis,
  Valkey
}

pub enum ValueType {
[...]
  BloomFilter(ServerType),
}
```
Is more representative, and will force people to consider which value they're looking for? I truly don't know.
I'd start by looking into other clients that solved this issue, and seeing what they did.

somechris: Checking the common libraries (see below the fold for details), it seems they either:

* act more low-level and return the raw string (which is not a good match for high-level interfaces), or
* return an enum that fails for anything outside the base types (which gives a sub-par user experience).

So there's nothing for us to follow.

I tried the `BloomFilter(ServerType)` appcoagd. But with ValKey being able to run Redis' modules and vice versa, the parameter is more a "module maintainer" than a `ServerType`. So it's not limited to Redis/ValKey, it might actually be _any_ maintainer... E.g.: `usmanm` has an alternative module with a t-Digest implementation. Maybe over time adding more and more such enum variants feels a bit awkward and might by more head-aches than it solves.

So I gave the `BloomFilterRedis`/`BloomFilterValKey` a shot. To me, that feels best for now.

It's not optimal. But it's dead simple. And it's free of additions that might maybe become a burden. And migrating to some other abstraction (if one is ever needed) is straight forward.

WDYT?

----
`valkey-glide`:
* Go: [`Type`](https://github.com/valkey-io/valkey-glide/blob/efd1992060654876d61301d6159a3357f34b55ed/go/base_client.go#L4515) returns the raw string (not an enum)
* Node: [`ObjectType`](https://github.com/valkey-io/valkey-glide/blob/efd1992060654876d61301d6159a3357f34b55ed/node/src/BaseClient.ts#L951) only contains base types, not types from modules.
* Java: [`type`](https://github.com/valkey-io/valkey-glide/blob/efd1992060654876d61301d6159a3357f34b55ed/java/client/src/main/java/glide/api/BaseClient.java#L4143) yield a plain string (for `String` and `GlideString` parameter variants)  (not an enum).
* Python: [`ObjectType`](https://github.com/valkey-io/valkey-glide/blob/efd1992060654876d61301d6159a3357f34b55ed/python/glide-shared/glide_shared/commands/command_args.py#L70) only contains base types, not types from modules.

`redis-py`: `BasicKeyCommands`' [`type`](https://github.com/redis/redis-py/blob/f83f8413af66f5b46ee306f635d5086a007ea2b3/redis/commands/core.py#L4495) returns the raw string (not an enum)
`go-redis`: [`Type`](https://github.com/redis/go-redis/blob/cd984fae99ba1374b05b3ac6a3eea2f6fd5310f0/generic_commands.go#L341) returns the raw result
`jedis`: `CommandObjects`' [`type`](https://github.com/redis/jedis/blob/7744cd4ecaf0a45b4b5df43f4ca3530582c01d81/src/main/java/redis/clients/jedis/CommandObjects.java#L131) yield a plain string (for `String` and `byte[]` parameter variants) (not an enum).
`node-redis`: [`TYPE`](https://github.com/redis/node-redis/blob/master/packages/client/lib/commands/TYPE.ts) yields a plain string (not an enum)

nihohit (approved):
Thanks!

PR #2042, opened 2026-04-08
json: Stop double-wrapping commands in `RedisResult`

somechris:
The `$body` parts yielded a `RedisResult<Cmd>` where a `Cmd` with a way
to deal with `?` was sufficient.

In all but the `impl Cmd` uses, the latter trivially holds. So the extra
wrapping in `RedisResult` was unnecessary and we drop it.

This makes JSON command implementations more similar to the other
commands, hence easier to read.


nihohit (approved):
thanks!

PR #2044, opened 2026-04-08
Add TLS certificate-based authentication support

StefanPalashev:
## Summary

This Pull Request implements support for Redis 8.6's new `TLS certificate-based authentication` feature, which allows automatic client authentication by mapping TLS certificate fields (like Common Name) to Redis ACL usernames.

## Motivation

Redis 8.6 introduced the `--tls-auth-clients-user <field>` configuration option that enables automatic user authentication based on TLS certificate fields. This eliminates the need for separate password-based authentication when using mTLS, providing a more seamless and secure authentication flow.

### Core Infrastructure (`redis-test/src/server.rs`)
- Added `cert_auth_field: Option<&str>` parameter to `RedisServer::new_with_addr_tls_modules_and_spawner`
- When `Some(field)` is provided, the server spawns with `--tls-auth-clients-user <field>` flag
- Defaults to "CN" (Common Name) but extensible to other fields like OIDs in the future

### Certificate Generation (`redis-test/src/utils.rs`)
- Created the `build_client_cert_with_custom_cn()` function, which generates client certificates with custom Common Names

### Test Utilities (`redis/tests/support/mod.rs`)
- Added `TestContext::new_with_cert_auth(tls_files)` - uses CN field by default
- Added `TestContext::new_with_cert_auth_field(tls_files, field)` - for custom certificate fields
- Updated certificate loading helpers to support separate client certificates

### Integration Tests (`redis/tests/test_tls_cert_auth.rs`)
- **Test 1**: Successful authentication when client cert CN matches Redis ACL username
- **Test 2**: Connection failure when no ACL user matches the certificate CN
- Both tests verify proper mTLS handshake and authentication behavior

## Relevant documentation:
- TLS: https://redis.io/docs/latest/operate/oss_and_stack/management/security/encryption/

### DIFF
--- a/redis-test/src/server.rs
+++ b/redis-test/src/server.rs
@@ -80,19 +80,24 @@ impl RedisServer {
         std::fs::read_to_string(self.log_file.clone()).ok()
     }
 
+    /// Get the default host to use for TCP connections.
+    pub fn get_default_host() -> String {
+        "127.0.0.1".to_string()
+    }
+
     pub fn get_addr(port: u16) -> ConnectionAddr {
         let server_type = ServerType::get_intended();
         match server_type {
             ServerType::Tcp { tls } => {
                 if tls {
                     redis::ConnectionAddr::TcpTls {
-                        host: "127.0.0.1".to_string(),
+                        host: Self::get_default_host(),
                         port,
                         insecure: true,
                         tls_params: None,
                     }
                 } else {
-                    redis::ConnectionAddr::Tcp("127.0.0.1".to_string(), port)
+                    redis::ConnectionAddr::Tcp(Self::get_default_host(), port)
                 }
             }
             ServerType::Unix => {
@@ -114,6 +119,7 @@ impl RedisServer {
             None,
             None,
             mtls_enabled,
+            None,
             modules,
             |cmd| {
                 cmd.spawn()
@@ -132,6 +138,7 @@ impl RedisServer {
             None,
             None,
             mtls_enabled,
+            None,
             modules,
             |cmd| {
                 cmd.spawn()
@@ -147,6 +154,7 @@ impl RedisServer {
         config_file: Option<&Path>,
         tls_paths: Option<TlsFilePaths>,
         mtls_enabled: bool,
+        cert_auth_field: Option<&str>,
         modules: &[Module],
         spawner: F,
     ) -> RedisServer {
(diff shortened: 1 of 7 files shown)


redis-test/src/server.rs
```diff
@@ -80,19 +80,24 @@ impl RedisServer {
         std::fs::read_to_string(self.log_file.clone()).ok()
     }
 
+    /// Get the default host to use for TCP connections.
+    pub fn get_default_host() -> String {
```
nihohit: nit: use a function outside of the `impl` scope.

redis/tests/support/mod.rs
```diff
@@ -399,6 +487,36 @@ fn get_version(conn: &mut impl redis::ConnectionLike) -> Version {
     parse_version(info)
 }
 
+/// Get the Redis server version by running `redis-server --version`.
+pub fn get_redis_binary_version() -> Version {
```
nihohit: How will you differentiate between a redis server and a valkey server?

redis/tests/test_tls_cert_auth.rs
```diff
@@ -0,0 +1,219 @@
...
+//! Redis 8.6 introduces automatic client authentication using TLS certificates.
+//! When configured with `tls-auth-clients-user CN`, Redis extracts the Common Name (CN)
+//! from the client's TLS certificate and uses it to authenticate the client as an ACL user.
+//!
+//! These tests verify:
+//! - Automatic authentication when CN matches an existing ACL user
+//! - Fallback to "default" user when CN doesn't match any ACL user
+//! - ACL permissions are correctly enforced based on the authenticated user
+
+#![cfg(feature = "tls-rustls")]
```
nihohit, line 12: don't you need to add the file in cargo.toml?

redis-test/src/server.rs
StefanPalashev: This resulted in the following change:

https://github.com/redis-rs/redis-rs/pull/2044/changes#diff-10169805b0709697d35a5b0eb336f195bbaced454d57c916ffd0269e1dc1ac6bR186

`RedisServer::get_default_host()` becomes `redis_test::server::get_default_host()`.

redis/tests/support/mod.rs
StefanPalashev: The function now returns an optional Version and is intended to work only with Redis instances. The macro has been updated accordingly, so these tests should be skipped when not running against Redis.

redis/tests/test_tls_cert_auth.rs
StefanPalashev: The file has been added to cargo.toml.

PR #2047, opened 2026-04-14
feat: add Slot type and functions to create Slot / Route with key

anatawa12:
Closes #2020

In addition, I deprecated existing `Route::new` function since it allows creating route with invalid slot number.

I've implemented both API proposal from the issue.

### DIFF
--- a/redis/src/cluster_handling/routing.rs
+++ b/redis/src/cluster_handling/routing.rs
@@ -944,10 +944,10 @@ pub struct Route(u16, SlotAddr);
 
 impl Route {
     /// Returns a new Route.
-    /// 
-    /// Since this function allows creating route for non-existing slot, 
+    ///
+    /// Since this function allows creating route for non-existing slot,
     /// it's recommended to use [`with_slot`] instead.
-    /// 
+    ///
     /// [`with_slot`]: Self::with_slot
     pub fn new(slot: u16, slot_addr: SlotAddr) -> Self {
         // TODO: Deprecate and remove this function in the future release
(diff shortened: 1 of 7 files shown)


redis/src/cluster_handling/routing.rs
```diff
@@ -958,10 +943,27 @@ pub struct Route(u16, SlotAddr);
...
 impl Route {
     /// Returns a new Route.
+    #[deprecated(
+        note = "This function is deprecated because this allows user to create Route with invalid slot number. Use with_slot instead."
+    )]
     pub fn new(slot: u16, slot_addr: SlotAddr) -> Self {
+        Self::new_unsafe(slot, slot_addr)
+    }
+
+    pub(crate) fn new_unsafe(slot: u16, slot_addr: SlotAddr) -> Self {
```
nihohit: please run the tests locally - this won't work, because the intergration tests can only use publicly available functions.

nihohit: also, this isn't what is commonly called "unsafe" in rust - this won't lead to undefined behaviors, just bad routing. probably should be `new_unchecked`. Maybe this is just unnecessary, and we can use `with_slot` or `with_key` throughout?

anatawa12:
Sorry I forgot to enable --features cluster-async for running tests.

I replaced all use cases of new_unsafe with ::with_slot(Slot::new().unwrap().

`Slot` struct in slot_map conflicted with new Slot api in test code so I renamed `Slot` to `SlotRange` since it holds range of slot rather than single slot.

anatawa12:
should I bump minor version in cargo.toml or remove deprecated attribute?

nihohit:
let's remove the deprecation and just add a comment and a non-public comment TODO to remove this.

nihohit:
@anatawa12 sorry, I lost track of this PR. can you please rebase it? the lints fixing this are already merged.

anatawa12:
I've rebased onto current main branch.

anatawa12:
I've fixed format error

nihohit (approved):
thanks!

Issue #2054, opened 2026-04-24
Backporting the fix for the never type fallback change


WaffleLapkin:
Hey! In https://github.com/rust-lang/rust/pull/155499#issuecomment-4290467341 we are experimenting with doing a change to the never type fallback, that we were planning for a while. We found that it breaks a lot crates that depend on old versions of `redis`.

The newer versions already have a fix for the problem, namely https://github.com/redis-rs/redis-rs/pull/1229.

Would you be willing to release backports of the fix as patch versions to older versions? That would mean affected crates can fix their issues by just running `cargo update`. I can help with the backports, but the releases need to be done by someone with permissions to make them :)

I see that #1229 is quite involved and changes public APIs, the backports would probably need to only fix the internal problems in `redis`, rather than changing the public APIs...

The distribution of failures per version of `redis` is: 0.13.0 (4), 0.14.0 (2), 0.15.1 (13), 0.16.0 (4), 0.17.0 (7), 0.18.0 (2), 0.19.0 (7), 0.20.0 (6), 0.20.1 (2), 0.20.2 (28), 0.21.0 (5), 0.21.1 (2), 0.21.2 (5), 0.21.3 (1), 0.21.4 (10), 0.21.5 (38), 0.21.6 (6), 0.21.7 (28), 0.22.1 (2), 0.22.3 (43), 0.23.0 (4), 0.23.1 (3), 0.23.2 (3), 0.23.3 (88), 0.23.4 (1), 0.24.0 (117), 0.25.0 (1), 0.25.2 (7), 0.25.3 (23), 0.25.4 (112); It would be most valuable to release backports as `0.21.x`, `0.22.x`, `0.23.x`,  `0.24.x` and `0.25.x`, although fixing even earlier versions might still be beneficial.

nihohit:
Hi @WaffleLapkin, of course I'd be happy to make the releases, but I don't have the time (or the tokens :) ) to go back to each version and make the fix. 
I did create branches for the requested versions, so if you're able to open the PRs, I'll merge & release them.

WaffleLapkin:
I've created PRs for the most important versions; if they are good I'll create PRs for all the others too :)

I assume it's okay if I update the changelog with all the backports in one go, after all PRs are merged?

nihohit:
> I assume it's okay if I update the changelog with all the backports in one go, after all PRs are merged?

[...]

PR #2057, opened 2026-04-24
0.22.4: fixup dependency on never type fallback to unit

WaffleLapkin:
cc https://github.com/redis-rs/redis-rs/issues/2054


nihohit:
@WaffleLapkin need to fix the
```
  push:
    branches: [ main, 0.21.x ]
  pull_request:
    branches: [ main, 0.21.x ]
```
statements in the .github/workflows/rust.yml file, in order to trigger CI.

nihohit:
@WaffleLapkin `ahash:0.7.6` was yanked, you'll need to upgrade to 0.7.7

update: and apparently more work after that - I updated the dependency and auto-formatted (in https://github.com/redis-rs/redis-rs/pull/2073), and then got plenty more lints.

WaffleLapkin:
@nihohit so (one of) the problem(s) is that `redis` doesn't commit the lockfile and it's apparently very hard to make a lockfile where all dependencies are locked to a version compatible with a very old stable... (`ahash` problem is that `0.7.6` is yanked, and newer versions have higher msrv; but there are similar problems with _so many_ dependencies)

would it be fine to just disable msrv CI for backports? I'm pretty confident that my changes don't raise the MSRV, but proving it appears hard :\

Otherwise I can try making a lockfile where all dependencies are downgraded enough to work with `1.59.0`, but that seems like a lot of wasted work...

nihohit:
@WaffleLapkin that sounds reasonable, but before that - would it be feasible to bump the MSRV to a more recent version, maybe 1.85 which we use in CI for `main`? That version is about a year old, so it should still be usable for users who can't updated the compiler very often, and all of the dependencies in main work with it.

WaffleLapkin:
@nihohit sure! should I update it only for CI or also in the `Cargo.toml`'s `rust-version`?

nihohit:
And, of course - bumping compiler version means that we have a ton of new lints to handle 🤦

WaffleLapkin:
Well, at the very least with this `cargo +nightly clippy --all-features` passes localy...

WaffleLapkin:
Fixing clippy warnings in benchmarks feels entirely pointless, I commented out the clippy check from CI.

.github/workflows/rust.yml
```diff
@@ -113,10 +113,10 @@ jobs:
         with:
           command: fmt
           args: --all -- --check
-      - uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --all-features --all-targets -- -D warnings
+      # - uses: actions-rs/cargo@v1
```
nihohit: no, I might still need to maintain this, and know that the next change is good code. you can skip the benchmarks in the lints by specifying targets or by adding `allow` statements, but clippy still needs to run.

WaffleLapkin: As you said yourself, you don't have time to maintain old versions, so this would probably be fine?

Regardless, I added the clippy check back, and removed `--all-targets` instead.

nihohit:
> As you said yourself, you don't have time to maintain old versions, so this would probably be fine?

Well, I assumed that we wouldn't need to update early versions, which is why we didn't keep version branches, and that assumption is quite evidently wrong :) I'd like to keep the option of maintaining those branches in a safe way possible.

WaffleLapkin:
(rustdoc is using `--document-private-items` which I haven't tested locally before, fixed a warning from it too; wish I didn't need to bother you with approving CI every time lol)

nihohit:
released. Thanks!

PR #2060, opened 2026-04-24
0.25.4: fixup dependency on never type fallback to unit

WaffleLapkin:
cc https://github.com/redis-rs/redis-rs/issues/2054


nihohit:
@WaffleLapkin need to fix the build here.

nihohit:
@WaffleLapkin some tests still need `never` type handling

nihohit:
published, thanks!

PR #2061, opened 2026-04-28
Add support for the HOTKEYS commands

StefanPalashev:
# Add support for the HOTKEYS commands

## Summary

Add support for the new `HOTKEYS` command suite introduced in **Redis 8.6**, which lets clients track the most accessed keys on a node by CPU time and/or network bytes. The implementation covers the four sub-commands (`START`, `GET`, `STOP`, `RESET`) for both synchronous and asynchronous connections, with strong typing for options and responses.

## Motivation

`HOTKEYS` is a stateful, node-local diagnostic tool for identifying hot keys in real-world workloads without the overhead of full key-space scans. Exposing it through the existing `Commands` / `AsyncCommands` style traits gives users an idiomatic way to drive sampling sessions and inspect the structured response, instead of hand-assembling `Cmd` invocations and parsing responses.

### New module `redis::commands::hotkeys`

- **`HotkeysOptions`** — builder for `HOTKEYS START` arguments. Exposes:
  - `new_with_cpu()` / `new_with_net()` constructors (at least one metric is required by the server, enforced at construction time).
  - `and_cpu()` / `and_net()` to combine metrics — `METRICS count` is auto-derived from the number of enabled metrics.
  - `with_count(k)` returning `Result<Self, String>`; `k` is range-checked against `[HOTKEYS_COUNT_MIN, HOTKEYS_COUNT_MAX]` (1..=64) before the command leaves the client.
  - `with_duration_secs(s)`, `with_sample_ratio(r)`, `with_slots(slots)`.

- **`HotkeysResponse`**, **`HotKeyEntry`**, **`SlotRange`** — typed view of the `HOTKEYS GET` payload, including:
  - Tracking metadata (`tracking_active`, `sample_ratio`, `collection_start_time_unix_ms`, `collection_duration_ms`).
  - Aggregate stats (`all_commands_all_slots_us`, `net_bytes_all_commands_all_slots`, totals for CPU user/sys and net bytes).
  - Per-metric top-K lists (`by_cpu_time_us`, `by_net_bytes`) gated `Option<...>` based on which metrics were requested.
  - Cluster-only fields populated when `SLOTS` was used (`*_selected_slots_us`, `net_bytes_*_selected_slots`).
- **`FromRedisValue` for `HotkeysResponse`** — handles the server's outer single-element array wrapping under both RESP2 and RESP3.

### New traits

- **`HotkeysCommands`** (sync) on `ConnectionLike`:
  - `hotkeys_start(opts) -> RedisResult<()>`
  - `hotkeys_get() -> RedisResult<Option<HotkeysResponse>>` (returns `None` when no session is available)
  - `hotkeys_stop() -> RedisResult<bool>` (true if a session was running, false otherwise)
  - `hotkeys_reset() -> RedisResult<()>`
- **`AsyncHotkeysCommands`** (gated on `feature = "aio"`) on `aio::ConnectionLike` — mirrors the sync surface with `RedisFuture<'a, T>` returns.

### Re-exports

`HotkeysCommands`, `HotkeysOptions`, `HotkeysResponse`, `HotKeyEntry`, `SlotRange` are exported from the crate root; `AsyncHotkeysCommands` is re-exported under `feature = "aio"`.

## Cluster usage

Notes:

The `HOTKEYS` command must not be exposed on clustered client abstractions or connection pooling clients. Exposing `HOTKEYS` on a `ClusterClient` or similar high-level abstraction would break session affinity because `START` might be routed to one node while `GET` is routed to a different node, resulting in errors or undefined behavior.

The `HOTKEYS` command with its typed response and arguments should only be implemented on standalone client instances that maintain a direct connection to a single Redis node. This ensures that the session affinity requirement is naturally satisfied, as all commands from that client instance will reach the same node.
 
`HOTKEYS` is **node-local** and is intentionally **NOT** wired into `ClusterConnection`'s aggregating path.  Using `HOTKEYS` against a specific cluster node can be done via `route_command` with explicit node routing.

## Tests

A new `redis/tests/test_hotkeys.rs` integration suite plus the unit tests inside `commands/hotkeys.rs` cover:

**Unit (`commands::hotkeys::tests`):**
- option building
- `ToRedisArgs` output
- `COUNT` range validation
- `FromRedisValue` parsing under several response shapes.

**Integration (gated on Redis ≥ 8.6 via `run_test_if_version_supported!`):**

- `hotkeys` mod (sync, RESP2/RESP3):
  - State-machine behavior (`STOP`/`RESET`/`GET`/`START` interactions and error cases).
  - Per-metric coverage (`CPU`, `NET`, `CPU`+`NET`) - verifies the right `Option` fields are populated.
  - `DURATION` + `COUNT` - confirms top-K limit and that `tracking_active` flips to false after the timer expires.
  - `SAMPLE` ratio round-trips through the response.
  - `SLOTS` on standalone is rejected with the expected error.
  
- `async_hotkeys` mod - mirrors the sync suite, parameterized over `RuntimeType` (`tokio` / `smol`) × `ProtocolVersion` (RESP2 / RESP3) via `rstest`.

- `hotkeys_cluster` mod — exercises `SLOTS` filtering both via a direct-node client and via `ClusterConnection::route_command`, asserts `selected_slots` round-trips and that the cluster-only response fields are populated.

## Reference documentation:
HOTKEYS main page: https://redis.io/docs/latest/commands/hotkeys/
[HOTKEYS START](https://redis.io/docs/latest/commands/hotkeys-start/) - Starts hotkeys tracking with specified metrics.
[HOTKEYS STOP](https://redis.io/docs/latest/commands/hotkeys-stop/) - Stops hotkeys tracking but preserves data.
[HOTKEYS GET](https://redis.io/docs/latest/commands/hotkeys-get/) - Returns tracking results and metadata.
[HOTKEYS RESET](https://redis.io/docs/latest/commands/hotkeys-reset/) - Releases resources used for tracking.

### DIFF
--- a/redis/src/commands/hotkeys.rs
+++ b/redis/src/commands/hotkeys.rs
@@ -1,0 +1,906 @@
+//! Defines types to use with the HOTKEYS commands.
+//!
+//! The HOTKEYS command is a stateful, node-local command requiring session affinity.
+//! It should only be used on standalone clients (Connection, MultiplexedConnection, etc.)
+//! and NOT on cluster clients (ClusterConnection).
+//!
+//! # Command Syntax (Redis 8.6.0+)
+//!
+//! ```text
+//! HOTKEYS START METRICS count [CPU] [NET] [COUNT k] [DURATION seconds] [SAMPLE ratio] [SLOTS count slot [slot ...]]
+//! HOTKEYS GET
+//! HOTKEYS STOP
+//! HOTKEYS RESET
+//! ```
+//!
+//! # Using HOTKEYS in Cluster Mode
+//!
+//! While the high-level `HotkeysCommands` trait is not available on `ClusterConnection`,
+//! HOTKEYS commands can still be used on individual cluster nodes by using `route_command` with
+//! explicit node routing:
+//!
+//! ```rust,no_run
+//! # #[cfg(feature = "cluster")]
+//! # {
+//! use redis::cluster::ClusterClient;
+//! use redis::cluster_routing::{RoutingInfo, SingleNodeRoutingInfo};
+//! use redis::{cmd, FromRedisValue, HotkeysOptions, HotkeysResponse};
+//!
+//! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/"];
+//! let client = ClusterClient::new(nodes).unwrap();
+//! let mut connection = client.get_connection().unwrap();
+//!
+//! // Route to a specific node
+//! let routing = RoutingInfo::SingleNode(SingleNodeRoutingInfo::ByAddress {
+//!     host: "127.0.0.1".to_string(),
+//!     port: 6379,
+//! });
+//!
+//! // Start tracking on that specific node - track keys by CPU time percentage
+//! let opts = HotkeysOptions::new_with_cpu();
+//! let _ = connection.route_command(
+//!     &cmd("HOTKEYS").arg("START").arg(opts),
+//!     routing.clone()
+//! ).unwrap();
+//!
+//! // ... perform operations ...
+//!
+//! // Get metrics from the same node
+//! let value = connection.route_command(
+//!     &cmd("HOTKEYS").arg("GET"),
+//!     routing.clone()
+//! ).unwrap();
+//! let response = HotkeysResponse::from_redis_value(value).unwrap();
+//!
+//! // Stop tracking on that node
+//! let _ = connection.route_command(
+//!     &cmd("HOTKEYS").arg("STOP"),
+//!     routing
+//! ).unwrap();
+//! # }
+//! ```
+
+use crate::errors::ParsingError;
+use crate::types::{FromRedisValue, RedisWrite, ToRedisArgs, Value};
+use std::collections::HashMap;
+
+/// Minimum value for the COUNT parameter of HOTKEYS START.
+pub const HOTKEYS_COUNT_MIN: u64 = 1;
+/// Maximum value for the COUNT parameter of HOTKEYS START.
+pub const HOTKEYS_COUNT_MAX: u64 = 64;
+
+/// Options for the HOTKEYS START command.
+///
+/// At least one of `cpu` or `net` must be enabled to specify which metrics to collect.
+/// The `METRICS count` is automatically derived from how many metric types are enabled.
+///
+/// Use [`HotkeysOptions::new_with_cpu()`] or [`HotkeysOptions::new_with_net()`] constructors to create
+/// valid options with at least one metric enabled.
+///
+/// # Example
+///
+/// ```rust,no_run
+/// use redis::{HotkeysOptions, HotkeysCommands};
+///
+/// # fn example() -> redis::RedisResult<()> {
+/// let client = redis::Client::open("redis://127.0.0.1/")?;
+/// let mut con = client.get_connection()?;
+///
+/// // Track hotkeys by both CPU and network usage for 60 seconds
+/// let opts = HotkeysOptions::new_with_cpu()
+///     .and_net()
+///     .with_duration_secs(60);
+///
+/// con.hotkeys_start(opts)?;
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Clone, Debug)]
+pub struct HotkeysOptions {
+    /// Track hotkeys by CPU time percentage
+    cpu: bool,
+    /// Track hotkeys by network bytes percentage
+    net: bool,
+    /// Value of K for top-K hotkeys tracking (optional COUNT parameter)
+    count_k: Option<u64>,
+    /// Duration in seconds for tracking (optional DURATION parameter)
+    duration_secs: Option<u64>,
+    /// Sampling ratio for probabilistic tracking (optional SAMPLE parameter)
+    sample_ratio: Option<u64>,
+    /// Specific slots to track in cluster mode (optional SLOTS parameter)
+    slots: Option<Vec<u16>>,
+}
+
+impl HotkeysOptions {
+    /// Creates options to track hotkeys by CPU time percentage.
+    ///
+    /// # Example
+    ///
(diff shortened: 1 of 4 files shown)


nihohit:
why are the tests serial?

StefanPalashev:
They shouldn't be marked as serial, thank you for catching it! It was a leftover from a local test.

redis/src/commands/hotkeys.rs
```diff
@@ -0,0 +1,906 @@
...
+//! explicit node routing:
+//!
+//! ```rust,no_run
+//! # #[cfg(feature = "cluster")]
+//! # {
+//! use redis::cluster::ClusterClient;
+//! use redis::cluster_routing::{RoutingInfo, SingleNodeRoutingInfo};
+//! use redis::{cmd, FromRedisValue, HotkeysOptions, HotkeysResponse};
+//!
+//! let nodes = vec!["redis://127.0.0.1:6379/", "redis://127.0.0.1:6378/"];
```
nihohit, line 33: please add an example for all nodes routing, too - that makes more sense IMO in cluster mode.
please also add a test for it in the cluster tests.

redis/src/commands/hotkeys.rs
```diff
@@ -0,0 +1,906 @@
...
+        self
+    }
+
+    /// Set specific hash slots to track in a cluster environment.
+    ///
+    /// Only keys that hash to the specified slots will be tracked.
+    /// Useful for tracking hotkeys on specific shards in a Redis cluster.
+    ///
+    /// Note: Using SLOTS when not in cluster mode will result in an error.
+    pub fn with_slots(mut self, slots: Vec<u16>) -> Self {
```
nihohit, line 306: can / should we use this to route the request in cluster mode?

redis/src/commands/hotkeys.rs
```diff
@@ -0,0 +1,906 @@
...
+            // SLOTS + count + one arg per slot
+            n += 2 + slots.len();
+        }
+        n
+    }
+}
+
+/// A single hotkey entry with its metric value.
+#[derive(Debug, Clone, PartialEq)]
+pub struct HotKeyEntry {
```
nihohit, line 382: #[non_exhaustive] on all types with public fields

redis/src/commands/hotkeys.rs
```diff
@@ -0,0 +1,906 @@
...
+        assert_eq!(args[8], b"SAMPLE");
+        assert_eq!(args[9], b"500");
+    }
+
+    #[test]
+    fn test_hotkeys_response_parsing_resp2() {
+        use crate::Value;
+
+        // Simulate RESP2 flat array response
+        let response = Value::Array(vec![
```
nihohit, line 801: consider using the `redis_value` macro to make this easier to read.

redis/src/commands/hotkeys.rs
```diff
@@ -0,0 +1,906 @@
...
+        assert_eq!(cpu_keys[1].key, "key2");
+        assert_eq!(cpu_keys[1].value, 750);
+    }
+
+    #[test]
+    fn test_hotkeys_response_parsing_resp3() {
+        use crate::Value;
+
+        // Simulate RESP3 map response.
+        let response = Value::Map(vec![
```
nihohit, line 856: same - redis_value!

redis/src/commands/hotkeys.rs
```diff
@@ -0,0 +1,906 @@
...
+        assert_eq!(cpu_keys[0].value, 1500);
+        assert_eq!(cpu_keys[1].key, "key2");
+        assert_eq!(cpu_keys[1].value, 750);
+    }
+
+    #[test]
+    fn test_hotkeys_response_parsing_with_net() {
+        use crate::Value;
+
+        let response = Value::Array(vec![
```
nihohit, line 937: same, and below

nihohit (commented):
reviewed the hotkeys.rs file, will continue the review later.

StefanPalashev:
I had the `redis_value!` macro in mind when implementing the tests, however things didn't go smoothly and I decided that the effort is not worth it. Here's why the refactoring to use `redis_value!` is not feasible:

The macro expands to `redis::Value` as seen by `redis-test` (a separate compilation of `redis` without `--test cfg`), which the compiler treats as a distinct type from the in-crate `Value`.”

The `redis_value!` macro is `#[macro_export]` from `redis-test` and expands to fully qualified `redis::Value::…` paths. When used from integration tests, as in the current usage, that crate depends on `redis` as a normal dependency and the path resolves naturally.

Using it from inside `redis`' own `#[cfg(test)]` mod tests, however, hits a compilation cycle:

The test binary for `redis` compiles `redis` with `--cfg test`, producing `Value definition A`.
Its `dev-dep redis-test` must be compiled, which depends on `redis` without `--cfg test`, producing `Value definition B`.
The macro `redis_value!` in `redis-test` was monomorphized against `definition B`'s `redis::Value`, so its expansion yields `B::Value::Array(…)`.
But `HotkeysResponse::from_redis_value` in the unit test wants `A`'s Value and the compiler rejects it with: "there are multiple different versions of crate redis in the dependency graph".

redis/src/commands/hotkeys.rs
StefanPalashev: Done

redis/src/commands/hotkeys.rs
StefanPalashev: `with_slots` is a server-side filter, not a routing hint. It is only valid with **single-node routing** to the **primary that owns the specified slots**. When used with fan-out routing, the same SLOTS list is broadcast to all primaries, causing non-owning nodes to return **errors**.

This behavior is now clarified in the `with_slots` docstring. Additionally, the final assertion in the new test demonstrates this failure mode: combining `with_slots([slot])` with `AllMasters` and `AllSucceeded` results in a top-level error.

nihohit (approved):
Thanks!

Issue #1955, opened 2026-02-10
Potential TCP deadlock with large Redis items (>=10mib) with ElastiCache


virratanasangpunth:
**Description**

I encountered a stall/deadlock when performing concurrent GETs and SETs of large items (>10 MiB) using redis-rs with Tokio on an AWS ElastiCache Redis cluster.

**Steps to Reproduce**

A minimal reproducer is attached: [reproducer commit](https://github.com/virratanasangpunth/redis-rs/commit/8d9da5140e7541ab64bf7183782758ed32463c66)

The reproducer spawns multiple Tokio tasks performing concurrent GETs and SETs. For example, to run with 10 MiB payloads:

```
PAYLOAD_SIZE_BYTES=10485760 \
REDIS_URL=rediss://vir-cluster-0001-001.vir-cluster.chwfvm.usw2.cache.amazonaws.com:6379/#insecure \
cargo run --example large-items-elasticache-deadlock --features tokio-comp,tokio-rustls-comp,tls-rustls-insecure
```

I ran the reproducer on an EC2 host (c7g.8xlarge) so that it can talk to Elasticache endpoint which is only accessible inside the VPC. So for large items (in this case 10mib), we can see that the reproducer ran for a bit then got stuck.

**Observed Behavior**
The program runs for a few iterations, then gets stuck without forward progress after:
```
Connecting to rediss://vir-cluster-0001-001.vir-cluster.chwfvm.usw2.cache.amazonaws.com:6379/#insecure...
Connected to rediss://vir-cluster-0001-001.vir-cluster.chwfvm.usw2.cache.amazonaws.com:6379/#insecure. Payload size: 10485760 bytes
Press Ctrl+C to stop.
[redis-get-task-0] GET item_no_0
[redis-get-task-0] GET item_no_0 OK (1.7ms)
[redis-get-task-0] GET item_no_1
[redis-get-task-0] GET item_no_1 OK (1.0ms)
[redis-get-task-0] GET item_no_2
[redis-set-task-0] SET item_no_0
[redis-get-task-0] GET item_no_2 OK (967.4µs)
[redis-get-task-0] GET item_no_3
[redis-get-task-0] GET item_no_3 OK (972.4µs)

....

[redis-set-task-0] SET item_no_11
[redis-get-task-0] GET item_no_19 OK (17.6ms)
[redis-get-task-0] GET item_no_0
[redis-set-task-0] SET item_no_11 OK (17.6ms)
[...]

PR #2070, opened 2026-05-01
Fix async connection handling to avoid TCP deadlocks

PDXKimani:
This work came out of #1955.

The classic TCP deadlock pattern requires two actors willing to participate in such a deadlock. Both client and server must exhibit a pattern of being unwilling to read from their receive buffers until they are able to write to their send buffer - that is, they "read only after write". In that case, if both receive buffers become full (due to e.g. large requests), neither client nor server will make progress.

To demonstrate that the async multiplexed connection is susceptible to TCP deadlock, I have begun by writing a unit test which implements a minimal "read only after write" server. Using tiny tokio duplex buffers allows us to demonstrate the bug at small scale, but the form of the bug is identical in the reported issue. We issue several concurrent "large" requests larger than the buffers, and observe that the client and server enter a deadlock.

As I understand the issue, when poll_flush pends on the write flush due to a full remote rcv buffer, we become susceptible to deadlock because we will not wake even if new bytes become available to read - only once the write can progress. Thus, we become a "read only after write" client.

To resolve the issue, we add a poll_read in both `poll_ready` and `poll_flush` before pending. This advances reads and critically registers a read waker in the context, so even if the remote rcv buffer is full, we will wake and continue to read new bytes from the server. This prevents even a susceptible server such as the one in the unit test from deadlocking, as we continue to read the bytes provided by the server, allowing it to progress.

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -307,7 +307,9 @@ where
         mut self: Pin<&mut Self>,
         cx: &mut task::Context,
     ) -> Poll<Result<(), Self::Error>> {
-        // Drain pending responses; see comment in poll_flush for rationale.
+        // Always service the read side of a connection concurrently with write to
+        // avoid TCP deadlock.
+        // See comment in poll_flush for more context.
         if matches!(self.as_mut().poll_read(cx), Poll::Ready(Err(()))) {
             return Poll::Ready(Err(()));
         }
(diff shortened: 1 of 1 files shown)


kvcache:
This is necessary because the sink_stream is driven forward via events from `stream::poll_fn()....forward()`. So the `Sink` trait is what drives the read() side of the stream. In order to not be vulnerable to tcp deadlock, you have to advance both buffers independently of the other's progress. So this change advances both whenever one advances.

redis/src/aio/multiplexed_connection.rs
```diff
@@ -307,6 +307,12 @@ where
         mut self: Pin<&mut Self>,
         cx: &mut task::Context,
     ) -> Poll<Result<(), Self::Error>> {
+        // Always service the read side of a connection concurrently with write to
+        // avoid TCP deadlock.
+        // See comment in poll_flush for more context.
```
nihohit: instead of commenting here with a reference to another comment, maybe add here and below
`ready!(self.read_to_prevent_tcp_deadlock(cx));`

and add the shared logic & comment in a shared function?

nihohit (commented):
Thanks! It's a pleasure to receive such a well-documented change. One comment.

redis/src/aio/multiplexed_connection.rs
PDXKimani: So we can't collapse to a `ready!` because we don't actually want to short-circuit return on a pending - we need to continue on to the write logic if the read is pending, or we'd open ourselves to a _different_ kind of lockup (write only after read). I think the code here is about as minimal as is reasonably achievable - trying to extract anything would actually add verbosity.

I'll edit that comment though so each spot adequately explains itself.

nihohit: sorry I wasn't clear - I intended that `read_to_prevent_tcp_deadlock` will return ready only on errors, not on a ready read, just like the current if statement.

PDXKimani: your snippet would expand to:
```rust
match self.read_to_prevent_tcp_deadlock(cx) {
    Poll::Ready(t) => t,
    Poll::Pending => return Poll::Pending, // This is the problem
};
```
We need to _continue_ to the write poll if the read poll is pending - we are in these functions to advance the write side, too. The unconditional read check first is to ensure that we have appropriately registered wakers in the event that writes will pend.

Unfortunately, that means there's not much to encapsulate. Assuming `read_to_prevent_tcp_deadlock` looks something like
```rust
match self.as_mut().poll_read() {
    Poll::Ready(Err(())) => Poll::Ready(()),
    _ => Poll::Pending,
}
```
we would still be required to call it like:
```rust
if let Poll::Ready(_) = self.read_to_prevent_tcp_deadlock(cx) {
    return Poll::Ready(Err(()));
}
```
or equivalently
```rust
if matches!(self.read_to_prevent_tcp_deadlock(cx), Poll::Ready(())) {
    return Poll::Ready(Err(()));
}
```
I don't think that does much in the way of code reduction, and I fear it actually buries the lede a bit as to what is happening.

PDXKimani: The critical element is that we must always poll both sides before returning Pending.

nihohit (approved):

nihohit:
Thanks for the investigation and the fix!

PR #2077, opened 2026-05-07
add changelog entries for backports

WaffleLapkin:
cc #2054 (I also added some previously missed versions)


nihohit (approved):
Thanks for being thorough!

PR #2086, opened 2026-05-21
add support for setting SO_LINGER on redis sockets

svix-jbrown:
This exposes the `SO_LINGER` sockopt in `TcpSettings`, which can be helpful in avoiding TIME-WAIT exhaustion in CI environments that might create and tear down a ton of sockets.

### DIFF
--- a/redis/src/io/tcp.rs
+++ b/redis/src/io/tcp.rs
@@ -50,6 +50,16 @@ impl TcpSettings {
         }
     }
 
+    /// Set the SO_LINGER time on this socket.
+    ///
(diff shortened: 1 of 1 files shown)


redis/src/io/tcp.rs
```diff
@@ -50,6 +50,13 @@ impl TcpSettings {
         }
     }
 
+    pub fn set_linger_time(self, linger: Duration) -> Self {
```
nihohit, line 63: document what this does and what's the default if it isn't set.

nihohit (approved):
Thanks!

PR #2087, opened 2026-05-21
Support bb8 pool for sentinel client

banthony42:
Goal is to enable for the user to use `bb8::Pool` with a `SentinelClient`

Usage:
```rust
    let client = SentinelClient::build(
        vec!["redis://127.0.0.1:26379/"],
        String::from("master1"),
        Some(SentinelNodeConnectionInfo::default()),
        SentinelServerType::Master,
    )
    .unwrap();

    let pool = bb8::Pool::builder()
        .max_size(5)
        .connection_timeout(Duration::from_secs(3))
        .build(AsyncLockedSentinelClient::new(client))
        .await
        .unwrap();
```

Support only `tokio`async runtime since `bb8` does not support `smol` because of internal use of `tokio::time::interval_at`

### DIFF
--- a/redis/tests/test_sentinel.rs
+++ b/redis/tests/test_sentinel.rs
@@ -1227,3 +1227,76 @@ pub mod pool_tests {
         assert!(try_conn.is_some());
     }
 }
+
+#[cfg(feature = "tokio-comp")]
+pub mod bb8_pool_tests {
+    use super::*;
+    use bb8::Pool;
+    use redis::sentinel::AsyncLockedSentinelClient;
+    use std::collections::HashSet;
+
+    fn parse_client_info(value: &str) -> HashMap<&str, &str> {
+        let info_map: std::collections::HashMap<&str, &str> = value
+            .split(" ")
+            .filter_map(|line| line.split_once('='))
+            .collect();
+        info_map
+    }
+
+    #[tokio::test]
+    async fn test_sentinel_client() {
+        let master_name = "master1";
+        let context = TestSentinelContext::new(2, 3, 3);
+        let master_client = SentinelClient::build(
+            context.sentinels_connection_info().clone(),
+            String::from(master_name),
+            Some(context.sentinel_node_connection_info()),
+            redis::sentinel::SentinelServerType::Master,
+        )
+        .unwrap();
+
+        let pool = Pool::builder()
+            .max_size(5)
+            .connection_timeout(Duration::from_secs(3))
+            .build(AsyncLockedSentinelClient::new(master_client))
+            .await
+            .unwrap();
+
+        let mut conns = Vec::new();
+        for i in 0..5 {
+            let conn = pool.get().await.unwrap();
+            assert_eq!(pool.state().connections, i + 1);
+            conns.push(conn);
+        }
+
+        // since max_size is 5 and we haven't freed any connection this try should fail
+        let try_conn = pool.get().await;
+        assert!(try_conn.is_err());
+
+        let mut client_id_set = HashSet::new();
+
+        for mut conn in conns {
+            let client_info_str: String = redis::cmd("CLIENT")
+                .arg("INFO")
+                .query_async(&mut *conn)
+                .await
+                .unwrap();
+
+            let client_info_parsed = parse_client_info(client_info_str.as_str());
+
+            // assert if all connections have different IDs
+            assert!(client_id_set.insert(client_info_parsed.get("id").unwrap().to_string()));
+
+            let info: String = redis::cmd("INFO")
+                .arg("REPLICATION")
+                .query_async(&mut *conn)
+                .await
+                .unwrap();
+            assert_is_master_role(info)
+        }
+
+        // since previous connections are freed, this should work
+        let try_conn = pool.get().await;
+        assert!(try_conn.is_ok());
+    }
+}
--- a/redis/src/bb8.rs
+++ b/redis/src/bb8.rs
@@ -4,6 +4,9 @@ use crate::{Client, Cmd, RedisError};
 #[cfg(feature = "cluster-async")]
 use crate::{cluster::ClusterClient, cluster_async::ClusterConnection};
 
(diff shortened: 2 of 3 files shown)


banthony42:
I have maybe a solution regarding `LockedSentinelClient` and `Bb8LockedSentinelClient` unification :
```rust
enum RuntimeMutex<SentinelClient> {
    #[cfg(all(feature = "bb8", feature = "tokio-comp"))]
    Tokio(tokio::sync::Mutex<SentinelClient>),
    #[cfg(all(feature = "bb8", feature = "smol-comp"))]
    Smol(smol::lock::Mutex<SentinelClient>),
    #[cfg(feature = "r2d2")]
    Sync(std::sync::Mutex<SentinelClient>),
}

pub struct LockedSentinelClient(pub(crate) RuntimeMutex<SentinelClient>);
```


But it mean break retrocompatibility of `r2d2` LockedSentinelClient 
And i have just realized that it will not solve the test issue (when all features are enabled)

banthony42:
Hello @nihohit 
I'd love to have your opinion when you have time  :slightly_smiling_face: 
See you soon

nihohit:
I'm not sure that there's a need for the various mutexes - I think that the tokio mutex is usable on all the async runtimes.

banthony42:
> I'm not sure that there's a need for the various mutexes - I think that the tokio mutex is usable on all the async runtimes.

You are right and i have finally found the answer :


I forget to check the compatibility with `bb8` regarding `smol`
And indeed it's seems not compatible, first hint the [bb8 README](https://github.com/djc/bb8) speak only about tokio
> A full-featured connection pool, designed for asynchronous connections (using tokio). 

Second stronger hint the direct use of `tokio::time::interval_at` within `bb8`:

##### bb8 user code :
```rust
    let pool = bb8::Pool::builder()
        .max_size(5)
        .connection_timeout(Duration::from_secs(3))
        .build(Bb8LockedSentinelClient::new(client)) // <-- 1 Dive here
        .await
        .unwrap();
```


##### bb8 [src/api.rs](https://github.com/djc/bb8/blob/main/bb8/src/api.rs#L435)
```rust
    pub async fn build(self, manager: M) -> Result<Pool<M>, M::Error> {
        let pool = self.build_inner(manager); // <-- 3 Dive here
        pool.inner.start_connections().await.map(|()| pool)
    }


    fn build_inner(self, manager: M) -> Pool<M> {
        if let Some(min_idle) = self.min_idle {
            assert!(
                self.max_size >= min_idle,
                "min_idle must be no larger than max_size"
            );
        }

        Pool {
            inner: PoolInner::new(self, manager),  // <-- 2 Dive here
        }
    }
```

##### bb8 [src/inner.rs](https://github.com/djc/bb8/blob/main/bb8/src/inner.rs#L27)

```rust
impl<M: ManageConnection + Send> PoolInner<M> {
    pub(crate) fn new(builder: Builder<M>, manager: M) -> Self {
        let inner = Arc::new(SharedPool::new(builder, manager));

        if inner.statics.max_lifetime.is_some() || inner.statics.idle_timeout.is_some() {
            let start = Instant::now() + inner.statics.reaper_rate;
            let interval = interval_at(start.into(), inner.statics.reaper_rate);  // <-- 4 use of tokio::time::interval_at

[...]
```


To confirm :

```rust
    fn tokio_interval_at_not_works_in_smol() {
        smol::block_on(async {
            let start = tokio::time::Instant::now();
            let mut interval = tokio::time::interval_at(start, Duration::from_millis(10));

            interval.tick().await;
        });
    }
```

Give me this error, which is the same error encounter on smol tests :

```
---- tests::tokio_interval_at_not_works_in_smol stdout ----

thread 'tests::tokio_interval_at_not_works_in_smol' (265030) panicked at /home/banthony42/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.52.3/src/time/interval.rs:138:26:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


Therefore i choose to simplify all the code, creating an `AsyncLockedSentinelClient`
And i have update the `async_test` macro to be able to run test on tokio runtime only.
Also i have fix the lint issue, all tests pass locally, so :crossed_fingers:

nihohit:
```
	Checking: ["hashbrown", "sentinel", "tokio-comp"]
redis/src/sentinel.rs:1097:5: error: unused import: `tokio::sync::Mutex as TokioMutex`

File: test-macros/src/lib.rs [hunk 1]
```diff
@@ -2,7 +2,20 @@ use proc_macro::TokenStream;
 use quote::quote;
 
 #[proc_macro_attribute]
-pub fn async_test(_: TokenStream, input: TokenStream) -> TokenStream {
+pub fn async_test(attr: TokenStream, input: TokenStream) -> TokenStream {
```
nihohit: this check is unnecessary. when we want a test to only run on tokio, we use 

```rs
#[tokio::test]
async fn some_test() {

Comment (banthony42) [reply test-macros/src/lib.rs]: :+1:  Didn't know, fixed ! thx

**Comment by banthony42** [comment]
> ```
> 	Checking: ["hashbrown", "sentinel", "tokio-comp"]
> redis/src/sentinel.rs:1097:5: error: unused import: `tokio::sync::Mutex as TokioMutex`
> ```

Fixed, thx :slightly_smiling_face:

File: redis/tests/test_sentinel.rs [hunk 2]
```diff
@@ -1227,3 +1227,77 @@ pub mod pool_tests {
         assert!(try_conn.is_some());
     }
 }
+
+#[cfg(feature = "bb8")]
```
nihohit: you can put a single #[cfg(feature = "tokio-comp")] here, instead of over each test

banthony42: Fixed :+1:  thx !

banthony42:
Last push was for some doc comment that was not up to date with the new struct name

nihohit (approved):
Thanks for the contribution!

PR #2091, opened 2026-05-26
feat: add ?Sized bound to impl AsyncPushSender for std::sync::Arc<T>

### LINKED ISSUE #2090: AsyncPushSender should be implemented for std::sync::Arc<T> for !T: Sized
Current blanket impl for `std::sync::Arc<T>` is guarded by `T : Sized` Implicit bound so we cannot use `Arc<dyn AsyncPushSender>` for `Arc<dyn crate::aio::AsyncPushSender>` but I think it should.

I'll create pull request.

anatawa12:
fix #2090


nihohit (approved):
thanks!

PR #2093, opened 2026-05-27
ci: Bump Redis patch versions

somechris:


nihohit:
@somechris is there a reason this is a draft?

somechris:
Yes: The original version of this PR failed unexpectedly, so I wanted to avoid pings until it went green ... and I did not get to the computer since.

Now that it's green, let me switch it to "Open"

nihohit:
Thanks!

Issue #2095, opened 2026-05-28
Tests broken against Redis 8.6.3+


somechris:
Tests (current `main` at e29a6c1) are broken against Redis 8.6.3 and Redis 8.8.0 both for me locally and also on CI

E.g.: https://github.com/redis-rs/redis-rs/actions/runs/26512781938/job/78082395864?pr=2093

(But I guess the above link will become unavailable soon)

The test runs always fail with either

```
    thread 'basic::test_vector_sets_auxiliary_commands' (37751) panicked at redis/tests/test_basic.rs:3931:13:
    assertion `left == right` failed
      left: Err(ResponseError: invalid vector specification)
     right: Ok(true)
```

or

```
    thread 'idempotency_tests::test_xcfgset_idempotency_behavior' (56775) panicked at redis/tests/test_streams.rs:3014:9:
    assertion `left != right` failed
      left: Some("1779976939980-0")
     right: Some("1779976939980-0")
```

Trying 10 runs, not a single passed for me.

If no one beats me to it, I'll have a look over the next days :+1:

nihohit:
I'm aware of the auxilliary commands test, and other PRs suggested this solution:
https://github.com/redis-rs/redis-rs/pull/2084/changes/9f948301ef4beb3ee79b114391291df582362d8a

The idempotency test failure is weird, though.

somechris:
> I'm aware of the auxilliary commands test, and other PRs suggested this solution: [9f94830](https://github.com/redis-rs/redis-rs/commit/9f948301ef4beb3ee79b114391291df582362d8a)

Thanks for that pointer.

I've amended that commit a bit in #2099 to make the vectors align better and linked the relevant commit in Redis.

 
> The idempotency test failure is weird, though.

That one is indeed weird.

It happened both in CI https://github.com/redis-rs/redis-rs/actions/runs/26512781938/job/78082395864?pr=2093#step:9:1970 and locally. But with the above fix, it vanished for me and I can no longer reproduce it. 🤷

PR #2099, opened 2026-05-30
tests: Stop inflating dimensions during vector set dimension reduction

somechris:
In `test_vector_sets_auxiliary_commands` the test data's vectors were only 2 entries long, which we "reduced" to 3.

Redis 8.6.3 received a guard against such inflating "reductions" in ea66422 and rejects inflations.

So we now test using vectors with 4 entries and reduce them to 3 entries. This makes tests against 8.6.3 pass (#2095).

As the numbers in the original test looked like they may had a purpose, did not align because of the flipping sign, and had unnecessary type specifiers, we cleaned that up while adding the needed entries.

### DIFF
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -3889,12 +3889,13 @@ mod basic {
         let key = "test_points_for_auxiliary_commands";
         let non_existent_key = "non_existent_key";
 
-        let points_data: Vec<(&'static str, [f32; 2])> = vec![
-            ("pt:A", [1.0f32, 1.0]),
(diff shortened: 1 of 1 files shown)


nihohit:
thanks!

StefanPalashev:
Thank you for fixing the test @somechris!

Just for your information, the initial values were taken from the documentation example.
https://redis.io/docs/latest/develop/data-types/vector-sets/#examples

somechris:
Hah! Cool. So they indeed had a hidden structure. Thanks for clarifying :grin:

PR #2103, opened 2026-06-01
ci: Use targets from Makefile instead of duplicating commands

somechris:
`clippy`'s invocation differed between CI and the Makefile differed. This
lead to needlessly failing CI runs although a local `make lint`
passed. We harmonize the linter calls to avoid such failures going
forward.

And as CI duplicated some commands from Makefile targets, we switch CI
to use these targets. This should help to keep local checking and CI
in sync.

### DIFF
--- a/.github/actions/lint-and-check/action.yml
+++ b/.github/actions/lint-and-check/action.yml
@@ -26,10 +26,9 @@ runs:
       shell: bash
       run: make lint
 
-    - name: Check features
+    - name: Check runtimes
       shell: bash
       run: |
-        cargo check --workspace --all-targets --all-features
         # Remove dev-dependencies so they do not enable features accidentally
         # https://github.com/rust-lang/cargo/issues/4664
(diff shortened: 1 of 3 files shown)


.github/actions/lint-and-check/action.yml
```diff
@@ -16,20 +16,19 @@ runs:
...
-      name: clippy
+    - name: Lint
       shell: bash
+      run: make lint
 
-    - name: Check features
+    - name: Check runtimes
       shell: bash
       run: |
-        cargo check --workspace --all-targets --all-features
```
nihohit, line 32: AFAIS this isn't run anywhere now. Am I missing an invocation?

somechris: See the commit message of c85c3d4

As far as I can see, `cargo clippy` runs `cargo check` internally.

So the `cargo check` does not add anything and only increases runtime.

Issue #1183, opened 2024-05-09
crate feature json does not support NX and XX args


DimiDumo:
When using the json module for redis and the crate feature json, I cannot use the NX and XX args when setting a JSON.

I would like to do the equivalent of this using `.json_set` using the json crate feature:

```
        redis::cmd("JSON.SET")
            .arg(&key)
            .arg(".")
            .arg(json_data)
            .arg("NX")
            .query_async(&mut conn).await?;
```

PR #2114, opened 2026-06-02
Add support for the new FPHA storage hint

StefanPalashev:
# Add `JSON.SET` options with `FPHA` storage-hint support

## Summary

Adds a typed options builder for `JSON.SET` that carries an optional `NX`/`XX` existence check and the optional `FPHA <TYPE>` storage hint introduced in `RedisJSON` for `Redis CE 8.8`.

The hint asks the server to pack any floating-point arrays inside the payload using a chosen lane precision (`BF16` / `FP16` / `FP32` / `FP64`), rather than the default storage. The wire format is:

```
JSON.SET <key> <path> <json_value> [NX|XX] [FPHA BF16|FP16|FP32|FP64]
```

## API

Two new public types are re-exported from the crate root behind the existing `json` feature:

| Type | Purpose |
|---|---|
| `JsonSetOptions` | Builder passed to the new `json_set_options` command. |
| `FphaType` | Storage-precision tag (`Bf16` / `Fp16` / `Fp32` / `Fp64`). |

`JsonSetOptions` is a plain struct with two `Option` fields and two builder methods:

```rust
use redis::json::{FphaType, JsonSetOptions};
use redis::{ExistenceCheck, JsonCommands};

// 1. Plain JSON document - parity with existing json_set.
con.json_set_options(key, "$", &json!({"item": 42}),
    &JsonSetOptions::default())?;

// 2. Float-array payload with a storage hint.
con.json_set_options(key, "$", &[1.0_f32, 2.0, -3.5],
    &JsonSetOptions::default().fpha(FphaType::Fp32))?;

// 3. Any Serialize payload (matrix, object, scalar) with a hint.
let matrix: &[&[f32]] = &[&[1.0, 2.0], &[3.0, 4.0]];
con.json_set_options(key, "$", matrix,
    &JsonSetOptions::default().fpha(FphaType::Bf16))?;

// 4. Existence check composes with any of the above.
con.json_set_options(key, "$", &value,
    &JsonSetOptions::default()
        .conditional_set(ExistenceCheck::NX)
        .fpha(FphaType::Fp16))?;
```

### Rules:
The value is passed as a separate argument `(any `V: Serialize`)`.
The options struct only carries command modifiers.
`JsonSetOptions` implements `ToRedisArgs` so the wire format lives in one place.
The existence check (if any) is written first, then `FPHA` + the type tag (if any).
The value is always emitted by `json_set_options` ahead of the options.

## Wire-format examples

| Call | Bytes on the wire |
|---|---|
| `json_set_options(k, "$", &{"a":1}, &default())` | `JSON.SET k $ {"a":1}` |
| `…, &{"a":1}, &default().conditional_set(NX)` | `JSON.SET k $ <json> NX` |
| `…, &[1.0,2.5], &default().fpha(Fp32)` | `JSON.SET k $ [1.0,2.5] FPHA FP32` |
| `…, &v, &default().conditional_set(XX).fpha(Fp32)` | `JSON.SET k $ <json> XX FPHA FP32` |
| `…, matrix, &default().fpha(Bf16)` | `JSON.SET k $ <matrix> FPHA BF16` |

## Server-side validation

`FPHA` range validation is performed by the server on every number inside any JSON array, regardless of nesting depth.
A single out-of-range lane causes the entire `JSON.SET` to be rejected with `value out of range for <TYPE> …` and the key is not created.
Numbers outside arrays (top-level or object-nested scalars) are stored as-is without range checking.

## Testing

- **8 unit tests** on the wire format (argument count, byte-level ordering of `<json>` / `NX`|`XX` / `FPHA <TYPE>`, empty payload, matrix / object serialization, builder order independence).
- **17 integration tests** against `RedisJSON` on `Redis CE 8.8`:
  - Round-trip per FPHA type.
  - Empty payload, NX / XX existence check.
  - Overflow rejections.
  - Boundary values: FP16 max (`±65504.0`), FP32 max (`±3.4028235e38`), BF16 above FP16 normal range.
  - Scientific notation round-trip.
  - Precision truncation: BF16 truncates `100.7 to 100.5` (7-bit mantissa); FP16 preserves `100.7` (10-bit mantissa).
  - Mixed payloads: 2D matrix, object with multiple FP-array fields, scalar, nested partial-overflow rejection.

## Compatibility

- New code paths are gated by the existing `json` cargo feature.
- No existing public API changes: `JsonCommands::json_set` and the `json_set` / `json_set_options` re-exports are additive.
- Server behavior gated to Redis CE 8.8+ via the new `REDIS_VERSION_CE_8_8` test constant

## Relevant documentation

https://redis.io/docs/latest/develop/data-types/json/
https://redis.io/docs/latest/commands/json.set

### DIFF
--- a/redis/src/commands/json.rs
+++ b/redis/src/commands/json.rs
@@ -1,7 +1,9 @@
 use crate::cmd::{Cmd, cmd};
 use crate::connection::ConnectionLike;
 use crate::pipeline::Pipeline;
-use crate::types::{FromRedisValue, RedisResult, ToRedisArgs, ToSingleRedisArg};
+use crate::types::{
+    ExistenceCheck, FromRedisValue, RedisResult, RedisWrite, ToRedisArgs, ToSingleRedisArg,
+};
 
 #[cfg(feature = "cluster")]
 use crate::commands::ClusterPipeline;
@@ -264,6 +266,14 @@ implement_json_commands! {
         cmd("JSON.SET").arg(key).arg(path).arg(serde_json::to_string(value)?).take()
     }
 
+    /// Sets the JSON Value at `path` in `key` with options.
+    ///
+    /// The value (JSON document or FPHA float array) is carried inside `options`
+    /// alongside the optional `NX`/`XX` existence check. See [`JsonSetOptions`].
+    fn json_set_options<K: ToSingleRedisArg, P: ToSingleRedisArg>(key: K, path: P, options: &'a JsonSetOptions) {
+        cmd("JSON.SET").arg(key).arg(path).arg(options).take()
+    }
+
         /// Sets the value at the path per key, for every given tuple.
     fn json_mset<K: ToSingleRedisArg, P: ToSingleRedisArg, V: Serialize>(key_path_values: &'a [(K,P,V)]) {
         let mut cmd = cmd("JSON.MSET");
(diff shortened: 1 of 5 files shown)


redis/src/commands/json.rs
```diff
@@ -264,6 +266,14 @@ implement_json_commands! {
         cmd("JSON.SET").arg(key).arg(path).arg(serde_json::to_string(value)?).take()
     }
 
+    /// Sets the JSON Value at `path` in `key` with options.
+    ///
+    /// The value (JSON document or FPHA float array) is carried inside `options`
+    /// alongside the optional `NX`/`XX` existence check. See [`JsonSetOptions`].
+    fn json_set_options<K: ToSingleRedisArg, P: ToSingleRedisArg>(key: K, path: P, options: &'a JsonSetOptions) {
```
somechris: Naively, I'd expect `json_set` and `json_set_options` to have the same signature with the `_options` variant taking additionally an options object.

But here `json_set`'s `value` moves into the `options`. So `json_set` and `json_set_options` work completely different. That'd be confusing to use for me.

Especially as `FPHA ...` here is only a type hint to Redis. Maybe I'm wrong here, but it seems `value` is still transferred in plain ASCII (not binary floating point data). So coupling them does not bring much here?

On the flip-side, the naming `JsonSetOptions` is somewhat difficult for me as well, as that does not only contain the options, but also the value.

I would find the interface easier to understand if the value and the real options were separate.

redis/tests/support/mod.rs
```diff
@@ -534,9 +534,22 @@ pub const REDIS_VERSION_CE_8_0: Version = (8, 0, 0);
 pub const REDIS_VERSION_CE_8_2: Version = (8, 1, 240);
 pub const REDIS_VERSION_CE_8_4: Version = (8, 3, 224);
 pub const REDIS_VERSION_CE_8_6: Version = (8, 6, 0);
+pub const REDIS_VERSION_CE_8_8: Version = (8, 7, 226);
```
somechris: I see that `8.7.226` is actually Redis' `8.8-m03` tag. So it is `8.8`-ish and might make sense :thumbsup:

But seeing `8.7` for the `8.8` constant ... these things always make me think twice.

I'd instead go with readability here and I'd prefer if we could just use `(8, 8, 0)` here.

That'd be very easy to understand without any second guesses. And we won't run the tests against `8.8-rc1` or some such versions where the difference might help to run a few more tests.

redis/tests/support/mod.rs
```diff
@@ -549,6 +562,18 @@ macro_rules! run_test_if_version_supported {
             return;
         }
 
+        ctx
+    }};
+    ($minimum_required_version:expr, $modules:expr) => {{
```
somechris: Nice :party:

But it duplicates all the logic ... :-(

We could do without the duplication if we let the `($minimum_required_version:expr)` branch call the other branch with an empty list as second parameter. (Like explicit currying)

And while we are extending this macro ... are there plans or options how we could extend this check to also check the version and flavour of the modules as well?

somechris (commented):
Just saw this change (Cool stuff!) and a few things popped up while looking over it. But I'm just a random coder, so feel free to ignore whatever I wrote. :+1:

redis/src/lib.rs
```diff
@@ -740,7 +740,7 @@ pub mod aio;
 
 #[cfg(feature = "json")]
 #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
-pub use crate::commands::JsonCommands;
+pub use crate::commands::{FphaInput, FphaType, JsonCommands, JsonSetOptions};
```
nihohit: I'd prefer not to add the new types to the general namespace. We haven't been very conscious about what goes where, but IMO fhpa is such an unclear name that adding it to a 'json' namespace will make things clearer.

redis/src/commands/json.rs
StefanPalashev: Thanks for taking the time to look into this. This time the API turned out to be slightly off from the ones that we had before. The aim was to have only one new API, while combining all of the new options (initially the plan was to extend the command with only the FPHA changes, but it turned out that the existence checks were also missing). And while simplicity was one part of it, the deeper driver was the `FphaInput` enum's compile-time pairing of lane element type with storage hint.

You are right about the hint part. The `JsonSetOptions` struct is misleading in its current state, since it actually contains the value rather than just configuration, and the `FPHA` type hint doesn't add meaningful value when the actual data is plain ASCII. The real benefit of having the value inside the struct is when the `FPHA` types are used.

The `FPHA` constructors need to serialize differently depending on whether they're handling float arrays, so the type information and the value need to be paired together to prevent mismatches. Separating them would lose compile-time safety.
The current implementation prevents the construction of `Fp64(&[1.0_f32])` or `Fp32(&[1.0_f64])` because the variant tag and the slice element type are paired at the type level and that pairing requires them to be created in the same expression.

These are all legitimate design issues and it would be wonderful if we can agree on a viable solution.

Maybe we can split the APIs like so:

`json_set_options(key, path, value, options)` for the generic + serialize path
`json_set_fpha(key, path, input: FphaInput<'_>)` and `json_set_fpha_options(key, path, input: FphaInput<'_>, options)` for the `FPHA` variants

redis/tests/support/mod.rs
StefanPalashev: I'll adjust all the versions accordingly.

redis/tests/support/mod.rs
StefanPalashev: Thanks again for suggesting an improvement.

I'll apply the suggestion to remove the duplication.
For now I do not plan to extend the macro further so that it checks the versions of the modules.

While I can see where this might come in handy, I'd rather avoid adding complexity until we actually have an use case for it. 

An approach to it would be to call `module list` and parse the version from there.

Do we actually need the extension as of now?

somechris: > I'll apply the suggestion to remove the duplication.

Awesome!

>> [versioning]
>
>Do we actually need the extension as of now?

No. It's not needed right now.
And it's totally not a blocker for anything.

I'm working on adding support for Bloom filters.
These need the `redisbloom` module.

And Valkey has their own module for Bloom filters.

And both come in multiple versions.

So, if we change the macro, I was thinking maybe we should keep
versioning in mind to not block further updates in that direction.

But you're right. Let's ignore that for now.
Maybe we never need it. We'll solve it when we get there. :thumbsup:

redis/src/commands/json.rs
nihohit: another option is to maintain a `(key, path, value, options)` structure, but to require for the value not `json::Value` but a `impl Into<FphaInput>`, and add `impl From<json::Value> for FphaInput` and `impl From<String> for FphaInput` (and other relevant implementations) in order to allow for seamless switches between the types?

somechris: Ok, cool. Help me understand your side of things. :thumbsup:

I can see only two real use cases for the FPHA hints.

1. User has data in a specific format and wants to inform Redis to store it in some (other) format.

   Regardless of whether the data and it's type gets paired or not, this use case will always require to manually give the needed FPHA hint.

2. User has data in a specific format and want to inform Redis to _automatically_ store data in that exact format to limit loss.

   Rust currently only offers 2 relevant native types (`f32`, `f64`), while Redis offers 4 FPHA hints (`FP16`, `BF16`, `F32`, and `F64`).

   The smaller ones (`FP16` and `BF16`) are more relevant as they allow to shrink the memory-footprint. But precisely those lack a simple Rust counterpart (`f16` is still in nightly, and `bfloat16` only comes in through various 3rd party packages)

   So for the near future, we cannot fully reliably and automatically map `f32`/`f64` to FPHA hints as `FP16` or `BF16`.

   And with the "automatically" part gone, we have to manually select the FPHA hint again. So this use-case collapses to the first use-case.

So regardless where we start, we end up having to manually select the FPHA hint.

This PR does its manual FPHA hint selection early in the pipeline (`FPHAInput`, `fpha_serialize`). So from that point on, we get safety and the code can re-use that manual selection during `json_set_options`. But although early, the relevant point is that the selection was manual and not compile-time-type-safety-guarded (Nothing prevents the same `f32` array to get tagged as either of `FPHAInput::FP16`, `FPHAInput::BF16`, `FPHAInput::F32`)

So to me, it boils down to: When should the manual FPHA hint selection happen? Early or late. 🤔 

As this PR's early selection couples `value` and `options` in an (for `redis-rs`) unusual way and still rquires to manually select the hint, I'd go with "late" and use a plain

```
json_set_options(key, path, value, options)
```

with an options like

```
pub struct JsonSetOptions {
  conditional_set: Option<ExistenceCheck>
  fpha_tag: Option<FphaType>
}
```

That approach:
* covers all use-cases,
* is in sync with the other commands,
* nicely aligns with the syntax of Redis' `JSON.SET`, and
* is intuitively to use (not too many extra types or special ways to set things up).

somechris: > Another option: `Into<FphaInput>` implementations

Sure. And that's really useful for `f64` arrays.

But it stops there, as the key issue is that there is no natively usable `f16` or `bfloat16` type in Rust (yet). And these are the more relevant FPHA hints. This PR maps `f32` to `FP16`, `BF16`, and `FP32`. And at some point it has to be _manually_ decided to which of the three FPHA hints an `f32` maps to.

We can add types and implement `From`s and more. This makes things a bit more complicated to use, but not more safe, because the `f32` to `FP16`, `BF16`, and `FP32` selection is still manual. Only at a different point in code.

So among these choices, I'd follow Occam's razor and pick the simplest one (plain Options struct with FPHA hint enum).

That's easy to use, in line with the current code base, and follows YAGNI. And it's open to transparently get changed to `Into<FphaInput>` (or other variants) once `f16` or `bfloat` materialize _natively_ in Rust one day.

StefanPalashev: Initially the `half` crate was used to represent the smaller floating-point types (`FP16` and `BF16`). However, these types only carried the user's intent and since every value of the `FP16` and `BF16` types can be represented exactly as an `f32`, and the server re-quantizes values based on the `FPHA` tag during processing, the dependency was removed.

Your suggested approach seems cleaner and adheres to the established API style, however it would drop the case when a `&[f64]` is tagged as a narrower type. This would now turn into a runtime error on the server.

I wanted to eliminate as many potential server-side errors as possible by leveraging compile-time type safety on the client. However, the current API's compile-time guarantees are narrow:
- They prevent `f64` to 'smaller-type' combinations entirely, but don't prevent precision loss within range or out-of-range errors for same-or-smaller-type pairs like `f32` to `FP16`.

StefanPalashev: @nihohit, should we proceed with the suggested approach by @somechris?

nihohit: what would be the concrete signature of json_set_options then?

StefanPalashev: `fn json_set_options<K, P, V: Serialize>(
    key: K, path: P, value: &V, options: &JsonSetOptions,
);`

StefanPalashev: With a slight modification of the type:

```
#[derive(Default)]
pub struct JsonSetOptions {
    conditional_set: Option<ExistenceCheck>,
    fpha_tag: Option<FphaType>,
}

impl JsonSetOptions {
    pub fn conditional_set(self, c: ExistenceCheck) -> Self;
    pub fn fpha(self, ty: FphaType) -> Self;
}
```

Which would lead to the following calls:
```
// Plain JSON
con.json_set(k, "$", &json!({"a": 1}))?;
con.json_set_options(k, "$", &json!({"a": 1}),
    &JsonSetOptions::default().conditional_set(ExistenceCheck::NX))?;

// Flat f32 vector + FPHA
con.json_set_options(k, "$", &[1.0_f32, 2.0],
    &JsonSetOptions::default().fpha(FphaType::Fp32))?;

// f64 vector tagged as Fp16 - server returns runtime error if out-of-range
con.json_set_options(k, "$", &[1.0_f64, 2.0],
    &JsonSetOptions::default().fpha(FphaType::Fp16))?;

// Matrix + FPHA + NX
con.json_set_options(k, "$", &matrix,
    &JsonSetOptions::default()
        .fpha(FphaType::Bf16)
        .conditional_set(ExistenceCheck::NX))?;
```

nihohit: LGTM

redis/tests/support/version.rs
```diff
@@ -365,10 +390,23 @@ macro_rules! skip_if_context_does_not_support {
...
+/// ```ignore
+/// let ctx = run_test_if_version_supported!(REDIS_VERSION_CE_8_0, &[Module::Search]);
+/// ```
 #[macro_export]
 macro_rules! run_test_if_version_supported {
-    ($component:expr) => {{
-        let ctx = $crate::support::TestContext::new();
+    ($component:expr) => {{ run_test_if_version_supported!($component, &[]) }};
+    ($component:expr, $modules:expr) => {{
+        let ctx = $crate::support::TestContext::with_modules($modules);
```
StefanPalashev, line 343: @somechris Could you please review the changes I introduced to the versioning logic to ensure it still properly guards execution for module-based versions, as before (`run_test_if_version_supported!(REDIS_VERSION_CE_8_8, &[Module::Json]);`)?

redis/tests/test_basic.rs
```diff
@@ -2567,7 +2567,7 @@ mod basic {
         assert_eq!(con.object_refcount("object_key_str").unwrap().unwrap(), 1);
 
         // Needed for OBJECT FREQ and can't be set before object_idletime
-        // since that will break getting the idletime before idletime adjuts
+        // since that will break getting the idletime before idletime adjusts
```
StefanPalashev: @nihohit There was this intermittent error:
https://github.com/redis-rs/redis-rs/actions/runs/27971931226/job/82780498145, which led me to fixing this typo. Should we keep an eye on this?

redis/tests/support/version.rs
```diff
@@ -3,17 +3,42 @@
...
+pub const REDIS_VERSION_CE_7_2: Component = redis_version!(7, 2);
+pub const REDIS_VERSION_CE_7_4: Component = redis_version!(7, 4);
+pub const REDIS_VERSION_CE_8_0: Component = redis_version!(8, 0);
+pub const REDIS_VERSION_CE_8_2: Component = redis_version!(8, 1, 240);
+pub const REDIS_VERSION_CE_8_4: Component = redis_version!(8, 3, 224);
+pub const REDIS_VERSION_CE_8_6: Component = redis_version!(8, 6);
+pub const REDIS_VERSION_CE_8_8: Component = redis_version!(8, 8);
 /// Numbered databases in cluster mode were introduced in Valkey 9.0.
-pub const VALKEY_VERSION_CE_9_0: Component = ("valkey", (9, 0, 0));
+pub const VALKEY_VERSION_CE_9_0: Component = valkey_version!(9, 0);
```
somechris: To me,

```
redis_version!(8, 3, 224);
````

is harder to read than

```
("redis", (8, 3, 224))
```

as I now need to first look and understand at the `redis_version` macro.

Granted, the macro is simple.

But we need the macro for `redis`, and `valkey`. Then for the bloom filters we'd need `valkey_bloom_version` and `redis_bloom_version`, same for the json modules.

We'd end up with lots of macros with somewhat limited value.

So I'd just keep the `("redis", (8, 3, 224))` and the like.

File: redis/tests/support/version.rs (hunk 5 above)
Comment (somechris) [reply redis/tests/support/version.rs:343]: The macro's logic still looks good :+1:  (assuming we want the module loading to happen here ... there are pros and cons)

---

But the current

```
    let ctx = run_test_if_version_supported!(REDIS_VERSION_CE_8_8, &[Module::Json]);
```

would run the test on Redis v8.8 with any Json module (even Valkey's -- we don't have that in CI, but still).

If I'm not mistaken, you need Redis v8.8+ and additionally also Redis' Json module 8.8+. So both have to be 8.8+.

So I think you should instead add

```
REDIS_JSON_8_8 = ("ReJSON", (8, 8, 0))
```

to `version.rs`

and then guard the test by

```
let ctx = run_test_if_version_supported!([&[REDIS_VERSION_CE_8_8], &[REDIS_JSON_8_8]], &[Module::Json]);
```

File: redis/tests/test_basic.rs (hunk 6 above)
Comment (StefanPalashev) [reply redis/tests/test_basic.rs]: Pull request that addresses the issue - https://github.com/redis-rs/redis-rs/pull/2160

Comment (nihohit) [reply redis/tests/test_basic.rs]: Thanks, that was (hopefully) fixed by https://github.com/redis-rs/redis-rs/pull/2160

**Comment by nihohit** [comment]
@StefanPalashev sorry for the hassle, can you please rebase?

**Comment by somechris** [comment]
Just to keep things linked and get number of open bugs down .... landing this change seems to also address #721 and #1183 🎉  Could you please add

```
Closes #721
Closes #1183
```

to your commit message?

(And maybe also mention there that you add support for NX and XX?)

File: redis/src/commands/json.rs (hunk 1 above)
Comment (nihohit) [reply redis/src/commands/json.rs]: do we want to use the structure used in https://github.com/redis-rs/redis-rs/pull/2193, slightly break the API for V2, and present a better API that will hold in the long term?

Comment (StefanPalashev) [reply redis/src/commands/json.rs]: Isn't this the very same structure? We have both `json_set` and `json_set_options`.

File: redis/tests/support/version.rs (hunk 7 above)
Comment (StefanPalashev) [reply redis/tests/support/version.rs]: Thank you for the comment! The aim of these macros and the change was to reduce the code duplication, specifically having to type `redis` / `valkey` multiple times. These changes have been discarded

File: redis/tests/support/version.rs (hunk 5 above)
Comment (StefanPalashev) [reply redis/tests/support/version.rs:343]: Guarding the test like this:
1. Doesn't quite work - The intent is guard for `server version ≥ 8.8` **AND** `ReJSON ≥ 8.8`,
which requires the `&[&[Component]]` conjunctive form whose `From` implementations live in `version.rs`.
The problem is the inner array references. When we write `&[REDIS_CE_8_8]`, its type is `&[Component; 1]` - a reference to a fixed-size array of length 1, not a slice `&[Component]`. 
So the literal `[&[REDIS_CE_8_8], &[REDIS_JSON_8_8]]` has a type of `[&[Component; 1]; 2]`.
As of now the only implementations that exist are for `[&[Component]; N]` (slices), via `matcher_array_impls!`. Rust's unsized coercion `(&[T; N] -> &[T])` is not applied to the inner elements inside an array literal so nothing at that position forces the coercion, so the array keeps its `&[Component; 1]` element type and matches no `From` implementation. This causes the compiler to report that there is an unsatisfied trait: 

> the trait bound ComponentMatcher<'_>: From<[&[(&str, (u32, u32, u32)); 1]; 2]> is not satisfied

If the inner &[] are dropped (i.e. `[REDIS_CE_8_8, REDIS_JSON_8_8]`), this would compile, but it would result in a disjunctive `OR`, which is not semantically correct in this case.

To fix it, the inner array-refs can be forced to coerce to slices by annotating the 1st element. Then the type will become `&[Component]` - (`run_test_if_version_supported!([&[REDIS_CE_8_8] as &[_], &[REDIS_JSON_8_8]], &[Module::Json]);`)

2. Introduces a code smell as there is a duplication in the `&[REDIS_JSON_8_8]], &[Module::Json]` part of the call. This doesn't seem a clear enough API. The module should be the single source of truth and it shouldn't be named twice. The general principle is that data that belongs together should be represented together. Grouping each `Module` with its corresponding version requirement will make the API both safer and easier to read.

I will come up with a follow-up pull request, that would suggest fixes for both of these problems. As of now, there is no need to specifically guard against the module's version, so this pull request can be merged as is if there are no other problems / comments to address.

Comment (somechris) [reply redis/tests/support/version.rs:343]: > Doesn't quite work - The intent is guard for server version ≥ 8.8 AND ReJSON ≥ 8.8,

Just tell Rust that the first inner part is of indefinite size by adding
`[..]` (same as it is used in the corresponding tests). With that hint it compiles fine.

```
run_test_if_version_supported!([&[REDIS_CE_8_8][..], &[REDIS_JSON_8_8]]);
```

> Introduces a code smell as there is a duplication in the &[REDIS_JSON_8_8]], &[Module::Json]

A guard of `[&[REDIS_JSON_8_8]]` of course implies `Module::Json`.

But what if the condition gets more complicated .. something like `[&SOME_COND, &[REDIS_JSON_8_8, SOME_OTHER_COND]]`?

I think that the current solution is probably good enough. Inferring modules reliably can be done. But I'm not sure the gain is woth the time/energy.

But if you find a good and simple solution of infering the modules, I'm all for it :+1:

Comment (StefanPalashev) [reply redis/tests/support/version.rs:343]: The idea is having the requirements for the modules wrapped in a type. Something like:
```
pub struct ModuleRequirement {
    module: Module,
    min_version: Option<Version>,
}
```

With it, the module part will be passed to the context's constructor, while the versioning part (if any) will be checked against the component matcher once the context has been built.

> A guard of `[&[REDIS_JSON_8_8]]` of course implies `Module::Json`.

It does, but it doesn't prevent from providing `REDIS_JSON_x_x` and `Module::Search` for example.

Comment (somechris) [reply redis/tests/support/version.rs:343]: If there was only Redis modules, you of course have a point.

But with Valkey, Redis (maybe at some point KeyDB/Dragonfly), we have multiple providers.

And `Module` with `Version` is not specific enough.

`Module::Json` means "load the json shared object", but is ignorant about flavor (Valkey/Redis) or version. Only the CI matrx knows. And that's outside of the testing framework. During loading all modules are the same.

`Redis_JSON_8_8` means the `json` module from Redis in version >=8.8. This is a specific but does not say how it gets loaded.

The current approach of separating version requirements and loading modules at some places duplicates a bit (like in the case of loading a `json` module and requiring it to be Redis' module >=8.8.0).

But it keeps it simple and treats all versions the same. Redis server, Valkey server, modules.

This allows to keep the overall logic simple.

To me KISS weighs heavier here than DRY.

That said, I'm not the one to decide :-)

---

What about splitting the version refactoring off?

What about preparing the JSON patch without messing with `version` (only add `Redis_JSON_8_8`).
That should allow to formulate the needed JSON code changes.
So the relevant code can land.

Once it did and you still feel strongly about the version refactor, do that in a separate PR.

That way, the JSON code can land and we still can discuss what you'd change.

WDYT?

Comment (StefanPalashev) [reply redis/tests/support/version.rs:343]: This is also my suggested approach, without even adding `Redis_JSON_8_8`, since these tests rely on the version of the server. The assumption is that if the server's version is high enough, the loaded JSON module will support these changes.

Comment (StefanPalashev) [reply redis/tests/support/version.rs:343]: Relevant pull request that suggests the discussed changes: https://github.com/redis-rs/redis-rs/pull/2197

Comment (somechris) [reply redis/tests/support/version.rs:343]: > This is also my suggested approach, without even adding Redis_JSON_8_8, since these tests rely on the version of the server.

I'm not sure I follow. If "these tests only rely on the version of the server" was true ... that would mean that they'd pass on Redis 8.8.0 with an older `rejson` module ... say `rejson` 8.6.0 :-) But that seems wrong, as the FPHA API only got added in [`f6e43a9`](https://github.com/RedisJSON/RedisJSON/commit/f6e43a97df8599581012151ffad796be397bf420). The first tag that includes that commit is `rejson` `8.7.80`.

So it depends on the version of `rejson`.

Hence, restricting on the Redis version alone is not good enough. One also needs to make sure that the `rejson` module is >=8.7.80 (or 8.8.0 for practical purposes)

And from my tests, it also depends on the version of the Redis server. Because Redis 7.0.0 does not load the `rejson` module v8.8.0. Redis 8.0.6 can load it (I have not tested any version in between)

YMMV, but from my point of view, you need to check both the Redis server version and the `rejson` module version.

File: redis/src/commands/json.rs (hunk 1 above)
Comment (nihohit) [reply redis/src/commands/json.rs]: no, there isn't a `zinterstore_options`, only `zinterstore` and `zinterstore_with_weights`, both take an options object.

File: redis/tests/support/version.rs (hunk 5 above)
Comment (StefanPalashev) [reply redis/tests/support/version.rs:343]: The last commit has been amended to incorporate the suggested changes.

File: redis/src/commands/json.rs [hunk 8]
```diff
@@ -302,3 +313,203 @@ impl<T> JsonCommands for T where T: ConnectionLike {}
 
 #[cfg(feature = "aio")]
 impl<T> JsonAsyncCommands for T where T: crate::aio::ConnectionLike + Send + Sized {}
+
+/// Storage-precision tag for the `FPHA` form of `JSON.SET`.
+///
+/// Applied via [`JsonSetOptions::fpha`] instructs the server to pack any floating-point arrays in the payload using the chosen lane precision.
+/// Values that fall outside the chosen type's representable range cause the server to reject the command with `ERR value out of range for <TYPE>`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum FphaType {
```
nihohit, line 323: #[non_exhaustive]

StefanPalashev: Applied with the latest amend (https://github.com/redis-rs/redis-rs/commit/c4089f75e1bc626b7cc16175a9df54213258e20f)

nihohit:
Thanks!

Issue #2085, opened 2026-05-20
Script Execution after Manual Failover in Cluster stopped working


kamulos:
Beginning with c5d4d276ebe5bd49badaa1ee7f36e77abb22f90a a failover test stopped for me. The test works roughly like this:

1. `redis-rs` opens a connection to a cluster of 3 shards, each having one replica
2. the test script logs onto each replica shard and executes `CLUSTER FAILOVER`
3. then `redis-rs` executes a script which results in `ReadOnly: You can't write against a read only replica. script: 6a73544f9b04a2523592c030c911170ad5315196, on @user_script:49.`

I am not sure about the exact mechanics here, but I have a theory:

- I do not have `read_from_replicas` set, so before no connection enabled `READONLY` for me
- Now `READONLY` is enabled for each connection
- When the failover occurs a primary will switch to replica, but `redis-rs` will still think it can write. Redis does not seem to close the connections on a manual failover.
- For some reason a script execution will return the Error answer `READONLY` instead of `MOVED` to which `redis-rs` does not react

I tested it with Valkey 8.0 and 9.0

nihohit:
Thanks for reporting this!

off the cuff, it sounds like `READONLY` errors should trigger a topology refresh, but it's a bit hard to say from this description. Do you think you can translate this scenario to a redis-rs test which will fail, and I'll try to make it pass?

kamulos:
I think this should be it:

```rust
    #[async_test]
    async fn async_cluster_failover_script() {
        let cluster = TestClusterContext::new_with_config(RedisClusterConfiguration {
            num_nodes: 2,
            num_replicas: 1,
            ..Default::default()
        });
        let mut connection = cluster.async_connection().await;

        let script = redis::Script::new(r#"return redis.call("SET", "key", "value")"#);

        let _: () = script
            .key("key")
            .invoke_async(&mut connection)
            .await
            .unwrap();

        let string: String = redis::cmd("CLUSTER")
            .arg("NODES")
            .query_async(&mut connection)
            .await
            .unwrap();

        println!("{string}");
        println!("performing failover now...");

        connection
            .route_command(
                redis::cmd("CLUSTER").arg("FAILOVER").clone(),
                RoutingInfo::SingleNode(SingleNodeRoutingInfo::SpecificNode(Route::new(
                    0,
                    SlotAddr::ReplicaRequired,
                ))),
            )
            .await
            .unwrap();

[...]

PR #2115, opened 2026-06-03
cluster: refresh topology and retry on READONLY errors

alexcole:
A write (e.g. EVAL) routed to a node demoted to a replica during failover returns READONLY. Previously this was NoRetry, so the client surfaced the error without refreshing its stale slot map. Now it refreshes topology and retries by slot (sync sleeps before refreshing so the retry uses the converged map), matching go-redis/predis behavior.

I hit this during a failover using Google Cloud MemoryStore and it caused an extended outage.

### DIFF
--- a/redis/src/cluster_handling/async_connection/request.rs
+++ b/redis/src/cluster_handling/async_connection/request.rs
@@ -263,6 +263,19 @@ pub(crate) fn choose_response<C>(
             (retry, PollFlushAction::RebuildSlots)
         }
 
+        (_, RetryMethod::RefreshSlotsAndRetry) => {
+            let retry = retry_or_send!(|mut request: PendingRequest<C>| {
+                // No redirect address is available (e.g. READONLY), so re-route by slot
+                // against the refreshed topology.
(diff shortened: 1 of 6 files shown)


somechris:
Just to link things up: This looks related to the issue #2085 and the potential fix at #2089.

alexcole:
oh nice thats basically the same fix! @nihohit lmk if you want to merge yours or mine. would be good to get one of these into main

nihohit:
Thanks for the suggested fix @alexcole. Did you test it with your code? How did you verify this?

alexcole:
I did! My setup was that I ran a script that simulated my production workload (Lua scripts with several writes) and ran them against a GCP memorystore valkey instance and triggered a [simulated maintenance event](https://docs.cloud.google.com/memorystore/docs/valkey/simulate-maintenance). Before the fix I could reproduce the READONLY errors and after the fix there were 0 errors.

nihohit (approved):
Sorry, this dropped off my radar.

nihohit:
Thanks!

PR #2116, opened 2026-06-04
Makefile: Drop noop `e` from module test selector

somechris:
The `e` got added in 0891022 when migrating to `nextest`.

But it seems it was here by accident, as `e` is not a flag; it simply selects all tests that contain `e`. As all our tests contain the word `test`, it selects all tests and does not add value.

So we drop the `e`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,12 +47,12 @@ test-module:
 	@echo "===================================================================="
 	@echo "Testing RESP2 with module support enabled (currently only RedisJSON)"
 	@echo "===================================================================="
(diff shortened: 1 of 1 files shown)


somechris:
I know, just a single letter, but it cost me over an hour. So let's get rid of it :-)

(While wrapping up the Bloom filter implementation, I wanted to re-use how the `json` module gets launched. So I wanted to understand what's going on here. The `e` looked like it must serve a purpose.)

nihohit:
🤦 🤦‍♂️ 🤦‍♀️ 
Thanks for the fix, and sorry for wasting your time!

PR #2117, opened 2026-06-05
Add Bloom filter support

### LINKED ISSUE #1398: bloom support
Using Redis's bloomfilter to predict data existence is a very common use case. I wonder if redis-rs can support this.
* https://redis.io/docs/latest/develop/data-types/probabilistic/bloom-filter/
* https://github.com/RedisBloom/RedisBloom

somechris:
Bloom filter support requires the redisbloom module.

Closes #1398

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,6 +9,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   REDISRS_REDIS_JSON_PATH: "/tmp/librejson.so"
+  REDISRS_REDIS_BLOOM_PATH: "/tmp/redisbloom.so"
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.head_ref || github.run_id || github.ref }}"
@@ -33,8 +34,8 @@ jobs:
             {  db-org: redis, db-name: redis, db-version: 6.2.22 },
             {  db-org: redis, db-name: redis, db-version: 7.0.15, json-module-version: v2.6.22 },
             {  db-org: redis, db-name: redis, db-version: 7.4.9, json-module-version: v2.6.22 },
-            {  db-org: redis, db-name: redis, db-version: 8.0.6, json-module-version: bundled },
-            {  db-org: redis, db-name: redis, db-version: 8.8.0, json-module-version: bundled },
+            {  db-org: redis, db-name: redis, db-version: 8.0.6, json-module-version: bundled, bloom-module-version: bundled },
+            {  db-org: redis, db-name: redis, db-version: 8.8.0, json-module-version: bundled, bloom-module-version: bundled },
             {  db-org: valkey-io, db-name: valkey, db-version: 7.2.13, json-module-version: v2.6.22 },
             {  db-org: valkey-io, db-name: valkey, db-version: 8.0.9, json-module-version: v2.6.22 },
             # Skipping ValKey 8.1 as it's too similar to 8.0
@@ -65,6 +66,11 @@ jobs:
         if: matrix.config.json-module-version == 'bundled'
         run: |
           cp "$HOME/db/rejson.so" "$REDISRS_REDIS_JSON_PATH"
+
+      - name: Copy bundled RedisBloom into place
+        if: matrix.config.bloom-module-version == 'bundled'
+        run: |
+          cp "$HOME/db/redisbloom.so" "$REDISRS_REDIS_BLOOM_PATH"
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain/@master
@@ -141,9 +147,13 @@ jobs:
             exit 1
           fi
 
-      - name: Run module-specific tests
+      - name: Run RedisJSON tests
         if: matrix.config.json-module-version != ''
-        run: make test-module
+        run: make test-module-json
+
+      - name: Run RedisBloom tests
+        if: matrix.config.bloom-module-version != ''
+        run: make test-module-bloom
 
       - name: Cache RedisJSON
         if: matrix.config.json-module-version != '' && matrix.config.json-module-version != 'bundled'
@@ -171,6 +181,11 @@ jobs:
           db-name: ${{ matrix.config.db-name }}
           db-version: ${{ matrix.config.db-version }}
 
+      - name: Copy bundled RedisBloom into place
+        if: matrix.config.bloom-module-version == 'bundled'
+        run: |
+          cp "$HOME/db/redisbloom.so" "$REDISRS_REDIS_BLOOM_PATH"
+
       - name: Install Rust
         uses: dtolnay/rust-toolchain/@master
         with:
@@ -178,8 +193,15 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - name: start-redis
-        run: redis-server &
+      - name: Start Redis
+        run: |
+          EXTRA_OPTS=()
+          if [ -e "$REDISRS_REDIS_BLOOM_PATH" ]
+          then
+            echo "Adding module '$REDISRS_REDIS_BLOOM_PATH'"
+            EXTRA_OPTS+=("--loadmodule" "$REDISRS_REDIS_BLOOM_PATH")
+          fi
+          redis-server "${EXTRA_OPTS[@]}" &
 
       - name: basic
         run: cargo run --example basic
@@ -205,6 +227,9 @@ jobs:
         run: cargo run --example async-resp3-pub-sub --features tokio-comp,connection-manager
       - name: async-scan
         run: cargo run --example async-scan --features tokio-comp
+      - name: module bloom
+        if: matrix.config.bloom-module-version != ''
+        run: cargo run --example bloom --features bloom
 
   lint:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -42,17 +42,29 @@ test:
 	@echo "===================================================================="
 	@RUSTFLAGS="-D warnings" RUST_BACKTRACE=1 cargo nextest run --locked -p redis-test
 
+test-module-json:
+	@echo "===================================================================="
+	@echo "Testing RESP2 with RedisJSON module"
+	@echo "===================================================================="
+	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 cargo nextest run -p redis --locked --all-features -E 'test(test_module_json)'
 
-test-module:
 	@echo "===================================================================="
-	@echo "Testing RESP2 with module support enabled (currently only RedisJSON)"
+	@echo "Testing RESP3 with RedisJSON module"
 	@echo "===================================================================="
-	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 cargo nextest run -p redis --locked --all-features -E 'test(test_module)'
+	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 PROTOCOL=RESP3 cargo nextest run -p redis --all-features -E 'test(test_module_json)'
+
+test-module-bloom:
+	@echo "===================================================================="
+	@echo "Testing RESP2 with RedisBloom module"
+	@echo "===================================================================="
+	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 cargo nextest run -p redis --locked --all-features -E 'test(test_module_bloom)'
 
 	@echo "===================================================================="
-	@echo "Testing RESP3 with module support enabled (currently only RedisJSON)"
+	@echo "Testing RESP3 with RedisBloom module"
 	@echo "===================================================================="
-	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 PROTOCOL=RESP3 cargo nextest run -p redis --all-features -E 'test(test_module)'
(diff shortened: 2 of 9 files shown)


redis-test/src/lib.rs
```diff
@@ -480,6 +480,35 @@ impl AioConnectionLike for MockRedisConnection {
     }
 }
 
+#[macro_export]
+macro_rules! assert_redis_result_err_code_eq {
```
nihohit: please document both macros

redis-test/src/lib.rs
```diff
@@ -480,6 +480,35 @@ impl AioConnectionLike for MockRedisConnection {
...
+
+#[macro_export]
+macro_rules! assert_redis_result_err_detail_contains {
+    ($redis_result:expr, $needle:expr) => {
+        let error = $redis_result.expect_err("RedisResult did not fail");
+        let error_detail = error.detail().expect("RedisError should have details");
+        let needle = $needle;
+        assert!(
+            error_detail.contains(&needle),
+            "Needle '{}' not found in '{}'",
```
nihohit: nit: Let's remove "Needle" - IMO `'Foo' not found in 'BAR'` is clear enough, and I'd prefer to avoid the usage of culture-specific expressions, even common ones like this, if they don't provide more context.

redis-test/src/lib.rs
```diff
@@ -480,6 +480,35 @@ impl AioConnectionLike for MockRedisConnection {
     }
 }
 
+#[macro_export]
+macro_rules! assert_redis_result_err_code_eq {
+    ($redis_result:expr, $expected_code:expr) => {
```
nihohit: why are macros needed instead of functions?

.github/workflows/rust.yml
nihohit, line 42: mabye we should use the valkey BF in CI too? AFAIS it's the same API.

redis/src/lib.rs
```diff
@@ -738,6 +739,13 @@ pub use commands::acl;
 #[cfg_attr(docsrs, doc(cfg(feature = "aio")))]
 pub mod aio;
 
+#[cfg(feature = "bloom")]
+#[cfg_attr(docsrs, doc(cfg(feature = "bloom")))]
+pub use crate::commands::{
```
nihohit: maybe we should add these under a `bloom` mod? I feel like we're pushing too many types to the top level module.

redis-test/src/lib.rs
somechris: Done in 6fe27c7

redis-test/src/lib.rs
somechris: Done in 6fe27c7

redis-test/src/lib.rs
somechris: Old habit because seeing `assert_...` calls with and without `!` side by side in tests looks like something is off.

But let's have them as functions instead :+1: 

Done in 6fe27c7

.github/workflows/rust.yml
somechris: Yes, it's a very similar API (e.g.: `BF.INFO` return types differ a bit and some such).

But the real reason I did not include that yet is some module/versioning/bundling mayhem.

Nothing that cannot be solved. But I'm still trying out which approach feels easiest while being concise enough.

Because at the end of the day, we might want to guard tests not only by the module being present or not, but whether it's a valkey or redis _module_ and their different versions.

E.g.: The PR for `JSON.SET`'s new FPHA support is functionality of the `redisjson` module (not Redis itself). So running the tests should be guarded by the maker of the JSON module (`redisjson` vs. `valkey-json`) and the module's version. But right now, the PR instead guards by the Redis version. We don't have tooling for anything else :shrug: 

For starters, a simple Redis version guard is probably good enough. That's why in the PR I asked Stefan if they have opinions around that.

But I'm not sure if we want more precise and explicit control over that than just implict requirements which server we can pair with which modules in CI.

So I'm toying a bit with guarding by module version and module name. It's not quite there yet and not quite done yet. So I held `valkey-bloom`'s support for now.

Do you have opinions or expectations around that? :thinking:

redis/src/lib.rs
somechris: Done in 6fe27c7

.github/workflows/rust.yml
nihohit: that's reasonable. I agree that valkey-bloom can wait for a separate change.

I think that using `MODULE LIST` to add similar checks to the module should be fairly simple, but I've never tried that. The bigger issue, which hampers us with valkey too, is how to test for the slight valkey/redis differences. Just like we should start checking for `valkey_version` and not only `redis_version`.

redis/src/bloom.rs
```diff
@@ -0,0 +1,555 @@
...
+///
+/// `RESP2` and `RESP3` respond with different structures, so we need to abstract that difference
+/// away to make the info type query usage simple. Note that it derefs to `i64` for ease of use.
+#[derive(Debug, PartialEq, Eq)]
+pub struct BloomFilterInfoTypeResponse {
+    value: i64,
+}
+
+impl Deref for BloomFilterInfoTypeResponse {
+    type Target = i64;
```
somechris: Marking as draft for now, because valkey_bloom requires a float here and I'd better change already here to avoid backwards-incompatible changes in the valkey-bloom commit.

.github/workflows/rust.yml
somechris: `MODULE LIST` partly works, just as a plain `INFO` (which we already run to get the version). But the thing is that some modules report the same name for the valkey and redis module.

E.g.: the `INFO` for redis-bloom v8.0.1 is
```
module:name=bf,ver=80001,api=1,filters=0,usedby=[],using=[],options=[]
```

and for valkey-bloom v1.0.0 it's

```
module:name=bf,ver=10000,api=1,filters=0,usedby=[],using=[],options=[]
```

So both report their name as `bf`. (Similarly for the `search` modules)

Simplest thing would be to assume that on Redis, we only run `redis*` modules and on Valkey only `valkey-*` modules. But we currently test Valkey with `redis-json`. So that assumption does not even hold right now.

Also tests would read strange if we would limit on `bf` twice in different contexts, like

```
let ctx = run_test_if_version_supported([
   ["redis:v8.8.0", "bf:v8.0.1"],
   ["valkey:v9.0.1", "bf:1.0.0"],
]);
```

Because, as reader, I'd wonder why `bf` requires so vastly different versions.

It would be simpler if valkey's modules prefix with `valkey` and Redis' with `redis`.

Then we could do something like
```
let ctx = run_test_if_version_supported([
   ["redis:v8.8.0", "redis-bf:v8.0.1"],
   ["valkey:v9.0.1", "valkey-bf:1.0.0"],
]);
```

which would read better. But currently, I don't see a good way to disambiguate between Redis' and Valkey's `bf`.

We could hide the naming issue behind constants:

```
const REDIS_BF_8_0_1 = ("bf", (8, 0, 1))
const VALKEY_BF_1_0_0 = ("bf", (1, 0, 0))

let ctx = run_test_if_version_supported([
   [REDIS_8_8_0, REDIS_BF_8_0_1],
   [VALKEY_9_0_1, VALKEY_BF_1_0_0],
]);
```

and choose the CI matrix accordingly. But that's not clean either.

(`MODULE LIST` allows to get the `.so` name, so we could disambiguate. But relying on that seems brittle, especially as we rename files in CI to simplify code paths)

Having thought about that for a few days, I think I'll ask the Valkey people if they have a good way :thinking:

somechris: As discussing that in this PR is probably too hidden, I've started #2118

redis/src/bloom.rs
somechris: Done in f83b265

somechris:
As this PR is hanging since a week, I wanted to briefly check in if there is still something missing? (I think I addressed all comments) :heart:

nihohit:
Sorry, I had other PRs (and the `time` crate fiasco) which demanded my review time. I'll get to completing the review, but just making sure - is this waiting for the test changes PR-chain? is this affected by it?

somechris:
> Sorry, I had other PRs [...]

Sure.

Sorry, I just wanted to avoid each one of us waiting on the other :smile: 

My apologies, if that came across too pushy :disappointed: 

I know that we're both doing that in our free time and also that the review load increased quite a bit as of late.

Thank you for your time, for coping with the reviews, and for runnning `redis-rs` in general :heart:

There is no need to rush things here.

> is this waiting for the test changes PR-chain? is this affected by it?

This PR here is only doing the Redis module and should be orthognal to the test/version PRs.

(The Valkey module counterpart is waiting on this PR and the test/version PRs, but I'm holding that back until the relevant parts landed)

redis-test/src/lib.rs
```diff
@@ -488,6 +489,52 @@ impl AioConnectionLike for MockRedisConnection {
...
+/// * `expected_code` - The error code that `redis_result` has to have.
+///
+/// # Panics
+///
+/// Panics unless `redis_result` is `Err` with `expected_code` as error code.
+pub fn assert_redis_result_err_code_eq<T: Debug>(
+    redis_result: RedisResult<T>,
+    expected_code: &str,
+) {
+    assert_eq!(
```
nihohit: assert_matches is simpler (and IMO doesn't need a function)

redis-test/src/lib.rs
```diff
@@ -488,6 +489,52 @@ impl AioConnectionLike for MockRedisConnection {
...
+/// # Panics
+///
+/// Panics unless `redis_result` is `Err` with details that contain `search_expr`.
+pub fn assert_redis_result_err_detail_contains<T: Debug>(
+    redis_result: RedisResult<T>,
+    search_expr: &str,
+) {
+    let error = redis_result.expect_err("RedisResult did not fail");
+    let error_detail = error.detail().expect("RedisError should have details");
+    assert!(
```
nihohit: same

Makefile
```diff
@@ -42,17 +42,29 @@ test:
...
+test-module-json:
+	@echo "===================================================================="
+	@echo "Testing RESP2 with RedisJSON module"
+	@echo "===================================================================="
+	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 cargo nextest run -p redis --locked --all-features -E 'test(test_module_json)'
+
+	@echo "===================================================================="
+	@echo "Testing RESP3 with RedisJSON module"
+	@echo "===================================================================="
+	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=tcp RUST_BACKTRACE=1 PROTOCOL=RESP3 cargo nextest run -p redis --all-features -E 'test(test_module_json)'
```
nihohit, line 54: @somechris definitely not for this PR, but the `REDISRS_SERVER_TYPE` and `PROTOCOL` env vars really don't need to be env vars - we can use macros to duplicate tests with different configurations, in a single run.

nihohit (commented):
looks good, minor comments

redis-test/src/lib.rs
somechris: Ha! Our Rust styles never align :-)

I find

```
assert_redis_result_err_code_eq(con.bf_card(KEY_2), "WRONGTYPE");
```

simpler to understand and more DRY than

```
assert_eq!(con.bf_card(KEY_2).unwrap_err().code(), Some("WRONGTYPE"));
```

or

```
assert_matches!(con.bf_card(KEY_2).unwrap_err().code(), Some(_));
```

But anyway. Done in e59ef99 :+1:

redis-test/src/lib.rs
somechris: Same. Done in e59ef99 :+1:

Makefile
somechris: Ok. That sounds like a bigger change. So we're migrating towards something like in `test_hotkeys.rs`?

nihohit: yes, exactly. and as I mentioned, not for this change.

nihohit: opened this, in order to track it
https://github.com/redis-rs/redis-rs/issues/2163

redis/tests/test_module_bloom.rs
```diff
@@ -0,0 +1,342 @@
...
+    // Insert more values (`bar` was already added before, `baz` is new)
+    assert_eq!(con.bf_insert(KEY_1, &["bar", "baz"]), Ok(vec![false, true]));
+    assert_eq!(
+        con.bf_mexists(KEY_1, &["foo", "bar", "baz"]),
+        Ok(vec![true, true, true])
+    );
+
+    // Inserting to a non-existing key while prohibiting key creation should fail
+    let options = BloomFilterInsertOptions::default().nocreate();
+    assert_matches!(con.bf_insert_options(KEY_2, &["foo"], options).unwrap_err().detail(), Some(d) => {assert!(d.contains("not found"))});
```
nihohit: I think that this needs to be 
`assert_matches!(con.bf_insert_options(KEY_2, &["foo"], options).unwrap_err().detail(), Some(d) if d.contains("not found"));` - if the internal `assert` fails, the logged error won't contain the full value.

nihohit: same below

redis-test/src/lib.rs
nihohit: Yeah, my problem is more function explosion - when a function is just a combination of 2-3 calls, and I know that by the laws of combinatorics I can get many functions each covering a slight variation of the check, I prefer to have the 2-3 actions visible and avoid the many functions.

redis/tests/test_module_bloom.rs
somechris: Oh! Great catch! Done in 410a5cd :+1:

nihohit (approved):
thanks!

PR #2119, opened 2026-06-08
Add XNACK command for Redis streams

StefanPalashev:
# Add XNACK command for Redis streams

## Summary

Add support for the `XNACK` command, introduced in `Redis CE 8.8`. 
`XNACK` lets a consumer negatively acknowledge pending stream messages without leaving them idle in the PEL and waiting for `XPENDING` + `XCLAIM` / `XAUTOCLAIM` redelivery. This makes the messages immediately available to other consumers in the group through `XREADGROUP CLAIM`. Depending on the `NACK`ing option, the delivery counter is adjusted to reflect the reason for the `NACK`.

## Motivation

Applications that use **streams** sometimes cannot handle or fail at handling specific time-sensitive stream messages:
- A consumer may fail to process a claimed message due to internal issues unrelated to the message itself, such as failing to connect to a required external service.
- A resource-constrained consumer may be unable to process certain messages within an acceptable time frame.
This consumer should leave the opportunity of handling this message to other consumers in the consumer group, with minimal delay.
- A consumer may need to shut down and immediately `NACK` all unprocessed messages it had claimed before doing so.

Currently, a consumer cannot `NACK` a message. It can either **ACK** it (`XACK` - marking it as `processed`) or leave it pending (which requires other consumers to handle via `XPENDING`, `XCLAIM / XAUTOCLAIM` or `XREADGROUP ... CLAIM`) and may introduce a long, unnecessary delay. 

`NACK` is beneficial in all of the cases above, as it allows a consumer to make the target messages immediately available for consumption by other consumers in the consumer group.

## Syntax

`XNACK key group <SILENT|FAIL|FATAL> IDS numids id [id ...]`

#### NACK modes

| `streams::StreamNackMode` | Purpose |
|---|---|
|`SILENT` | Used when there are internal errors or the consumer shuts down, not because the message is problematic. The delivery counter of each specified message is decremented by 1 (to undo the increment that occurred when the message was added to the group's PEL).|
|`FAIL` | Used when the message causes problems for the current consumer but may succeed for other consumers (e.g., requires more resources than available). The delivery counter stays the same (it was already incremented by 1 when added to the group's PEL).|
|`FATAL` | Used for invalid or suspected malicious messages. The delivery counter is set to `LLONG_MAX`.|

#### Response:
Returns the number of messages successfully `NACK`ed, regardless of the applied `NackMode`.

## Behavior
`NACK`ed messages are moved to the head of the consumer group's PEL in FIFO order and marked as unowned (last consumer set to ""). This makes them immediately available for claiming by other consumers via `XREADGROUP ... CLAIM` and ensures they are always prioritized over idle pending messages.

When `CLAIM` is specified, `XREADGROUP` delivers messages in the following order:

1. `NACK`ed messages, in `NACK` FIFO order
2. Pending messages idle longer than `min-idle-time`, in stream order
3. Never-delivered entries (only when reading with `>`), in stream order

## Testing
14 integration tests gated on `REDIS_VERSION_CE_8_8` that verify:
- the behavior of the command
- the state of the delivery counter after each of the `NackMode`s
- the PEL ordering
- edge cases (such as non-existent streams or groups or passing invalid message identifiers)

## Compatibility
New code paths are gated on the existing `streams` feature.
No existing public API changes.
Server behavior gated to `Redis CE 8.8+` via the `REDIS_VERSION_CE_8_8` version constant.

## Relevant documentation
https://redis.io/docs/latest/commands/xnack/

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -2044,8 +2044,18 @@ implement_commands! {
     /// the next `XREADGROUP ... CLAIM`. `ids` not present in the PEL are
     /// silently skipped; the returned count reflects only ids actually NACKed.
     ///
-    /// `mode` selects how the per-message delivery counter is adjusted:
-    /// see [`streams::StreamNackMode`].
+    /// `options` carries the required NACK mode, which selects how the per-message
+    /// delivery counter is adjusted. See [`streams::StreamNackMode`] and [`streams::StreamNackOptions`].
+    ///
+    /// ```no_run
+    /// use redis::{Commands, RedisResult};
+    /// use redis::streams::{StreamNackMode, StreamNackOptions};
+    /// let client = redis::Client::open("redis://127.0.0.1/0").unwrap();
+    /// let mut con = client.get_connection().unwrap();
+    ///
+    /// let opts = StreamNackOptions::new(StreamNackMode::Fail);
+    /// let nacked: RedisResult<usize> = con.xnack("k1", "g1", &["1-1", "1-2"], &opts);
+    /// ```
     ///
     /// ```text
     /// XNACK <key> <group> <SILENT|FAIL|FATAL> IDS <numids> <id> [<id> ...]
@@ -2053,16 +2063,16 @@ implement_commands! {
     /// [Redis Docs](https://redis.io/commands/XNACK)
     #[cfg(feature = "streams")]
     #[cfg_attr(docsrs, doc(cfg(feature = "streams")))]
-    fn xnack<K: ToRedisArgs, G: ToRedisArgs, ID: ToRedisArgs>(
+    fn xnack<K: ToSingleRedisArg, G: ToSingleRedisArg, ID: ToSingleRedisArg>(
         key: K,
         group: G,
-        mode: streams::StreamNackMode,
-        ids: &'a [ID]
+        ids: &'a [ID],
+        options: &'a streams::StreamNackOptions
     ) -> (usize) {
         cmd("XNACK")
             .arg(key)
             .arg(group)
-            .arg(mode)
+            .arg(options)
             .arg("IDS")
             .arg(ids.len())
             .arg(ids)
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -2036,6 +2036,39 @@ implement_commands! {
             .take()
     }
 
+    /// Negatively acknowledge (NACK) one or more pending stream messages.
+    ///
+    /// `ids` that are present in the consumer group's PEL are moved to the
+    /// head of the PEL and marked as unowned (last consumer is set to an
+    /// empty string), so they are prioritized over idle pending messages on
+    /// the next `XREADGROUP ... CLAIM`. `ids` not present in the PEL are
+    /// silently skipped; the returned count reflects only ids actually NACKed.
+    ///
+    /// `mode` selects how the per-message delivery counter is adjusted:
+    /// see [`streams::StreamNackMode`].
+    ///
+    /// ```text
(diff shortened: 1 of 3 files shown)


redis/src/commands/mod.rs
```diff
@@ -2031,6 +2031,39 @@ implement_commands! {
...
+    /// `mode` selects how the per-message delivery counter is adjusted:
+    /// see [`streams::StreamNackMode`].
+    ///
+    /// ```text
+    /// XNACK <key> <group> <SILENT|FAIL|FATAL> IDS <numids> <id> [<id> ...]
+    /// ```
+    /// [Redis Docs](https://redis.io/commands/XNACK)
+    #[cfg(feature = "streams")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "streams")))]
+    fn xnack<K: ToRedisArgs, G: ToRedisArgs, ID: ToRedisArgs>(
```
somechris: I did not do a review, but while glancing over this PR because of #2124, it kind of looked like if  `K` and `G` should be `ToSingleRedisArg`s?

redis/src/commands/mod.rs
```diff
@@ -2031,6 +2031,39 @@ implement_commands! {
...
+        key: K,
+        group: G,
+        mode: streams::StreamNackMode,
```
nihohit: let's wrap this with an options object, so that if/when more options will be added, we won't need to add another function.

redis/tests/test_streams.rs
```diff
@@ -3026,3 +3026,396 @@ mod idempotency_tests {
...
+    use super::*;
+    use rstest::rstest;
+
+    const GROUP: &str = "xnack_group";
+    const CONSUMER: &str = "xnack_consumer1";
+    const CONSUMER2: &str = "xnack_consumer2";
+    const NON_EXISTENT_MESSAGE_ID: &str = "9999999-0";
+
+    // Create a stream with `n` entries delivered to `consumer` and return the list of delivered ids.
+    fn xnack_setup_pending(
```
nihohit, line 3057: this looks similar to `create_group_add_and_read`. why not reuse that?
and I see that multiple tests have the 
xgroup_create_mkstream, xadd_keyrange, xread_options order - maybe they should be converted to use `create_group_add_and_read`, too?

nihohit (commented):
I need to go over the tests in more details, but from a high level this looks ok

redis/tests/test_streams.rs
StefanPalashev: The `create_group_add_and_read ` helper is hardcoded with fixed group, consumer and 10 entries. The `xnack` tests intentionally vary the number of entries depending on the scenario. Furthermore, the ordering tests (fifo_ordering, three_tier_delivery_order) are much easier to understand with less entries.

If you strongly prefer using the helper for consistency, we can probably refactor it and update these tests accordingly.

redis/src/commands/mod.rs
StefanPalashev: Wouldn't this make this API inconsistent with the others that expose `_options` signatures?

redis/src/commands/mod.rs
StefanPalashev: The change has been applied to all of the arguments. It makes sense to also have the `ID` parameter as `ToSingleRedisArg`.

redis/src/commands/mod.rs
nihohit: it would, and that's acceptable. See the change to `zinterstore` in v2 - the `with_options` variant makes sense for "small" functions with obvious default, like SET. Here, a command that is built with complex options from the get-go, doesn't need the 2 variants.

StefanPalashev: The mode parameter has been wrapped into an options (`StreamNackOptions`) struct.

nihohit (approved):
Thanks!

PR #2120, opened 2026-06-09
Keep dispatching requests when repairing a connection to a replica

virratanasangpunth:
Previously, the cluster client would stop dispatching new requests when performing a recovery.
A recovery can be many things: repairing a connection, or refreshing a topology.

In this commit, when repairing a connection to a replica, the cluster client will continue 
dispatching requests. Requests to an affected replica will be routed around to another replica or a primary.

If a primary is dead within a shard however, we fallback to the prev behavior where we stop dispatch,
and perform a topology refresh. (prev behavior was to fallback to a random node that may not be in the same shard)

To achieve this, we modify the `Sink::poll_flush` loop to not block on the recovery future. Previously in this central loop we would complete the recovery future first before dispatching requests from pending to in-flight. Another important change to PR is to modify `get_connection`: it will do a smarter fallback to a node within a shard as opposed to falling back to a random connection.


virratanasangpunth:
Hi @nihohit this is my first PR here. Could you approve the workflow run and take a look when you have a chance? Thanks!

nihohit:
can you please add to the description a high-level description of the architectural change?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -431,7 +431,58 @@ where
...
+    fn into_conn(self) -> Option<C> {
+        match self {
+            ConnState::Connected(conn) | ConnState::Reconnecting(conn) => Some(conn),
+            ConnState::Connecting => None,
+        }
+    }
+}
+
+fn mark_reconnecting<C>(connections: &mut ConnectionMap<C>, addr: NodeAddress) {
+    let state = match connections.remove(&addr) {
```
nihohit: instead of `remove` and `insert`, you can use [`entry`](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#method.entry) and modify it in-place.

virratanasangpunth:
> can you please add to the description a high-level description of the architectural change?

@nihohit I've added under the "To achieve this..." paragraph.

virratanasangpunth:
`make test` is now passing locally after the [fix](https://github.com/redis-rs/redis-rs/pull/2120/changes/9d9127d534dfcba5ac18c6ae25243b57f544f005).

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -788,67 +843,72 @@ where
...
-        let addr_conn_option = match conn {
-            Some((addr, Some(conn))) => Some((addr, conn)),
-            Some((addr, None)) => self
-                .connect_check_and_add(&addr)
-                .await
-                .ok()
-                .map(|conn| (addr, conn)),
-            None => None,
+            InternalSingleNodeRouting::Random => return self.random_connection().await,
+            InternalSingleNodeRouting::SpecificNode(route) => route,
```
nihohit: was there a reason to reoder the variants? this increases the size of the git diff, and makes review more difficult

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -788,67 +843,72 @@ where
...
-                let read_guard = self.conn_lock.read().await;
-                if let Some((random_addr, random_conn)) = get_random_connection(&read_guard.0) {
-                    drop(read_guard);
-                    (random_addr, random_conn)
-                } else {
-                    return Err(
-                        (ErrorKind::ClusterConnectionNotFound, "No connections found").into(),
-                    );
-                }
+        let read_guard = self.conn_lock.read().await;
```
nihohit, line 885: why move the read guard into the match statement, instead of taking the read guard over the whole function?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -788,67 +843,72 @@ where
...
+            .slot_addr_for_route(&route, self.routing_strategy.as_deref())
+            .cloned();
+
+        if let Some(ref addr) = preferred
+            && let Some(conn) = read_guard.0.get(addr).and_then(ConnState::connected)
+        {
+            return Ok((addr.clone(), conn));
+        }
+
+        for cand in read_guard.1.shard_fallback_addrs(&route) {
```
nihohit: cand -> candidate

nihohit: use `filter` and `find` instead of a for loop - it conveys intention better.

nihohit (commented):
I started reviewing, and then reached the tests - you've added almost 150 of testing infra with hardly any documentation or justification, and it isn't mentioned in the PR's description. Please explain what is it you've added, and why this large addition is necessary beyond the existing testing infra.

redis/tests/test_cluster_async.rs
```diff
@@ -1125,6 +1118,7 @@ mod cluster_async {
             name,
             {
                 move |cmd: &[u8], port| {
+                    println!("cmd: {} port: {port}", String::from_utf8_lossy(cmd));
```
nihohit: remove

redis/tests/test_cluster_async.rs
```diff
@@ -1951,40 +1945,172 @@ mod cluster_async {
     }
 
     #[test]
-    fn test_async_cluster_io_error() {
```
nihohit, line 2078: why was the original test removed?

redis/tests/test_cluster_async.rs
```diff
@@ -1951,40 +1945,172 @@ mod cluster_async {
...
+            handler: _handler,
+            ..
+        } = MockEnv::with_client_builder(
+            ClusterClient::builder(vec![&*format!("redis://{name}")]).retries(0),
+            name,
+            MockCluster::new(name, respond_startup_two_nodes)
+                .with(
+                    node(6379)
+                        .respond_startup_calls()
+                        .then_broken_pipe_error()
```
nihohit: so this node always succeeds reconnecting, but then disconnects on the first request? weird behavior. why not succeed connection once and then disconnect forever?

redis/tests/test_cluster_async.rs
```diff
@@ -1951,40 +1945,172 @@ mod cluster_async {
...
+            async_connection: mut connection,
+            handler: _handler,
+            ..
+        } = MockEnv::with_client_builder(
+            ClusterClient::builder(vec![&*format!("redis://{name}")])
+                .retries(1)
+                .read_routing_strategy(RandomReplicaStrategy),
+            name,
+            MockCluster::new(name, respond_startup_with_replica)
+                .with(
```
nihohit: where are the replicas in this setup?

nihohit (commented):
some comments on the tests

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: The old code requires manual `drop(read_guard)` calls for certain match branches (Random and Redirect) otherwise you may end up in a deadlock because you re-acquire a read-lock across awaits (`random_connection` and `get_redirected_connection` re-acquire the read lock)

Scoping the guard inside each arm removes this foot-gun.

redis/tests/test_cluster_async.rs
virratanasangpunth: Oh whooops 😓; I had to include this to help with test client behavior and forgot to remove it.

redis/tests/test_cluster_async.rs
virratanasangpunth: Previously on main, each command would retry the reconnect. This could cause a reconnect storm. In this PR we've moved the reconnect to the recovery future so we do one reconnect-at-a-time per node. The original test `test_async_cluster_io_error` was removed since it was depending on the "per-request" retry behavior. Instead of this test I've added `test_async_cluster_io_error_recovers_via_repair` which demonstrates that the subsequent "GET test" command succeeded.

So I think this is a regression from main: on main with `retries(2)`, a command succeeded transparently after reconnect. However in csr branch there is a race where retries might happen faster than the connection can be repaired.

With that said the connection will be repaired eventually and GET command will eventually succeed.

I can think about this a bit more; I'm just sharing my train of thoughts here.

redis/tests/test_cluster_async.rs
virratanasangpunth: Replica is configured in `respond_startup_with_replica` at line 2092, where it starts primary at 6379 and replica at 6380.

redis/tests/test_cluster_async.rs
virratanasangpunth: > why not succeed connection once and then disconnect forever?

This is in fact the mock behavior: the first `respond_startup_calls` lets the initial handshake through, then `.then_broken_pipe_error().forever()` makes it unavailable forever.

redis/tests/support/mock_cluster.rs
```diff
@@ -202,6 +202,150 @@ pub fn respond_startup_with_replica_using_config(
...
+    dead: bool,
+}
+
+pub struct MockCluster {
+    name: String,
+    startup: fn(&str, &[u8]) -> Result<(), RedisResult<Value>>,
+    nodes: Vec<NodeBehavior>,
+}
+
+impl MockCluster {
```
virratanasangpunth: > I started reviewing, and then reached the tests - you've added almost 150 of testing infra with hardly any documentation or justification, and it isn't mentioned in the PR's description. Please explain what is it you've added, and why this large addition is necessary beyond the existing testing infra.

@nihohit Fair point, I should have called this out in the description. I ran into a readability issue when trying to mock the node behavior, hence the need for this new test code.

To demonstrate the motivation, this is the mock code that I need to write for `test_async_cluster_reconnect_does_not_stall_healthy_shard` test. In this test I need two nodes where: 
1. both nodes respond to the initial handshake successfully.
2. but then the first node (6379) has to return an error for subsequent commands/handshakes.

```
let primary_down = Arc::new(AtomicBool::new(false));
let MockEnv { runtime, async_connection: mut connection, .. } =
    MockEnv::with_client_builder(
        ClusterClient::builder(vec![&*format!("redis://{name}")]).retries(0),
        name,
        {
            let primary_down = primary_down.clone();
            move |cmd: &[u8], port| {
                if port == 6379 && primary_down.load(Ordering::SeqCst) {
                    return Err(Err(broken_pipe_error()));
                }
                respond_startup_two_nodes(name, cmd)?;
                match port {
                    6379 => {
                        primary_down.store(true, Ordering::SeqCst);
                        Err(Err(broken_pipe_error()))
                    }
                    6380 => Err(Ok(redis_value!("123"))),
                    other => panic!("unexpected data command on node {other}"),
                }
            }
        },
    );
```

compare that with this:

```
MockCluster::new(name, respond_startup_two_nodes)
    .with(
        node(6379)
            .respond_startup_calls()
            .then_broken_pipe_error()
            .forever(),
    )
    .with(
        node(6380)
            .respond_startup_calls()
            .then_value(redis_value!("123"))
            .forever(),
    )
    .into_handler(),
```

I can add doc comments to this Struct to make it more clear. Additionally I think this can probably be filed as its own PR before this PR#2120 if you'd like.

redis/tests/test_cluster_async.rs
nihohit: yeah, these things happen often, that's why I added
https://github.com/redis-rs/redis-rs/pull/2133

redis/tests/test_cluster_async.rs
nihohit: I see, so the combination of `then_broken_pipe_error` and `forever` has an implicit side effect.

nihohit:
> Additionally I think this can probably be filed as its own PR before this PR#2120 if you'd like

Thanks, I would appreciate that very much. Sadly I have limited time for reviews during the week, and smaller PRs are much easier and faster to review.
Additionally, if in this PR you could convert existing tests to this format, it would really help in demonstrating its value.
And thanks for the effort - I know I'm slow to respond, but it doesn't mean that I don't appreciate the time and effort this takes.

redis/tests/test_cluster_async.rs
virratanasangpunth: nice!

redis/tests/support/mock_cluster.rs
virratanasangpunth: >  Additionally, if in this PR you could convert existing tests to this format, it would really help in demonstrating its value.

I've tried adapting `MockCluster` to other tests and not all of them fit that pattern that well. So I think `MockCluster` might have been an overkill abstraction. I've removed them from this PR.

virratanasangpunth:
>  Additionally, if in this PR you could convert existing tests to this format, it would really help in demonstrating its value.

I've actually decided to remove MockCluster in this PR and use the existing pattern for writing tests. This PR's scope is smaller now.

redis/tests/test_cluster_async.rs
virratanasangpunth: Okay I've added the test `test_async_cluster_io_error` back to maintain `.retries()` semantic on main branch.

The test is passing now. The key here is to [restore](https://github.com/redis-rs/redis-rs/pull/2120/changes/f2ece8a176ed253d0259ee140b9cc8a51516f762) the per-command re-connect semantic on the main branch.

virratanasangpunth:
Okay I believe the PR is in a good shape to review now, some highlights since you've last seen it:

1. unnecessary test infra code removed in favor of existing pattern.
2. no existing test removed.

virratanasangpunth:
@nihohit gentle ping for this PR review. Thank you!

nihohit:
> I've actually decided to remove MockCluster in this PR and use the existing pattern for writing tests. This PR's scope is smaller now.

That's fine. I don't object to that proposal, so feel free to propose that PR later.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -1159,14 +1284,76 @@ where
...
 
-            inner
-                .refresh_connections_locked(&mut write_guard.0, addrs)
+                let params = &inner.cluster_params;
+                let results = stream::iter(work.into_iter().map(|(addr, prev)| async move {
+                    let res = get_or_create_conn(&addr, prev, params).await;
+                    (addr, res)
+                }))
+                .buffer_unordered(8)
+                .collect::<Vec<_>>()
```
nihohit: this means that if 1 connection has issues or is slow to reconnect, it will block all of the reconnecting connections

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -1159,14 +1284,76 @@ where
...
+                            }
+                        }
+                    }
+                }
+
+                boxed_sleep(backoff).await;
+                backoff = (backoff * 2).min(Duration::from_secs(5));
+            }
+
+            inner.resubscribe();
```
nihohit: if I understand the loop that runs before this, as long as there's even 1 reconnecting connection, `resubscribe` won't be called - so the reconnected connections will be usable, but without pubsub until they all complete?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -1159,14 +1284,76 @@ where
...
         &mut self,
         addrs: HashSet<NodeAddress>,
-    ) -> impl Future<Output = ()> + use<C> {
+    ) -> impl Future<Output = RedisResult<()>> + use<C> {
         let inner = self.inner.clone();
         async move {
-            let mut write_guard = inner.conn_lock.write().await;
+            {
+                let mut guard = inner.conn_lock.write().await;
+                for addr in addrs {
```
nihohit: if I understand this code correctly, if we entered `refresh_connection` with some connections to reconnect, and before the function completed more connections disconnected, then we won't reconnect the newly disconnected until all previous reconnected?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -1403,35 +1599,59 @@ where
...
-
-                // Give other tasks a chance to progress before we try to recover
-                // again. Since the future may not have registered a wake up we do so
-                // now so the task is not forgotten
-                cx.waker().wake_by_ref();
-                return Poll::Pending;
+            let block_dispatch = match self.as_mut().poll_recover(cx) {
+                Poll::Ready(Err(err)) => {
+                    // We failed to reconnect, while we will try again we will report the
+                    // error if we can to avoid getting trapped in an infinite loop of
```
nihohit: but AFAIS refresh_connections doesn't return errors - it does loop forever.

nihohit (commented):
I am far from finishing reviewing this - it's a complicated change. but AFAIS, the current fix has some big issues.

nihohit:
I think that this PR is made more complex and harder to review because it contains 2 changes - both an attempt to make reconnects non-blocking, and the shard-based fallback. I know that both are related and entangled, but it might be easier to review only 1 change at a time.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: Hmm yes looks like this old code comment needs to be updated.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: Correct. This is already an existing behavior on main: https://github.com/redis-rs/redis-rs/blob/main/redis/src/cluster_handling/async_connection/mod.rs#L946C1-L946C48

Happy to look more into this HoL blocking. But that will add scope to this PR.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: Not quite. The code should gracefully handle additional broken connections here: https://github.com/virratanasangpunth/redis-rs/blob/csr/redis/src/cluster_handling/async_connection/mod.rs#L1639

But there is a potential HoL blocking with `buffere_unoredered(8)` (again, main branch behavior) say like all first 8 connection attempts take forever.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: Hmm you might be right about this one. I have to admit that I don't know the pubsub code in this repo that well. Let me read up on what the fix is here.

nihohit:
@virratanasangpunth you're right that a lot of the issues exist today in the code, but my intention is to merge something that will be a considerable improvement. Considering how much complexity this fix adds, it needs to be a more complete fix than it is now.

I'm not sure that the current direction is the right one. maybe connection attempts need to be spawned, andthe reconnecting state should hole a handle or a future, and never have a reconnection in the recovery flow?
maybe `resubscribe` isn't granular enough, and we need to find per-connection subscriptions?

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: comment updated. FWIW this comment was already stale in `main` branch: the reconnect recovery future is infallible. RecoverSlots errors are broadcasted however.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: Fixed. We now `resubscribe` after each connection repair success.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: This behavior is updated in the latest revision: connection repairs now happen concurrently.

virratanasangpunth:
@nihohit sorry for getting back to you late; but I have iterated more on this PR. I believe I've addressed your earlier feedbacks regarding HoL blocking with reconnecting, plus a resubscribe concern.

To prevent HoL blocking, each reconnect has its own future that gets polled by the sink.

The PR is ready for review now.

redis/src/cluster_handling/slot_map.rs
```diff
@@ -46,6 +46,21 @@ impl SlotMap {
             .map(|addrs| addrs.slot_addr(slot, &route.slot_addr(), strategy))
     }
 
+    #[cfg(feature = "cluster-async")]
+    pub(crate) fn shard_fallback_addrs(&self, route: &Route) -> Vec<NodeAddress> {
+        let Some(addrs) = self.slots.get(route.slot()) else {
+            return Vec::new();
+        };
+        match route.slot_addr() {
+            SlotAddr::Master => vec![addrs.primary.clone()],
```
nihohit: is this a fallback? doesn't seem to match the name of the function.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -482,25 +538,29 @@ where
...
-                                cmd,
-                                routing: InternalSingleNodeRouting::Connection {
-                                    identifier: addr,
-                                    conn,
-                                }
-                                .into(),
+                read_guard
+                    .0
+                    .get(addr)
+                    .and_then(ConnState::connected_or_reconnecting)
```
nihohit, line 546: why do we use here `connected_or_reconnecting`? isn't the connection in reconnecting state assumed to be disconnected?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -788,67 +848,73 @@ where
...
                 // redirected requests shouldn't use a random connection, so they have a separate codepath.
                 return self.get_redirected_connection(redirect).await;
             }
             InternalSingleNodeRouting::ByAddress(address) => {
-                if let Some(conn) = read_guard.0.get(&address).cloned() {
-                    return Ok((address, conn));
-                } else {
-                    return Err((
+                let read_guard = self.conn_lock.read().await;
+                return match read_guard.0.get(&address).and_then(ConnState::connected) {
```
nihohit, line 873: why in execute_on_multiple_nodes you sent on reconnecting state, and here only on connected state?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -788,67 +848,73 @@ where
...
+            && let Some(conn) = read_guard.0.get(addr).and_then(ConnState::connected)
+        {
+            return Ok((addr.clone(), conn));
         }
-        .map(|addr| {
-            let conn = read_guard.0.get(&addr).cloned();
-            (addr, conn)
-        });
+
+        let fallback = read_guard
```
nihohit, line 899: please add a one-line comment explaining this

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -788,67 +848,73 @@ where
...
+            .1
+            .shard_fallback_addrs(&route)
+            .into_iter()
+            .find_map(|candidate| {
```
nihohit, line 903: can we reorder these based on the used `ReadRoutingStrategy`? ATM the fallback order will be constant.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -877,48 +943,83 @@ where
...
             }
-            .await;
-            if result.is_ok() {
-                break;
+        }
+
+        // Second pass is to attempt to repair connections on-demand as we iterate thru each one.
+        if !found {
+            for (addr, state) in &mut *connections {
+                let prev = std::mem::replace(state, ConnState::Connecting).into_conn();
```
nihohit, line 997: shouldn't we skip here the connected connections, since we already tested them?

redis/src/cluster_handling/slot_map.rs
```diff
@@ -46,6 +46,21 @@ impl SlotMap {
...
+            SlotAddr::ReplicaOptional | SlotAddr::ReplicaRequired => {
```
nihohit: ReplicaRequired's fallback shouldn't contain the primary

redis/src/cluster_handling/slot_map.rs
```diff
@@ -46,6 +46,21 @@ impl SlotMap {
             .map(|addrs| addrs.slot_addr(slot, &route.slot_addr(), strategy))
     }
 
+    #[cfg(feature = "cluster-async")]
+    pub(crate) fn shard_fallback_addrs(&self, route: &Route) -> Vec<NodeAddress> {
```
nihohit, line 51: please add unit tests for this function

nihohit:
Claude comment #1:

PR description vs. implementation for dead primaries. The description says a dead primary "fallback[s] to the prev behavior where we stop dispatch, and perform a topology refresh." In the code, a dead primary takes the same background-repair path (PollFlushAction::Reconnect → start_repairs); the request bounces to a random node, gets MOVED, and only then (indirectly) triggers a slot rebuild. There's no explicit primary-vs-replica branch in poll_flush. The behavior may be acceptable, but the description oversells a "stop dispatch + topology refresh" path that doesn't exist as written.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -431,7 +431,63 @@ where
     }
 }
 
-type ConnectionMap<C> = HashMap<NodeAddress, C>;
+type ConnectionMap<C> = HashMap<NodeAddress, ConnState<C>>;
+
+enum ConnState<C> {
+    // Connection has been succefully established at some point.
```
nihohit: nit: succefully -> successfully

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -431,7 +431,63 @@ where
...
+    }
+}
+
+fn mark_reconnecting<C>(connections: &mut ConnectionMap<C>, addr: NodeAddress) {
+    connections
+        .entry(addr.clone())
+        .and_modify(|state| {
+            *state = match std::mem::replace(state, ConnState::Connecting) {
+                ConnState::Connected(conn) => {
+                    log::warn!(
```
nihohit: nit: just `warn!`, no need for `log::`

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -991,10 +1091,76 @@ struct ClusterConnInner<C> {
...
+    loop {
+        let prev = {
+            let guard = core.conn_lock.read().await;
+            let in_topology = guard.1.addresses_for_all_nodes().contains(&addr);
+            let entry = guard.0.get(&addr);
+            let healthy = matches!(entry, Some(ConnState::Connected(_)));
+            let prev_conn = entry.and_then(ConnState::connected_or_reconnecting);
+            drop(guard);
+
+            if healthy {
```
nihohit: why not move this to where `healthy` is declared?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -991,10 +1091,76 @@ struct ClusterConnInner<C> {
...
+            if !in_topology {
+                core.conn_lock.write().await.0.remove(&addr);
+                break;
+            }
+            prev_conn
+        };
+
+        match get_or_create_conn(&addr, prev, &core.cluster_params).await {
+            Ok(conn) => {
+                let installed = {
```
nihohit: maybe `new` or `newly_installed`? because it's only positive if the connection was replaced, not if it was previously installed and unchanged.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -991,10 +1091,76 @@ struct ClusterConnInner<C> {
...
+                    if matches!(guard.0.get(&addr), Some(ConnState::Connected(_))) {
+                        false
+                    } else {
+                        guard.0.insert(addr.clone(), ConnState::Connected(conn));
+                        true
+                    }
+                };
+                if installed {
+                    info!("ClusterConnInner: reconnected to {addr:?}");
+                    core.resubscribe();
```
nihohit, line 1158: this means that resubscription will happen on every reconnection individually. Please add a TODO to tailor the resubscription to the reconnected node.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -1156,27 +1323,35 @@ where
...
-        async move {
-            let mut write_guard = inner.conn_lock.write().await;
-
-            inner
-                .refresh_connections_locked(&mut write_guard.0, addrs)
-                .await;
+    // Create reconnecting futures and register them into `reconnect_futures`
+    // so we can poll make at the next poll_flush iteration.
+    fn register_reconnect_futures(&mut self, addrs: HashSet<NodeAddress>) {
+        let in_progress = self.inner.conn_lock.try_read().ok();
```
nihohit: shouldn't we leave if we failed to get the lock, instead of iterating on the values?

also, the `inprogress` name is unclear here.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -1156,27 +1323,35 @@ where
...
+        for addr in addrs {
+            if let Some(guard) = &in_progress
+                && matches!(
+                    guard.0.get(&addr),
+                    Some(ConnState::Reconnecting(_) | ConnState::Connecting)
```
nihohit, line 1345: won't this mean that the same connections will be added to the loop multiple times?

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -1156,27 +1323,35 @@ where
...
+            {
+                continue;
+            }
+            self.reconnect_futures
+                .push(Box::pin(reconnect_loop(self.inner.clone(), addr)));
         }
     }
 
+    fn poll_reconnects(&mut self, cx: &mut task::Context<'_>) {
+        while let Poll::Ready(Some(())) = Pin::new(&mut self.reconnect_futures).poll_next(cx) {}
```
nihohit, line 1356: why is a loop needed here?

nihohit (commented):
a big improvement on the previous round, IMO. still some details to flesh out. I also didn't review the tests yet.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: `reconnect_futures` can have multiple futures that are ready to be polled next. By doing a loop here we ensure that we advance the futures that can be advanced.

This also mirrors how `poll_in_flight` is currently done.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: Good question on this asymmetry. This is reflecting different trade-offs between a fan-out vs a single command. For the fan-out case I don't want to return an incomplete response by omitting a node. I'd rather fail the whole batch (e.g. due to a timeout) rather than returning an incomplete response (users would have no way to distinguish between complete vs incomplete responses)

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: I think this is similar to the discussion here: https://github.com/redis-rs/redis-rs/pull/2120#discussion_r3532734882 ?

A connection in `Reconnecting` state could still work if it happens to be a transient network issue.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: In the first pass we call `query_slots` right away. In the second pass we call `get_or_create_conn` before we get to `query_slots`. That way we do an inline connection repair during refresh_slots.

A connection's state is always best-effort. A Connected connection can be actually broken. "Connected" here means the last observed state.

So basically the second pass was all the main branch used to do. In this PR where we have more info about the connection states we opportunistically use the likely-good-ones first.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: The same address can get multiple futures pushed (when the lock is contended), but there will never be multiple live repair loops. The dedupe here is opportunistic  if we can acquire the lock. The actual dedupe enforcement is done inside the `reconnect_loop` future itself where we return early if connection is being repaired by another future.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: renamed to `opportunistic_read_guard`. This is the same 🧵 as this one: https://github.com/redis-rs/redis-rs/pull/2120#discussion_r3539085795

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: Hmm so basically we want strategy-aware fallbacks. I did a quick estimate and the diff would be ~100 lines. I think we can defer this and I can file an issue for it or leave a TODO comment?

I suspect that this is a rare cluster setup where during fallbacks you have so many (fallback) nodes in a shard.

virratanasangpunth:
> Claude comment #1:
> 
> PR description vs. implementation for dead primaries. The description says a dead primary "fallback[s] to the prev behavior where we stop dispatch, and perform a topology refresh." In the code, a dead primary takes the same background-repair path (PollFlushAction::Reconnect → start_repairs); the request bounces to a random node, gets MOVED, and only then (indirectly) triggers a slot rebuild. There's no explicit primary-vs-replica branch in poll_flush. The behavior may be acceptable, but the description oversells a "stop dispatch + topology refresh" path that doesn't exist as written.

Hmmm I think Claude is splitting hair here 😅 ? You will get into slot_refresh if a primary is dead, in a subsequent poll_flush iteration. 

I still think my description, "fallback[s] to the prev behavior where we stop dispatch, and perform a topology refresh.", is still accurate. I think Claude think we're doing that as a part of get_connection fallback within the same poll_flush iteration or something?

redis/src/cluster_handling/async_connection/mod.rs
nihohit: please add a single-line doc explaining this, with the explanation from https://github.com/redis-rs/redis-rs/pull/2120/changes#r3532734882

redis/src/cluster_handling/async_connection/mod.rs
nihohit: please add a short doc explaining the difference

redis/src/cluster_handling/async_connection/mod.rs
nihohit: yes, please add a TODO (and if you don't mind, also follow through and implement it in a later PR, please :) )

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -788,67 +846,75 @@ where
...
-                        (ErrorKind::ClusterConnectionNotFound, "No connections found").into(),
-                    );
-                }
-            }
-        };
+        Err((ErrorKind::ClusterConnectionNotFound, "No connections found").into())
+    }
 
-        Ok((addr, conn))
+    async fn random_connection(&self) -> RedisResult<(NodeAddress, C)> {
```
nihohit: having both a `random_connection` and `get_random_connection` functions is confusing. choose one to keep - either replace the single call to `random_connection` with `get_random_connection` + `map_err`, or change the one other call to `get_random_connection` to take a result instead of an option.

redis/src/cluster_handling/async_connection/mod.rs
```diff
@@ -877,48 +943,83 @@ where
...
+                    break;
+                }
+                Err(err) => last_err = Some(err),
             }
-            .await;
-            if result.is_ok() {
-                break;
+        }
+
+        // Second pass is to attempt to repair connections on-demand as we iterate thru each one.
```
nihohit: thru -> through
and add a line explaining why we also try again with connected connections

redis/src/cluster_handling/async_connection/mod.rs
nihohit: If there's an answer here to 

> shouldn't we leave if we failed to get the lock, instead of iterating on the values?

I missed it. care to explain?

redis/src/cluster_handling/async_connection/mod.rs
nihohit: please add a short comment explaining this above the
```rs
if matches!(
            guard.0.get(&addr),
            Some(ConnState::Reconnecting(_) | ConnState::Connecting)
        ) {
            return;
        }
```
in `reconnect_loop`

redis/tests/test_cluster_async.rs
```diff
@@ -2112,6 +2105,199 @@ mod cluster_async {
...
+    #[test]
+    fn test_async_cluster_io_error_recovers_via_repair() {
+        let name = "test_async_cluster_io_error_recovers_via_repair";
+        let MockEnv {
+            runtime,
+            async_connection: mut connection,
+            handler: _handler,
+            ..
+        } = MockEnv::with_client_builder(
+            ClusterClient::builder(vec![&*format!("redis://{name}")]).retries(0),
```
nihohit, line 2170: IMO testing with retries and ensuring that the errors aren't surfaced is IMO more important - but we can have both tests.

redis/tests/test_cluster_async.rs
```diff
@@ -2112,6 +2105,199 @@ mod cluster_async {
...
+                .timeout(futures_time::time::Duration::from_secs(5)),
+        );
+        assert_eq!(
+            healthy.expect("healthy shard stalled behind the dead shard's reconnect"),
+            Ok(Some(123)),
+        );
+    }
+
+    #[test]
+    fn test_async_cluster_read_routes_around_dead_replica() {
```
nihohit, line 2253: we should also test ReplicaRequired routing, and ReplicaOptional when there's another replica to route to.

nihohit (commented):
Ok, the design is fine, the final comments are very shallow. I believe we're close to merging this. Thanks for keeping through the long review process :)

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: If we cannot acquire the lock, that is okay. The actual dedupe mechanism is enforced inside the `reconnect_loop` future itself.

So in the worst case when lock is not acquired, we may push more than one `reconnect_loop` futures (for the same address) into `self.reconnect_futures`. But after we poll them all will exit early except for one.

redis/src/cluster_handling/async_connection/mod.rs
virratanasangpunth: I believe the comment is already included in the current revision: https://github.com/virratanasangpunth/redis-rs/blob/csr/redis/src/cluster_handling/async_connection/mod.rs#L1331

redis/tests/test_cluster_async.rs
virratanasangpunth: We do have the test that exercises retries (and no errors are surfaced): https://github.com/virratanasangpunth/redis-rs/blob/csr/redis/tests/test_cluster_async.rs#L2071

I believe `test_async_cluster_io_error` is an already existing test.

nihohit (approved):
Thanks for the contribution!

nihohit:
@virratanasangpunth if you still want to contribute the change to the mock cluster tests, please feel welcome!

virratanasangpunth:
@nihohit thank you for your patience! I appreciate your effort reviewing this.

PR #2121, opened 2026-06-09
docs: Straighten out documentation of `InfoType`

somechris:
The documentation was spread out over the `struct` and `impl` which made it hard for IDEs to properly find and render it. We collect it into the `struct` documentation.

We document the caveat that only the last value is retained for keys that occur multiple times.


nihohit (approved):
Thanks!

PR #2125, opened 2026-06-10
Fix the windows build

nihohit:


somechris:
While we certainly want a real, proper fix :+1:, #2127 for the time being would switch the runner to `windows-2022`. This allowed to build in my [own playground](https://github.com/somechris/redis-rs/actions/runs/27276395467/job/80558745179?pr=11)

This should let CI do its thing on the other PRs until there is a real, proper fix of the windows build.

PR #2128, opened 2026-06-10
Add integration tests for subkey keyspace notifications

StefanPalashev:
# Add integration tests for Redis 8.8 subkey keyspace notifications

## Summary

This pull request adds a new test file that covers the four subkey notification channel families introduced in Redis 8.8.

## Background

Standard Redis keyspace notifications, enabled with `notify-keyspace-events`, publish on two channel families:

| Flag | Channel | Payload |
|------|---------|---------|
| `K` | `__keyspace@<db>__:<key>` | the event name (e.g. `hset`) |
| `E` | `__keyevent@<db>__:<event>` | the key name |

These notifications reveal that a key was touched and by which command, but provide no visibility into the specific **subkey(s)** involved. This means that a subscriber interested in a single field must react to every `HSET` on the hash and perform an additional read to determine whether the field it cares about was actually modified.

## What are subkey keyspace notifications?

Redis 8.8 adds four new flag letters to `notify-keyspace-events` that publish the affected subkey(s) inline, removing that round-trip:

| Flag | Channel pattern | Payload format |
|------|-----------------|----------------|
| `S`  | `__subkeyspace@<db>__:<key>`               | `<event>\|<len>:<sk>[,<len>:<sk>...]` |
| `T`  | `__subkeyevent@<db>__:<event>`             | `<keylen>:<key>\|<len>:<sk>[,...]`    |
| `I`  | `__subkeyspaceitem@<db>__:<key>\n<sk>`     | `<event>`                              |
| `V`  | `__subkeyspaceevent@<db>__:<event>\|<key>` | `<len>:<sk>[,...]`                     |

- `S` / `T` mirror the existing `K` / `E` axis (per-key channel vs. per-event channel) but enrich the payload with the affected subkeys
- `I` produces one channel per *(key, subkey)* pair so subscribers can target a single field with a single `(P)SUBSCRIBE`
- `V` factors the event and key into the channel name so the payload only carries the subkey list.

### Payload encoding

All four families use the same length-prefix grammar for the subkey list in both single-subkey and multi-subkey operations. As a result, there is no single-subkey shortcut form - a `HSET` on one field still emits `hset|6:field1`, not `hset|field1`.

#### Parsing rules:

1. Split at the first `|` (bar), which separates the header from the subkey list. The `I` (flag) family is the only exception, as it carries no subkey list.
2. Read `<len>:<subkey>` items. Subkey boundaries are determined strictly by `<len>`, so commas and colons inside a subkey value are preserved verbatim.
3. Items are separated by `,`.

Example: `hset|5:a,b:c,3:xyz` decodes to event `hset` with subkeys `["a,b:c", "xyz"]` - the `,` inside the first subkey is part of the value, not a separator.

### Typical use cases:

- **`S` (`__subkeyspace@<db>__:<key>`)** - "Send a notification about what changed *on this hash/stream/etc.*". Replaces the common pattern of subscribing to `__keyspace@0__:hash` and then issuing an `HGETALL` to diff the state.
The changed fields arrive inline with the notification.
- **`T` (`__subkeyevent@<db>__:<event>`)** - "Send a notification about every `hset` (or `hdel`, …) anywhere, including which key and which fields were touched".
Useful for centralized auditors and replication-style consumers.
- **`I` (`__subkeyspaceitem@<db>__:<key>\n<sk>`)** - "Send a notification when *this exact field* of this key changes". Each `(key, subkey)` pair gets its own channel, so a field-granular cache can `PSUBSCRIBE __subkeyspaceitem@0__:user:42\nemail` and receive a message only when `user:42.email` is written or deleted.
Generally cheaper than parsing `S` payloads when subscribers care about one field out of many.
- **`V` (`__subkeyspaceevent@<db>__:<event>|<key>`)** - combines `T` and `S`. The channel name encodes both the event and the key, leaving the payload as just the affected subkey list.
Useful when you want to filter at `PSUBSCRIBE` time (`__subkeyspaceevent@0__:hset|user:*`) rather than at the payload-parse stage.

### new file - `redis/tests/test_subkeyspace_notifications.rs`

The file contains the necessary parsers, which are all built on a single shared helper (`parse_length_prefixed_list`) along with
eleven tests, split into two groups:
- **7 parser unit tests**
- **4 integration tests** - gated on `REDIS_VERSION_CE_8_8`.
  - `test_subkeyspace_flag_s` - enables `KEAS` - `HMSET` then `HDEL`, asserts the two messages decode to the expected `(event, subkeys)` tuples.
  - `test_subkeyevent_flag_t` - enables `KEAT` - asserts the channel is exactly `__subkeyevent@0__:hset` and the payload's key length-prefix matches the mutated key.
  - `test_subkeyspaceitem_flag_i` - enables `KEAI` - asserts one notification arrives per field with the channel encoding `(key, subkey)` and payload `hset`.
  - `test_subkeyspaceevent_flag_v` - enables `KEAV` - asserts the channel encodes `(event, key)` and the payload is the length-prefixed subkey list.
The subscriptions are done through the existing `Connection::as_pubsub()` / `PubSub::psubscribe()` APIs.

## Relevant documentation:
Keyspace notifications - https://redis.io/docs/latest/develop/pubsub/keyspace-notifications/
Subkey notifications - https://redis.io/docs/latest/develop/pubsub/subkeyspace-notifications/

### DIFF
--- a/redis/tests/support/mod.rs
+++ b/redis/tests/support/mod.rs
@@ -534,6 +534,7 @@ pub const REDIS_VERSION_CE_8_0: Version = (8, 0, 0);
 pub const REDIS_VERSION_CE_8_2: Version = (8, 1, 240);
 pub const REDIS_VERSION_CE_8_4: Version = (8, 3, 224);
 pub const REDIS_VERSION_CE_8_6: Version = (8, 6, 0);
+pub const REDIS_VERSION_CE_8_8: Version = (8, 8, 0);
 
 /// Macro to run tests only if the Redis version meets the minimum requirement.
 /// If the version is insufficient, the test is skipped with a message.
--- a/redis/tests/test_subkeyspace_notifications.rs
+++ b/redis/tests/test_subkeyspace_notifications.rs
@@ -1,0 +1,293 @@
+#![allow(clippy::let_unit_value)]
+
+//! Integration tests for Redis 8.8 subkey notifications.
+//!
+//! Four channel families are exercised.
+//! Each is gated by a single letter in `notify-keyspace-events`:
+//!
+//! | Flag | Channel pattern                              | Payload                              |
+//! |------|----------------------------------------------|--------------------------------------|
+//! | `S`  | `__subkeyspace@<db>__:<key>`                 | `<event>|<len>:<sk>[,<len>:<sk>...]` |
+//! | `T`  | `__subkeyevent@<db>__:<event>`               | `<keylen>:<key>|<len>:<sk>[,...]`    |
+//! | `I`  | `__subkeyspaceitem@<db>__:<key>\n<sk>`       | `<event>`                            |
+//! | `V`  | `__subkeyspaceevent@<db>__:<event>|<key>`    | `<len>:<sk>[,...]`                   |
+
+#[macro_use]
+mod support;
+
+#[cfg(test)]
+mod subkeyspace_notifications_tests {
+    use std::collections::BTreeMap;
+    use std::time::Duration;
+
+    use crate::support::*;
+    use redis::{Commands, cmd};
+
+    // ---------- payload / channel parsers ---------------------------------
+
+    /// Parses length-prefixed, comma-separated lists (e.g. `<len>:<sk>[,<len>:<sk>...]`).
+    fn parse_length_prefixed_list(mut rest: &[u8]) -> Option<Vec<Vec<u8>>> {
+        let mut out = Vec::new();
+        while !rest.is_empty() {
+            let colon = rest.iter().position(|&b| b == b':')?;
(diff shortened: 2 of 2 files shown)


nihohit:
why are these tests needed in this repo? Maybe I'm missing something, but AFAIS this checks functionality on the server-side, not on the client side.

StefanPalashev:
Sorry for making noise with this pull request. I thought it might be good to have the tests in here as well, since they can serve as an example on how to use these type of subscriptions.

nihohit:
if you want to add an example, there's the example folder. Sadly, most people don't look in the tests folders for examples :)

But in general, I don't think that we need to add to the CI / maintenance burden specific functionality, unless it relates to specific implementation in the library.

PR #2134, opened 2026-06-12
tests: Allow tests to detect (and skip on) versions of all components (Redis, Valkey, modules)

### LINKED ISSUE #2118: Tests: Find a way to detect and check Redis/Valkey versions along with modules
We currently check version `redis_version` to determine Redis' version and guards tests by it.

But we lack a way to properly detect Valkeys version or identify modules and their versions.

So the goal is to find a way to obtain the above information and make it accessible to CI/tests to allow effective guarding.

somechris:
This extracts versions from all `*_version` and `module` keys from `INFO`.

And it allows to use a list of components to `OR` them and write
version guards like

```
let ctx = run_test_if_version_supported!(&[&REDIS_8_8, &VALKEY_9_0]);
```

to run the test only on `Redis v8.8 || Valkey v9.0` and skip otherwise.

It also allows to use list of lists of components to `OR` the inner lists
and `AND` the outer lists to write version guards like

```
let ctx = run_test_if_version_supported!(&[&[&REDIS_8_8, &VALKEY_9_0][..], &[&REJSON_MODULE_8_8]]);
```

to match `(Redis v8.8 || Valkey v9.0) && REJSON module v8.8`.

Fixes #2118

### DIFF
--- a/redis/tests/support/cluster.rs
+++ b/redis/tests/support/cluster.rs
@@ -202,18 +202,6 @@ impl TestClusterContext {
         }
     }
 
-    /// Gets the Redis version of the first server in the cluster
-    ///
-    /// # Panics
-    ///
-    /// As this function is only meant to be used during testing, it panics upon any issues.
-    pub fn get_version(&self) -> super::Version {
-        let server = self.cluster.servers.first().unwrap();
-        let mut conn = self.build_single_client_connection(server).unwrap();
-
-        super::get_version(&mut conn)
-    }
-
     pub fn get_ports(&self) -> Vec<u16> {
         self.nodes
             .iter()
@@ -225,5 +213,14 @@ impl TestClusterContext {
                 }
             })
             .collect()
+    }
+}
+
+impl super::TestContextVersioning for TestClusterContext {
+    fn get_version(&self) -> super::Version {
+        let server = self.cluster.servers.first().unwrap();
+        let mut conn = self.build_single_client_connection(server).unwrap();
+
+        super::get_version(&mut conn)
     }
 }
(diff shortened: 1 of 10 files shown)


somechris:
@nihohit: As this is PR is a bit lengthy ... I typically do atomic commits. So if you do not squash them, but go commit by commit, each of them does is simple to review and self-contained.

somechris:
Mhmm. The build failure looks unrelated.
(I tried other unrelated changes, and they now fail on GitHub CI as well although they passed before)

nihohit:
The build failures are unrelated but consistent, and I'm not sure if it's from the rust Azure SDK or the `time` crate.

nihohit:
Are the commits self-contained? can this PR be split into multiple PRs, each performing a part of the task?

somechris:
> Are the commits self-contained? can this PR be split into multiple PRs, each performing a part of the task?

Mhmm. Yup, let me try that. 8 PRs coming along ... :grin: 
(#2138, #2139, #2140, #2141, #2142, #2143, #2144, #2145)


But be warned that they are self-contained as in [atomic commits](https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention) (like in the Linux kernel, or Gerrit review style). So they do one thing and one thing only. Hence, they are smaller and easier to review. And they lead from one consistent state to another consistent state.

But they are of course not independent from another (e.g.: The 3rd commit moves the versioning code into a dedicated module. The 4th commit changes things in that new module, so it builds on the 3rd commit.)

Let me know which variant you prefer.

somechris:
This PR is getting handled through PRs for each individual commit.

PR #2136, opened 2026-06-12
Work around for `cargo-semver-check` issues with `time v0.3.48`

somechris:


nihohit:
0.3.48 was yanked. running CI in other PRs to check whether the issue is solved.

nihohit:
The problematic version was yanked.

PR #2138, opened 2026-06-16
tests/cluster_async: Switch to positive version checking (Version refactor 1/8)

somechris:
In some parts of the `cluster_sync` test, functionality that is only available on Redis 7+ is tested.

But instead of checking for Redis 7+, the code checked for Redis 6 and negated the result. This was harder to read and did not match the other code where we have version guards. So we switch to testing for Redis 7+.

This will make upcoming version refactorings easier to read.


nihohit (approved):
thanks!

PR #2142, opened 2026-06-16
tests: Delegate version comparison to `TestClusterVersioning` (Version refactor 5/10)

somechris:
By delegating version comparison to `TestClusterVersioning`, all version
refactorings can now happen withtin `test::support::version` and no
longer need to mess with the tests' logic.

### DIFF
--- a/redis/tests/support/version.rs
+++ b/redis/tests/support/version.rs
@@ -70,6 +70,32 @@ pub trait TestContextVersioning {
     ///
     /// As this function is only meant to be used during testing, it panics upon any issues.
     fn get_version(&self) -> Version;
+
+    /// Returns whether the context's server has at least the given Redis version
+    fn supports(&self, version: &Version) -> bool {
+        self.get_version() >= *version
+    }
+}
+
+/// Skips the current test if it does not support the given Redis version
+///
+/// # Arguments
+///
+/// * `$ctx` - The context the test uses for its servers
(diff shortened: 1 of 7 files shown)


redis/tests/support/version.rs
```diff
@@ -70,6 +70,32 @@ pub trait TestContextVersioning {
     ///
     /// As this function is only meant to be used during testing, it panics upon any issues.
     fn get_version(&self) -> Version;
+
+    /// Returns whether the context's server has at least the given Redis version
+    fn supports(&self, version: &Version) -> bool {
```
nihohit, line 75: nit: why pass version by reference?

nihohit (approved):
the minor comment can be fixed in another PR when rebasing

redis/tests/support/version.rs
somechris: I'll look into that tomorrow.

Thanks for today's merges! :heart:

somechris: Why require ownership, if it's not needed?

Additionally, `run_test_if_version_supported(&REDIS_VERSION_CE_...)` (with `&`) was used widely in the tests, so I figured that it's your prefered way of writing it.

I see benefits both ways and do not see a clear winner.

So #2152 removes the reference and adapts all tests. :thumbsup:

PR #2143, opened 2026-06-16
tests/version: Parse all available versions (Valkey, modules, ...) (Version refactor 6/10)

### LINKED ISSUE #2118: Tests: Find a way to detect and check Redis/Valkey versions along with modules

somechris:
This allows tests to run/skip if it's running on Valkey, on a module,
and on specific versions thereof.

Fixes #2118

### DIFF
--- a/redis/tests/support/version.rs
+++ b/redis/tests/support/version.rs
@@ -1,31 +1,186 @@
 //! Tooling to handle server version extraction and matching
 
-use redis::InfoDict;
+use redis::ConnectionLike;
+use std::collections::HashMap;
 
 // Redis version constants for version-gated tests
-pub const REDIS_VERSION_CE_7_0: Version = (7, 0, 0);
-pub const REDIS_VERSION_CE_7_2: Version = (7, 2, 0);
-pub const REDIS_VERSION_CE_7_4: Version = (7, 4, 0);
-pub const REDIS_VERSION_CE_8_0: Version = (8, 0, 0);
-pub const REDIS_VERSION_CE_8_2: Version = (8, 1, 240);
-pub const REDIS_VERSION_CE_8_4: Version = (8, 3, 224);
-pub const REDIS_VERSION_CE_8_6: Version = (8, 6, 0);
+pub const REDIS_VERSION_CE_6_0: Component = ("redis", (6, 0, 0));
+pub const REDIS_VERSION_CE_7_0: Component = ("redis", (7, 0, 0));
+pub const REDIS_VERSION_CE_7_2: Component = ("redis", (7, 2, 0));
+pub const REDIS_VERSION_CE_7_4: Component = ("redis", (7, 4, 0));
+pub const REDIS_VERSION_CE_8_0: Component = ("redis", (8, 0, 0));
+pub const REDIS_VERSION_CE_8_2: Component = ("redis", (8, 1, 240));
+pub const REDIS_VERSION_CE_8_4: Component = ("redis", (8, 3, 224));
+pub const REDIS_VERSION_CE_8_6: Component = ("redis", (8, 6, 0));
 
-pub type Version = (u16, u16, u16);
+/// Version of a software component
+pub type Version = (u32, u32, u32);
 
-pub fn parse_version(info: InfoDict) -> Version {
-    let version: String = info.get("redis_version").unwrap();
-    let versions: Vec<u16> = version
-        .split('.')
-        .map(|version| version.parse::<u16>().unwrap())
-        .collect();
-    assert_eq!(versions.len(), 3);
-    (versions[0], versions[1], versions[2])
+/// Software component name along with its version
+pub type Component<'a> = (&'a str, Version);
+
+#[derive(Clone)]
+pub struct AvailableComponents {
+    /// The available components' versions by their name
+    components: HashMap<String, Version>,
 }
 
-pub fn get_version(conn: &mut impl redis::ConnectionLike) -> Version {
-    let info: InfoDict = redis::Cmd::new().arg("INFO").query(conn).unwrap();
-    parse_version(info)
+impl AvailableComponents {
+    /// Extracts the available/used/usable software components from an `INFO` response
+    ///
+    /// # Panics
+    ///
+    /// This method panics upon issues
+    fn parse_info(info_response: &str) -> HashMap<String, Version> {
+        let mut ret = HashMap::new();
+        for raw_line in info_response.lines() {
+            // Strip off comments
+            let line = raw_line.split("#").next().unwrap();
+
+            // Extract key and value
+            let mut split = line.splitn(2, ":");
+            let (Some(key), Some(value)) = (split.next(), split.next()) else {
+                continue;
+            };
+
+            // Turn into raw component name and version
+            let Some((name, version)) = Self::parse_info_kv(key.trim(), value.trim()) else {
+                continue;
+            };
+
+            // Store them
+            ret.insert(name.to_owned(), version);
+        }
+        ret
+    }
+
+    /// Parses an `INFO` key/value into component raw name and version
+    ///
+    /// # Panics
+    ///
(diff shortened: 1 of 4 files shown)


nihohit:
@somechris yet another rebase :)

nihohit:
@somechris please also add https://github.com/redis-rs/redis-rs/pull/2142#discussion_r3429530118 here or in one of the next PRs

redis/tests/support/version.rs
```diff
@@ -1,31 +1,169 @@
...
-pub fn parse_version(info: InfoDict) -> Version {
-    let version: String = info.get("redis_version").unwrap();
-    let versions: Vec<u16> = version
-        .split('.')
-        .map(|version| version.parse::<u16>().unwrap())
-        .collect();
-    assert_eq!(versions.len(), 3);
-    (versions[0], versions[1], versions[2])
+/// Software component name along with its version
+pub type Component<'a> = (&'a str, Version);
```
nihohit, line 20: why not a static str? or better yet, enum?

redis/tests/support/version.rs
```diff
@@ -1,31 +1,169 @@
...
+    ///
+    /// This method panics upon issues
+    fn parse_info(info_response: &str) -> HashMap<String, Version> {
+        let mut ret = HashMap::new();
+        for raw_line in info_response.lines() {
+            // Strip off comments
+            let line = raw_line.split("#").next().unwrap();
+
+            let mut split = line.splitn(2, ":");
+            let (name, version) = match (split.next(), split.next()) {
```
nihohit: nit: `let (Some(key), Some(value)) =  (split.next(), split.next()) else {continue }`

redis/tests/support/version.rs
```diff
@@ -1,31 +1,169 @@
...
+        let number_str = value.split("-").next().unwrap();
+
+        // Convert each number part to a number
+        let numbers = number_str
+            .split('.')
+            .map(|version| version.parse::<u32>().unwrap())
+            .collect::<Vec<_>>();
+
+        // Massage the numbers into a [`Version`]
+        let (major, minor, patch) = match numbers.len() {
```
nihohit: nit: 
```rs
match numbers.as_slice() { 
  [number] => {...}
  [major, minor] => {...}
  [major, minor, patch] => { .. }
  _ => panic
}

File: redis/tests/support/version.rs [hunk 4]
```diff
@@ -1,31 +1,169 @@
...
+    ///
+    /// This method panics upon issues
+    fn extract_version(value: &str) -> Version {
+        // Cut-off suffixes (e.g.: the trailing `-rc1` in `1.2.3-rc1`)
+        let number_str = value.split("-").next().unwrap();
+
+        // Convert each number part to a number
+        let numbers = number_str
+            .split('.')
+            .map(|version| version.parse::<u32>().unwrap())
```
nihohit, line 111: why parse as u32 if later you convert to u16?

redis/tests/support/version.rs
```diff
@@ -64,48 +202,57 @@ pub fn get_redis_binary_version() -> Option<Version> {
...
         }
     }};
 }
 
-/// Macro to run tests only if the Redis version meets the minimum requirement.
+/// Macro to run tests only if the default [`TestContext`] supports the given component
+///
 /// If the version is insufficient, the test is skipped with a message.
+///
+/// # Arguments
```
nihohit: I don't find this generic documentation fields helpful.

redis/tests/support/version.rs
```diff
@@ -1,31 +1,169 @@
...
+
+                let patch = rest % 100;
+                rest = (rest - patch) / 100;
+
+                let minor = rest % 100;
+                let major = (rest - minor) / 100;
+
+                (major, minor, patch)
+            }
+            2 => (numbers[0], numbers[1], 0),
```
nihohit: when should we expect receiving the version in only 1 or 2 numbers?

redis/tests/support/version.rs
somechris: I initially started with enums, because modelling-wise, obviously enums are the first choice. But

```
pub const REDIS_CE_6_0: Component = (Components.REDIS, (6, 0, 0));
pub const REDIS_CE_7_0: Component = (Components.REDIS, (7, 0, 0));
pub const REDIS_CE_7_2: Component = (Components.REDIS, (7, 2, 0));
pub const REDIS_CE_7_4: Component = (Components.REDIS, (7, 4, 0));
```

looked so unreadable, compared to

```
pub const REDIS_CE_6_0: Component = ("redis", (6, 0, 0));
pub const REDIS_CE_7_0: Component = ("redis", (7, 0, 0));
pub const REDIS_CE_7_2: Component = ("redis", (7, 2, 0));
pub const REDIS_CE_7_4: Component = ("redis", (7, 4, 0));
```

(Especially when later on using the bloom modules `Components.REDIS_BLOOM_FILTER_MODULE` and `Components.VALKEY_BLOOM_FILTER_MODULE` compared to `redis:bf` and `valkey:bf`)

So I ended up not using enums.

Sure it's repetive. But readability weighted heavier for me here.

Seeing the above two variants side by side, do you still think enums would be preferable? (Then I'll gladly adapt it)

(`'static` would make things trickier when getting the parsed Redis `INFO` (which is of course not `'static`) response into `Vec<Component>`.)

nihohit: I actually find the first perfectly readable, but if you find it unreadable, I'm ok with going with your preference, I don't object to it strongly.

somechris:
> @somechris please also add [#2142 (comment)](https://github.com/redis-rs/redis-rs/pull/2142#discussion_r3429530118) here or in one of the next PRs

Done in #2152

redis/tests/support/version.rs
somechris: `Version` used `u16`before the refactor.

But with version Redis v8.8.0, the single number versions for modules are >=80000 which is too big for `u16`. So they are parsed as `u32`. But when deconstructed to triples (8.x.y) in the end, each of them fits into `u16`. So I figured bumping `Version` to `u32`s was unneeded.

Not sure. Do you prefer to bump `Version` to `u32` and spare the casts?

(I've added a corresponding comment to the parsing)

redis/tests/support/version.rs
somechris: Done in 19768bc :+1:

redis/tests/support/version.rs
somechris: Done in 19768bc 👍

redis/tests/support/version.rs
nihohit: > Do you prefer to bump Version to u32 and spare the casts?

Sure, this is test-only, we can pay the cost :)

redis/tests/support/version.rs
somechris: 1 number version is used for modules. They report like:

```
module:name=search,ver=80001,[...]
```

I've added a comment to the code.


2 number version is not used at all. But leaving out 2 number version while covering 1 number, and 3 number felt like a hole. But you're right, it's not needed right now. I removed it.

redis/tests/support/version.rs
somechris: I removed it here. But I left it with the above macro, as that has two signatures -- For me, the `# Arguments` allows to easier see that the signatures actually do the same, and cover only the optional argument.

Should I remove it there as well?

redis/tests/support/version.rs
somechris: Some `u32`s coming you way :-)

Done in a780469 :+1:

nihohit (approved):
Thanks!

PR #2144, opened 2026-06-16
tests/version: Add disjunctive (`OR`) and conjunctive (`AND`) matchers (Version refactor 7/10)

somechris:
This allows to use list of components to `OR` them and write version
guards like

```
let ctx = run_test_if_version_supported!(&[&REDIS_8_8, &VALKEY_9_0]);
```

to run the test only on `Redis v8.8 || Valkey v9.0` and skip otherwise.

And it allows to use list of lists of components to `OR` the inner lists
and `AND` the outer lists to write version guards like

```
let ctx = run_test_if_version_supported!(&[&[&REDIS_8_8, &VALKEY_9_0][..], &[&REJSON_MODULE_8_8]]);
```

to match `(Redis v8.8 || Valkey v9.0) && REJSON module v8.8`.

### DIFF
--- a/redis/tests/support/version.rs
+++ b/redis/tests/support/version.rs
@@ -21,6 +21,90 @@ pub type Version = (u32, u32, u32);
 /// Software component name along with its version
 pub type Component<'a> = (&'a str, Version);
 
+/// Matcher for [`Component`]s
+pub struct ComponentMatcher<'a> {
+    conjunctive_parts: Vec<Vec<Component<'a>>>,
+}
+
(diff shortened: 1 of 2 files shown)


nihohit:
@somechris looks ok, please rebase

somechris:
> @somechris looks ok, please rebase

@nihohit Done

(rephrased a bit, as you said that we should get rid of `&`s. The outer `&`s are gone, but the inner have to stay as array elements need to be `Sized`)

nihohit (approved):

PR #2146, opened 2026-06-16
Add numbered databases support for the cluster client

virratanasangpunth:
## Motivation

Valkey 9.0 has added numbered database support for clusters: https://valkey.io/blog/numbered-databases/

In this PR we allow the users of redis-rs to be able to configure the database number for each cluster client.

Note that the database number is a property of a connection.

## Example Usage

```
let client = ClusterClientBuilder::new(nodes.to_vec())
    .database_id(database_id)
    .build()
    .expect("ClusterClient::build to work");
let connection = client.get_async_connection().await.expect("get_async_connection work")
```

## Testing

Unit tests added.

Also tested with real docker images:
1. happy path: seems to work fine.
2. when valkey9 server is not initialized, you get `Io: Redis server refused to switch database - Server(ResponseError): DB index is out of range` when calling `get_async_connection()`
3. when used with redis (which doesnt support cluster numbered databases), you get `Io: Redis server refused to switch database - Server(ResponseError): SELECT is not allowed in cluster mode` when calling `get_async_connection()`


virratanasangpunth:
Actually this PR is incomplete so I'm closing this. There is a footgun where a user can issue a `SELECT` command which only gets routed to one node and is not persisted across reconnects. 

Will be back with a more complete PR.

UPDATE: here is the companion PR for correctness: https://github.com/redis-rs/redis-rs/pull/2150

nihohit (approved):
this is fine in general, but it needs to also run against a real cluster to see that this change actually works.

virratanasangpunth:
> this is fine in general, but it needs to also run against a real cluster to see that this change actually works.

@nihohit thank you for the review. Yes I have tested this against a real cluster (using multiple redis docker containers as a cluster).

nihohit:
I need a test to be added either to test_cluster.rs and to test_cluster_async.rs, targeting real clusters, in order to continuously verify that this behavior works with any cluster that supports it. Such API changes can't rely on one-time tests on the developer's machine.

virratanasangpunth:
Thanks for the tips on testing. A test against real valkey9 added.

virratanasangpunth:
@nihohit sorry about the churn. I've fixed the formatting mistakes. I've asked Claude to run the "PR checks" locally and they are passing now. This PR is RFR now.

redis/src/cluster_handling/client.rs
```diff
@@ -297,6 +306,15 @@ impl ClusterClientBuilder {
         } else {
             None
         };
+        // An explicit builder value (even 0) takes precedence. Otherwise inherit the
+        // database from the first node's URL (e.g. `redis://host:6379/4`); when
+        // inherited, all initial nodes must agree.
+        let database_id = if explicit_database_id.is_some() {
```
nihohit: this line is confusing - explicit_database_id means that database_id is None?

maybe replace `database_id` with `database_id_from_url`?

redis/src/cluster_handling/client.rs
```diff
@@ -326,6 +344,14 @@ impl ClusterClientBuilder {
                     "Cannot use different protocol among initial nodes.",
                 )));
             }
+            if let Some(database_id) = database_id
```
nihohit: what about if the database_id was set from `explicit_database_id` differently than it was set in any URL?

redis/src/cluster_handling/client.rs
```diff
@@ -881,6 +924,68 @@ mod tests {
...
+    fn give_database_id_by_method_overrides_initial_nodes() {
+        let client = ClusterClientBuilder::new(vec![
+            "redis://127.0.0.1:6379/2",
+            "redis://127.0.0.1:6378/2",
+            "redis://127.0.0.1:6377/2",
+        ])
+        .database_id(7)
+        .build()
+        .unwrap();
+        assert_eq!(client.cluster_params.database_id, 7);
```
nihohit: this should be an error

redis/src/cluster_handling/client.rs
```diff
@@ -881,6 +924,68 @@ mod tests {
...
+    fn explicit_database_id_zero_overrides_initial_nodes() {
+        let client = ClusterClientBuilder::new(vec![
+            "redis://127.0.0.1:6379/4",
+            "redis://127.0.0.1:6378/4",
+            "redis://127.0.0.1:6377/4",
+        ])
+        .database_id(0)
+        .build()
+        .unwrap();
+        assert_eq!(client.cluster_params.database_id, 0);
```
nihohit: same

redis/tests/support/version.rs
```diff
@@ -62,6 +64,23 @@ pub fn get_redis_binary_version() -> Option<Version> {
     Some((versions[0], versions[1], versions[2]))
 }
 
+/// Returns `true` if the server binary is Valkey (as opposed to Redis).
+///
+/// Inspects the `redis-server --version` output rather than a running server,
+/// so it can gate a test before any server is started. Returns `false` if the
+/// binary cannot be located.
+pub fn is_valkey_binary() -> bool {
```
nihohit, line 227: @somechris should this PR wait for your work? or will you clean this up once the components are fully merged?

redis/tests/test_cluster.rs
```diff
@@ -48,6 +49,54 @@ mod cluster {
...
+        let cluster = TestClusterContext::new_with_config_and_builder(
+            RedisClusterConfiguration {
+                cluster_databases: Some(16),
+                tls_insecure: false,
+                ..Default::default()
+            },
+            |builder| builder.database_id(4),
+        );
+
+        let mut con_db4 = cluster.connection();
```
nihohit: just run CLIENT INFO and check the `db` entry

redis/tests/test_cluster_async.rs
```diff
@@ -69,6 +70,66 @@ mod cluster_async {
...
+        let cluster = TestClusterContext::new_with_config_and_builder(
+            RedisClusterConfiguration {
+                cluster_databases: Some(16),
+                tls_insecure: false,
+                ..Default::default()
+            },
+            |builder| builder.database_id(4),
+        );
+
+        let mut con_db4 = cluster.async_connection().await;
```
nihohit: same

nihohit (commented):
small fixes, no major issues

redis/src/cluster_handling/client.rs
virratanasangpunth: if database in the url and the explicit one disagree we will throw an error

redis/src/cluster_handling/client.rs
```diff
@@ -306,14 +306,11 @@ impl ClusterClientBuilder {
         } else {
             None
         };
-        // An explicit builder value (even 0) takes precedence. Otherwise inherit the
-        // database from the first node's URL (e.g. `redis://host:6379/4`); when
-        // inherited, all initial nodes must agree.
-        let database_id = if explicit_database_id.is_some() {
-            None
+        let database_id_from_url = if cluster_params.database_id.is_none() {
```
nihohit: why make this optional? why not just abandon the split between DB from url / from config?
```rs
let database_id = match cluster_params.database_id {Some(id) => id, None => first_node.redis.db };
```

and then below you just check 
```rs
if node.redis.db != database_id {
  return Err(RedisError::from((
      "Cannot use different database among initial nodes.",
  )));
}
```
without the huge match block.

nihohit:
Thanks for the contribution!

virratanasangpunth:
@nihohit thank you for reviewing merging this!

virratanasangpunth:
@nihohit no rush, but I'm curious if we are going to release a new version at some point? so this change can be consumed by the public?

nihohit:
Hi @virratanasangpunth, I assume this is a slip of the tongue, but the phrase "at some point" implies that this should have already happened, and isn't very polite in this context. I assume that this is just an honest mistake :).

Yes, a new version will be released. I usually release around the weekend, so probably this weekend or the next after, when I'll have time. If you need this now, you can always point your Cargo.toml file at a git commit instead of a release.

virratanasangpunth:
@nihohit Yes my apology for that language; I didn't mean to be impolite. I appreciate your time/effort :)

PR #2147, opened 2026-06-17
tests: Drop `_VERSION_` from version constants (version refactor 9/10)

somechris:
Follow-up commits will add version constants for Valkey servers. If we
followed the Redis' constants' lead as-is, they'd end up like
`VALKEY_VERSION_9_0`, which would give guards like:

```
run_test_if_version_supported!(&[&REDIS_VERSION_CE_8_6, &VALKEY_VERSION_9_0]);
```

That's squishy, because of the repetitive `version`.

As `REDIS_VERSION_CE_8_6` is lengthy already an its own, we drop the
`_VERSION_` to arrive at `REDIS_CE_8_6` which is much shorter and also
clear; especially as test guard. This will give us more compact guards
in follow-up commits, like:

```
run_test_if_version_supported!(&[&REDIS_CE_8_6, &VALKEY_9_0]);

### DIFF
--- a/redis/tests/support/version.rs
+++ b/redis/tests/support/version.rs
@@ -4,16 +4,16 @@ use redis::ConnectionLike;
 use std::collections::HashMap;
 
 // Redis version constants for version-gated tests
-pub const REDIS_VERSION_CE_6_0: Component = ("redis", (6, 0, 0));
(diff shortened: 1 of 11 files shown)


nihohit:
@somechris rebase, please

somechris:
> @somechris rebase, please

@nihohit Done :+1:

nihohit (approved):

PR #2148, opened 2026-06-17
tests: Add test guards for Valkey servers (Version refactor 10/10)

somechris:
This allows to run more tests against newer Valkey servers.

Although Valkey 9.1 is not tested in CI, we nonetheless add
corresponding guards, as they will automatically enable tests if CI
adds newer Valkey servers.

### DIFF
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -369,8 +369,7 @@ mod basic {
 
     #[test]
     fn test_hash_expiration() {
-        // Hash expiration is only supported in Redis 7.4.0 and later.
-        let ctx = run_test_if_version_supported!(REDIS_CE_7_4);
+        let ctx = run_test_if_version_supported!([REDIS_CE_7_4, VALKEY_9_0]);
 
         let mut con = ctx.connection();
         redis::cmd("HMSET")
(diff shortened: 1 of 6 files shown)


nihohit:
@somechris rebase please

redis/tests/test_basic.rs
```diff
@@ -484,7 +484,8 @@ mod basic {
     /// 5. Attempting to delete a field from a non-existing hash results in a NIL response.
     #[test]
     fn test_hget_del() {
-        let ctx = run_test_if_version_supported!(REDIS_CE_8_0);
+        // `HGETDEL` is only supported in Redis 8.0+ and Valkey 9.1+
```
nihohit: IMO these comments are unnecessary - they repeat the next line

somechris: There already were a few of these comments. So I tried following that
lead.

And yes, they do duplicate things a bit, but they mention the actual
Redis command (which the test function name not always does) and
against which versions a human checked them. This looked useful for
maintenance. Because as they are on a single line, they are very easy
to grep. E.g.:

```
grep -rn -B 1 REDIS_CE | grep //
```

Anyway, I've now removed all these duplicating comments (old and new ones) :+1:

---

However! I did keep the ones where Redis supports something that Valkey
does not. The “(but not Valkey<=9.1)” there will help us in the future
to see which Valkey versions have already been checked for the
respective command.

E.g.: The following shows what is missing in Valkey to allow enabling
the test. And it shows which versions we checked against already.

```
fn test_set_value_comparison_value_not_equals() {
    // `SET` option `IFNE` is only supported in Redis 8.4+ (but not Valkey<=9.1)
    let ctx = run_test_if_version_supported!(REDIS_CE_8_4);
...
```

Dropping the comment would give us

```
fn test_set_value_comparison_value_not_equals() {
    let ctx = run_test_if_version_supported!(REDIS_CE_8_4);
...
```

Here we don't know whether we checked against _some_ Valkey version
or not.

Are you ok with keeping the “(but not Valkey<=9.1)” comments or should I
remove them too? :thinking:

nihohit (approved):

nihohit:
Thanks for this change, and the whole journey through this refactor :)

PR #2149, opened 2026-06-17
tests: Use server binary from env to detect major version

somechris:
While we try to respect the `REDIS_SERVER_BIN` environment variable to start the server, the command used to detect the major version was still hard wired to `redis-server`.

We switch it to the environment variable to get more reliable results when running tests locally.


nihohit (approved):
Thanks

PR #2150, opened 2026-06-17
Forbid SELECT command for the cluster client

virratanasangpunth:
## Motivation

This is a companion PR to https://github.com/redis-rs/redis-rs/pull/2146. I'm splitting them up to keep each scope tight.

In this PR we introduce a new command-routing semantic for the cluster client: `RouteBy::Unsupported`. It's used to reject a command that cannot be meaningfully executed in the context of a cluster client.

So why is `SELECT` not allowed? Two reasons: the first is that Redis clusters don't support numbered databases anyway; redis-server returns an error if you try to `SELECT n` (for any `n` other than `0`). 

The second reason is for correctness. The companion PR adds a numbered dbs support to cluster client. To enforce that the configured db stays immutable throughout the life of a connection we require that a user cannot override with `SELECT`.

## Example Usage

```rust
let err = cmd("SELECT")
    .arg(1)
    .query_async::<Value>(&mut conn)
    .await
    .unwrap_err();
assert_eq!(err.kind(), ErrorKind::Client);
```

## Callouts!!

1. This will break any existing cluster clients that are issuing `SELECT 0` command. Previously in main this will succeed silently (it will get routed to a random node and does a no-op).
2. This is a design choice that we choose to go with "forbid". In GLIDE, they have chosen with fan-out-to-all-nodes routing for `SELECT`. We could go that route but I think making numbered db immutable is a simpler solution. With fan-out you need to consider partial failures, which could cause a split-brain cluster client.

### DIFF
--- a/redis/src/cluster_handling/async_connection/mod.rs
+++ b/redis/src/cluster_handling/async_connection/mod.rs
@@ -966,6 +966,8 @@ where
         // we send request per cmd, instead of sending the pipe together, in order to send each command to the relevant node, instead of all together to a single node.
         let requests = subscription_pipe.into_cmd_iter().map(|cmd| {
             let routing = RoutingInfo::for_routable(&cmd)
+                .ok()
+                .flatten()
                 .unwrap_or(RoutingInfo::SingleNode(SingleNodeRoutingInfo::Random))
                 .into();
             PendingRequest {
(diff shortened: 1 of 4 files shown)


nihohit:
The issue with forbidding commands is that this means that the library then needs to release a version in order to handle new server implementations that support that command. Why not let the server return the error, instead of pre-emptively rejecting it?

PR #2151, opened 2026-06-18
fix(aio): bound permit allocation by concurrency limit, not pipeline length

Ali2Arslan:
## Context
`MultiplexedConnection::send_packed_commands` sized its permit vector with `Vec::with_capacity(count)`, where `count` is the pipeline length.

## Solution
Remember the configured limit alongside the semaphore and size the allocation by `count.min(limit).max(1)` instead of `count`.

### DIFF
--- a/redis/src/aio/multiplexed_connection.rs
+++ b/redis/src/aio/multiplexed_connection.rs
@@ -7,7 +7,7 @@ use crate::{
     aio::setup_connection,
     check_resp3, cmd,
     cmd::Cmd,
-    errors::{RedisError, closed_connection_error},
+    errors::{ErrorKind, RedisError, closed_connection_error},
     parser::ValueCodec,
     types::{RedisFuture, RedisResult, Value},
 };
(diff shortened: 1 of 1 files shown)


redis/src/aio/multiplexed_connection.rs
```diff
@@ -654,9 +667,10 @@ impl MultiplexedConnection {
             Pipeline::resolve_buffer_size(config.pipeline_buffer_size),
         );
 
-        let concurrency_limiter = config
-            .concurrency_limit
-            .map(|n| Arc::new(async_lock::Semaphore::new(n)));
+        let concurrency_limiter = config.concurrency_limit.map(|n| ConcurrencyLimiter {
```
nihohit: I know that this is a pre-existing issue, but we should fail if n == 0

Ali2Arslan: agreed, thanks for taking a look, updated to return a `ErrorKind::InvalidClientConfig` when limit is (explicitly) 0

nihohit (approved):

nihohit:
Thanks!

PR #2152, opened 2026-06-18
tests/version: Require ownership to check version support (Version refactor 5b/10)

somechris:
Ownership is not necessary to check version support. But callers of `supports` can automatically get ownership and it frees `supports` calls from the ampersand. This makes the call a bit easier to read as it turns.


nihohit (approved):

nihohit:
Thanks!

PR #2154, opened 2026-06-18
Add support for the `COUNT` aggregator for sorted sets

StefanPalashev:
# Add support for the `COUNT` aggregator to sorted set intersection/union commands

## Summary

Introduce the missing `ZINTER` / `ZUNION` commands.

Add support for the `COUNT` aggregator (new in **Redis 8.8**) to the existing set operation commands (**`ZINTERSTORE` / `ZUNIONSTORE`**)

## Background - what is the COUNT aggregator?

`ZUNIONSTORE` / `ZINTERSTORE` / `ZUNION` / `ZINTER` combine sorted sets and compute each output element's score by aggregating its per-input scores.

Until version 8.8 the aggregator was one of:

| Aggregator | Output score for an element |
|------------|-----------------------------|
| `SUM` (default) | `Σ (score × weight)` over the input sets containing it |
| `MIN` | `min(score × weight)` |
| `MAX` | `max(score × weight)` |

These all operate on the **stored scores**.
A common need is to score an element purely by **how many input sets contain it** (e.g. "for each player, the number of leaderboards they appear on"), ignoring the stored scores entirely.

Redis 8.8 adds `COUNT` for exactly this. It ignores the input scores and sums the **weights** instead (default weight `1`):

| Mode | Output score for an element |
|------|-----------------------------|
| `COUNT` without `WEIGHTS` | number of input sets that contain it |
| `COUNT` with `WEIGHTS` | sum of the weights of the input sets that contain it |

For an **intersection**, the element is by definition in every input set, so the no-weights score equals the number of input keys.

### Example
Input:
```
s1 = { foo: 1, bar: 1 }
s2 = { foo: 2, bar: 2 }
s3 = { foo: 3 }
```
Command and resulting sorted set:
```
ZUNIONSTORE t1 3 s1 s2 s3 AGGREGATE COUNT
  -> bar: 2, foo: 3                      (membership count)

ZINTERSTORE t1 3 s1 s2 s3 AGGREGATE COUNT
  -> foo: 3                              (foo is the only common element)

ZUNIONSTORE t1 3 s1 s2 s3 WEIGHTS 10 5 3 AGGREGATE COUNT
  -> bar: 15, foo: 18                    (Σ weights of containing sets)

ZINTERSTORE t1 3 s1 s2 s3 WEIGHTS 10 5 3 AGGREGATE COUNT
  -> foo: 18
```

Before this command-level support, the only way to get this was the documented workaround: 
Copy each set with all scores rewritten to `1`, then `ZUNIONSTORE` with `SUM`. `COUNT` removes that entirely.

## Testing

| Test | Gating | Covers |
|------|--------|--------|
| `test_zinterstore_zunionstore` | none | `sum`, `min` and `max` aggregators, weighted and uweighted variants |
| `test_zinterstore_zunionstore_count` | 8.8 | the new `COUNT `aggregator for `zinterstore` /`zunionstore` |
| `test_zinter_zunion` | none | `sum`, `min` and `max` aggregators, weighted and unweighted variants, members only and `WITHSCORES` |
| `test_zinter_zunion_count` | 8.8 | the new `COUNT` aggregator for `zinter` / `zunion` |

## Relevant documentation:
ZINTER - https://redis.io/docs/latest/commands/zinter/
ZUNION - https://redis.io/docs/latest/commands/zunion/
ZINTERSTORE - https://redis.io/docs/latest/commands/zinterstore/
ZUNIONSTORE - https://redis.io/docs/latest/commands/zunionstore/

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -1106,7 +1106,8 @@ implement_commands! {
     }
 
     /// Intersect multiple sorted sets and store the resulting sorted set in
-    /// a new key using SUM as aggregation function.
+    /// a new key, optionally applying an `AGGREGATE` modifier.
+    /// If no `AGGREGATE` modifier is provided, `SUM` is used.
     /// [Redis Docs](https://redis.io/commands/ZINTERSTORE)
     fn zinterstore<D: ToSingleRedisArg, K: ToRedisArgs>(dstkey: D, keys: K, options: SortedSetOperationOptions) -> (usize) {
         cmd("ZINTERSTORE").arg(dstkey).arg(keys.num_of_args()).arg(keys).arg(options).take()
@@ -1118,6 +1119,34 @@ implement_commands! {
     fn zinterstore_with_weights<D: ToSingleRedisArg, K: ToRedisArgs, W: ToRedisArgs>(dstkey: D, keys: &'a [(K, W)], options: SortedSetOperationOptions) -> (usize) {
         let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(k, w)| (k, w)).unzip();
         cmd("ZINTERSTORE").arg(dstkey).arg(keys.num_of_args()).arg(keys).arg("WEIGHTS").arg(weights).arg(options).take()
+    }
+
+    /// Intersect multiple sorted sets and return the resulting members,
+    /// applying an optional `AGGREGATE` modifier (SUM by default).
+    /// [Redis Docs](https://redis.io/commands/ZINTER)
+    fn zinter<K: ToRedisArgs>(keys: K, options: SortedSetOperationOptions) -> (Vec<String>) {
+        cmd("ZINTER").arg(keys.num_of_args()).arg(keys).arg(options).take()
+    }
+
+    /// [`Commands::zinter`], but with the ability to specify a weight for each
+    /// sorted set by pairing each key with its corresponding weight.
+    /// [Redis Docs](https://redis.io/commands/ZINTER)
+    fn zinter_with_weights<K: ToRedisArgs, W: ToRedisArgs>(keys: &'a [(K, W)], options: SortedSetOperationOptions) -> (Vec<String>) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(k, w)| (k, w)).unzip();
+        cmd("ZINTER").arg(keys.num_of_args()).arg(keys).arg("WEIGHTS").arg(weights).arg(options).take()
+    }
+
+    /// [`Commands::zinter`], but additionally returns the score of each member.
+    /// [Redis Docs](https://redis.io/commands/ZINTER)
+    fn zinter_withscores<K: ToRedisArgs>(keys: K, options: SortedSetOperationOptions) -> (Vec<(String, f64)>) {
+        cmd("ZINTER").arg(keys.num_of_args()).arg(keys).arg(options).arg("WITHSCORES").take()
+    }
+
+    /// [`Commands::zinter_with_weights`], but additionally returns the score of each member.
+    /// [Redis Docs](https://redis.io/commands/ZINTER)
+    fn zinter_with_weights_withscores<K: ToRedisArgs, W: ToRedisArgs>(keys: &'a [(K, W)], options: SortedSetOperationOptions) -> (Vec<(String, f64)>) {
+        let (keys, weights): (Vec<&K>, Vec<&W>) = keys.iter().map(|(k, w)| (k, w)).unzip();
+        cmd("ZINTER").arg(keys.num_of_args()).arg(keys).arg("WEIGHTS").arg(weights).arg(options).arg("WITHSCORES").take()
     }
 
     /// Count the number of members in a sorted set between a given lexicographical range.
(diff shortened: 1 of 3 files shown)


redis/src/commands/mod.rs
```diff
@@ -1148,6 +1148,138 @@ implement_commands! {
         cmd("ZINTERSTORE").arg(dstkey).arg(keys.num_of_args()).arg(keys).arg("AGGREGATE").arg("MAX").arg("WEIGHTS").arg(weights).take()
     }
 
+    /// Intersect multiple sorted sets and store the resulting sorted set in a new key using COUNT as aggregation function.
+    /// Unlike the other aggregators, COUNT ignores the input scores:
+    /// the score of each element is set to the number of input sets that contain it (which, for an intersection, is the number of input sets).
+    ///
+    /// Requires Redis 8.8 or later.
+    /// [Redis Docs](https://redis.io/commands/ZINTERSTORE)
+    fn zinterstore_count<D: ToSingleRedisArg, K: ToRedisArgs>(dstkey: D, keys: K) -> (usize) {
```
nihohit: why different functions and not an option object?

I'm really asking, I don't have a strong opinion on this.

redis/src/commands/mod.rs
```diff
@@ -1148,6 +1148,138 @@ implement_commands! {
...
+        cmd("ZINTERSTORE").arg(dstkey).arg(keys.num_of_args()).arg(keys).arg("AGGREGATE").arg("COUNT").take()
+    }
+
+    /// [`Commands::zinterstore_count`], but with the ability to specify a weight for each sorted set by pairing one with each key in a tuple.
+    /// The score of each element is set to the sum of the weights of the input sets that contain it.
+    ///
+    /// Requires Redis 8.8 or later.
+    /// [Redis Docs](https://redis.io/commands/ZINTERSTORE)
+    fn zinterstore_count_weights<D: ToSingleRedisArg, K: ToRedisArgs, W: ToRedisArgs>(dstkey: D, keys: &'a [(K, W)]) -> (usize) {
```
nihohit: why is the sum an integer and not floating point? can't the weights be floating points?

nihohit:
```
ZINTERSTORE t1 3 s1 s2 s3 WEIGHTS 10 5 3 AGGREGATE COUNT
  -> foo: 18
```

this isn't the behavior I see [in the redis REPL](https://redis.io/docs/latest/commands/zinterstore/). can you please explain? 

```
redis> ZINTERSTORE out 2 zset1 zset2 WEIGHTS 2 3 aggregate count
(integer) 2
```

nihohit (commented):
not a full review yet, initial comments

redis/src/commands/mod.rs
StefanPalashev: The signatures are consistent with the pattern established by the existing functions.

redis/src/commands/mod.rs
StefanPalashev: The command returns an integer indicating the number of members in the resulting sorted set at the destination key. The weights can be floating points.
Here's an example (executed on the same sets from the example above):
```
ZUNIONSTORE result 3 s1 s2 s3 WEIGHTS 10.5 5.3 3 AGGREGATE COUNT
(integer) 2

ZRANGE result 0 -1 WITHSCORES
1) "bar"
2) "15.8"
3) "foo"
4) "18.8"
```

StefanPalashev:
> ```
> ZINTERSTORE t1 3 s1 s2 s3 WEIGHTS 10 5 3 AGGREGATE COUNT
>   -> foo: 18
> ```
> 
> this isn't the behavior I see [in the redis REPL](https://redis.io/docs/latest/commands/zinterstore/). can you please explain?
> 
> ```
> redis> ZINTERSTORE out 2 zset1 zset2 WEIGHTS 2 3 aggregate count
> (integer) 2
> ```

The behavior of `COUNT` isn't clearly documented, and the "Command and output:" section above may be somewhat confusing because it does not show the direct output of the command. The command itself returns `2`, which is the number of members in the resulting sorted set (`foo` and `bar`). If the resulting set (`t1`) is then inspected, the score for `foo` will be `18`, derived from the specified weights: `10` (from `s1`) + `5` (from `s2`) + `3` (from `s3`).

redis/src/commands/mod.rs
nihohit: true, but this change adds 32 new functions, and the next argument will, what? add 64 new functions?

nihohit:
@StefanPalashev please rebase over https://github.com/redis-rs/redis-rs/pull/2193 - adding the COUNT to the options types will prevent the explosion of new functions.

StefanPalashev:
@nihohit, the changes are backwards incompatible, should this pull request target version 2 instead?
Why can't we have those commands in as they are and have the APIs changed in the future?

nihohit:
> the changes are backwards incompatible, should this pull request target version 2 instead?

yes.

> Why can't we have those commands in as they are and have the APIs changed in the future?

Because I'm not intereseted in doing the work twice. I have limited maintenance time, and I'd like to make sure that the time is utilized in the best way.

nihohit:
#2193 was merged, can you please rebase?

StefanPalashev:
Rebased

nihohit (approved):
Thanks!

PR #2162, opened 2026-06-26
tests: Set `test_basic::test_object_freq_command`'s comments straight

somechris:
`test_basic::test_object_freq_command` was split into two in 7d066c6 to address flakyness in CI. While the refactor worked, it left a few dangling comments (`object_idletime` argument is no longer valid, assert checking for 2 while it's comment said to check for 1).

We set these comments straight while details are still fresh, and as `OBJECT FREQ` itself is severly underdocumented also explain why the test is supposed to work in first place.

### DIFF
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -2552,10 +2552,11 @@ mod basic {
         let ctx = TestContext::new();
         let mut con = ctx.connection();
 
+        // Set a key to later check access frequency on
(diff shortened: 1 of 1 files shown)


somechris:
Aborting for now, as closer inspection showed that the test's setup was relying on the LFU's abuse of the object's LRU field of Redis. I'll see if I can re-do the test in a less abusing fashion.

PR #2164, opened 2026-06-26
tests: Avoid relying on Redis bug in `test_object_freq_command`

somechris:
`test_basic::test_object_freq_command` was split into two in 7d066c6 to address flakyness in CI. While the refactor worked, it left a few dangling comments (`object_idletime` argument is no longer valid, assert checking for 2 while it's comment said to check for 1).

Also, by first setting the value and switching to the LFU policy afterwards, the test relied on a Redis bug [1].

We set these comments straight while details are still fresh, and re-formulate to not longer rely on the Redis bug.

[1]: https://github.com/redis/redis/issues/15375

### DIFF
--- a/redis/tests/test_basic.rs
+++ b/redis/tests/test_basic.rs
@@ -2552,10 +2552,7 @@ mod basic {
         let ctx = TestContext::new();
         let mut con = ctx.connection();
 
-        con.set("object_key_str", "object_value_str").unwrap();
-
-        // Needed for OBJECT FREQ and can't be set before object_idletime
-        // since that will break getting the idletime before idletime adjuts
+        // Enable an LFU `maxmemory-policy`. This is required for `OBJECT FREQ` to work.
         redis::cmd("CONFIG")
             .arg("SET")
             .arg(b"maxmemory-policy")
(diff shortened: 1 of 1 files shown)


nihohit:
is this behavior documented anywhere? I prefer to avoid relying on undocmented behavior of external services.

somechris:
> is this behavior documented anywhere?

I can read that question in three ways :thinking: ... I'll try to respond to all of them :+1:

You mean the buggy behaviour? Yesterday, I created the linked bug (redis/redis#15375). That already saw a confirmation.

Or do you mean the intended `LFU_INIT_VALUE` behaviour? That's documented in Redis itself (and quite a few blog posts from people, including [the blog of `antirez` themselves](https://antirez.com/news/109)). For example in Redis' [evict.c](https://github.com/redis/redis/blob/unstable/src/evict.c#L228), they explain their strategy. They explicitly [mention starting at `LFU_INIT_VALUE`](https://github.com/redis/redis/blob/unstable/src/evict.c#L251). Valkey moved the strategy documentation into [`lrflfu.c`](https://github.com/valkey-io/valkey/blob/unstable/src/lrulfu.c#L54) but it is otherwise the same. They also explicitly [mention starting at `LFU_INIT_VALUE`](https://github.com/valkey-io/valkey/blob/unstable/src/lrulfu.c#L73)

Or do you mean in the documentation of the `OBJECT FREQ` command itself? That command only guarantees to give ["the logarithmic access frequency counter"](https://redis.io/docs/latest/commands/object-freq/). But it makes no claim on its value whatsoever. If you only go by that page, any test is undefined behaviour.

At any rate: Before this PR, the test relies on Redis behaviour that is actively contradicts their documentation (`CONFIG SET` at runtime does not properly change the eviction policy) and their code comments. After the PR, the test uses `CONFIG SET` in a save way that is in sync with documentation and intent.

> I prefer to avoid relying on undocmented behavior of external services.

Same here.
I think this PR improves the situation.
A really clean solution relies on both Redis and Valkey improving their `OBJECT FREQ` documentations (or fixing the policy switching, which seems unlikely, as it would require much work for little outcome).

nihohit (approved):
Thanks!

Issue #2161, opened 2026-06-25
Proposal: allow ordered async address selection for custom DNS resolvers


honeyspoon:
## Problem

`AsyncDNSResolver` lets callers provide custom async DNS resolution, but redis-rs currently races every returned `SocketAddr` with `select_ok` when opening async TCP/TLS connections.

That is a good default for normal DNS resolution, but it means resolver ordering is not observable. A resolver cannot express: "try this selected address first, then fall back to the remaining resolved addresses in this order."

This matters for applications that pre-resolve one Redis hostname to many A/AAAA records and create multiple connections, such as a connection pool. The application may want each new connection to start with a deliberately selected address from the full answer set, while still preserving same-attempt fallback if that first address fails.

## Current workaround

A custom resolver can return only one `SocketAddr` per `resolve` call. That forces deterministic dialing, but it loses fallback to the other resolved addresses within the same connection attempt.

A custom resolver can also return the full ordered answer set, but redis-rs races the addresses, so the resolver's preferred order is not honored.

## Prior art in this repository

- #871 reported that async connections only tried the first resolved address while sync connections tried multiple addresses.
- #872 fixed that by trying async resolved addresses sequentially.
- #1578, #1581, and #1595 added the current `AsyncDNSResolver` hook.

This proposal keeps the current racing behavior as the default, but adds ordered fallback as an explicit async connection option.

## Proposed API

Add an async connection address selection policy:

```rust
#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
pub enum AsyncConnectionAddrSelection {
    #[default]
    Race,
    Sequential,
}
```

Expose it on `AsyncConnectionConfig`:

```rust
let config = AsyncConnectionConfig::new()
    .set_dns_resolver(my_resolver)
    .set_connection_addr_selection(AsyncConnectionAddrSelection::Sequential);
```

Behavior:

- `Race`: current behavior, race all addresses returned by `AsyncDNSResolver` with `select_ok`.
- `Sequential`: try addresses in iterator order and return the first successful connection.

## Why this belongs in redis-rs

`AsyncDNSResolver` already makes address resolution pluggable. This change makes that existing hook expressive enough for ordered fallback without adding provider-specific behavior or replacing TCP/TLS connector internals.

The resolver remains responsible for DNS policy. redis-rs only decides how to consume the resolver's ordered address iterator.

## Non-goals

- No provider-specific DNS logic.
- No new DNS dependency in redis-rs.
[...]

PR #2165, opened 2026-06-26
Add async address selection policy

honeyspoon:
## Summary

Adds an async connection address selection policy so callers can choose whether redis-rs races all DNS results or connects to resolved addresses sequentially.

`AsyncConnectionAddrSelection::Race` remains the default and preserves existing behavior. `Sequential` tries resolved socket addresses in iterator order, which lets custom DNS resolvers control address distribution for `AsyncConnectionConfig` users.

The setting is wired through multiplexed async connections. The policy type lives next to `AsyncConnectionConfig` and is re-exported as `redis::AsyncConnectionAddrSelection`.

Refs #2161.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p redis --features tokio-comp --lib addr_selection`
- `cargo test -p redis --features tokio-comp --lib connect_sequential`
- `cargo clippy -p redis --features tokio-comp --lib -- -D warnings`
- `cargo check -p redis --features cluster`
- `cargo check -p redis --features tokio-comp,connection-manager`
- `cargo check -p redis --features tokio-rustls-comp`
- `cargo check -p redis --features tokio-native-tls-comp`
- `cargo check -p redis --no-default-features --features tokio-comp`
- `cargo check -p redis --no-default-features --features cluster-async,tokio-comp`
- `cargo check -p redis --no-default-features --features cluster-async,smol-comp`
- `cargo check -p redis --no-default-features --features smol-comp,connection-manager`
- `cargo doc -p redis --no-deps --features tokio-comp,connection-manager,cluster-async`

### DIFF
--- a/redis/src/aio/connection.rs
+++ b/redis/src/aio/connection.rs
@@ -1,6 +1,7 @@
 use super::AsyncDNSResolver;
 use super::RedisRuntime;
 
+use crate::AsyncConnectionAddrSelection;
 use crate::connection::{ConnectionAddr, ConnectionInfo};
 #[cfg(feature = "aio")]
 use crate::types::RedisResult;
@@ -11,20 +12,20 @@ use crate::pipeline::Pipeline;
 use crate::{FromRedisValue, RedisError, ToRedisArgs};
 use futures_util::future::select_ok;
 use std::future::Future;
+use std::net::SocketAddr;
 
 pub(crate) async fn connect_simple<T: RedisRuntime>(
     connection_info: &ConnectionInfo,
     dns_resolver: &dyn AsyncDNSResolver,
+    connection_addr_selection: &AsyncConnectionAddrSelection,
 ) -> RedisResult<T> {
     Ok(match connection_info.addr {
         ConnectionAddr::Tcp(ref host, port) => {
             let socket_addrs = dns_resolver.resolve(host, port).await?;
-            select_ok(
-                socket_addrs
-                    .map(|addr| Box::pin(<T>::connect_tcp(addr, &connection_info.tcp_settings))),
-            )
+            connect_with_addr_selection(socket_addrs, connection_addr_selection, |addr| {
+                <T>::connect_tcp(addr, &connection_info.tcp_settings)
+            })
             .await?
-            .0
         }
 
         #[cfg(any(feature = "tls-native-tls", feature = "tls-rustls"))]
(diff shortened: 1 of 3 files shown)


redis/src/aio/connection_manager.rs
```diff
@@ -862,8 +886,35 @@ mod tests {
...
+            .set_connection_addr_selection(AsyncConnectionAddrSelection::Sequential);
+        assert_eq!(
+            config.connection_addr_selection,
+            AsyncConnectionAddrSelection::Sequential
+        );
+    }
+
     #[test]
     fn test_lazy_connection_manager_with_config() {
+        #[cfg(feature = "tokio-comp")]
```
nihohit: why was this needed?

redis/src/cluster_handling/sync_connection/mod.rs
```diff
@@ -293,6 +295,19 @@ impl ClusterConfig {
         self.async_dns_resolver = Some(std::sync::Arc::new(resolver));
         self
     }
+
+    /// Set how async TCP connection attempts use addresses returned by DNS resolution.
+    ///
+    /// The default is [`crate::AsyncConnectionAddrSelection::Race`], which preserves the
+    /// existing redis-rs behavior of racing all returned socket addresses.
+    #[cfg(feature = "cluster-async")]
+    pub fn set_connection_addr_selection(
```
nihohit: why is this an option in sync connections?

redis/src/io/dns.rs
```diff
@@ -11,3 +15,55 @@ pub trait AsyncDNSResolver: Send + Sync + 'static {
         port: u16,
     ) -> crate::RedisFuture<'a, Box<dyn Iterator<Item = SocketAddr> + Send + 'a>>;
 }
+
+/// Controls how async TCP connection attempts use addresses returned by DNS resolution.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum AsyncConnectionAddrSelection {
```
nihohit: why were these moved?

redis/src/aio/connection_manager.rs
honeyspoon: not needed

redis/src/io/dns.rs
honeyspoon: I'll put it back

redis/src/cluster_handling/sync_connection/mod.rs
honeyspoon: yeah out of scope

honeyspoon:
updated

redis/src/aio/connection.rs
```diff
@@ -70,6 +70,49 @@ pub(crate) async fn connect_simple<T: RedisRuntime>(
...
+    socket_addrs: impl Iterator<Item = SocketAddr>,
+    mut connect: F,
+) -> RedisResult<T>
+where
+    F: FnMut(SocketAddr) -> Fut,
+    Fut: Future<Output = RedisResult<T>>,
+{
+    let mut last_error = None;
+    for socket_addr in socket_addrs {
+        match connect(socket_addr).await {
```
nihohit, line 102: this being sequential doesn't mean that the requests can't be sent concurrently, only that the results are chosen by order. why not use `join_all` here?

honeyspoon: actually this comment made me realize my pr makes no senes
I was assuming the iterable was saved in the state, so each new connection would take the next one in the iterable
but instead it just creates a new iterable all the time
so if it's sequential it will just always pick the first
instead I can keep the state for the sequence/round robin in the async dns resolver

nihohit: Oh, I didn't realize that was your intention 😅

PR #2167, opened 2026-06-28
Makefile: Select module tests directly to reduce output

somechris:
We selected module tests by their test name, which first collected all tests into the candidate set and then skipped those that were not for the given module. This resulted in 100s of `SKIPPED` lines and made it hard to spot the signal.

By selecting the module tests directly, the output of the module tests easily fits on a single screen and better exposes what's going on.


nihohit (approved):
thanks!

PR #2168, opened 2026-06-28
Add support for Valkey's `bloom` module

somechris:


.github/workflows/rust.yml
```diff
@@ -181,6 +213,20 @@ jobs:
...
+        with:
+          path: |
+            /tmp/valkey-bloom/target/release/libvalkey_bloom.so
+          key: ${{ runner.os }}-valkey-bloom-${{ matrix.config.bloom-module-version }}
+
+      - name: Copy Valkey Bloom into place
+        if: startsWith(matrix.config.bloom-module-version, 'valkey-bloom:')
+        run: |
+          cp /tmp/valkey-bloom/target/release/libvalkey_bloom.so "$REDISRS_REDIS_BLOOM_PATH"
+
```
somechris, line 229: Having to duplicate these 2/3 steps from the `build-and-test` job is somewhat not ideal. But I cannot see a better way. Do you?

In the long run ... should building and setting up modules eventually become part of the `install-db` action?

redis/examples/bloom.rs
```diff
@@ -4,6 +4,10 @@
 
 use redis::bloom::{BloomFilterDumpChunk, BloomFilterDumpIterator};
 use redis::{Connection, RedisResult, TypedCommands};
+use crate::support::{AvailableComponents, REDIS_BLOOM_ANY};
+
+#[path = "../tests/support/mod.rs"]
+mod support;
```
somechris: That `mod` with `path` is kind of nasty. But as `SCANDUMP`/`LOADCHUNK` is only necessary and available with Redis' `bloom` module, we need a version check, which right now is only available in the tests.

Here, we'd profit, if `AvailableComponents` was more generally available.

So I wondered ... do you think that would apply also elsewhere?
Would general `redis-rs` users find `AvailableComponents` useful?
(E.g.: To ease planning/testing when switching from Redis to Valkey or vice versa)?

.github/workflows/rust.yml
nihohit: Yes, setting up the modules should be a part of installing the DB

somechris: As this includes messing with Redis' JSON module which seems unrelated, I hope it's ok to uncouple it and I've filed #2170 for it.

redis/examples/bloom.rs
nihohit: I think I'll need a clearer problem space and solution design in order to answer that.
Without considering it deeply, I can't think of something which is easy to use and informative.

nihohit (approved):

PR #2184, opened 2026-07-01
perf(parser): fast-path RESP decode, bypassing combine for common replies

fcostaoliveira:
## Problem

Profiling the client read path — a pipelined `GET` workload over a multiplexed connection — showed **~60% of client CPU inside `combine`'s partial-state parser-combinator machinery** (`parse_with_state` / `AnySendSyncPartialStateParser::parse_partial` / `PartialMode` / `ThenPartial` / `AndThen` / `Recognize`). That machinery exists to resume a parse across a buffer boundary, but for the overwhelmingly common case — a complete reply already sitting in the buffer — it is pure overhead.

## Change

A small hand-written decoder (`fast_parse_value`) for the high-frequency RESP types (`+ - : $ * _`) that runs directly over the byte slice. It returns `Some((value, consumed))` **only** when it fully parses a supported, well-formed value with semantics identical to the `combine` grammar (including `combine`'s UTF-8 validation of line bodies). For anything else — unsupported RESP3 types (`% ~ | , # ! = ( >`), incomplete buffers, or malformed input — it returns `None` and the caller **falls back to `combine`**, which stays the single source of truth for behavior and error reporting.

Wired into `parse_redis_value` (slice) and the async `ValueCodec` (guarded by a `partial` flag so a fast-path success never interleaves with a `combine` resume-across-reads sequence; the flag is also reset on the error path so the invariant is self-enforcing).

**Why keep `combine` rather than replace it:** the fast path is a pure, low-risk accelerator for the common types; `combine` remains authoritative for the full RESP3 grammar, partial reads, and all error reporting, so behavior can only ever match it.

## Benchmarks

`cargo bench -p redis --bench bench_basic -- decode`, base = this PR's merge-base, on a quiet Cascade Lake box (`critcmp`):

```
group            base                          changes
decode/decode    4.40   1051.6±26.77ns         1.00   239.0±1.51ns
```

**~4.4× faster** on the pure decode microbench. End-to-end, a pipelined `GET` over a multiplexed connection improves **~+20%** (the sync single-`GET` path, which is dominated by the network round-trip, is unchanged — a useful control that the gain is the parser).

## Correctness

The fast path is a pure accelerator that defers to `combine` for anything it does not fully handle, so it can only ever agree with `combine`. Tests:

- **`fast_path_matches_combine`** — supported types + edge cases (`$-1`/`*-1`→`Nil`, `+OK`→`Okay`, `$0`, recursion-depth boundary), and that unsupported RESP3 types **and non-UTF-8 line bodies** are declined (each cross-checked against `combine`).
- **`fast_path_never_diverges_from_combine_fuzz`** — 100k RESP-shaped random inputs; `parse_redis_value` (fast path + fallback) must be **observationally identical** to pure `combine` — same value on `Ok`, and `Err ⇔ Err`.
- **`fast_path_roundtrip_matches_value`** — arbitrary RESP value trees round-trip through `parse_redis_value` back to the original value.
- **`codec_fast_path_and_partial_interleave`** — the async `ValueCodec` decoded across **every chunk boundary** (fully-buffered = fast path, byte-at-a-time = `combine` partial-resume, and every mix) yields the exact value sequence with no stranded bytes.

The existing `afl/parser` target fuzzes `parse_redis_value`, so it now also exercises the fast path.

## Notes

- No new dependencies; single file of production code (`redis/src/parser.rs`) plus tests.
- MSRV: uses an `if let`-chain, stabilized in 1.88 (the crate's MSRV).

### DIFF
--- a/redis/src/parser.rs
+++ b/redis/src/parser.rs
@@ -68,6 +68,139 @@ pub fn get_push_kind(kind: String) -> PushKind {
     }
 }
 
+// ---------------------------------------------------------------------------
+// Hand-written RESP fast-path decoder.
+//
+// Profiling the client read path showed most of the CPU spent inside `combine`'s
+// partial-state parser-combinator machinery. That machinery exists to resume a
+// parse across a buffer boundary,
+// but for the overwhelmingly common case — a *complete* reply already sitting
+// in the buffer — it is pure overhead.
+//
+// `fast_parse_value` decodes one complete RESP value directly from a byte
+// slice for the high-frequency types (`+ - : $ * _`). It returns
+// `Some((value, consumed))` ONLY when it fully parses a supported, well-formed
+// value with semantics identical to the `combine` grammar below. For anything
+// else — an incomplete buffer, an unsupported RESP3 type (`% ~ | , # ! = ( >`),
+// or malformed input — it returns `None`, and the caller falls back to the
+// `combine` parser, which remains the single source of truth for correctness
+// and error reporting. This keeps the fast path a pure accelerator that can
+// only ever agree with `combine`, never diverge from it.
+
+/// Index of the `\r` of the first `\r\n` at or after `from`, or `None` if the
+/// buffer does not yet contain a line terminator.
+#[inline]
+fn find_crlf(buf: &[u8], from: usize) -> Option<usize> {
+    let mut i = from;
+    while let Some(off) = buf[i..].iter().position(|&b| b == b'\r') {
+        let cr = i + off;
+        match buf.get(cr + 1) {
+            Some(b'\n') => return Some(cr),
+            Some(_) => i = cr + 1,
+            None => return None,
+        }
+    }
+    None
+}
+
+/// Read a `\r\n`-terminated line starting at `pos`. Returns the content (without
+/// the terminator) and the position just past the `\r\n`.
+#[inline]
+fn read_line(buf: &[u8], pos: usize) -> Option<(&[u8], usize)> {
+    let cr = find_crlf(buf, pos)?;
+    Some((&buf[pos..cr], cr + 2))
+}
+
+/// Parse one RESP value from `buf` starting at `pos`. Returns `(value, new_pos)`
+/// on success. `depth` guards against unbounded recursion, matching the
+/// `combine` grammar's `MAX_RECURSE_DEPTH` behavior (fall back on excess).
+fn fast_parse_at(buf: &[u8], pos: usize, depth: usize) -> Option<(Value, usize)> {
+    if depth > MAX_RECURSE_DEPTH {
+        return None;
+    }
+    let marker = *buf.get(pos)?;
+    let body = pos + 1;
+    match marker {
+        b'+' => {
+            let (line, np) = read_line(buf, body)?;
+            let s = str::from_utf8(line).ok()?;
+            let v = if s == "OK" {
+                Value::Okay
+            } else {
+                Value::SimpleString(s.into())
+            };
+            Some((v, np))
+        }
+        b'-' => {
+            let (line, np) = read_line(buf, body)?;
+            let s = str::from_utf8(line).ok()?;
+            Some((Value::ServerError(err_parser(s)), np))
+        }
+        b':' => {
+            let (line, np) = read_line(buf, body)?;
+            let s = str::from_utf8(line).ok()?;
+            let n = s.trim().parse::<i64>().ok()?;
+            Some((Value::Int(n), np))
+        }
+        b'_' => {
+            // RESP3 null. combine's `null()` runs through the same UTF-8-validating
+            // `line()` as every other type, so a non-UTF-8 body is rejected there;
+            // validate here too (else the fast path would accept `_\xff\r\n` as Nil
+            // while combine errors). The content is otherwise ignored.
+            let (line, np) = read_line(buf, body)?;
+            str::from_utf8(line).ok()?;
+            Some((Value::Nil, np))
+        }
+        b'$' => {
+            let (line, np) = read_line(buf, body)?;
+            let s = str::from_utf8(line).ok()?;
+            let size = s.trim().parse::<i64>().ok()?;
+            if size < 0 {
+                return Some((Value::Nil, np));
+            }
+            let size = size as usize;
+            let end = np.checked_add(size)?;
+            // Need the payload plus its trailing CRLF.
+            if end + 2 > buf.len() {
+                return None;
+            }
+            if &buf[end..end + 2] != b"\r\n" {
+                return None; // malformed → let combine report it
+            }
+            Some((Value::BulkString(buf[np..end].to_vec()), end + 2))
+        }
+        b'*' => {
+            let (line, np) = read_line(buf, body)?;
+            let s = str::from_utf8(line).ok()?;
+            let len = s.trim().parse::<i64>().ok()?;
+            if len < 0 {
+                return Some((Value::Nil, np));
+            }
+            let len = len as usize;
+            // Cap the pre-allocation so a bogus length can't trigger a huge
+            // up-front allocation; the real bound is the buffer contents.
+            let mut out = Vec::with_capacity(len.min(1024));
+            let mut cur = np;
+            for _ in 0..len {
+                let (v, next) = fast_parse_at(buf, cur, depth + 1)?;
+                out.push(v);
+                cur = next;
+            }
+            Some((Value::Array(out), cur))
+        }
+        _ => None, // unsupported RESP3 type → fall back to combine
+    }
+}
+
+/// Fast-path entry point: decode one complete supported RESP value from the
+/// start of `buf`. See the module note above for the contract.
+#[inline]
+fn fast_parse_value(buf: &[u8]) -> Option<(Value, usize)> {
+    // Start at depth 1 to match the `combine` grammar, whose recursion `count`
+    // starts at 1 and errors when it exceeds MAX_RECURSE_DEPTH.
+    fast_parse_at(buf, 0, 1)
+}
+
 fn value<'a, I>(
     count: Option<usize>,
 ) -> impl combine::Parser<I, Output = Value, PartialState = AnySendSyncPartialState>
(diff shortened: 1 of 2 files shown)


fcostaoliveira:
Follow-up that builds on this PR: #2185 extends the same fast path to the synchronous connection (+14–17% on a pipelined sync GET). It includes this PR's commit as its base; happy to rebase it to just the sync delta once this lands.

nihohit:
Hi, thanks for the contribution!
I still didn't look at the code, but this sounds like a good contribution, but I'm worried about the benchmarks here - real network communications provide partial values, and I'd like to see how this code performs against a remote DB, with very large values that will arrive over multiple TCP packets.

Also, please run `make lint` and `make test` locally before pushing fixes, to be sure that your commits will pass CI.

fcostaoliveira:
Thanks for the quick feedback!

**`make lint` / `make test`** — good call: running `make lint` (`clippy --workspace --all-targets --all-features`) caught an unused `use` in a test module that a narrower `cargo clippy -p redis` had missed. Fixed and force-pushed. Both are green now on the updated branch:
- `make lint` → clean.
- `make test` → `713 tests run: 713 passed` + `56 passed` (2 flaky sentinel retries, unrelated).

**Remote DB with large multi-packet values** — agreed this is the important case, since the fast path only helps when a *complete* reply is already buffered and must fall back to `combine` for anything that spans multiple `recv()`s. I ran the client and the server on **two separate physical hosts** (real TCP, so large replies arrive over many segments), single `GET`, base vs this PR:

| value size | base sync / async | this PR sync / async |
|-----------:|-------------------|----------------------|
| 16 KB  | 375.8 / 379.6 µs | 374.6 / 379.9 µs |
| 256 KB | 2658 / 2652 µs   | 2655 / 2640 µs |
| 4 MB   | 38444 / 38450 µs | 38668 / 38703 µs |

i.e. **no regression** on large multi-packet values (within run-to-run noise; these are network-transfer-bound at ~100 MB/s, where parsing is not the bottleneck). The wins I reported are on **small values under pipelining / multiplexing**, where per-reply parsing dominates and the reply is already buffered — that's exactly the case the fast path targets, and it defers to `combine` (with a one-time cheap check) otherwise.

Happy to add a partial-read regression test or run a different size/shape if useful.

fcostaoliveira:
Extended the remote (two-host, real TCP) single-`GET` sweep down to small values as well, base vs this PR:

| value size | base sync / async | this PR sync / async |
|-----------:|-------------------|----------------------|
| 10 B   | 86.3 / 89.3 µs   | 85.6 / 89.1 µs |
| 100 B  | 151.7 / 153.0 µs | 153.3 / 152.3 µs |
| 1 KB   | 159.4 / 158.4 µs | 157.8 / 158.5 µs |
| 16 KB  | 373.7 / 379.4 µs | 374.5 / 378.9 µs |
| 256 KB | 2658 / 2622 µs   | 2658 / 2659 µs |
| 4 MB   | 38663 / 38706 µs | 38469 / 38531 µs |

A single `GET` over the network is **round-trip-bound at every size** (10 B is already ~86 µs), so the parser change — a few hundred ns saved per reply — is in the noise here: **no regression, and no meaningful gain, at any size**. The measurable win is on the CPU-bound path where the round-trip is amortized (pipelined / multiplexed small values), which is what the `decode`/`get_pipeline` numbers in the description show. This is just to confirm the real-network end-to-end behavior across the full size range is neutral.

fcostaoliveira:
For completeness, the setup behind the remote numbers above (I should have included this):

- **Two identical hosts**: Intel Xeon Gold 6248 (Cascade Lake), Ubuntu 22.04 — redis on one, the benchmark client on the other.
- **Link**: 1 GbE, MTU 1500, LAN RTT ≈ 0.13–0.38 ms. Single connection, one in-flight `GET`.

Caveat worth flagging: on 1 GbE the **large-value** cases are bandwidth-bound (4 MB GET ≈ 109 MB/s ≈ line rate), which is *why* the parser change is invisible there — the client is waiting on the wire, not the CPU. On a faster (10/25 GbE) link a large `GET` would be far less bandwidth-bound and the client-CPU fraction (RESP parse + the bulk-string copy) proportionally larger, so that setup would be more sensitive for large values. The change's measurable win is on the CPU-bound path (pipelined/multiplexed, RTT amortized), independent of link speed; the remote runs here are to confirm **no regression** end-to-end, which holds across 10 B → 4 MB.

nihohit:
> so the parser change — a few hundred ns saved per reply — is in the noise here

This what I'm worried about. I'm not sure that this is an actual improvement for the average user, but it does incur a maintenance cost. Considering that I'm in the process of refactoring the parsing pipeline (https://github.com/nihohit/redis-rs/tree/shachar/use-bytes), I'm not sure that I'm willing to approve this.
I'll go over the code and get a more informed opinion.

fcostaoliveira:
Totally fair concerns — let me address both, and I looked at `shachar/use-bytes` since it's directly relevant.

**On "in the noise for the average user":** agreed for a single blocking `GET` — that's round-trip-bound, and I said as much. The change isn't aimed at single-command latency; it's per-reply **CPU**, which only becomes the bottleneck once the round-trip is amortized — i.e. pipelines and the multiplexed connection under load. On the pure decode microbench (`bench_basic` `decode`, no I/O), on a quiet Cascade Lake box:

| | decode |
|---|---|
| base (`combine`) | 1.04 µs |
| this PR (fast-path) | **0.25 µs** (~4×) |

which shows up end-to-end as ~+20% on a pipelined multiplexed `GET`. So the beneficiaries are throughput-bound users (pipelining / high-concurrency multiplexing), not the single-`GET` case.

**On the `use-bytes` refactor (and maintenance cost):** I think these are orthogonal rather than competing. `use-bytes` keeps `combine` as the grammar and makes *materialization* zero-copy (slicing the frozen `Bytes`) — that removes the `to_vec` copy, which is the win on **large** values. The fast-path instead skips `combine`'s partial-state parser-combinator machinery (the ~60% of client CPU the profile showed) for a complete buffered reply — a win on **all** sizes, biggest on small/pipelined. For reference I ran decode on your branch too: base 1.04 µs → use-bytes 1.27 µs → fast-path 0.25 µs (the branch is WIP, so not a fair perf comparison — I only point it out because it shows the two target *different* costs: the copy vs. the parser).

So they compose: I'd be glad to rebase the fast-path to produce your `ValueRange` representation instead of owned `Value`s, so you get both the zero-copy materialization *and* the combine-bypass. Happy to follow your direction there.

**If the maintenance surface is the blocker**, I can also scope this down. The fast-path is a pure accelerator — `combine` stays the single source of truth, and a 100k-input differential fuzz asserts `parse_redis_value` == `combine` on every input — but it is still added code. One option is to land only the piece with the clearest per-reply payoff: **RESP3 coverage** (map/set/double/push/…), which gives e.g. `HGETALL`/`CONFIG GET` decode ~3.5× and client-tracking invalidation push ~4× — real CPU per reply for RESP3 (7.x default), independent of latency.

No worries at all if you'd rather fold this into your refactor yourself — your call as maintainer. Just let me know which direction is most useful and I'll adapt (rebase onto `use-bytes`, split out RESP3, or close in favor of your work).

nihohit:
I suggest waiting until my branch is merged, and then checking if you can improve on the change there, without a lot of complex code.

fcostaoliveira:
Sounds good — that's sensible sequencing, and I'd much rather build on your refactor than duplicate it. I'll wait for `use-bytes` to land, then see what a fast-path adds on top of it, kept minimal (your "without a lot of complex code" is well taken).

Happy to close this and #2185 now to keep your queue clean and reopen against the new baseline once your branch merges — or leave them as reference if you'd prefer. Your call.

(And thanks for merging #2186 and #2187!)

fcostaoliveira:
Hi @nihohit 👋 — friendly follow-up on the plan we agreed here (wait for `use-bytes` to land, then rebuild a minimal fast-path on top of it).

Is there anything we can do on our end to help move `use-bytes` forward? Happy to pitch in wherever it's actually useful, e.g.:

- **Benchmarking the branch** against the decode/workload benches we've built up (pipelined small-value GETs, SMEMBERS/HGETALL/ZRANGE-shaped replies, MGET at various sizes) on the same two-host setup used for #2195 — so it lands with numbers attached;
- **Differential testing** it against current main — we have a quickcheck + AFL harness that compares parser outputs input-by-input, which transfers directly to validating the `Bytes`-based materialization;
- **Drafting the fast-path rebuild on top of the branch** as a proof that the two compose, so you can see the end state before committing to either half;
- or simply **reviewing** when it becomes a PR.

No rush at all — the plan still makes sense to us and waiting is fine. Just wanted to make sure it isn't stalled on something we could be helping with. And thanks for the quick turnaround on #2194 and #2196! 🙏

nihohit:
I generated `use-bytes` using Claude, and still didn't finish reviewing it, specifically the last commit, where the main logic change sits. I'm also unsure what testing I need to add in order to verify that the low-level behavior around socket management and async I/O didn't change.

So, feel free to pull the branch, test it, enhance it, and even open it as a PR. I don't care about the credit :)
If you have spare tokens to throw at it, I'd love hearing what Fable / Codex thinks :)

fcostaoliveira:
Fantastic — on it! 🚀 Plan: (1) a deep review of the branch with extra attention on the last commit and specifically on socket-management / async-I/O behavioral equivalence, (2) our differential fuzz harness (quickcheck + AFL) comparing parser outputs old-vs-new input-by-input, (3) two-host benchmarks (pipelined small values, SMEMBERS/HGETALL/ZRANGE/MGET shapes, plaintext + mTLS) so it lands with numbers, then (4) fixes as needed and a PR with the findings written up. Will report back here.

fcostaoliveira:
Done — opened as #2199 (targeting `2.0.x`, rebased onto current head incl. #2196). Summary of what we did with it:

**Review (the "what Fable thinks" part):** the design is sound. The key insight holding it together is that `ValueCodec` became stateless — parsing always restarts with fresh state over the current buffer, which is exactly what makes the pointer-derived offsets safe across buffer reallocations, and `split_to().freeze()` guarantees every recorded range lands inside the frozen frame. The `unsafe` surface (`Str::from_utf8_unchecked`) was audited twice independently: every range that becomes a `Str` provably passes `str::from_utf8` first; raw bulk bytes only ever become `Bytes`.

**Your socket/async-I/O question:** behavior is preserved — the `Framed` read-buffer management and the multiplexed sink/flush logic are untouched (deadlock-regression tests still cover them), and `read_buf` on a `BytesMut` can never see zero spare capacity, so `Ok(0)` still means EOF. The testing that locks this in: your `partial_io_parse` quickcheck (fragmented/would-block reads), a new test we added for **multiple pipelined values through one reused buffer in 7-byte chunks** (the real multiplexed read shape), the codec boundary/EOF unit tests, and the full suite.

**Differential testing:** we replayed **1,896 coverage-guided inputs** (an AFL corpus evolved against the parser grammar) through the old parser and yours, comparing canonically-normalized output. **Zero unintended divergences** — the only 4 diffs are your documented lossy→strict UTF-8 change for verbatim/blob-error payloads.

**Benchmarks:** independently reproduced on an idle x86 server — 1.36×–2.73× across your payload set, biggest on many-element responses (the `MGET`/`SMEMBERS`/`HGETALL` shape).

**Three small commits on top** (details in the PR): error strings are now copied out of the frame so a stored `ServerError` never pins a whole reply buffer; `parse_redis_value_async` reserves before `read_buf` so reads don't degrade to tiny chunks after a freeze; plus the pipelined-fragmentation test. Also documented the trade-offs in `version2.md` (buffer retention, where the `Vec<u8>` copy moved, fragmented-arrival re-parse) and moved `PR_DESCRIPTION.md` out of the tree into the PR body.

`make test` is fully green on the rebased branch. Happy to iterate on anything!

PR #2185, opened 2026-07-01
perf(parser): extend the RESP fast-path to the sync connection

fcostaoliveira:
Follow-up to #2184, which adds the RESP fast-path for the async `ValueCodec` and `parse_redis_value`. **This PR builds on #2184** (its commit is the first of the two here) and extends the same fast path to the **synchronous connection**. Please review/merge #2184 first; once it lands this rebases to just the second commit.

## Change

`Parser::parse_value` (the sync read path) now tries `fast_parse_value` over the decoder's **already-buffered** bytes before invoking `combine`:

```rust
if let Some((value, consumed)) = fast_parse_value(self.decoder.buffer()) {
    self.decoder.advance(&mut reader, consumed);
    return Ok(value);
}
// … unchanged combine::decode! fallback
```

In a pipelined sync read, one `read` syscall buffers many replies, so only the first `parse_value` blocks on I/O and calls 2..N decode straight from the buffer, skipping `combine`'s partial-state machinery. Anything the fast path doesn't fully handle (empty/incomplete buffer, unsupported RESP3 type) falls through to `combine` unchanged.

## Why it's safe to bypass `combine` here

- `combine`'s `Buffer::advance(_read, len)` just drops `len` bytes from the front of the buffer; it ignores the reader.
- `combine`'s partial `state` is only non-empty when `combine` returns an incomplete result. `parse_value` is blocking — its `combine::decode!` loop only returns `Ok` on a **complete** parse (state consumed/reset) or `Err`. So whenever the fast path runs (at the top of a fresh call), `state` is empty; a fast-path success can never be interleaved into a `combine` resume-across-reads sequence.
- `combine`'s `position` is only read to seed the stream (the redis-rs init closure ignores it) and to compute an error-message offset against the **current** buffer (`BytesMut::advance` only bumps the start pointer within the same allocation), so offsets stay correct — never more than cosmetic, and here in fact correct. No panic path.

## Benchmarks

Pipelined GET over a sync connection, quiet Cascade Lake box (base = this PR's merge-base with #2184):

|---|---|---|---|
| `get_pipeline/sync_tls/16` | 1.337 Melem/s | 1.558 | **+16.6%** |
| `get_pipeline/sync_plaintext/16` | 1.391 Melem/s | 1.583 | **+13.9%** |

The async multiplexed path (already fast-pathed in #2184) and the pure `decode` bench are unchanged — a control that the gain is the sync change.

## Correctness

- `sync_parser_buffered_fast_path` — several concatenated replies (bulk, simple, int, an unsupported `%` map mid-stream, array, null) sharing one decoder buffer decode to the exact sequence (exercises the fast↔combine handoff).
- `sync_parser_error_after_fast_path` — after the fast path advances the shared buffer, a malformed value still makes `combine` return an error (no panic).
- Existing sync end-to-end tests (`test_basic`) and the full parser suite pass.

No new dependencies; the change is confined to `redis/src/parser.rs`.

### DIFF
--- a/redis/src/parser.rs
+++ b/redis/src/parser.rs
@@ -617,6 +617,17 @@ impl Parser {
 
     /// Parses synchronously into a single value from the reader.
     pub fn parse_value<T: Read>(&mut self, mut reader: T) -> RedisResult<Value> {
+        // Fast path: if the decoder already holds a complete supported value,
+        // decode it straight from the buffer and advance, skipping combine. This
+        // is the common case for pipelined reads, where a single `read` buffers
+        // many replies and only the first `parse_value` call actually blocks on
+        // I/O. Anything else — empty/incomplete buffer, or an unsupported RESP3
+        // type — falls through to combine, which handles reading more, partial
+        // state, EOF, and error reporting exactly as before.
+        if let Some((value, consumed)) = fast_parse_value(self.decoder.buffer()) {
+            self.decoder.advance(&mut reader, consumed);
+            return Ok(value);
+        }
         let mut decoder = &mut self.decoder;
         let result = combine::decode!(decoder, reader, value(None), |input, _| {
             combine::stream::easy::Stream::from(input)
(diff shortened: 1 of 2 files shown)


fcostaoliveira:
Ran the same checks as on #2184:

- `make lint` → clean; `make test` → `709 passed` + `56 passed` (flaky sentinel retry, unrelated).
- **Remote DB, large multi-packet values** (client and server on separate hosts, single `GET`, base vs this PR):

| size | base sync / async | this PR sync / async |
|-----:|-------------------|----------------------|
| 16 KB  | 373.7 / 379.4 µs | 372.0 / 378.9 µs |
| 256 KB | 2652 / 2634 µs   | 2667 / 2656 µs |
| 4 MB   | 38679 / 38658 µs | 38403 / 38543 µs |

No regression on large values that span many TCP segments (network-bound). This PR's win is on the **sync** pipelined/small-value path where per-reply parsing dominates; the sync fast path only fires when a complete reply is already buffered and otherwise defers to `combine`.

fcostaoliveira:
For completeness, the setup behind the remote numbers above (I should have included this):

- **Two identical hosts**: Intel Xeon Gold 6248 (Cascade Lake), Ubuntu 22.04 — redis on one, the benchmark client on the other.
- **Link**: 1 GbE, MTU 1500, LAN RTT ≈ 0.13–0.38 ms. Single connection, one in-flight `GET`.

Caveat worth flagging: on 1 GbE the **large-value** cases are bandwidth-bound (4 MB GET ≈ 109 MB/s ≈ line rate), which is *why* the parser change is invisible there — the client is waiting on the wire, not the CPU. On a faster (10/25 GbE) link a large `GET` would be far less bandwidth-bound and the client-CPU fraction (RESP parse + the bulk-string copy) proportionally larger, so that setup would be more sensitive for large values. The change's measurable win is on the CPU-bound path (pipelined/multiplexed, RTT amortized), independent of link speed; the remote runs here are to confirm **no regression** end-to-end, which holds across 10 B → 4 MB.

PR #2187, opened 2026-07-01
perf(cmd): encode into the Vec buffer instead of via io::Write

fcostaoliveira:
## Change

`write_command` is only ever called with a pre-reserved `Vec<u8>` (both call sites). It went through the generic `io::Write` trait — a `Result`-wrapped `write_all` for each of ~5 pieces per argument. This writes directly with `extend_from_slice` instead.

Output is byte-identical: `Vec<u8>`'s `write_all` *is* `extend_from_slice`; the buffer is already sized by `args_len` before writing, so no reallocation happens on the hot path.

## Benchmarks

Quiet Cascade Lake box, `bench_basic`:

|---|---|---|---|
| `encode/small` | 252 ns | 219 ns | **−13%** |
| `encode/integer` / `encode/pipeline` | — | — | flat |

End-to-end, a pipelined `GET` over the multiplexed connection is **~+2–3%** (each query re-encodes its commands, so `encode_pipeline`/`write_command` are a real share of the client CPU).

## Correctness

Byte-identical output; `make lint` clean and `make test` green (975+ tests across the tcp/tls/unix × RESP2/RESP3 matrices). Independent change — `cmd.rs` only.

### DIFF
--- a/redis/src/cmd.rs
+++ b/redis/src/cmd.rs
@@ -8,7 +8,7 @@ use futures_util::{
 use std::pin::Pin;
 #[cfg(feature = "cache-aio")]
 use std::time::Duration;
-use std::{fmt, io, io::Write};
+use std::{fmt, io::Write};
 
 use crate::pipeline::Pipeline;
 use crate::types::{FromRedisValue, RedisResult, RedisWrite, ToRedisArgs, from_redis_value};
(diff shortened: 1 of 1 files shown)


fcostaoliveira:
Ran `make lint` (clean) and `make test` (968 passed across the redis/valkey × RESP2/RESP3 × tcp/tls/unix matrices, plus 56; a few flaky sentinel retries unrelated) locally before pushing. Byte-identical output, so no behavioural change — purely dropping the `io::Write` `Result` wrapping on a path that only ever targets a pre-reserved `Vec`.

redis/src/cmd.rs
```diff
@@ -330,19 +330,23 @@ where
 
     cmd.reserve(totlen);
 
-    write_command(cmd, args, cursor).unwrap()
+    write_command(cmd, args, cursor)
 }
 
-fn write_command<'a, I>(cmd: &mut (impl ?Sized + Write), args: I, cursor: u64) -> io::Result<()>
+// Both encoding entry points target an in-memory, pre-reserved `Vec<u8>`, so
```
nihohit: remove the unnecessary doc

fcostaoliveira:
Done — removed the comment. Thanks for the review!

PR #2188, opened 2026-07-01
Make backpressure boundary configurable

kushudai:
We're using redis with larger payloads (~16MB P99s)
We see significant churn from CPU profiles that we were able to attribute to the tokio framed reallocations on writes because it frequently flushes and grows to a bounded set.

I was hoping we could simply reconfigure the backpressure boundary but that lever is not exposed by redis-rs
This PR is an attempt to change that. Please let me know if this direction is acceptable or if the knob exposes too much of the internals of how the crate works.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -44,7 +44,7 @@ futures-util = { version = "0.3.31", default-features = false, features = [
   "sink",
 ], optional = true }
 pin-project-lite = { version = "0.2", optional = true }
-tokio-util = { version = "0.7", optional = true }
+tokio-util = { version = "0.7.5", optional = true }
 async-lock = { version = "3", optional = true }
 tokio = { version = "1", features = [
   "sync",
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -890,8 +890,8 @@ mod tests {
         assert_eq!(config.write_backpressure_boundary, Some(16 * 1024 * 1024));
     }
 
-    #[test]
-    fn test_lazy_connection_manager_with_config() {
+    #[tokio::test]
+    async fn test_lazy_connection_manager_with_config() {
         // Test that lazy connection manager can be created with custom config
         let client = Client::open("redis://127.0.0.1/").unwrap();
         let config = ConnectionManagerConfig::new()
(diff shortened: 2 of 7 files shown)


nihohit (commented):
the change looks reasonable, but 
1. it isn't tested
2. you should probably add it also to the async cluster client

kushudai:
Thanks, let me do both and update this PR

kushudai:
@nihohit I think I've addressed the feedback. Please let me know if I missed something - thank you!

nihohit:
I apologize, I wasn't clear - you added tests that the configuration is saved, which is nice, but what I wanted was tests with a real connection, testing that setting the value doesn't break the connection.

redis/src/aio/connection_manager.rs
```diff
@@ -863,7 +878,20 @@ mod tests {
...
+
+    #[test]
+    fn test_connection_manager_config_write_backpressure_boundary_custom() {
+        let config =
+            ConnectionManagerConfig::new().set_write_backpressure_boundary(16 * 1024 * 1024);
+        assert_eq!(config.write_backpressure_boundary, Some(16 * 1024 * 1024));
+    }
+
+    #[tokio::test]
+    async fn test_lazy_connection_manager_with_config() {
```
kushudai, line 894: Sorry, this is unrelated to my change. I changed this to be a tokio test because it fails for me when running in isolation - I believe it works in CI because `all-features` pulls in the smol reactor - `new_lazy_with_config` needs an async runtime because it calls `runtime.spawn`. I can revert it if this is undesirable.

redis/Cargo.toml
```diff
@@ -44,7 +44,7 @@ futures-util = { version = "0.3.31", default-features = false, features = [
   "sink",
 ], optional = true }
 pin-project-lite = { version = "0.2", optional = true }
-tokio-util = { version = "0.7", optional = true }
+tokio-util = { version = "0.7.5", optional = true }
```
kushudai, line 47: This is a no-op for the lock file (which uses 0.7.18) but `backpressure_boundary` was added to `tokio-util` in 0.7.5 hence the bump (minimum version)

kushudai:
Very reasonable ask 😄 
Added and tested locally.

nihohit (approved):
Thanks!

PR #2191, opened 2026-07-02
Add support for the Redis Array data type

StefanPalashev:
# Add support for the Redis Array data type

## Summary

Add a high-level client interface for the new **Array** data type introduced in Redis 8.8. 
Note that this is currently a **_preview_** type and may be subject to change.

## Background - the Array data type

Redis 8.8 adds `array`, a **sparse, index-addressed** value type - values live at explicit integer indices, indices may have gaps and the type also maintains an internal **write cursor** for append-style insertion.

It is important to distinguish between two notions of length:
| Concept | Command | Meaning |
|---|---|---|
| Length | `ARLEN` | highest populated index + 1 (spans gaps) |
| Count | `ARCOUNT` | number of populated slots |

Small example:
```
ARSET arr 0 a          -> 1     # 1 new slot
ARMSET arr 2 c 5 f     -> 2     # 2 new slots
ARLEN arr              -> 6     # max index 5, + 1
ARCOUNT arr            -> 3     # three populated slots
ARGET arr 1            -> nil   # gap
```

## APIs

| Command | Syntax | Reply |
|---|---|---|
| `ARSET` | `ARSET key index value [value ...]` | int (new slots filled) |
| `ARGET` | `ARGET key index` | bulk string / nil |
| `ARLEN` | `ARLEN key` | int (max index + 1) |
| `ARCOUNT` | `ARCOUNT key` | int (populated count) |
| `ARDEL` | `ARDEL key index [index ...]` | int (deleted) |
| `ARMSET` | `ARMSET key index value [index value ...]` | int (new slots filled) |
| `ARMGET` | `ARMGET key index [index ...]` | array (nil for gaps) |
| `ARGETRANGE` | `ARGETRANGE key start end` | array (nil for gaps) |
| `ARDELRANGE` | `ARDELRANGE key start end [start end ...]` | int (deleted) |
| `ARINSERT` | `ARINSERT key value [value ...]` | int (last index written) |
| `ARNEXT` | `ARNEXT key` | int (next insert index) |
| `ARSEEK` | `ARSEEK key index` | int `1` |
| `ARRING` | `ARRING key size value [value ...]` | int (last index written) |
| `ARLASTITEMS` | `ARLASTITEMS key count [REV]` | array (nil for deleted) |
| `ARSCAN` | `ARSCAN key start end [LIMIT limit]` | array of `index value` pairs |
| `AROP` | `AROP key start end <SUM\|MIN\|MAX\|AND\|OR\|XOR\|MATCH value\|USED>` | varies (see Behavior) |
| `ARGREP` | `ARGREP key start end <predicate...> [AND\|OR] [LIMIT n] [WITHVALUES] [NOCASE]` | indices, or `index value` pairs with `WITHVALUES` |
| `ARINFO` | `ARINFO key [FULL]` | field/value map |

## Behavior

- **Sparse, non-shifting deletes.** `ARDEL` and `ARDELRANGE` clear slots **in place** - surviving elements are **not** shifted or compacted thus `ARLEN` only shrinks when the top index is removed.
- **Index validation.** - Negative indices are rejected server-side, so index parameters are of type `usize`.
- **Write cursor.**:
- `ARINSERT` writes at the cursor and returns the last written index
- `ARNEXT` reads it
- `ARSEEK` moves it
- `ARRING` wraps within a fixed size. 
The cursor is shared by `ARINSERT`/`ARRING`/`ARSEEK` but is **not** moved by `ARSET`. `ARLASTITEMS` can return `nil` for recently-inserted slots that were later deleted.
- **`AROP` replies vary by operation:**:
- `SUM` is integer or float
- `MIN`/`MAX` return the element value or `nil` for a non-numeric range
- `AND`/`OR`/`XOR` are integer bitwise reductions
- `MATCH value`/`USED` are integer counts.
- **`ARGREP` combines predicates with a single flat combinator** (default `OR`). Mixed `AND`/`OR` is unsupported and matches nothing, so grouped logic must be expressed with a single `RE` predicate. Its `start`/`end` accept the `-`/`+` sentinels (and reverse iteration).
- **`ARGREP` regex engine is RE2-style** - no backreferences or lookaround (both error) and `\w`/`\d` plus `NOCASE`/`(?i)` case folding are **ASCII-only** (non-ASCII such as `é`/`É` is matched literally).
- **Positive-only parameters.** `LIMIT` and ring `size` must be positive. The option builders reject a zero `LIMIT` and `arring`'s `size` is a `NonZeroUsize`, so invalid values never reach the wire.

## What the pull request adds

The full `AR*` command family - 18 commands across 20 trait methods with their argument/option types living in a dedicated `redis_arrays` module. Everything is gated behind a new opt-in `redis-arrays` Cargo feature.

All 20 methods are added as single skeletons in the shared `implement_commands!` block, so the sync/async and generic/typed traits are all generated. Each of them is annotated with both `#[cfg(feature = "redis-arrays")]` and `#[cfg_attr(docsrs, doc(cfg(...)))]`.

| Group | Methods |
|---|---|
| Random access | `arset`, `arget`, `arlen`, `arcount`, `ardel` |
| Bulk | `armset`, `armget` |
| Ranges | `argetrange`, `ardelrange` |
| Write cursor | `arinsert`, `arnext`, `arseek`, `arring`, `arlastitems` |
| Scan / search / aggregate | `arscan`, `arscan_options`, `arop`, `argrep`, `argrep_options` |
| Introspection | `arinfo` |

### Types module (`redis/src/commands/redis_arrays.rs`)

New module (re-exported from `lib.rs`) providing the argument/option types, each with a hand-written `ToRedisArgs`:

- `ArrayBound` (`Start`/`End`/`Index`, with `From<usize>`) - models `ARGREP`'s `-`/`+` open-ended range sentinels, which only `ARGREP` accepts.
- `ArrayPredicate` (`Exact` / `Match` / `Glob` / `Re`) - the `ARGREP` predicate kinds.
- `ArrayGrepCombinator` (`And` / `Or`) and `ArrayGrepOptions` (combinator, limit, `WITHVALUES`, `NOCASE`).
- `ArrayScanOptions` (`LIMIT`).
- `ArrayAggregateOp` (`Sum`/`Min`/`Max`/`And`/`Or`/`Xor`/`Match`/`Used`) for `AROP`.

### Testing

Eight feature-gated test groups covering the following areas:
- sparse semantics
- reverse ranges
- the write-cursor model
- ring wrapping
- per-operation AROP typing
- ARGREP boolean composition + regex and case-folding limitations
- edge cases.

| Test | Gating | Covers |
|---|---|---|
| `test_arrays_basic_operations` | `REDIS_CE_8_8` | `ARSET`/`ARGET`/`ARLEN`/`ARCOUNT`/`ARDEL`, sparse semantics |
| `test_arrays_multi_operations` | `REDIS_CE_8_8` | `ARMSET`/`ARMGET`, non-contiguous pairs, order-preserving reads |
| `test_arrays_range_operations` | `REDIS_CE_8_8` | `ARGETRANGE`/`ARDELRANGE`, reverse ranges, non-compacting deletes |
| `test_arrays_cursor_operations` | `REDIS_CE_8_8` | `ARINSERT`/`ARNEXT`/`ARSEEK`/`ARRING`/`ARLASTITEMS` |
| `test_arrays_scan_search_aggregate` | `REDIS_CE_8_8` | `ARSCAN`/`AROP`/`ARGREP`/`ARINFO` |
| `test_arrays_argrep_boolean_composition` | `REDIS_CE_8_8` | flat combinator; grouped logic via `RE` |
| `test_arrays_argrep_engine` | `REDIS_CE_8_8` | `NOCASE` scope, ASCII-only folding, no backrefs/lookaround, ASCII `\d` |
| `test_arrays_edge_cases` | `REDIS_CE_8_8` | missing keys, empty values, `WRONGTYPE` |

## Compatibility

- New code paths are gated behind the opt-in `redis-arrays` cargo feature (no dependencies, not in `default`).
- No existing public API changes.
- No extra dependencies - the type module is pure RESP.
- Server behavior is gated to `Redis CE 8.8+` via the `REDIS_CE_8_8` versioning constant.

## Relevant documentation

[`ARCOUNT`](https://redis.io/docs/latest/commands/arcount/), [`ARDEL`](https://redis.io/docs/latest/commands/ardel/), [`ARDELRANGE`](https://redis.io/docs/latest/commands/ardelrange/), [`ARGET`](https://redis.io/docs/latest/commands/arget/), [`ARGETRANGE`](https://redis.io/docs/latest/commands/argetrange/), [`ARGREP`](https://redis.io/docs/latest/commands/argrep/), [`ARINFO`](https://redis.io/docs/latest/commands/arinfo/), [`ARINSERT`](https://redis.io/docs/latest/commands/arinsert/), [`ARLASTITEMS`](https://redis.io/docs/latest/commands/arlastitems/), [`ARLEN`](https://redis.io/docs/latest/commands/arlen/), [`ARMGET`](https://redis.io/docs/latest/commands/armget/), [`ARMSET`](https://redis.io/docs/latest/commands/armset/), [`ARNEXT`](https://redis.io/docs/latest/commands/arnext/), [`AROP`](https://redis.io/docs/latest/commands/arop/), [`ARRING`](https://redis.io/docs/latest/commands/arring/), [`ARSCAN`](https://redis.io/docs/latest/commands/arscan/), [`ARSEEK`](https://redis.io/docs/latest/commands/arseek/), [`ARSET`](https://redis.io/docs/latest/commands/arset/)


StefanPalashev:
@nihohit, @somechris Could you please take a look? I'd really appreciate your feedback whenever you have time. Thank you!

somechris:
When I saw th PR, I held back with any remark, as I'm not sure the time is now to add Redis' arrays.

`redis-rs` is somewhat strict with semantic versioning. Changing function signatures means version trouble.

Yet Redis' arrays are a preview thing. The [commands' webpage](https://redis.io/docs/latest/commands/arlen/) carry a prominent yellow banner saying:

```
Array is a new data type that is currently in preview and may be subject to change.
```

(I know you mention it the PRs summary, but still. It should certainly make its way into the commit message and documentation too)

On second thought ... maybe a different feature name would help me ... maybe something like "array-preview", so the name itself makes it clearer that this is not production grade. Along with documentation of the feature that this is subject to change?

--

That said, I think the PR is a bit on the lengthy side (with 18 completely nw commands added in one go, it's hard to keep all changes in mind when reviewing things). Having it broken down would certainly help me. Like 1 PR introducing the feature and basic commands like ARGET,ARSET,ARDEL,ARLEN,ARCOUNT. That's where the main bike shedding about design would happen. Changing things should not be too expensive here.

Once that's settled, maybe 2nd PR adding more commands. and a 3rd adding the last ones?

--

But it's of course for @nihohit to decide :+1:

redis/src/commands/mod.rs
```diff
@@ -1582,6 +1588,235 @@ implement_commands! {
...
+    }
+
+    /// Return metadata about an array as a field-value map.
+    ///
+    /// With `full` set, additional per-slice statistics are included (at O(N) cost).
+    /// Querying a key that does not exist is an error, not an empty map.
+    /// [Redis Docs](https://redis.io/commands/ARINFO)
+    #[cfg(feature = "redis-arrays")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "redis-arrays")))]
+    fn arinfo<K: ToSingleRedisArg>(key: K, full: bool) -> (std::collections::HashMap<String, crate::Value>) {
```
somechris, line 1816: When glancing over the code, `full` felt like it should be in an `Option`.

The [documentation](https://redis.io/docs/latest/commands/arinfo/) also lists `full` as option.

So I'd probably switch that to an option :shrugs:

redis/src/commands/mod.rs
```diff
@@ -1582,6 +1588,235 @@ implement_commands! {
...
+    /// [Redis Docs](https://redis.io/commands/ARGREP)
+    #[cfg(feature = "redis-arrays")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "redis-arrays")))]
+    fn argrep<K: ToSingleRedisArg>(key: K, start: redis_arrays::ArrayBound, end: redis_arrays::ArrayBound, predicates: &'a [redis_arrays::ArrayPredicate<'a>]) -> (Vec<usize>) {
+        cmd("ARGREP").arg(key).arg(start).arg(end).arg(predicates).take()
+    }
+
+    /// Like `argrep`, but with options controlling the predicate combinator, a result `LIMIT`, `WITHVALUES`, and case-insensitive (`NOCASE`) matching.
+    ///
+    /// With `WITHVALUES` the reply is `(index, value)` pairs, so choose the matching return type (`Vec<(usize, String)>`), otherwise it is a list of indices (`Vec<usize>`).
```
somechris, line 1801: Trying some examples against a live Redis, it seems that this description might hold; `WITHVALUES` yielding an array of pairs:

```
127.0.0.1:6379> ARMSET log 0 "boot: ok" 1 "warn: disk" 2 "ERROR: cpu" 3 "info: ready" 4 "error: net"
(integer) 5
127.0.0.1:6379> ARGREP log - + MATCH "error" NOCASE WITHVALUES
1) 1) (integer) 2
   2) "ERROR: cpu"
2) 1) (integer) 4
   2) "error: net"
```

But the [documentation says](https://redis.io/docs/latest/commands/argrep/):

> When `WITHVALUES` is given, a flat array of alternating index-value pairs: `[idx1, val1, idx2, val2, ...]`

It's hard to see what to check against :-(

somechris (commented):
Meh. I'm stopping here. `ARGETRANGE`'s documentation just says `Array reply` for RESP2 and `Map reply` for RESP3. No details, no nothing. Same for `ARINFO`.

It's somewhat tricky to model against an undocumented moving target.

I'd hold my horses.

StefanPalashev:
@somechris Thank you for taking the time to review and for the feedback.
Based on it and the fact that the feature is indeed in a **preview** state, I recommend postponing the review until the feature is out of **preview**, so we can evaluate the final behavior and API.

@nihohit, should the pull request be marked as a draft?

somechris:
Maybe my remarks came across too negative. I did not mean to kill this off.

I just think:

* This needs to be marked as moving target more clearly.
* We'd need to prepare for a way to land breaking changes around arrays.

  Otherwise, we'll get stuck.

  But the repo's current branching structure does not allow for such things (unless we're cutting a new major version).

  But then the branching structure is not too friendly anyways ... Not sure how sold the project is on the current structure ... or maybe we could find a way that allows for that?

* We'd need to somehow agree against what we review/model.

  If it's internal Redis documentation ... that's not ideal. (Maybe nihohit has access ... I don't.)

  If it's actual Redis behaviour ... we're doing the documentation work for Redis in our free time. I'd much rather spend my OSS time on other parts of `redis-rs`. ... like bringing in Valkey features (e.g.: valkey-io/valkey#3077 :-) )

* The PR should be more review friendly

nihohit:
Hi @StefanPalashev, thanks for this PR! I apologize I haven't gotten around to reviewing this yet.

I agree with @somechris - let's not open PRs for features that haven't been released. There's enough work on features that have been released and still aren't supported with an explicit function. I also very much agree with the point about PR size - it's hard to review 1200 lines. It's even harder to review [4300 lines](https://github.com/redis-rs/redis-rs/pull/2015). I very much appreciate the effort you put in, for the benefit of everyone, but as an OSS maintainer that reviews everything manually, sans tokens, I'd appreciate it if the PRs were smaller, and focused more on maintenance ([especially the features that require special testing](https://github.com/redis-rs/redis-rs/pull/2028#issuecomment-4895495192)) and less on the shiny new preview features that can be used using the existing `cmd` interface.

That being said, I want to be clear that these contributions are very much valuable and appreciated, and I don't want you to feel like they're not wanted. They very much are.

PR #2193, opened 2026-07-03
zinterstore / zunionstore - use options instead of separate commands

nihohit:


nihohit:
@StefanPalashev @somechris any comments on this, or does this look ok?

redis/src/commands/mod.rs
```diff
@@ -3929,6 +3863,78 @@ impl ToRedisArgs for SortedSetAddOptions {
...
+/// }
+/// fn intersect_min(con: &mut redis::Connection) -> RedisResult<usize> {
+///     con.zinterstore("out", &["zset1", "zset2"], SortedSetStoreOptions::default().aggregate(Aggregate::Min))
+/// }
+/// fn intersect_weighted(con: &mut redis::Connection) -> RedisResult<usize> {
+///     con.zinterstore_with_weights("out", &[("zset1", 2), ("zset2", 3)], SortedSetStoreOptions::default().aggregate(Aggregate::Min))
+/// }
+/// ```
+#[derive(Clone, Default)]
+pub struct SortedSetStoreOptions {
```
somechris: We could re-use this `struct` once we add support for `ZUNION` and `ZINTER` (no `STORE` at the end).

So I'd suggest to use a name that would also fit `ZUNION` and `ZINTER` (i.e. drop the `Store` from the name) to allow simple forward migration.

What about something that `SortedSetCombinationOptions`, `SortedSetCombineOptions`, `SortedSetOperationOptions`?
Maybe `SortedSetAggregationOptions` (if we do not include `WEIGTHS`)?
Or since Redis uses `Z` to designate the commands ... maybe `ZCombineOptions` or something the like?

My first choice would be `SortedSetCombineOptions`. It's descriptive and cleary expresses that it's for combining sorted sets (no cryptic `Z`) and allows to include weights.

 `SortedSetCombineOptions` is a bit lengthy ... but I'm having a Java background, so I'm used to long names 😁

redis/src/commands/mod.rs
```diff
@@ -3929,6 +3863,78 @@ impl ToRedisArgs for SortedSetAddOptions {
...
+    aggregate: Option<Aggregate>,
```
somechris, line 3914: `ZUNIONSTORE`, `ZINTERSTORE`, `ZUNION`, and `ZINTER` all also take `WEIGHTS`. And weights are also a crucial option how the sets get combined. So I wondered if we should add the weigths here as well?
But weight's are more tied to the provided keys than an abstract property of the combination. I'm torn. 🤷

redis/src/commands/mod.rs
```diff
@@ -3929,6 +3863,78 @@ impl ToRedisArgs for SortedSetAddOptions {
...
+/// combine the scores of members that appear in more than one input sorted set.
+#[derive(Clone, Copy)]
+#[non_exhaustive]
+pub enum Aggregate {
+    /// Sum the scores of matching members (the Redis default).
+    Sum,
+    /// Use the minimum score of matching members.
+    Min,
+    /// Use the maximum score of matching members.
+    Max,
```
somechris, line 3876: Just double-checking ... `COUNT` is missing as it is left to #2154, right?

somechris:
> @StefanPalashev @somechris any comments on this, or does this look ok?

LGTM.

I think the `SortedSetStoreOptions` `struct` would benefit from a [forward compatible name](https://github.com/redis-rs/redis-rs/pull/2193/changes#r3558661562).

(My other comments are just random thoughts.)

StefanPalashev:
How does this pull request differ from the already merged https://github.com/redis-rs/redis-rs/pull/2192?
I based the work of https://github.com/redis-rs/redis-rs/pull/2154 on it. In my pull request, the struct (`SortedSetOptions`) is aliased and the non-storing commands use it instead.

nihohit:
> How does this pull request differ from the already merged https://github.com/redis-rs/redis-rs/pull/2192?

That one was accidentally merged, and was removed from the branch. I hoped nobody will notice it 🤦 .

redis/src/commands/mod.rs
nihohit: It goes back to the discussion we had for FPHA, but here we can break backwards compatiblity. I considered having just a single options argument  , with 2 constructors - keys, and keys with weights. That design, sadly, required either being annoyingly generic, or unnecessary allocations - and AFAIS that's true for every design that puts the weights in the options, as long as the weights are generic.

somechris: Sounds reasonable. :+1:

Is it worth documenting that decision here to spare others from trying to bring weights to `SortedSetOperationOptions`?

redis/src/commands/mod.rs
somechris: When I click "Resolve comment", GitHub answers with "Failed to resolve thread" ... consider it resolved

PR #2194, opened 2026-07-03
perf(cmd): encode expiry options via itoa instead of format!/to_string

fcostaoliveira:
## What

`SetExpiry` (`SET … EX/PX/EXAT/PXAT`) and `Expiry` (`GETEX …`) serialized their `u64` arguments with `format!("{n}")` / `n.to_string()`, each of which heap-allocates a `String`. This switches both to a stack `itoa::Buffer` — the same primitive already used for every other integer argument in the crate (`itoa_based_to_redis_impl!` in `types.rs`, and `write_command` in `cmd.rs`).

The encoded bytes are identical: both emit the base-10 ASCII digits of the `u64`, with no signs, padding, grouping, or locale involvement. All fields of both enums are `u64`.

## Why

Expiry options are on a hot encode path for expiry-heavy workloads (`SET`-with-TTL, `GETEX`). This removes one heap allocation per expiry argument, matching how the rest of the crate encodes integers.

## Benchmark

Added `encode/set_ex` to `benches/bench_basic.rs` (`SET key val EX 3600` × 1000, packed). This is a pure client-side encode microbenchmark (no server / no network), run locally:

| | time (1000 commands) |
|---|---|
| before (`format!`) | 334 µs |
| after (`itoa`) | **276 µs** (−17%) |

## Correctness

Output is byte-identical, so behavior is unchanged. Verified against a live server:

- `test_getex`, `test_set_options_*` (4), and the expiry tests (`test_*expir*`, `test_mset_ex_*`) — all pass.
- `make lint` clean.

### DIFF
--- a/redis/benches/bench_basic.rs
+++ b/redis/benches/bench_basic.rs
@@ -214,6 +214,22 @@ fn bench_encode_integer(b: &mut Bencher) {
     });
 }
 
+fn bench_encode_set_ex(b: &mut Bencher) {
+    // `SET key val EX <secs>` — exercises the `SetExpiry` option encoder.
+    b.iter(|| {
+        let mut pipe = redis::pipe();
+
+        for _ in 0..1_000 {
+            pipe.cmd("SET")
+                .arg("session:abc123")
+                .arg("some-value")
+                .arg(redis::SetExpiry::EX(3600))
+                .ignore();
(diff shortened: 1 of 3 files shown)


nihohit (approved):
thanks!

PR #2196, opened 2026-07-03
perf: enable TCP_NODELAY by default

### LINKED ISSUE #2195: TCP_NODELAY is off by default: Nagle costs 39-68% multiplexed throughput and ~2x p50
## Summary

`TcpSettings::default()` leaves `nodelay: false`, so every TCP connection the crate creates (sync and async, plaintext and TLS) runs with **Nagle's algorithm enabled**. For strictly sequential request-response traffic this is invisible — there is never unACKed data at write time, so Nagle never delays anything. But on a **multiplexed connection under concurrency** (the primary async usage), a command written while a previous command is still unACKed is held by the kernel until that ACK returns. Writes become ACK-clocked: roughly one wire-write per server turnaround.

Measured impact on a real network: **+39–68% throughput and 27–46% lower p50** just from `set_nodelay(true)`, sustained over long runs, on plaintext and mTLS alike.

For comparison, most major Redis clients ship with TCP_NODELAY on: Jedis (`setTcpNoDelay(true)`), Lettuce (netty `TCP_NODELAY=true`), go-redis (Go enables it by default on TCP), StackExchange.Redis (`NoDelay=true`).

## Environment

- Two identical bare-metal x86 servers (Cascade Lake Xeon), Ubuntu 22.04 — client on one, redis on the other
- Dedicated 1 GbE link, MTU 1500, LAN RTT ≈ 0.13–0.38 ms
- Probe: N tokio tasks sharing **one** multiplexed async connection, each running closed-loop `GET`s of a small value; the only variable changed between runs is `TcpSettings::set_nodelay`
- Loopback control shows **no difference** between modes (instant ACKs never engage Nagle) — which is why localhost testing never surfaces this

## Evidence 1 — concurrency sweep (plaintext, 128 B values, 64k samples/cell, 3 repetitions)

Throughput (ops/s, median of 3 reps):

| tasks | default (Nagle) | nodelay=true | Δ |
|---:|---:|---:|---:|
| 1 | 7,060 | 7,162 | +1% (control: sequential is unaffected) |
| 2 | 7,044 | 9,768 | **+39%** |
| 4 | 11,895 | 19,193 | **+61%** |
| 8 | 20,610 | 33,950 | **+65%** |
| 32 | 68,366 | 114,787 | **+68%** |

The signature: with Nagle, 2 concurrent tasks deliver the same throughput as 1 — writes serialize to one per ACK round-trip. Tail latencies are also consistently worse under the default (e.g. at 32 tasks, p9999 spiked to 4.6 ms in one rep vs ≤930 µs in every nodelay rep).

## Evidence 2 — long run over mTLS (client-cert auth, 64 B values, 4 tasks, 300 continuous seconds per mode)

| | default (Nagle) | nodelay=true | Δ |
|---|---|---|---|
| ops in 300 s | 3,011,664 | 4,729,102 | **+57%** |
| p50 / p99 | 381 / 545 µs | 254 / 398 µs | **−33% / −27%** |
| p999 / p9999 | 946 / 1,129 µs | 722 / 854 µs | −24% / −24% |
| per-second ops, min–max over 300 s | 9,698–10,523 | 15,336–16,169 | **zero overlap** |

The per-second timelines never overlap: the best Nagle second is below the worst nodelay second, for all 300 seconds. The deficit is continuous, not transient.

## Evidence 3 — kernel-level stall counter

Polling `ss -ti` at 20 Hz during the 300 s runs and counting samples where the socket reports `notsent` bytes (data the application has written that the kernel is holding back):

| | default (Nagle) | nodelay=true |
|---|---|---|
| samples with `notsent > 0` | **88.8%** (5,012 / 5,646) | 15.4% (869 / 5,645) |

Under the default, the kernel is holding application-written bytes in ~9 out of 10 samples, for the entire run.

## Evidence 4 — packet capture (tcpdump, all data segments over 300 s)

| | default (Nagle) | nodelay=true |
|---|---|---|
| outbound data segments | 1,575,323 | 3,630,442 |
| mean payload | 98 B (~1.9 commands coalesced/segment) | 64 B |
| segments departing <50 µs after an inbound packet | **99.9986%** (all but 22) | 88.7% |
| independent departures (>50 µs after last inbound) | 22 (0.0014%) | 410,097 (11.3%) |

Closed-loop traffic is naturally reply-triggered, so a high ACK-adjacent fraction is expected in both modes. The discriminator is the absoluteness: under the default, an outbound segment essentially **cannot leave the NIC except immediately after an ACK arrives** — 22 exceptions in 1.58 M segments over 5 minutes. That is Nagle's ACK-clocking captured on the wire. The 11.3% of segments that depart mid-flight with nodelay are exactly the writes Nagle was delaying.

## Why the crate's own write paths never benefit from Nagle

- The sync connection serializes each command/pipeline into one buffer and writes it once.
- The multiplexed sink drains its request channel and flushes only when the channel momentarily empties (`forward()` semantics), so bursts already coalesce at the application layer.

There is no redis-rs write pattern that produces the many-tiny-writes-per-message behavior Nagle exists to fix; the algorithm only ever adds delay here.

## Trade-offs (why this is a default change, not unconditional)

- With nodelay, moderate-concurrency multiplexed use emits **more, smaller packets** (see Evidence 4: 2.3× the segments). Deployments near packets-per-second limits (small cloud instances) or on WAN/metered links may prefer to keep Nagle; `TcpSettings::set_nodelay(false)` remains a one-line opt-out.
- Anyone driving `ConnectionLike` directly with many tiny unbatched writes could also prefer the opt-out.

## Reproduction

The probe used for all measurements: [`redis/examples/nagle_probe.rs`](https://github.com/fcostaoliveira/redis-rs/blob/main/redis/examples/nagle_probe.rs) — plaintext or mTLS, prints per-second throughput and latency percentiles:

```
nagle_probe <redis-url> <tasks> <seconds> <nodelay: true|false> [<ca.pem> <client.crt> <client.key>]
```

## Proposal

Change `TcpSettings::default()` from `nodelay: false` to `nodelay: true`, with a release-note flag on the behavioral change and the opt-out documented. PR to follow.

fcostaoliveira:
## What

One-line behavioral change: `TcpSettings::default()` now sets `nodelay: true` (TCP_NODELAY on, Nagle's algorithm off) for every connection the crate creates, plus documentation on the field explaining the default and the opt-out.

Full evidence and methodology in #2195. Highlights, measured on two identical bare-metal x86 hosts over a dedicated 1 GbE link (LAN RTT ≈ 0.13–0.38 ms):

- **+39–68% throughput** at 2–32 concurrent tasks on one multiplexed async connection (3 repetitions, stable), plaintext and mTLS alike
- **p50 −27…−46%**, all tail percentiles improve (p9999 under the default spiked to 4.6 ms; ≤930 µs with nodelay)
- 300-second mTLS run: per-second throughput distributions **never overlap** — the best Nagle second is below the worst nodelay second
- Kernel evidence: `ss -ti` shows application bytes held back (`notsent > 0`) in **88.8%** of 20 Hz samples under the default vs 15.4% with nodelay
- Wire evidence: **99.9986%** of outbound data segments depart only immediately after an inbound packet under the default (22 free departures in 1.58 M segments over 5 minutes) — Nagle's ACK-clocking on the wire
- Sequential request-response traffic is unaffected (no unACKed data at write time), which is why single-command benchmarks and loopback testing never surface this

## Why the crate never benefits from Nagle

Every write path already batches: the sync connection serializes each command/pipeline into one buffer and writes it once; the multiplexed sink drains its request channel and only flushes when the channel momentarily empties. There is no many-tiny-writes pattern for Nagle to fix — it only ever adds latency here.

Jedis, Lettuce, go-redis, and StackExchange.Redis all ship with TCP_NODELAY on.

## Trade-offs / opt-out

With nodelay the client emits more, smaller packets at moderate concurrency (~2.3× segments in the 300 s capture). Deployments near packets-per-second caps (small cloud instances) or on metered/WAN links can keep Nagle with a one-line opt-out, now documented on `set_nodelay`:

```rust
TcpSettings::default().set_nodelay(false)
```

This is a behavioral (not API) change; suggest a release-note callout.

## Verification

- `make lint` clean; unit tests and connection smoke tests pass (the change is protocol-invisible — only the socket option differs)
- Reproduction probe (per-second timeline, latency percentiles, plaintext or mTLS): [`redis/examples/nagle_probe.rs` on the fork](https://github.com/fcostaoliveira/redis-rs/blob/main/redis/examples/nagle_probe.rs)

Closes #2195

### DIFF
--- a/version2.md
+++ b/version2.md
@@ -12,7 +12,7 @@ redis = "2"
 
 ### TCP_NODELAY is now enabled by default (Breaking Change)
 
-`TcpSettings::default()` now sets `nodelay: true`, disabling Nagle's algorithm on every TCP connection the crate creates (sync and async, plaintext and TLS). Previously Nagle was left enabled, which serialized writes on a multiplexed connection to one per ACK round-trip under concurrency — measured at 39–68% lower throughput and roughly double the p50 latency on a real network (see [#2195](https://github.com/redis-rs/redis-rs/issues/2195) for the full evidence). Sequential request-response traffic is unaffected, and Redis clients in other ecosystems already ship with TCP_NODELAY enabled.
+By default, Nagle's algorithm is now disabled on every TCP connection the crate creates (sync and async, plaintext and TLS). Previously it was left enabled, which serialized writes on a multiplexed connection to one per ACK round-trip under concurrency — measured at 39–68% lower throughput and roughly double the p50 latency on a real network (see [#2195](https://github.com/redis-rs/redis-rs/issues/2195) for the full evidence). Sequential request-response traffic is unaffected, and Redis clients in other ecosystems already ship with TCP_NODELAY enabled.
 
 No API changed, but the wire behavior did: the client now emits more, smaller packets at moderate concurrency. Deployments close to packets-per-second limits (small cloud instances) or on metered/WAN links may prefer the old behavior.
 
--- a/version2.md
+++ b/version2.md
@@ -10,6 +10,21 @@ redis = "2"
 
 ## Breaking Changes
 
+### TCP_NODELAY is now enabled by default (Breaking Change)
+
+`TcpSettings::default()` now sets `nodelay: true`, disabling Nagle's algorithm on every TCP connection the crate creates (sync and async, plaintext and TLS). Previously Nagle was left enabled, which serialized writes on a multiplexed connection to one per ACK round-trip under concurrency — measured at 39–68% lower throughput and roughly double the p50 latency on a real network (see [#2195](https://github.com/redis-rs/redis-rs/issues/2195) for the full evidence). Sequential request-response traffic is unaffected, and Redis clients in other ecosystems already ship with TCP_NODELAY enabled.
+
+No API changed, but the wire behavior did: the client now emits more, smaller packets at moderate concurrency. Deployments close to packets-per-second limits (small cloud instances) or on metered/WAN links may prefer the old behavior.
+
+**Migration:** nothing to do for most users — expect lower latency and higher multiplexed throughput. To keep Nagle's algorithm:
+
+```rust
+use redis::{IntoConnectionInfo, io::tcp::TcpSettings};
+
+let info = "redis://127.0.0.1/".into_connection_info()?
+    .set_tcp_settings(TcpSettings::default().set_nodelay(false));
+```
+
 ### `cmd_iter` yields `CmdRef` instead of `&Cmd` (Breaking Change)
 
 **Most users can upgrade to 2.0.0 with no code changes.** The flattening is an internal representation change; the pipeline builder API (`cmd`, `arg`, `add_command`, `ignore`, `query`, `query_async`, `exec`, …) is unchanged. The only adjustments are needed if you iterate a pipeline's commands or call `with_capacity` directly.
--- a/redis/src/io/tcp.rs
+++ b/redis/src/io/tcp.rs
@@ -35,6 +35,14 @@ impl TcpSettings {
     }
 
     /// Sets the value of the `TCP_NODELAY` option on this socket.
+    ///
+    /// Enabled (`true`) by default, disabling Nagle's algorithm, since Redis
+    /// traffic is strictly request-response: coalescing small writes only adds
+    /// latency (most visibly on multiplexed connections under concurrency,
+    /// where Nagle serializes writes to one per ACK round-trip). Set to `false`
+    /// to re-enable Nagle's algorithm, e.g. on packets-per-second-limited
+    /// cloud instances or metered/WAN links where fewer, larger packets are
+    /// preferable to lower latency.
     pub fn set_nodelay(self, nodelay: bool) -> Self {
         Self { nodelay, ..self }
     }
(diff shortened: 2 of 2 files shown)


nihohit:
This is a breaking change in behavior, and so it can't be accepted in this version. Please retarget to branch 2.0.x, and update the documentation with this breaking change.

fcostaoliveira:
Done — retargeted to `2.0.x` and rebased (clean cherry-pick, no conflicts). Also added a **"TCP_NODELAY is now enabled by default (Breaking Change)"** section to `version2.md`, following the format of the existing entries: what changed, the wire-behavior implications (more, smaller packets at moderate concurrency), and the one-line migration for anyone who wants to keep Nagle (`TcpSettings::default().set_nodelay(false)` — relevant for PPS-limited or metered/WAN deployments).

Agreed this is the right home for it — it's exactly the kind of behavioral change a major version exists for.

version2.md
```diff
@@ -10,6 +10,21 @@ redis = "2"
 
 ## Breaking Changes
 
+### TCP_NODELAY is now enabled by default (Breaking Change)
+
+`TcpSettings::default()` now sets `nodelay: true`, disabling Nagle's algorithm on every TCP connection the crate creates (sync and async, plaintext and TLS). Previously Nagle was left enabled, which serialized writes on a multiplexed connection to one per ACK round-trip under concurrency — measured at 39–68% lower throughput and roughly double the p50 latency on a real network (see [#2195](https://github.com/redis-rs/redis-rs/issues/2195) for the full evidence). Sequential request-response traffic is unaffected, and Redis clients in other ecosystems already ship with TCP_NODELAY enabled.
```
nihohit: By default, Nagle's algorithm is disabled on every [...]

no need to name the struct that most users are unaware of.

fcostaoliveira: Done — reworded to lead with the behavior ("By default, Nagle's algorithm is now disabled…"); the struct now only appears in the migration snippet where it's actually needed.

nihohit (approved):
Thanks!

PR #2197, opened 2026-07-07
Versioning improvements

StefanPalashev:
# Versioning improvements

## Summary

This pull request replaces the `matcher_array_impls!` `macro_rules!` with **const-generic `From` impls**.
This makes array matchers of any length work (previously only lengths 1, 2, and 3 were implemented).
Which solves the following problem:
Coercing `From<[Component; N]>` / `From<[&[Component]; N]>` impls were generated by a macro invoked for `N = 1, 2, 3`.
A 4-element matcher would have failed to compile until `matcher_array_impls!(4);` was added and so on.

### `redis/tests/support/version.rs`

- **Const generics replace the array macro:** 
`matcher_array_impls!` and its three invocations are gone; `From<[Component; N]>` and `From<[&[Component]; N]>` are now single const-generic impls that delegate to the slice `From` via `as_slice()`.
Behavior is identical for existing call sites and arrays of any length now work.

## Compatibility

There are **no changes to the `redis` client library or its public API** - the work is confined to the test support module (`redis/tests/support/version.rs`).

### DIFF
--- a/redis/tests/support/version.rs
+++ b/redis/tests/support/version.rs
@@ -82,35 +82,22 @@ impl<'a> From<&[&[Component<'a>]]> for ComponentMatcher<'a> {
     }
 }
 
-/// Macros to provide array implementations for matchers' slice implementations
+/// Coercing array implementations for matchers' slice implementations
 ///
-/// Rust can auto-coerce array to slices. But with generic arguments, this
+/// Rust can auto-coerce arrays to slices. But with generic arguments, this
 /// array-to-slice-auto-coercion does not kick in. So one would have to convert manually. To avoid
-/// this for the common cases, this macro implements coercing `From`s. for a given array length
-///
-/// # Arguments
-///
-/// * `$n` - The array lengths to implement coercing `From`s for.
-macro_rules! matcher_array_impls {
-    ($n:expr) => {
-        impl<'a> From<[Component<'a>; $n]> for ComponentMatcher<'a> {
-            fn from(value: [Component<'a>; $n]) -> Self {
-                let coerced_value: &[Component<'a>] = &value;
-                Self::from(coerced_value)
-            }
-        }
-
-        impl<'a> From<[&[Component<'a>]; $n]> for ComponentMatcher<'a> {
-            fn from(value: [&[Component<'a>]; $n]) -> Self {
-                let coerced_value: &[&[Component<'a>]] = &value;
-                Self::from(coerced_value)
-            }
-        }
-    };
+/// this, these const-generic `From`s coerce arrays of any length to the corresponding slice matcher.
+impl<'a, const N: usize> From<[Component<'a>; N]> for ComponentMatcher<'a> {
+    fn from(value: [Component<'a>; N]) -> Self {
+        Self::from(value.as_slice())
+    }
 }
-matcher_array_impls!(1);
-matcher_array_impls!(2);
-matcher_array_impls!(3);
+
+impl<'a, const N: usize> From<[&[Component<'a>]; N]> for ComponentMatcher<'a> {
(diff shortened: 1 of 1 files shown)


redis/tests/support/version.rs
```diff
@@ -82,35 +84,75 @@ impl<'a> From<&[&[Component<'a>]]> for ComponentMatcher<'a> {
...
+    /// The disjunctive (`OR`) group of components that satisfy this requirement.
+    ///
+    /// A module may report under more than one name (see [`Module::info_names`]), so any of them at the required version is a match.
+    /// A `min_version` of `None` becomes `(0, 0, 0)`, which any loaded version satisfies.
+    fn components(&self) -> Vec<Component<'static>> {
+        let version = self.min_version.unwrap_or((0, 0, 0));
+        self.module
+            .info_names()
+            .iter()
+            .map(|&name| (name, version))
```
somechris: This line assumes that modules from different vendors (Valkey, Redis) do the same thing in the same version. But that assumption does not hold. Module versioning of the same module between vendors is typically vastly different.

Let's have an example. Let's say Redis' Bloom filter module adds a new data structure in v8.10.0. And Valkey's Bloom filter module adds the same data structure in v1.5.0.

In this setting, you need to load the `bloom.so` provided by the environment. But you can run the test only if the loaded `bloom.so`'s version is high enough.

```
run_test_if_version_supported!([REDIS_CE_8_0, VALKEY_9_0], &[(Module::Bloom, (8,10,0))])
```
^ does not work as it would mistakenly not run on Valkey's Bloom filter v1.5.0

```
run_test_if_version_supported!([REDIS_CE_8_0, VALKEY_9_0], &[(Module::Bloom, (1,5,0))])
```
^ does not work as it would mistakenly run on Redis's Bloom filter v8.8.0

So we have to resort to

```
run_test_if_version_supported!([[REDIS_CE_8_0, VALKEY_9_0], [REDIS_BLOOM_8_10, VALKEY_BLOOM_1_5]], &[(Module::Bloom, (0,0,0))])
```

This was just the thing we tried to get away from.

Sure one can address this in many ways, but it only gets more and more complicated. So I'd live with the duplication for now and keep the working, simple status quo.

StefanPalashev: This was anticipated and written in the behavior notes. The code got a bit more complex than I expected, but I decided to give it a go anyway. Maybe we can keep only the 1st commit?

somechris: > Maybe we can keep only the 1st commit?

Yup. The 1st commit LGTM :+1:

nihohit (commented):
I need to go over the 2nd commit again, but the 1st commit looks good.

redis/tests/support/version.rs
nihohit: I agree that the 2nd commit is problematic, but maybe we can add to the module constraints both a version and a (optional?) server-type argument, which states whether it's supported by Redis or Valkey?

somechris: Maybe I'm misunderstanding things. But to me, the goal would be to have shorthands and to make things simpler.

In CI, we run Valkey 7.2 with Redis's Json module. Same for Valkey 8.0.

Maybe in the long run, we can phase such things out.

Maybe in the long run, such cross-vendor use will even stop working.

But for now, coupling the server vendor with the module vendor looks it might give rise to more confusion that it'd benefit us.

These free-form version conditions are still fresh. So I wondered ... WWould it be better to revisit creating short-hands after some time? After we saw more uses of these conditions, once we can better assess what short-hands would benefit most?

StefanPalashev:
Removed commit 2

nihohit:
Thanks!

PR #2199, opened 2026-07-08
Zero-copy response parsing: Value backed by Bytes/Str

fcostaoliveira:
# Zero-copy response parsing: `Value` backed by `Bytes`/`Str`

## Summary

Reworks `Value` to store its textual and binary payloads in cheaply-cloneable, reference-counted buffers (`bytes::Bytes` and a new UTF-8 `Str` type) instead of owned `Vec<u8>`/`String`, and rewrites the RESP parser to be **zero-copy**. Parsing a response no longer allocates once per element — leaves are produced as refcount-bumped slices into the response buffer.

Result: **1.4×–4.1× faster parsing and 10×–470× fewer heap allocations**, with the biggest wins on multi-element responses (large arrays/maps). Cloning a `Value` is now a refcount bump rather than a deep copy.

## Motivation

The old `Value` owned its data (`BulkString(Vec<u8>)`, `SimpleString(String)`, …), so the parser allocated and memcpy'd a fresh buffer for **every leaf** of a response. For replies with many elements (`MGET`, `HGETALL`, `SCAN`, stream reads), that's thousands of small allocations per response, and every `Value::clone()` is a deep copy. Backing the payloads with `Bytes` lets the parser slice the read buffer instead of copying it, and makes clones O(1).

### `Value` payloads → `Bytes` / `Str`
- `BulkString(Vec<u8>)` → `BulkString(bytes::Bytes)`
- `SimpleString(String)` → `SimpleString(Str)`
- `VerbatimString { text: String, .. }` → `{ text: Str, .. }`
- `BigNumber(Vec<u8>)` → `BigNumber(bytes::Bytes)` (the `num-bigint` variant is unchanged)
- `PushKind::Other(String)` / `VerbatimFormat::Unknown(String)` → `Str`
- `ServerError` code/detail: `ArcStr` → `Str`

### New `Str` type
A UTF-8-guaranteed string backed by `bytes::Bytes`:
- `Deref<Target = str>` (+ `AsRef<str>`/`AsRef<[u8]>`/`Borrow<str>`), `Display`, `Debug`, `PartialEq`/`Eq`/`Hash`
- `From<&str>`/`From<String>`, `Into<String>`/`Into<Bytes>`, `const fn from_static`
- so most code that treated the inner value as `&str` keeps compiling unchanged.

### Zero-copy parser rewrite (`parser.rs`)
- The `combine`-based parser now produces an intermediate `ValueRange` whose leaves hold `Range<usize>` offsets into the buffer rather than copied data.
- After a complete value is parsed, the consumed prefix is frozen into a single `Bytes` and `materialize()` turns each range into a cheap `Bytes::slice`. The async `ValueCodec` path performs **no copy at all**; the borrowed-`&[u8]` entry points (`parse_redis_value`) copy the input once and then slice it.
- All entry points (`ValueCodec`, `Parser::parse_value`, `parse_redis_value`, `parse_redis_value_async`) were unified onto this path. A reply spanning several reads resumes where it stopped rather than re-parsing from the start, via a `ParseState` that `ValueCodec` owns and the two read loops keep call-locally.

### Dependency / features
- `bytes` is now a mandatory dependency (it was optional behind `aio`), and the `bytes` feature is gone along with all nine `#[cfg(feature = "bytes")]` gates, so those impls are unconditional. Note this was Cargo's implicit optional-dependency feature rather than a declared one, so `features = ["bytes"]` now errors; documented in `version2.md`.
- `aio` now enables `tokio/io-util` (the async reader uses `AsyncReadExt::read_buf`).

### Consumer migration
Updated `FromRedisValue` impls, accessors, `Debug`, cluster handling, commands (`acl`, `geo`, `streams`, `hotkeys`), pubsub, `redis-test`, and the test suite for the new payload types. Cluster topology/`NodeAddress` deliberately **stay `ArcStr`** — they're long-lived and deduplicated, so a compact interned string is better than pinning the response buffer.

## Benchmarks

New bench: `cargo bench -p redis --bench bench_decode` (no server required). Allocation counts come from a counting global allocator and are deterministic, so they reproduce anywhere; wall-clock is machine-dependent and is in the second table below. Both compare against the pre-refactor baseline (`upstream/2.0.x`).

| Response | Allocations (before → after) |
|---|---|
| Single 1 MiB bulk string | 154 → **2** (77×) |
| Array of 5000 small bulks | 7509 → **16** (469×) |
| Array of 500 × 1 KiB bulks | 2022 → **11** (184×) |
| Array of 5000 simple strings | 7152 → **16** (447×) |
| Array of 1000 key/value pairs | 2933 → **13** (226×) |

The old parser allocated ~one heap block per leaf; the new one allocates a small constant regardless of element count. (The last row is a flat `*2000` array, not a RESP3 map — it was mislabelled "Map" in an earlier revision.)

## Breaking changes & migration

Most code going through `FromRedisValue`/`from_redis_value` is unaffected. Code that matches on `Value` directly needs minor updates:

```rust
// Before:
if let Value::BulkString(bytes) = v {
    let s = String::from_utf8(bytes)?;        // bytes: Vec<u8>
}
// After:
if let Value::BulkString(bytes) = v {
    let s = String::from_utf8(bytes.into())?; // bytes: Bytes  (or use &bytes as &[u8])
}
```

`Str` derefs to `&str`, so reads keep working; construction takes a `Str` (e.g. `Value::SimpleString("OK".into())`).

Other notes:
- `parse_redis_value_async` now takes `&mut BytesMut` (the reusable read buffer) instead of a `combine::stream::Decoder`.
- Verbatim strings / blob errors are now validated as UTF-8 (previously lossy) — they are UTF-8 in practice.
- `Value::BulkString` `Debug` output for binary data uses `Bytes`'s formatting.

## Provenance

The core refactor (all three base commits) is @nihohit's work from `shachar/use-bytes`, opened as a PR per [his invitation on #2184](https://github.com/redis-rs/redis-rs/pull/2184#issuecomment-4914847816). On top of it: an independent review, a cross-version differential test campaign, independent benchmark confirmation, and three follow-up commits (below).

## Review summary

Focused on the last base commit (the parser rewrite) and on socket/async-I/O behavioral equivalence, per the open question on #2184:

- **Offset soundness:** the parser records `Range<usize>` offsets via pointer arithmetic against the buffer base. They are *indices from the buffer's logical start*, which is what makes them survive the reallocation a resumed parse implies: growing a `BytesMut` preserves indices, and while a parse is in flight the buffer is only ever appended to. Bytes are dropped from the front exactly once, by `split_to(consumed).freeze()`, after the value is complete and the state has been reset — so every recorded range lies inside the frozen frame.
- **UTF-8 soundness:** `ValueRange`'s string variants carry a `StrRange`, whose only constructor takes a `&str`, so the single `Str::from_utf8_unchecked` call site cannot be reached without a UTF-8 check. It lives in its own module so the field is unreachable from the rest of `parser.rs`. The offsets themselves are pointer arithmetic that no type can check, so `slice_str` re-validates under `debug_assert!` — worth noting because **Miri does not check `from_utf8_unchecked` preconditions** (confirmed with a deliberately-bad canary that Miri passed), so a clean Miri run is not evidence here.
- **`unsafe` surface (audited twice, independently):** every `Str::from_utf8_unchecked` range originates from `str::from_utf8`-validated input (`line()`, blob/verbatim validation); raw bulk bytes only ever become `Bytes`, never `Str`. No validate-A-slice-B mismatches found.
- **EOF semantics preserved:** `decode` (partial) vs `decode_eof` (complete) reproduce the old boundary/mid-message distinction; mid-message EOF errors, boundary EOF yields `None`.
- **Socket-facing behavior:** the `Framed`-based async path manages its read buffer exactly as before; the multiplexed sink/flush logic is untouched, and the deadlock-regression tests still cover it. `read_buf` on a `BytesMut` can never see zero capacity (the `BufMut` impl grows on demand), so `Ok(0)` still means EOF.
- **`FromRedisValue for Bytes` now shares instead of copying.** It returns `bytes.clone()` (a refcount bump) where 1.x did `copy_from_slice`. Cheaper, but it keeps the *whole* reply buffer alive: pulling a 10-byte bulk string out of a 1 MiB reply and holding it now pins 1 MiB. Copy out (`Vec::from(&b[..])`) if you retain a small field from a large reply.
- **Known trade-offs** (documented in `version2.md`): leaf slices pin the buffer they arrived in, and replies delivered in one read share that buffer; the `Vec<u8>`/`String` extraction copy moved from parse time to conversion time (net copies unchanged); peak memory during a large reply is lower overall (one copy resident instead of two) but the read buffer must now be contiguous and as large as the reply.

### Performance: not a uniform win

The gains are on large, multi-leaf replies; small replies pay a fixed per-reply cost and get slightly slower. Measured through `ValueCodec` with the codec reused across replies, which is the shape `MultiplexedConnection` actually uses:

| through `ValueCodec` | vs 2.0.x |
| --- | --- |
| 1 MiB bulk string | **241x faster** (the copy disappears entirely) |
| array of 500 x 1 KiB bulks | 1.5x faster |
| pipelines of small bulks | 1 allocation per reply -> **0** |
| `+OK`, `:1` | **~19-25% slower** |
| nested arrays, depth 10-90 | **~37-51% slower** |

Cause: per-leaf `malloc`/`free` is replaced by atomic refcount traffic on the shared frame, which loses on tiny replies; and the intermediate `ValueRange` tree costs one allocation per *container*, so deep nesting allocates roughly 2x. Against a loopback round trip the scalar cases are ~0.6% of wall time; the cost is real only for CPU-bound pipelined workloads.

Worth flagging that the original `bench_decode` could not see any of this: it only exercised `parse_redis_value`, which builds a fresh parser per call (~3 allocations of setup a live connection never pays). That flattered the branch and **inverted the sign** on deep nesting — 2.1x *faster* through `parse_redis_value`, 1.5x *slower* through the codec. Small-reply, pipeline and codec-level cases have been added.

## Follow-up commits on top of the base

1. **`parser: detach error strings`** — `materialize_error` copies error code/detail out of the response frame instead of slicing it. Errors are rare, small, and long-lived (stored by users, carried by retry loops and cluster redirects); slicing would pin the entire reply buffer for as long as the error lives — e.g. one error among thousands of pipelined results would retain the whole reply. The hot data path stays zero-copy.
2. **`parser: minimum async read granularity`** — `parse_redis_value_async` reserves spare capacity before `read_buf`; after `split_to`/`freeze` the spare capacity can be arbitrarily small, which would degrade reads to tiny chunks.
3. **`test: pipelined values across fragmented reads`** — multiple replies flowing through one reused buffer in 7-byte chunks: leftover preservation after each frozen value, mid-value refills. The read-path shape a multiplexed connection produces on a real socket.

## Testing

- **Cross-version differential replay:** 1,896 coverage-guided inputs (an AFL corpus evolved against the parser grammar) replayed through the old parser (base `2.0.x`) and this branch, comparing canonically-normalized results. **Zero unintended divergences.** The only 4 diffs are the documented lossy→strict UTF-8 validation change for verbatim strings/blob errors (all four inputs contain bytes the old parser lossy-replaced).
- **Full test suite** (`make test`: nextest across connection types and RESP2/RESP3) — green on the rebased branch.
- `partial_io_parse` quickcheck (arbitrary values through a would-block/fragmented async reader) validates offset correctness across fragmented reads; the new pipelined-fragmentation test extends this to multiple values per buffer.
- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean.

## Benchmarks: wall-clock, second machine

The allocation table above is machine-independent. These are wall-clock timings, run separately on an idle x86 server (Cascade Lake), old vs new back-to-back — a different machine from the one used elsewhere in this description, which is why the absolute µs differ:

| Payload | old | new | speedup |
|---|---:|---:|---:|
| single 1 MiB bulk | 272.8 µs | 169.4 µs | **1.61×** |
| 5000 small bulks | 1,318 µs | 972 µs | **1.36×** |
| 500 × 1 KiB bulks | 314.9 µs | 115.2 µs | **2.73×** |
| 5000 simple strings | 1,070 µs | 770 µs | **1.39×** |
| 1000-pair map | 516 µs | 378 µs | **1.37×** |

The shapes with many elements (large arrays/maps) are exactly the profile of multi-key read commands (`MGET`, `SMEMBERS`, `HGETALL`, `ZRANGE`), where per-element allocation used to dominate client CPU.


fcostaoliveira:
Hi @nihohit 👋 — update after a deep review pass focused on exactly what you flagged: **low-level socket / async-I/O behavioral equivalence** vs the pre-refactor parser.

**The one real behavioral change we found there, now fixed:** a reply truncated by a dropped connection (server closes mid-response) was surfacing as `ErrorKind::Parse` instead of `UnexpectedEof`, so `is_connection_dropped()` / `is_io_error()` returned `false`. Cause: the EOF branch in the sync + async `parse_*` entry points did `parse_buffer(buf, true)? { Some => ok, None => UnexpectedEof }`, but under `complete=true` combine turns a half-read value into a parse error, so `?` short-circuited and the `UnexpectedEof` arm was dead. Pre-2.0 those paths returned `UnexpectedEof`. Fixed by classifying an end-of-input parse failure as a truncated reply; **the codec/multiplexed path is untouched** (it already returned the parse error pre-2.0, so it's equivalent). Pinned with regression tests on the sync bulk, truncated-aggregate, and async paths — and I separately probed a batch of complete-but-malformed frames to confirm the classifier never mistakes corruption for a dropped connection.

**On "what testing to add":** the branch now carries, on top of your work — partial-I/O / fragmented-read reassembly, the truncated-EOF connection-drop tests above, error-frame detachment (errors are copied out so they don't pin the reply buffer), and strict-UTF-8 negatives — plus the old-vs-new differential harness (quickcheck + AFL).

A few other correctness/ergonomics fixes while here: re-enabling the `bytes` feature via `aio` (it had stopped being activated, silently dropping `FromRedisValue for Bytes` + `bufmut_for_next_arg` for async users); `Str` upholding the `Borrow`/`Hash` contract, gaining `Ord`, and `FromRedisValue`/`ToRedisArgs`; re-exporting `bytes`; and a couple of `version2.md` precision fixes. All independent commits — squash/drop whatever doesn't fit "without a lot of complex code."

Green on `make lint` / `make test`. Happy to reshape any of it.

nihohit:
@fcostaoliveira I haven't forgotten this, I hope I'll get to reviewing this soon.

fcostaoliveira:
> @fcostaoliveira I haven't forgotten this, I hope I'll get to reviewing this soon.

thank you @nihohit !

redis/examples/streams.rs
```diff
@@ -222,7 +222,11 @@ fn read_records(client: &redis::Client) -> RedisResult<()> {
             println!("\tID {id}");
             for (n, s) in map {
                 if let Value::BulkString(bytes) = s {
-                    println!("\t\t{}: {}", n, String::from_utf8(bytes).expect("utf8"))
+                    println!(
+                        "\t\t{}: {}",
```
nihohit: nit: \t\t{n}

redis/src/lib.rs
```diff
@@ -717,6 +718,11 @@ pub use crate::types::{
 
 pub use crate::types::{calculate_value_digest, is_valid_16_bytes_hex_digest};
 
+// `Value` exposes `bytes::Bytes` in its public API, so re-export the crate to
+// spare downstreams from having to add (and version-match) `bytes` themselves.
+pub use bytes;
+pub use bytes::Bytes;
```
nihohit: remove these, we don't add unnecessary dependencies

redis/Cargo.toml
```diff
@@ -162,13 +165,18 @@ entra-id = [
 
 # Instead of specifying "aio", use either "tokio-comp" or "smol-comp".
 aio = [
+  # `bytes` is a mandatory dependency now, but the `bytes` feature still gates
```
nihohit: instead of this doc, just remove the feature gate

redis/Cargo.toml
```diff
@@ -124,6 +124,9 @@ default = ["acl", "streams", "geospatial", "script", "num-bigint"]
 acl = []
 bloom = []
 geospatial = []
+# `bytes` is now a mandatory dependency; this no-op feature is kept so existing
```
nihohit: remove doc

redis/Cargo.toml
```diff
@@ -124,6 +124,9 @@ default = ["acl", "streams", "geospatial", "script", "num-bigint"]
...
+# `#[cfg(feature = "bytes")]`-gated impls and downstream `features = ["bytes"]` keep working.
+bytes = []
```
nihohit: no new feature

redis/Cargo.toml
```diff
@@ -162,13 +165,18 @@ entra-id = [
...
+  # transitively before `bytes` became mandatory; keep doing so, otherwise those
+  # impls silently disappear for every async downstream.
   "bytes",
   "dep:pin-project-lite",
   "dep:futures-util",
   "dep:tokio",
   "dep:tokio-util",
   "dep:async-lock",
   "tokio-util/codec",
+  "tokio/io-util",
```
nihohit, line 171: why is this needed?

redis/Cargo.toml
```diff
@@ -36,8 +36,8 @@ sha1_smol = { version = "1.0", optional = true }
 
 combine = { version = "4.6", default-features = false, features = ["std"] }
 
-# Only needed for AIO
-bytes = { version = "1", optional = true }
+# `Value` stores its textual/binary payloads as `Bytes`, so this is a core dependency.
```
nihohit: remove doc

redis/src/types.rs
```diff
@@ -138,6 +139,179 @@ pub enum NumericBehavior {
     NumberIsFloat,
 }
 
+/// A cheaply-cloneable, UTF-8 string backed by [`bytes::Bytes`].
```
nihohit: nit: add a `types` folder, and move these ~100 lines to types/str.rs, and also the unit tests

redis/src/parser.rs
```diff
@@ -5,77 +5,257 @@
...
+// ------------------------------------------------------------------
+// Materialization: `ValueRange` + the response buffer -> `Value`.
+// ------------------------------------------------------------------
+
+/// Slices `frame` at `range`, wrapping it as a [`Str`] without re-validating.
+///
+/// All ranges that reach this helper were validated as UTF-8 during parsing.
+#[inline]
+fn slice_str(frame: &Bytes, range: Range<usize>) -> Str {
+    // SAFETY: the parser only produces `Str` ranges from input that has already
```
nihohit: where is this verified? is there a unit test that checks this? the inputs to this function don't provide any assurances - it would've been better if this wasn't a separate function, but called from the ValueRange match, so it would be evident that this is only called on checked variants.

redis/src/parser.rs
```diff
@@ -329,64 +497,88 @@ where
...
-                } else {
-                    let err = err
-                        .map_range(|range| format!("{range:?}"))
-                        .map_position(|pos| pos.translate_position($decoder.buffer()))
-                        .to_string();
-                    RedisError::from(ParsingError::from(err))
-                }
+/// Attempts to parse a single value from `buffer`.
+///
+/// Parsing always starts from the beginning of `buffer` with fresh state, so
```
nihohit: why do we need to re-parse partially parsed data?

redis/src/parser.rs
```diff
@@ -453,21 +660,29 @@ impl Parser {
...
-            combine::stream::easy::Stream::from(input)
-        });
-        match result {
-            Err(err) => Err(to_redis_err!(err, decoder)),
-            Ok(result) => Ok(result),
+        loop {
+            if let Some(value) = parse_buffer(&mut self.buffer, false)? {
+                return Ok(value);
+            }
+            let mut chunk = [0u8; 8 * 1024];
```
nihohit: why copy the data? if the self.buffer is large enough, shouldn't we read into the buffer?

redis/src/parser.rs
```diff
@@ -622,12 +842,95 @@ mod tests {
...
+    }
+
+    #[test]
+    fn invalid_utf8_in_verbatim_and_blob_error_is_rejected() {
+        // Both verbatim strings and blob errors are now decoded with strict
+        // `str::from_utf8` (previously `from_utf8_lossy`), so a non-UTF-8 body
+        // must surface as a parse error rather than silently gaining U+FFFD
+        // replacement characters.
+        assert_matches!(parse_redis_value(b"=5\r\ntxt:\xff\r\n"), Err(_));
+        assert_matches!(parse_redis_value(b"!4\r\nER\xff\xff\r\n"), Err(_));
```
nihohit: need to verify that the correct error is returned

nihohit (commented):
first pass

nihohit:
@fcostaoliveira are you still interested in pushing this?

fcostaoliveira:
Yes, still very much interested — thanks for the ping. I'll address all the review comments soon.

redis/examples/streams.rs
fcostaoliveira: Done — bound the decoded string to a local so the args inline: `println!("\t\t{n}: {value}")`. Swept the rest of the PR diff; this was the only inlinable site (the one in `redis-test` takes a method call).

redis/src/lib.rs
fcostaoliveira: Done — both `pub use bytes;` and `pub use bytes::Bytes;` are gone. Nothing in the repo used `redis::Bytes`, so nothing broke.

redis/Cargo.toml
fcostaoliveira: Done — removed the feature gate rather than documenting it: `"bytes"` is out of the `aio` list, and the nine `#[cfg(feature = "bytes")]` gates are gone so those impls are now unconditional.

redis/Cargo.toml
fcostaoliveira: Done — doc comment removed.

redis/Cargo.toml
fcostaoliveira: Done — `bytes = []` deleted. One thing worth flagging: it wasn't a feature this PR added, it was Cargo's implicit optional-dependency feature (base had `bytes = { optional = true }`), which making the dep mandatory destroyed. So deleting it is a real break for anyone passing `features = ["bytes"]` — documented in `version2.md` rather than kept as a no-op alias. Say the word if you'd rather have the alias for one release.

redis/Cargo.toml
fcostaoliveira: It's load-bearing: `parse_redis_value_async` reads via `AsyncReadExt::read_buf`, and both the trait and the method are gated behind tokio's `io-util`. Nothing else in `aio` enables it — `tokio-util/codec` is `codec = []`, and `combine/tokio` takes tokio with `default-features = false`. Removing the line gives `error[E0432]: unresolved import tokio::io::AsyncReadExt` plus `E0599` on `read_buf` from `cargo check -p redis --no-default-features --features tokio-comp`. Added a comment recording one trap: `cargo tree` shows `io-util` enabled either way, because our dev-dependency on tokio turns it on — so tests and benches build fine and only a library build breaks.

redis/Cargo.toml
fcostaoliveira: Done — doc comment removed; `bytes` is just `bytes = { version = "1" }` now.

redis/src/types.rs
fcostaoliveira: Done — `types.rs` → `types/mod.rs` (matching the `mod.rs` convention the other nine directory modules use) plus a new `types/str.rs` holding `Str` and its unit tests. `crate::types::Str` and the `lib.rs` re-export are unchanged. I also added 11 more unit tests while they were being moved — including one that catches the hand-written `Hash` impl being replaced by a derive, which the existing tests didn't cover.

redis/src/parser.rs
fcostaoliveira: You were right that nothing verified it. It *was* sound — all four callers get their range from a `&str`, since `+` and `-` go through `line()`'s strict `str::from_utf8`, `=`/`!` do their own, and the `split_once(':')`/`splitn(2, ' ')` splits are on ASCII — but that was a prose claim, and worse: **Miri does not check `from_utf8_unchecked` preconditions**. I confirmed that with a deliberately-bad canary, which Miri happily passed, so there was no safety net at all. Rather than inline the `unsafe` into each arm (which would turn one `unsafe` into four, since `detached_str` needs it too) I made the type carry the proof: the string variants now hold a `StrRange`, whose only constructor takes `&str`, so a new `Str`-valued reply kind can't reach the `unsafe` without a UTF-8 check. The offsets are still pointer arithmetic, which no type can check, so `slice_str` also carries a `debug_assert!` that re-validates (free in release, live in tests and under `afl/parser`) — verified non-vacuous: a one-byte off-by-one in `StrRange::of` fails 8 tests. Five new tests cover non-UTF-8 rejection at every construction site, multi-byte round-trips, a 256-value byte sweep, and a multi-byte-at-every-offset sweep to pin `range_of`. If you'd rather have zero `unsafe` here, dropping it costs +5% on ASCII and +40% on non-ASCII text — happy to take that instead, it's a 5-line change.

redis/src/parser.rs
fcostaoliveira: We don't — that was a real regression, thanks for catching it. The reason it happened: `combine`'s resume protocol is "caller removes the bytes I committed and hands me only what follows", which is what the old `decode_stream` did with `bytes.advance(removed_len)`. But its partial state also holds already-produced output (`CountMinMax::PartialState` is `(usize, Vec<ValueRange>, ..)`), and those `ValueRange`s now hold offsets into the reply buffer that the finished value slices — so `advance` would both invalidate the offsets and throw away bytes the value needs. I resolved that the wrong way, by dropping the state, which made every `decode` re-parse from byte 0. Fixed by keeping both: don't advance, and hand `combine` `&buffer[committed..]`. It sees a byte-identical token stream, the consumed bytes stay alive to be frozen and sliced, and offsets stay valid because they're indices from the buffer's logical start (which `BytesMut` growth preserves). `ValueCodec` carries the state again; the two read loops keep a call-local one, so no public API change. On a 1 MiB array of 5000 bulk strings: 1 KiB reads 718ms → 2.67ms, 4 KiB 181ms → 2.17ms, 8 KiB 88.5ms → 2.01ms, i.e. back to parity with 2.0.x. Added `chunked_decode_resumes_instead_of_rescanning`, which poisons the consumed prefix after each `decode` so a re-parse fails loudly, plus a `parse_chunked_1mb_array` bench — the existing `bench_decode` only parsed whole buffers, so this was invisible to it.

redis/src/parser.rs
fcostaoliveira: You're right the loop wastes work, but measuring it put the cost somewhere else. `std::io::Read::read` takes `&mut [u8]`, so reading into `self.buffer` means `resize`-ing the tail with zeros first — trading a memcpy of *bytes received* for a memset of *bytes requested*, which loses whenever the socket short-reads (a 1400-byte read still costs an 8 KiB memset). The async path only gets away with it because `read_buf` may legally touch uninitialized memory; the sync equivalent (`Read::read_buf` + `BorrowedBuf`) is still unstable (rust-lang/rust#117693), so it's out at MSRV 1.88. What the measurement did find is that `let mut chunk = [0u8; 8 * 1024]` zero-fills 8 KiB on *every* call (~356ns, not elided by LLVM), which dominates small replies. Hoisting the scratch chunk into `Parser` removes that and beats reading-into-buffer on every case I measured: non-pipelined `+OK` 760ns → 410ns (-45%), 64 KiB bulk at 1400-byte reads 41.0µs → 24.8µs (-39%), 1 MiB at 1400-byte reads 627µs → 347µs (-45%), 1 MiB at 8 KiB reads 213µs → 163µs (-24%). Reading into the buffer instead measures -3%/-39%→-11%/-24% on those same four. Allocation counts are unchanged; the cost is one 8 KiB buffer per `Parser`, i.e. per sync connection. I also pulled the `8 * 1024` into a shared `READ_CHUNK_SIZE` used by both loops — they still differ in that async can fill more than 8 KiB at once, for the initialized-memory reason above. Happy to switch to the `resize` version if you'd rather read straight into the buffer; just wanted the numbers on the table first.

redis/src/parser.rs
fcostaoliveira: Done — both now assert `ErrorKind::Parse`, `!is_connection_dropped()` and that the message contains `invalid utf-8 sequence`. The `is_connection_dropped()` check is the one that matters: bare `Err(_)` also passed if the parser regressed into returning an EOF/`Io` error, which is exactly what the neighbouring truncation tests assert for. Rolled into a broader test that covers every `Str`-carrying position (simple string, inline error code/detail, blob error code/detail, verbatim text, verbatim format prefix, push kind), since they're separate code paths with different messages.

redis/src/types.rs
nihohit: please don't rename types.rs to types/mod.rs .
I understand the logic behind the choice, but this will make future merges from main significantly harder. I will make this change after the work will be merged into `main`.

fcostaoliveira: Reverted in 7c2bf8d — `types.rs` keeps its path. You were right about the cost: `types/mod.rs` was one of six conflicts in the upstream merge just before this.

`Str` still lives in `types/str.rs` with its tests, since `types.rs` and `types/` can coexist — that keeps your original comment satisfied without the rename. If you'd rather it went back inline until you do the split yourself, say so and I'll fold it in.

fcostaoliveira:
@nihohit rebased onto 2.0.x, conflicts resolved, CI green (23/23). `make lint`, `make test` and the unit suite pass locally.

Reverted the `types.rs` rename as you asked — `Str` stays in `types/str.rs`, which coexists with `types.rs`, so no rename. You were right about the cost: `types/mod.rs` was one of six conflicts in the merge.

Two things worth knowing before you spend time on it:

**The async codec is slower on small replies.** ~19-25% on `+OK`/`:1` and ~37-51% on deep nesting, against 241x faster on a 1 MiB bulk and zero per-reply allocations in pipelines. Cause is atomic refcount traffic replacing per-leaf malloc, plus the intermediate `ValueRange` tree allocating per container. This answers the doubt you raised on #2184 — small replies are the average user, so it's in the description rather than buried. Note the original bench only measured `parse_redis_value`, which builds a fresh parser per call and inverted the sign on nesting; codec-level benches are added.

**Two retention bugs found and fixed since your review**, both from payloads now being slices of the read buffer: the client-side cache pinned whole buffers (23.5 MB vs 638 KB on 2.0.x for 200 cached values), and `parse_redis_value_async` reserved a fresh 8 KiB chunk per reply (28x worse than the function it replaced). `caching/` had never been adapted.

Also fixed from your first pass: the dropped parse state (real regression, 718ms -> 2.67ms), and the `unsafe` precondition is now type-enforced rather than asserted in prose — worth flagging that Miri does **not** check `from_utf8_unchecked`, confirmed with a bad canary it happily passed.

Anything you'd rather see cut before you look? The sync-read change is the obvious candidate for #2185 — I kept it only because it repairs a regression this rewrite introduced, not as a separate optimization.

nihohit:
Thanks @fcostaoliveira, sounds great! A couple of questions, before I start reading through the code - 

> ~19-25% on +OK/:1 and ~37-51% on deep nesting [..]. Cause is atomic refcount traffic replacing per-leaf malloc

What's the atomic refcounting, and why is this needed? What are we synchronizing that requires atomics?
Is the difference between the current state and the last round, or between the PR and the baseline? This is pretty worrying.
Also, what's "deep" nesting? how deep did you test?

> `parse_redis_value_async` reserved a fresh 8 KiB chunk per reply 

Should we reserve only if the buffer doesn't have enough free room? or never reserve at all?


Also, reading over the PR's description again - there are 2 "Benchmarks" sections, with different number sets - is that because the benchmarks were run on 2 different machines?

nihohit:
as for the cost of allocation, maybe we can use replace Vec with something like SmallVec / TinyVec which will allocate only for larger arrays - we'll probably only want that in the parser, not in the external API, but it might remove the need for an additional allocation in ValueRange.

fcostaoliveira:
Good questions — several of them turned up things we had wrong. Taking them in order.

**"What's the atomic refcounting, and why is this needed?"**

First, a correction: our sentence conflated two effects and half of it was wrong. `+OK`, `:1` and the nested-array case contain **zero `Bytes`/`Str` leaves**, so nothing "replaced per-leaf malloc" in any of the payloads that regressed. The atomics are a new *fixed per-reply* cost, not a replacement.

Where they come from: every reply paid `buffer.split_to(n).freeze()`, which makes the buffer shared — `ref_cnt.fetch_add(1, Relaxed)` on clone, `fetch_sub(1, Release)` plus an `Acquire` load before dealloc on drop (`bytes-1.12/src/bytes.rs:1442`, `:1530`, `:1554`). ~22ns/reply here.

Why atomics at all: `Value: Send` is load-bearing, and it's genuinely concurrent rather than just moved. Pipelined replies decoded from one read share **one** refcount and are dropped by different worker tasks, so `Rc` would be UB, not just a `Send` violation.

**Fixed in `4d6500c`.** When a reply is a bare scalar it now `advance`s instead of freezing. `+OK`/`:1`/`_`/`#t`/`,1.5` and a 1000-reply pipeline are ~11% faster than the previous tip (improved in 9/9 runs); the regression vs 2.0.x drops from ~25% to ~10%. Frame-carrying replies are unchanged.

**"Current state vs last round, or PR vs baseline?"**

PR vs `2.0.x`. Verified rather than assumed — head-vs-previous-round is 1.00 on every workload, checked against two candidate "last rounds".

**"What's 'deep' nesting? how deep did you test?"**

`*1` chains at depth 10/50/90. 90 is 90% of `MAX_RECURSE_DEPTH`, so it's near the deepest reply the parser accepts at all — not real traffic. Two corrections, and the framing was the bigger error:

- **37-51% was too high.** It came from criterion wall-clock; under thread-CPU paired sampling it's +25/34/35% at 10/50/90 (call it +30-40%, the error bar is ~5pp).
- At depths 1-3 — everything real, LRANGE/HGETALL/XREAD — it's 1.22/1.21/1.26, i.e. **identical to a flat small reply**. Nesting adds nothing at realistic depths; the extra only appears at depth ≥20. Presenting it next to the small-reply number implied a second user-visible case, and that was misleading.

Cause isn't depth: we allocate 2 Vecs per aggregate (`Vec<ValueRange>` then `Vec<Value>`) where 2.0.x allocates 1, so it scales with aggregates-per-leaf. Real replies have many leaves per aggregate and come out ahead — HGETALL 10 pairs 21→5 allocations, LRANGE 100 32B 101→7.

End-to-end against a real server: request/response unchanged (0.99); 1000-command GET pipeline +9%. That one's real and I won't hand-wave it with an RTT argument.

**"reserve only if the buffer doesn't have enough free room, or never reserve at all?"**

The former, already pushed: `if buffer.capacity() - buffer.len() < READ_CHUNK_SIZE / 8`. Never reserving isn't safe — after a `split_to`/`freeze` the spare capacity can be arbitrarily small, and reads degrade to tiny chunks.

**"2 Benchmarks sections — 2 different machines?"**

Yes, and that was confusing. Fixed: the second is retitled to say so, the first now carries only the deterministic allocation counts (the wall-clock column didn't reproduce on the second machine), and a stale row labelled "Map of 1000 pairs" that actually measures a flat `*2000` array is corrected.

**SmallVec/TinyVec**

Good instinct — that's exactly where the cost is — but it can't be expressed: the inline buffer is stored by value, so `Array(SmallVec<[Self; N]>)` is an infinitely-sized recursive type (E0072) for any N. `SmallVec<[Box<ValueRange>; N]>` trades 1 allocation for N. (FWIW `smallvec` is already in our tree transitively via `url`, so the dependency wouldn't have been the objection.)

What does work, no new dependency: recycle the intermediate `Vec<ValueRange>` through a thread-local pool. That's `1 × containers`, parity with 2.0.x on nesting, and it also removes growth reallocs — 100-element array 8→2 allocations, 5000-element 15→4. Nesting regression shrinks 35-63%; leaf-heavy cases get ~5% faster; nothing regresses. It's ~70 lines with a `thread_local!` and a `Drop` guard, so I'd rather land it as a follow-up than grow this PR — happy either way, your call.

One more thing worth flagging: our own `bench_decode` goes through `parse_redis_value`, which builds a fresh parser per call and charges 2.0.x for setup a live connection never pays. That's what inverted the sign on nesting, and it may be flattering us on many-small-leaf arrays too. Codec-level benches with a reused codec are added; I'd treat the older numbers with suspicion.

nihohit:
> so I'd rather land it as a follow-up than grow this PR

sounds reasonable.

> 1000-command GET pipeline +9%. That one's real and I won't hand-wave it with an RTT argument.

Just to be clear - this means that for large pipelines of flat values there's a regression in real world testing? Isn't this the scenario equivalent to either `5000 small bulks` or `500 × 1 KiB bulk`, both of which show improvement in the benchmarks?
If so, what's the cause of this?
What other real-world scenarios did you benchmark, and what were the results there?

> SmallVec/TinyVec

I think your response misses something there. Both crates have growable types, where you define a stack size of N, and only once the number of items passes the stack size the container allocates and passes the values to the heap. So for small collections / maps this could save the allocation.

> At depths 1-3 — everything real, LRANGE/HGETALL/XREAD — it's 1.22/1.21/1.26, i.e. identical to a flat small reply

Just verifying - is this a 1.2* speedup or slow down? Because "a flat small reply" could be either an OK or the `5000 small bulks` benchmark scenario.

fcostaoliveira:
> is this a 1.2* speedup or slow down?

Slowdown — and that's on me, I published bare ratios without saying. Everything below is a **signed percentage, + = slower**.

> Isn't this the scenario equivalent to either `5000 small bulks` or `500 × 1 KiB bulk`, both of which show improvement?

You've found something worse than an inconsistency: **`5000 small bulks` is not actually an improvement, and our benchmark was measuring the wrong thing.**

Same bytes, same machine, same process, two entry points:

| payload | via `parse_redis_value` (what `bench_decode` measures) | via reused `ValueCodec` (what `MultiplexedConnection` uses) |
|---|---|---|
| 5000 small bulks | **−16%** | **+5%** |
| 500 × 1 KiB | −55% | **−23%** |
| 1 MiB bulk | −59% | −98% |
| GET 20 B | −10% | **+8%** |

2.0.x's `parse_redis_value` builds a fresh `Parser` and goes through `combine::decode!`, which re-enters the parser after every refill (`combine-4.6.7/src/stream/mod.rs:1421-1449`). On identical bytes, **2.0.x's own `parse_redis_value` is 1.2–2.3× slower than 2.0.x's own codec** — so `bench_decode` compares us against an inflated baseline that no real connection pays. `500 × 1 KiB` survives as a genuine win; `5000 small bulks` is roughly parity through the codec, and the 1.36× we published is an artifact. I'd treat every number in that table as suspect.

So the pipeline isn't a different scenario from the array — it's the same small-leaf regime, about 15pp worse because it pays the per-reply frame 1000× instead of once.

**The real variable is bytes per leaf, not element count.** 500 leaves, codec:

| bytes/leaf | one array of 500 | pipeline of 500 |
|---|---|---|
| 8 B | +12% | +24% |
| 128 B | +5% | +23% |
| 512 B | −6% | +11% |
| 1 KiB | −16% | +3% |
| 2 KiB | −77% | −15% |

Crossover ~300 B/leaf in one array, ~1.5 KiB/leaf in a pipeline. We swap "malloc + memcpy(len)" for "atomic refcount + slice": the memcpy scales with leaf size, the refcount doesn't. Per-reply it's a fixed cost, +16ns when `needs_frame` is false and +57ns when it's true — the 41ns gap is `split_to().freeze()`, which I measured standalone at 43.5ns. And you're right that `needs_frame` doesn't help here: `$3\r\nfoo` still needs the frame, which is why `pipe 1000 SET` is +10% and `pipe 1000 GET` is +22%.

> What other real-world scenarios did you benchmark?

Codec / end-to-end vs real redis, + = slower:

| | codec | e2e |
|---|---|---|
| GET / SET / INCR / EXISTS | +21 / +10 / +10 / +11% | no measurable effect (RTT-dominated) |
| MGET 10 / 100 | +19 / +5% | +4 / +5% |
| HGETALL 10 / 100 | +9 / +8% | +4 / +8% |
| LRANGE 100 / 1000 | +5 / +12% | +5 / +11% |
| SMEMBERS 50 / ZRANGE 50 | +5 / +5% | +4 / +8% |
| pipe 1000 SET / GET | +10 / +22% | +4 / **+8%** |

**And one thing we didn't expect: most of the end-to-end loss on multi-element reads isn't the parser at all.** Stopping at `Value` instead of converting:

| | → `Vec<String>` | → `Value` |
|---|---|---|
| MGET 100 | +5% | −1% |
| LRANGE 1000 | +11% | **−3%** |
| HGETALL 100 | +8% | **−3%** |

Parsing is at parity or better; the loss is the leaf copy that moved from parse time to conversion time. `version2.md` claims net copies are unchanged — true, but it's not time-neutral for these shapes, and I'll fix that wording.

The crossover data suggests a concrete fix: **copy leaves below ~256 B instead of slicing them** — the same trick already used to detach errors. It targets exactly the small-GET-pipeline case without touching the large-reply wins, and it would stop small values pinning whole reply buffers. Want me to try it in this PR, or keep it with the `ValueRange` pooling as follow-up work?

Caveats: this box is clock-throttled and was under load, so these are ratios only, nothing finer than ~5pp, and I wouldn't defend the single-command e2e rows as anything but noise. Our earlier "241× on 1 MiB" should be read as ">50×" — it's a function of reply size, not a constant.

Still owe you a proper answer on SmallVec — you're right that I addressed the wrong thing. Coming separately.

fcostaoliveira:
On SmallVec — you're right that both are growable and only spill past N, and I never meant to dispute that. My wording read as "SmallVec can't grow", which isn't what I meant and isn't true.

The blocker is narrower: the inline buffer is held **by value**, so `[Self; N]` makes `ValueRange` contain itself. rustc rejects every N, including 1:

```
error[E0072]: recursive type `ValueRange` has infinite size
  |     Array(SmallVec<[Self; 1]>),
  |           ------------------- recursive without indirection
```

`size_of::<SmallVec<[T; N]>>() >= N * size_of::<T>()`, so with `S = size_of::<ValueRange>()` you need `S >= N*S`. Same for TinyVec/ArrayVec, for `SmallVec<[(Self, Self); N]>`, and through a newtype — rustc just names both types in that last case. Only a pointer breaks the cycle, and `SmallVec<[Box<Self>; 4]>` does compile but is worse: measured 102 allocations for a 100-element array against 7, and 8-12% slower.

**But your instinct was right — there was an allocation to save, just not where either of us was looking.** Chasing it properly turned up this, for a 100-element array on the current branch:

```
alloc   4848          <- intermediate Vec<ValueRange>: ONE alloc, correctly sized
alloc    288          <- output Vec<Value>, capacity 4 ...
realloc  288 ->  576  <- ... doubling five times
realloc  576 -> 1152
realloc 1152 -> 2304
realloc 2304 -> 4608
realloc 4608 -> 9216
```

Six of the seven allocations are the **output** vector, not the parser. `materialize_vec` ended in `collect()` into a `Result<Vec<_>, _>`, and collecting into a `Result` reports a size-hint lower bound of 0 (it may short-circuit), so it starts at capacity 4 and doubles.

The intermediate `Vec<ValueRange>` — the one SmallVec was aimed at — is the one allocation that was **already optimal**, because combine's `count_min_max` feeds `extend` a size hint.

Fixed in `6e5f922` with `Vec::with_capacity`. On the existing bench:

| payload | before | after |
|---|---|---|
| 5000 small bulks | 16 | **5** |
| 500 × 1 KiB bulks | 11 | **4** |
| 5000 simple strings | 16 | **5** |
| 1000 key/value pairs | 13 | **4** |

I also prototyped the formulation that *does* capture your intent — `SmallVec<[u32; N]>` indexing a reused arena, which is finite-sized and gets the inline win for small collections. It compiles and removes the intermediate allocation, but measured no faster than plain buffer reuse, and it needs the arena parked in `ParseState` (I hit a bug where the grown and empty buffers trade places every parse and neither is reused). Between that and the thread-local pool, the pool is simpler and already reaches zero container allocations, which leaves the inline storage nothing to win. Happy to send either as the follow-up.

FWIW `smallvec` is already in `Cargo.lock` via `url → idna → idna_adapter → icu_normalizer`, so adding it as a direct dep would have cost nothing — it just doesn't buy anything here.

nihohit:
I assume that the latest posted benchmarks include `6e5f922`.

I still don't understand these two lines:

| payload | via `parse_redis_value` (what `bench_decode` measures) | via reused `ValueCodec` (what `MultiplexedConnection` uses) |
|---|---|---|
| 500 × 1 KiB | −55% | **−23%** |

which seems to contradict 

| bytes/leaf | one array of 500 | pipeline of 500 |
|---|---|---|
| 1 KiB | −16% | +3% |


In any case, these numbers aren't good - this looks like a regressions for most scenarios. I wonder if we can salvage this, pivot to another change (maybe add a LargeBulkString variant?), or whether we should abandon this PR.

Another option is to save the allocation in the first parse run - all of the recursive variants will just define a range (So `Set(Vec<Self>)` becomes `Set(Range<usize>)` or maybe `Set { count: usize, data: Range<usize> }`), and then in the materialization phase perform both parsing steps. Although, this should only save one allocation, so I don't see it bringing a large change in perf. Any other idea?

nihohit:
@fcostaoliveira are you still interested in moving this forward?

PR #2200, opened 2026-07-10
commands/macros: Move `RV` to last position for Generic typed commands

somechris:
While untyped commands have their return value's type (`RV`) in the last
position, `Generic` typed commands had it in the first position.

This difference is unexpected. So we harmonize on the last position, as
this is what the untyped commands, which were around longer, use.

As the `Generic` handling was duplicating logic, we also switch to
delegating to the general case.

### DIFF
--- a/redis/src/commands/macros.rs
+++ b/redis/src/commands/macros.rs
@@ -8,16 +8,12 @@ macro_rules! implement_command_async {
         fn $name:ident<$($tyargs:ident : $ty:ident),*>(
             $($argname:ident: $argty:ty),*) $body:block Generic
     ) => {
-        $(#[$attr])*
-        #[inline]
-        #[allow(clippy::extra_unused_lifetimes, clippy::needless_lifetimes)]
-        fn $name<$lifetime, RV: FromRedisValue, $($tyargs: $ty + Send + Sync + $lifetime,)*>(
-            & $lifetime mut self
-            $(, $argname: $argty)*
-        ) -> crate::types::RedisFuture<$lifetime, RV>
-        {
-            Box::pin(async move { $body.query_async(self).await })
-        }
+        implement_command_async!(
+            $lifetime
+            $(#[$attr])+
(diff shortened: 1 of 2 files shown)


somechris:
As messing with such macros is always tricky to review, I've again
split it into two separate, isolated atomic commits.

If you prefer to have them submitted individually, please let me know.

nihohit:
IIRC the reason that the RV is in the first position is in order to make declaring generics easier. While I agree with the aesthetics of this change, IMO breaking existing code just for the sake of harmony isn't worth the effort - there should be some concrete benefit.

somechris:
When is the time to fix tech debt, if not with a new major version? :grin:

I assume only few users would need to migrate anything, as it's easier to write

```
let result: TYPE = con.operation(...)
```

than any of

```
let result = con.operation::<TYPE, ...>(...)
let result = con.operation::<..., TYPE>(...)
```

> there should be some concrete benefit.

The hard ground truth is that I started back when `TypedCommands` was not yet around. So I used `Commands` and learnt "RV is last". Now, I try to use `TypedCommands` wherever I can and have to relearn "RV is first". But from time to time, `TypedCommands` is too opinionated about a return type, and I have to resort to `Commands` and then "RV is last" again.

That's frustrating.

I do not care whether `RV` is first or last. Anything is fine.

But I do care about consistency. About not having to think.

So the benefit for me would be to improve on the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment); hence maturing `redis-rs`.

Much in the spirit of D. Golden, who wrote in ["A Plea for Friendly Software"](https://dl.acm.org/doi/epdf/10.1145/1010884.1010885):

> [...], I would point out that it is perfectly possible to have a program which is structured, modular, readable, flexible, self-documenting, maintainable, which performs its specified function, and which is a source of constant frustration and irritation to its users.

:heart:

nihohit:
> But I do care about [...] not having to think.

❤️ 

that convinces me :)

nihohit (approved):

PR #2205, opened 2026-07-11
Improve `RedisServer` readability

somechris:
Over time, `RedisServer::new_with_addr_tls_modules_and_spawner` grew hard to read.

The instantiation of `RedisServer` was duplicated over all `match addr` branches. Different names are used for the same things (`host` vs. `bind`). Etc.

We clean the them up to improve readability.

### DIFF
--- a/redis-test/src/cluster.rs
+++ b/redis-test/src/cluster.rs
@@ -171,7 +171,7 @@ impl RedisCluster {
         let max_attempts = 5;
 
         let mut make_server = |port| {
-            RedisServer::new_with_addr_tls_modules_and_spawner(
+            RedisServer::new_with_addr_tls_modules_and_arg_refiner(
                 ClusterType::build_addr(port),
                 None,
                 tls_paths.clone(),
@@ -210,7 +210,6 @@ impl RedisCluster {
                     }
                     cmd.current_dir(tempdir.path());
                     folders.push(tempdir);
-                    cmd.spawn().unwrap()
                 },
             )
         };
--- a/redis-test/src/cluster.rs
+++ b/redis-test/src/cluster.rs
@@ -190,22 +190,17 @@ impl RedisCluster {
                         Self::password()
                     );
                     std::fs::write(&acl_path, acl_content).expect("failed to write acl file");
-                    cmd.arg("--cluster-enabled")
-                        .arg("yes")
-                        .arg("--cluster-config-file")
-                        .arg(tempdir.path().join("nodes.conf"))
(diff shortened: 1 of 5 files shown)


somechris:
The 7 commits here are individually simple to review, but harder as a diff over the whole PR. If you prefer individual PRs, let me know.

nihohit:
instead of a massive function, wouldn't it make more sense to use a builder type?

somechris:
> instead of a massive function, wouldn't it make more sense to use a builder type?

You're taking the words right out of my mouth! :+1: 

The follow-up would be add a builder and switch to that. That would give a nicer user experience.

But the builder needs a function that actually builds `RedisServer`. That'd essentially boil down to `new_with_addr_tls_modules_and_spawner`, as many parts in there are not optional. But that method is longish and not too readable. So I figured it's best to clean it up beforehand and not have everything in one big PR.

Would you be ok if I aborted this, add the builder, and then upload each commit again as individual PR (as we did with the version refactoring) against 2.x?

One design question though: :thinking: 

Our integration tests currently mostly do not use `RedisServer` directly, but indirectly through `TestContext` (and its relatives).

So a builder for `RedisServer` does not benefit our integration tests directly. We'd instead need a builder for `TestContext`.

But `TestContext` (which is greatly useful when writing tests) is not visible to 3rd party users as it's in `redis::tests`, not `redis-test`.

So instead of adding a builder for `RedisServer`. I'd propose to
* Add a builder for `RedisServer`,
* Move `TestContext` into `redis-test` (as it would be useful for 3rd party users ... I would loved to have that available on $LAST_DAYJOB), and
* Add a builder for `TestContext`.

WDYT?

nihohit:
> Would you be ok if I aborted this, add the builder, and then upload each commit again as individual PR (as we did with the version refactoring) against 2.x?

sure, no problem.

> So instead of adding a builder for RedisServer. I'd propose to

> Add a builder for RedisServer,
> Move TestContext into redis-test (as it would be useful for 3rd party users ... I would loved to have that available on $LAST_DAYJOB), and
> Add a builder for TestContext.

Switch the order of the last 2 - IIRC there were reasons that TestContext wasn't moved, maybe something about circular dependencies, or versioning when making a breaking change? In any case, moving TestContext will be a large change, so let's delay it to the end.

somechris:
Closing, as it will be redone with adding `RedisServer` and `TestContext` builders, and (if there are no dependency issues etc) moving `TestContext` to `redis-test`.

PR #2208, opened 2026-07-13
tests: Stop running sentinel tests when testing on unix sockets

somechris:
When using `REDISRS_SERVER_TYPE=unix`, sentinel tests were run. But (as
sentinels require TCP) they did not test on unix sockets, but fell back
to TCP.

But that's was a dupe, as sentinels on TCP got tested already during the
`REDISRS_SERVER_TYPE=tcp` parts of `make test`.

So we tested the same thing twice.

This only slowed test runs down and we therefore now stop running
sentinels during unix sockets testing.

We also turn the unix sockets -> TCP fallback off in `get_addr`, as we
should not second guess user choices. This would have allowed us to
detect the "running tests twice" earlier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ test:
 	@echo "===================================================================="
 	@echo "Testing Connection Type UNIX SOCKETS"
 	@echo "===================================================================="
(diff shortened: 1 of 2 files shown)


nihohit (approved):
thanks!

PR #2209, opened 2026-07-13
nextest: Clean up overrides for no longer existing tests

somechris:


nihohit:
from this PR's test runs:
```
        SLOW [>  3.000s] (───────) redis::test_cluster_async cluster_async::test_async_cluster_basic_failover::case_1_tokio
        SLOW [>  3.000s] (───────) redis::test_cluster_async cluster_async::test_async_cluster_basic_failover::case_2_smol

somechris:
Argh! My bad!

The filter is for `test_async_cluster_...` so I grepped for that in the code and did not match anything.

Now I see that the function is called `async_cluster_...` (no leading `test_`) ... so nextest prefixes a `test_` here? Huh! That's ... nasty!

Anyway closing as invalid.

Thanks for catching that!

nihohit:
> so nextest prefixes a test_ here? Huh! That's ... nasty!

That's actually our test macros (https://github.com/redis-rs/redis-rs/blob/main/test-macros/src/lib.rs) that add the prefix. The reason is that some tests have multiple prefixes (`test_multiplexed_connection_`, `test_connection_manager_`), but you're right that it might be better to restore the `test_` prefix into the code, and add all other prefixes as `::multiplexed_connection`|`::connection_manager` suffixes

nihohit:
https://github.com/redis-rs/redis-rs/pull/2211

PR #2210, opened 2026-07-13
test: Switch to selecting tests through profiles

somechris:
We needlessly ran a few tests more than once (e.g.: `test_tls_cert_auth`
was running for `tcp+tls`. But unwarranted also for `unix`, and for
`tcp` without TLS).

Updating all selectors in-situ would have resulted in a maintenance
burden.

So nextest is now run through profiles, which take care to not run tests
needlessly. This gives us a better readable `Makefile`, allows to re-use
the profiles when running manually. And the updating the filtersets is
easy too.

We re-enable the tests that went missing in a16079a and c34fb33 due to
non-grep-able test names.


nihohit (approved):
thanks!

PR #2214, opened 2026-07-15
tests/acl: Add the missing requirements for token based authentication

somechris:
`aio` gets added to the `test_acl::token_based_authentication_acl_tests` guard, as it imports from `redis::aio`.

The mod also relied on `futures_channel`, which is available only for the `connection-manager` feature. But as that is not needed, we add it as general dev dependency.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -194,6 +194,7 @@ rustls = { version = "0.23", features = ["ring"] }
 test-macros = { path = "../test-macros" }
 assert_matches = "1.5.0"
 env_logger = "0.11.10"
+futures-channel = "0.3.31"
 
 [[test]]
 name = "test_async"
--- a/redis/tests/test_acl.rs
+++ b/redis/tests/test_acl.rs
@@ -308,7 +308,11 @@ fn test_acl_sample_info() {
     );
 }
 
-#[cfg(all(feature = "acl", feature = "token-based-authentication"))]
+#[cfg(all(
+    feature = "acl",
+    feature = "aio",
+    feature = "token-based-authentication"
+))]
 mod token_based_authentication_acl_tests {
     use crate::support::*;
     use futures_channel::oneshot;


nihohit:
can you please explain what's broken? is there a `cargo test` command with the right features that fails compilation?

somechris:
> can you please explain what's broken? is there a `cargo test` command with the right features that fails compilation?

On `main`

```
cargo nextest run -p redis --features acl,token-based-authentication --test 'test_acl'
```

(the exact guard conditions of `token_based_authentication_acl_tests`) fails to compile with 80 errors. Basically: `aio` itelf is required but not available, and `futures_channel` is missing.

After the fix, it runs 8 tests:

```
cargo nextest run -p redis --features acl,token-based-authentication --test 'test_acl'
   Compiling redis v1.4.0 (/home/christian/rust/redis-rs/redis)
   Compiling redis-test v1.0.4 (/home/christian/rust/redis-rs/redis-test)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 19.54s
────────────
 Nextest run ID 40ef03ea-6736-420f-8ddf-d574c0534509 with nextest profile: default
    Starting 8 tests across 1 binary (1 test skipped)
        SKIP [         ] redis::test_acl test_acl_getsetdel_users
        PASS [   0.076s] redis::test_acl test_acl_cat
        PASS [   0.073s] redis::test_acl test_acl_sample_info
        PASS [   0.088s] redis::test_acl test_acl_genpass
        PASS [   0.091s] redis::test_acl test_acl_log
        PASS [   0.091s] redis::test_acl test_acl_help
        PASS [   0.094s] redis::test_acl test_acl_dryrun
        PASS [   0.093s] redis::test_acl test_acl_info
        PASS [   0.096s] redis::test_acl test_acl_whoami
────────────
     Summary [   0.099s] 8 tests run: 8 passed, 1 skipped
```

The issue is that the guards of `token_based_authentication_acl_tests` sub-`mod` are too weak.

In CI, we run either with

* `--all-features`
* `--features=json,tokio-native-tls-comp,smol-native-tls-comp,connection-manager,cluster-async`
* `--no-default-features`

The first two have the guard's required `acl` and `token-based-authentication`, but also the additional `aio` (with runtime) and `connection-manager` (which pulls in `futures_channel`), so these tests pass in CI.

The last fails to meet the guard's requirements, so the sub-`mod` is omitted and it passes in CI.

We simply never tested with only guard's requirements.

But when working on the `TestContext` refactoring, I went over things manually and this triggered.

nihohit:
Do you think that we should build with tests in the flag-frenzy step, to catch these things?

somechris:
> Do you think that we should build with tests in the flag-frenzy step, to catch these things?

Naw, I'd probably not; the tests are not for the general users.

People, who run them with custom features, can probably figure things out or open a ticket about it.

CI already takes long. And the biggest part of that is flag-frenzy. Adding more things there would make things only take even longer.

But in some other branch I'm adding features to `redis-test` ... I was thinking about enabling it for `redis-test` :-) But that's not for here to discuss ;-)

PR #2216, opened 2026-07-15
Add support for the `INCREX` command

StefanPalashev:
# Add support for the `INCREX` command

## Summary

Add support for the `INCREX` command introduced in **Redis 8.8**.
`INCREX` increments a key's numeric value and sets its expiration in a single call, with optional lower/upper bounds and a
choice between rejecting or **saturating** out-of-bounds results.

## Background

Before `INCREX`, incrementing a counter and (re)setting its `TTL` required two round-trips - 
`INCRBY` / `INCRBYFLOAT` followed by `EXPIRE`, which is not atomic and cannot express "clamp to a ceiling" in one step.
`INCREX` folds all of this into one command and adds a bounds policy.

| Concern | Before (`INCRBY` + `EXPIRE`) | After (`INCREX`) |
| --- | --- | --- |
| Increment + set TTL | Two commands, non-atomic | One atomic command |
| Bounded increment (cap/floor) | Manual read-modify-write | Done through `LBOUND` / `UBOUND` |
| Overflow handling | Errors out (`INCRBY`) | Reject silently or `SATURATE` to the bound |
| Was it applied? | Inferred by the caller | Reported directly as `actual_increment` |

Example:

```
> SET counter 98
> INCREX counter BYINT 5 SATURATE UBOUND 100
1) (integer) 100   ; value after the (clamped) increment
2) (integer) 2     ; actual increment applied (100 - 98), not the requested 5
```

## Syntax

```
INCREX key [BYFLOAT increment | BYINT increment]
           [SATURATE]
           [LBOUND lower-bound] [UBOUND upper-bound]
           [EX seconds | PX milliseconds | EXAT unix-time-seconds
            | PXAT unix-time-milliseconds | PERSIST]
           [ENX]
```

The reply is always a two-element array `[value, actual_increment]`:

- **`value`** - the key's value after the increment.
- **`actual_increment`** - the increment actually applied. It is `0` when the default policy rejected an out-of-bounds operation and equals the clamped delta (which may differ from the requested increment) under `SATURATE`.

## Behavior

- **Default policy (no `SATURATE`), out of bounds** - the operation is rejected silently.
The reply is `[current_value, 0]`, which is a regular successful reply and not an error.
The key's value **and TTL** are left unchanged (a supplied expiration is **not** applied).
- **`SATURATE`** - the result is clamped to the explicit `LBOUND`/`UBOUND` or to the type limit when no bound is given (`i64::MAX`/`i64::MIN` for `BYINT`; `±LDBL_MAX` for `BYFLOAT`). `actual_increment` reports the clamped delta.
- **Zero-delta `SATURATE`** - a clamp that lands on the current value has an `actual_increment == 0` but still counts as *applied*, so a supplied expiration **is** applied (subject to `ENX`). This differs from the default-policy reject above.
- **`ENX`** - only sets the expiration if the key has no TTL. It requires one of `EX`/`PX`/`EXAT`/`PXAT` and cannot be combined with `PERSIST` because the server rejects the combination.
- **Server errors** are forwarded verbatim through `RedisError::code()` / `detail()`.

### Numeric type fidelity

| Rust type | Server type | Notes |
| --- | --- | --- |
| `i64` | 64-bit `long long` | **Exact** - clamp limits are `i64::MAX`/`i64::MIN`. Out-of-range arguments are rejected server-side. |
| `f64` | C `long double` | Matches the crate-wide convention. The generic `Commands` variant with a wider `FromRedisValue` type should be used for full precision. |

## Testing

| Test | Gating | Covers |
| --- | --- | --- |
| `test_increx_options_args` | none | unit test coverage of `IncrexOptions`. |
| `test_increx_with_integers` | `REDIS_CE_8_8` | integration test that tests the behavior of the increment by integer method with scenarios such as - normal apply, default rejection policy `[value, 0]`, explicit bounds clamp, implicit saturating clamp and more. |
| `test_increx_with_floats` | `REDIS_CE_8_8` | same as the one above, but for the increment by float method and without the implicit saturating clamp.  |
| `test_increx_server_errors_forwarded_verbatim` | `REDIS_CE_8_8` | verbatim `code()`/`detail()` for `WRONGTYPE`, non-numeric value (BYINT/BYFLOAT) and `ENX`-without-expiration. |

## Compatibility

- The changes are additive and no existing public APIs were changed.
- No new cargo feature was added, since `INCREX` is a core string command.
- Server behavior gated to **Redis CE 8.8+** in tests via the `REDIS_CE_8_8` constant.
- The one change to existing code is adding `#[derive(Clone)]` to `Expiry`, which is backward compatible.

## Relevant documentation

- [`INCREX`](https://redis.io/commands/increx/), [`INCRBY`](https://redis.io/commands/incrby/), [`INCRBYFLOAT`](https://redis.io/commands/incrbyfloat/)

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -519,33 +519,34 @@ implement_commands! {
         cmd("DECRBY").arg(key).arg(delta).take()
     }
 
-    /// Increment the integer value of a key by the given amount and set its expiration.
+    /// Increment the numeric value of a key by the given amount and set its expiration.
     ///
     /// Uses 0 as the initial value if the key does not exist.
-    /// The reply is an [`IncrexResult`] holding the value after the increment and the increment that was actually applied
+    /// The increment's type determines the operation: integer increments use BYINT, while floating-point ones use BYFLOAT.
+    /// The reply is an [`IncrexResult<V>`] holding the value after the increment and the increment that was actually applied
     /// (see [`IncrexOptions`] for the bounds and `SATURATE` behavior).
+    /// Note that `V` also types the `LBOUND`/`UBOUND` bounds.
     ///
-    /// The server operates on 64-bit signed integers, so `i64` is an exact match.
+    /// For `BYINT`, the server operates on 64-bit signed integers, so `i64` is an exact match.
     /// Every storable or clamped value fits in `i64`, the implicit `SATURATE` limits are `i64::MAX`/`i64::MIN`.
     /// Out-of-range increment or bound is rejected by the server.
-    /// [Redis Docs](https://redis.io/commands/INCREX)
-    fn increx_by_int<K: ToSingleRedisArg>(key: K, increment: i64, options: IncrexOptions<i64>) -> (IncrexResult<i64>) {
-        cmd("INCREX").arg(key).arg("BYINT").arg(increment).arg(options).take()
-    }
-
-    /// Increment the floating-point value of a key by the given amount and set its expiration.
     ///
-    /// Uses 0 as the initial value if the key does not exist.
-    /// The reply is an [`IncrexResult`] holding the value after the increment and the increment that was actually applied
-    /// (see [`IncrexOptions`] for the bounds and `SATURATE` behavior).
-    ///
-    /// The server computes in C `long double`, whose range and precision exceed `f64`.
+    /// For `BYFLOAT`, the server computes in C `long double`, whose range and precision exceed `f64`.
     /// A magnitude beyond `f64::MAX`, including the implicit `SATURATE` limit of `±LDBL_MAX`, therefore decodes to `±f64::INFINITY` rather than the exact value.
-    /// This is lossy saturation, not an error. If the exact value is needed, call the generic variant instead and choose a wider return type
-    /// (e.g. `(String, String)`, or a `bigdecimal::BigDecimal` pair with the `bigdecimal` feature).
+    /// This is lossy saturation, not an error. If the exact value, keep the `f64` increment, but call the generic variant and decode into a wider return type, e.g.
+    /// `let r: (String, String) = con.increx(key, 1.0f64, opts)?;` (or a `(BigDecimal, BigDecimal)` pair with the `bigdecimal` feature).
     /// [Redis Docs](https://redis.io/commands/INCREX)
-    fn increx_by_float<K: ToSingleRedisArg>(key: K, increment: f64, options: IncrexOptions<f64>) -> (IncrexResult<f64>) {
-        cmd("INCREX").arg(key).arg("BYFLOAT").arg(increment).arg(options).take()
+    fn increx<K: ToSingleRedisArg, V: IncrexNumber>(key: K, increment: V, options: IncrexOptions<V>) -> (IncrexResult<V>) {
+        cmd("INCREX")
+            .arg(key)
+            .arg(if increment.describe_numeric_behavior() == NumericBehavior::NumberIsFloat {
+                "BYFLOAT"
+            } else {
+                "BYINT"
+            })
+            .arg(increment)
+            .arg(options)
+            .take()
     }
 
     /// Sets or clears the bit at offset in the string value stored at key.
@@ -3858,10 +3859,20 @@ impl ToRedisArgs for Expiry {
     }
 }
 
+/// Numeric types accepted as an [`increx`](crate::TypedCommands::increx) increment.
+///
+/// A blanket implementation covers every type that is both
+/// a single Redis argument ([`ToSingleRedisArg`]) and decodable from the reply ([`FromRedisValue`]).
+/// Whether an increment issues `BYINT` or `BYFLOAT` is decided at runtime from its [`describe_numeric_behavior`](ToRedisArgs::describe_numeric_behavior).
+pub trait IncrexNumber: ToSingleRedisArg + FromRedisValue {}
+
+impl<T: ToSingleRedisArg + FromRedisValue> IncrexNumber for T {}
+
 /// Options for the [INCREX](https://redis.io/commands/increx) command.
 ///
 /// `T` is the type of the increment and of the `LBOUND`/`UBOUND` bounds.
-///  Use `IncrexOptions<i64>` with [`increx_by_int`] and `IncrexOptions<f64>` with [`increx_by_float`].
+/// It matches the increment passed to [`increx`](crate::TypedCommands::increx).
+/// (e.g. `IncrexOptions<i64>` for an `i64` increment, `IncrexOptions<f64>` for an `f64` one).
 ///
 /// # Example
 /// ```rust,no_run
@@ -3871,12 +3882,9 @@ impl ToRedisArgs for Expiry {
 ///         .saturate()
 ///         .upper_bound(100)
 ///         .with_expiration(Expiry::EX(60));
-///     con.increx_by_int("counter", 5, opts)
+///     con.increx("counter", 5i64, opts)
 /// }
 /// ```
-///
-/// [`increx_by_int`]: crate::TypedCommands::increx_by_int
-/// [`increx_by_float`]: crate::TypedCommands::increx_by_float
 #[derive(Clone, Default)]
 pub struct IncrexOptions<T> {
     saturate: bool,
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -4,9 +4,9 @@ use crate::cmd::{Cmd, Iter, cmd};
 use crate::connection::{Connection, ConnectionLike, Msg, RedisConnectionInfo};
 use crate::pipeline::Pipeline;
 use crate::types::{
-    ExistenceCheck, ExpireOption, Expiry, FieldExistenceCheck, FromRedisValue, IntegerReplyOrNoOp,
-    NumericBehavior, RedisResult, RedisWrite, SetExpiry, ToRedisArgs, ToSingleRedisArg,
-    ValueComparison,
+    ExistenceCheck, ExpireOption, Expiry, FieldExistenceCheck, FromRedisValue, IncrexResult,
+    IntegerReplyOrNoOp, NumericBehavior, RedisResult, RedisWrite, SetExpiry, ToRedisArgs,
+    ToSingleRedisArg, ValueComparison,
 };
 
 #[cfg(feature = "vector-sets")]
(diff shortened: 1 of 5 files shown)


redis/src/commands/mod.rs
```diff
@@ -519,6 +519,35 @@ implement_commands! {
...
+    ///
+    /// Uses 0 as the initial value if the key does not exist.
+    /// The reply is an [`IncrexResult`] holding the value after the increment and the increment that was actually applied
+    /// (see [`IncrexOptions`] for the bounds and `SATURATE` behavior).
+    ///
+    /// The server operates on 64-bit signed integers (C `long long`), so `i64` is an exact match.
+    /// Every storable or clamped value fits in `i64`, the implicit `SATURATE` limits are `i64::MAX`/`i64::MIN`.
+    /// Out-of-range increment or bound is rejected by the server.
+    /// [Redis Docs](https://redis.io/commands/INCREX)
+    fn increx_by_int<K: ToSingleRedisArg>(key: K, increment: i64, options: IncrexOptions<i64>) -> (IncrexResult<i64>) {
```
somechris: Not sure about the general design direction, but `incr` is in the same position and uses `NumericBehaviour` to switch between `INCR` and `INCRBYFLOAT`. Should we do the same thing here and only have `increx` that auto-switches between `BYFLOAT` and `BYINT`?

redis/src/commands/mod.rs
```diff
@@ -519,6 +519,35 @@ implement_commands! {
         cmd("DECRBY").arg(key).arg(delta).take()
     }
 
+    /// Increment the integer value of a key by the given amount and set its expiration.
+    ///
+    /// Uses 0 as the initial value if the key does not exist.
+    /// The reply is an [`IncrexResult`] holding the value after the increment and the increment that was actually applied
+    /// (see [`IncrexOptions`] for the bounds and `SATURATE` behavior).
+    ///
+    /// The server operates on 64-bit signed integers (C `long long`), so `i64` is an exact match.
```
somechris: `long long` is _at_ _least_ 64-bits. You're probably looking for `int64_t` (C99), which is exactly 64-bits. But I'd drop that C reference altogether. It's not C but Rust :-) I'd argue that mentioning the bits is good enough and less tricky than C types.

redis/src/commands/mod.rs
```diff
@@ -519,6 +519,35 @@ implement_commands! {
...
+    ///
+    /// The server computes in C `long double`, whose range and precision exceed `f64`.
+    /// A magnitude beyond `f64::MAX`, including the implicit `SATURATE` limit of `±LDBL_MAX`, therefore decodes to `±f64::INFINITY` rather than the exact value.
+    /// This is lossy saturation, not an error. If the exact value is needed, call the generic variant instead and choose a wider return type
+    /// (e.g. `(String, String)`, or a `bigdecimal::BigDecimal` pair with the `bigdecimal` feature).
+    /// [Redis Docs](https://redis.io/commands/INCREX)
+    fn increx_by_float<K: ToSingleRedisArg>(key: K, increment: f64, options: IncrexOptions<f64>) -> (IncrexResult<f64>) {
+        cmd("INCREX").arg(key).arg("BYFLOAT").arg(increment).arg(options).take()
+    }
+
```
somechris, line 560: I guess it's ok, but just to make it deliberately, not by accident: This approach does not cover the

> If neither BYFLOAT nor BYINT is specified, the key is incremented by 1 in integer mode.

That's fine, just want to make sure we do it deliberately.

redis/src/commands/mod.rs
```diff
@@ -3830,6 +3859,95 @@ impl ToRedisArgs for Expiry {
...
+    /// Set the upper bound for the operation (`UBOUND`).
+    pub fn upper_bound(mut self, upper_bound: T) -> Self {
+        self.upper_bound = Some(upper_bound);
+        self
+    }
+
+    /// Set the expiration to apply to the key (`EX`/`PX`/`EXAT`/`PXAT`/`PERSIST`).
+    pub fn with_expiration(mut self, expiration: Expiry) -> Self {
```
somechris, line 4009: Most methods of this builder come without the `with_` prefix. But here `with_` is used. That's inconsistent. 🤔 

But I also see that `expiration` in other parts is set by `with_expiration`. So using only `expiration` is inconsistent with that. 🤷 

Do we have some direction whether builders should use a `with_` prefix?

(My feeling is that we mostly do not have a `with_` prefix)

redis/src/commands/mod.rs
StefanPalashev: I saw that and it's a valid point, I decided to follow the proposed design document, but it can be changed to match the other signature and its functionality.

somechris: > I decided to follow the proposed design document, [...]

We have a design document for `redis-rs` ? :scream: Awesome! :clap: I did not know about it. Where can I find it?

redis/src/commands/mod.rs
StefanPalashev: Thank you. The C reference has been dropped.

redis/src/commands/mod.rs
StefanPalashev: It was actually an internal document focused specifically on the `INCREX` command.

Some of its contents are reflected in the description of the [original server pull request](https://github.com/redis/redis/pull/15045) (which my description also summarizes). There was also a [follow-up pull request](https://github.com/redis/redis/pull/15237) that revised the overflow policy and this pull request already reflects those changes.

StefanPalashev: Both approaches could work, I chose the explicit method names since they seemed more descriptive and make the selected variant obvious.

@nihohit, I'd appreciate your thoughts on this. If you think the other approach would be better, I'd be happy to rework it.

redis/src/commands/mod.rs
StefanPalashev: The expiration struct was reused similarly to how it was reused for previous commands implementations, thus it followed the same naming for the builder.
We might need some naming convention for it.

redis/src/commands/mod.rs
somechris: > It was actually an internal document focused specifically on the INCREX command.

Ah ok. Gotcha. Thanks.

And just to make sure: I'm not asking for a change. I only saw that it's different to `incr` and hence asked about it.

redis/src/commands/mod.rs
nihohit: sadly, I think that at this point we have builders using `with_` prefix, `use_` prefix, and no prefix. The only thing we can do without breakage is to find the closest existing builder and be consistent with its naming pattern.

redis/src/commands/mod.rs
nihohit: IMO we should be consistent with the `incr` pattern. It's not the best, because we can't pass a float in string form, but IMO consistency here is the less confusing pattern. If we believe that this is the wrong pattern, we could change both `incr` and `increx` in v2.

redis/src/commands/mod.rs
```diff
@@ -3829,6 +3859,102 @@ impl ToRedisArgs for Expiry {
     }
 }
 
+/// Numeric types accepted as an [`increx`](crate::TypedCommands::increx) increment.
+///
+/// A blanket implementation covers every type that is both
+/// a single Redis argument ([`ToSingleRedisArg`]) and decodable from the reply ([`FromRedisValue`]).
+/// Whether an increment issues `BYINT` or `BYFLOAT` is decided at runtime from its [`describe_numeric_behavior`](ToRedisArgs::describe_numeric_behavior).
+pub trait IncrexNumber: ToSingleRedisArg + FromRedisValue {}
```
nihohit: why is a new trait needed? can't we just use the `ToSingleRedisArg + FromRedisValue` bound in the functions?

StefanPalashev: No, `the implement_commands!` macro parses the bound as a single token `($ty:ident)`, so the `+` in `ToSingleRedisArg + FromRedisValue` doesn't match any macro rule. `IncrexNumber` is just those two bundled into one name.

We need both of them, because the result is generic.

The new signature changes `increx` to return `IncrexResult<V>` where `V` is the same type as the increment.
So `V` must implement:
1. `ToSingleRedisArg` - so it can be sent to Redis (the increment argument)
2. `FromRedisValue` - so it can be read back from Redis (to build the result)

somechris: I gave it a quick try to switch `$ty:ident` to `$ty:ty` but that does not pick up the `+`.

I tried `$ty:tt` with `(ToSingleRedisArg + FromRedisValue)`. That of course parsed the `+`, but during expansion, adding the `+ Send ...` failed due to the parentheses. Stripping them through a macro fails, as that get evaluated too late for me and Rust complained about not expecting `!` in a type. Stripping them in parsing does not work as closing parentheses are part of the repetition matching and I cannot escape them.

Parsing the `+` is in the way of the `at-least-once` parsing and does not work either.

So we'd be down to switching `implement_commands!` to a `where` or using different symbols (e.g.: `/` instead of `+` as in `ToSingleRedisArg/FromRedisValue`). The latter [of course works](https://github.com/somechris/redis-rs/commit/99a1eb65217862b3e983121b3839446a1f3d83a4#diff-e08d2f40a2c2d8561bd22f617620d780d893a15197020381b8e22eb5e31662f7R264) (`$lifetime` would need discussion) and is simple ... but the command definition with `/` looks really ugly to me.

But maybe there's a way simpler way out?
Do we need to couple the types of input and output?
Because to me, that looks unnecessary.

That should already help to get the typing working.

Also (if the docs are correct) the result to INCREX allows to determine whether it's a result for a BYFLOAT or a BYINT.

That would allow to shift the parsing and the result could change from `IncrexResult<V>` to `IncrexResult` killing the problem for good.

StefanPalashev: If the result is to be inferred from the reply type of the elements in the response array, the generic parameter from the `IncrexResult `could probably be dropped. Then the type has to be converted to an `enum` like so:

```
#[derive(Clone, Copy, Debug, PartialEq)]
pub enum IncrexResult {
    /// Reply of a BYINT operation (both elements are integers).
    Int   { value: i64, actual_increment: i64 },
    /// Reply of a BYFLOAT operation (bulk string on RESP2, double on RESP3).
    Float { value: f64, actual_increment: f64 },
}

impl FromRedisValue for IncrexResult {
    fn from_redis_value(v: Value) -> Result<Self, ParsingError> {
        let [a, b] = <[Value; 2]>::from_redis_value(v)?;
        match a {
            Value::Int(_) => Ok(Self::Int   { value: i64::from_redis_value_ref(&a)?,
                                              actual_increment: i64::from_redis_value_ref(&b)? }),
            _             => Ok(Self::Float { value: f64::from_redis_value_ref(&a)?,
                                              actual_increment: f64::from_redis_value_ref(&b)? }),
        }
    }
}
```

Then the signature will change to:
```
fn increx<K: ToSingleRedisArg, V: ToSingleRedisArg>(key: K, increment: V, options: IncrexOptions<V>) -> (IncrexResult)
```

But the call sites will need to be matched:
```
match con.increx("counter", 5, IncrexOptions::default().saturate().upper_bound(100))? {
    IncrexResult::Int { value, actual_increment } => { /* ... */ }
    IncrexResult::Float { .. } => unreachable!(),
}
```
+an _ arm, because the `enum` will have to be marked as `non_exhaustive`

somechris: I'd lazily parse to the target type.

So, I'd roll with a simple:

```
pub struct IncrexResult {
  value: Value,
  increment: Value
}
```

and to get the values out, offer dedicated functions that convert to the needed types.

That's simple, gets the job done, and does not need hard typing / `enum`s / etc

StefanPalashev: I tried to apply the suggested changes. The dedicated functions look like this:
``` rust
impl IncrexResult {
    pub fn value_as<T: FromRedisValue>(&self) -> Result<T, ParsingError> {
        T::from_redis_value_ref(&self.value)
    }
    pub fn actual_increment_as<T: FromRedisValue>(&self) -> Result<T, ParsingError> {
        T::from_redis_value_ref(&self.actual_increment)
    }
}
```

While it eliminates the matching from the previous `enum-based` approach, this one has its own implications. Nothing stops an user to try and parse a `BYFLOAT` operation's result into an integer. Furthermore, there is an inconsistency between RESP2 and RESP3. In RESP3, `Value::Double(2.5)` can be parsed as an `i64`, producing `2` via truncation. In RESP2, the equivalent value would be represented as a `BulkString` and attempting to parse "2.5" as an `i64` would instead return an error (`Err ("2.5" isn't an integer)`).

This approach seems clean, but it isn't as restrictive as the ones before. The `enum` one prevents such scenarios with the cost of a more complex response parsing and the initial one with separate functions for each of the operations seemed cleanest, but it differed from the original `incr` command's behavior.

StefanPalashev: Should I commit the recent changes or should we discuss them first?

somechris: > Nothing stops an user to try and parse a `BYFLOAT` operation's result into an integer.

*cough* You make it sound like a bad thing.

I think that it's a good thing :grin: Check the term [guard digit](https://en.wikipedia.org/wiki/Guard_digit). It's a standard technique in numerical computations.

But if you want to prevent that ... that's easy to forbid :wink:  something along these lines should do

```
pub fn as_f64(self) -> (f64, f64) { ... }
pub fn as_i64(self) -> (i64, i64) { ... }
```

But is that easier/better for users?



> Furthermore, there is an inconsistency between RESP2 and RESP3.

Since the conversion is _not_ automatic and you're already in a
function, you can simply code that inconsistency away.

To me, it's even the job of `redis-rs` to smooth such inconsistencies out, where possible.


> This approach seems clean, but it isn't as restrictive as the ones before.

The approach is as restrictive as you choose to make it.

You used a lenient implementation, so the result is lenient. That's ok.
If you think it should be stricter, implement it stricter.

But check a few use cases, and document your decisions.

That allows to understand what/why and it makes it clear what to challenge and why.

nihohit:
Thanks!

Issue #2215, opened 2026-07-15
RedisServer cleanup & builder, TestContext builder


somechris:
In #2205, it was said that we should

1. Clean-up `RedisServer`'s creation function
1. Add a builder for `RedisServer` to get the creation function explosion under control
1. Add a builder for `TestContext` for the same reason.
1. If versioning requirements are not in the way, move `TestContext` into `redis-test`to make it available to others.

PR #2220, opened 2026-07-16
redis-test/RedisServer: Dedupe server creation (1/19)

somechris:
Similar `RedisServer` instances are created in all arches of the `addr` match. Most parameters are duplicated. And only `redis_cmd` massaging is crucial to the branches, not creating/spawning the server.

So we extract the actual server creation. This immediately dedupes. And it will allow simplification of different, duplicated `spawner`s in follow-up commits.

This is the first part towards #2215


somechris:
If (for better context) you want to take a peek at the full branch, it's at https://github.com/somechris/redis-rs/pull/22/changes

nihohit:
1/19 😱 🤣

nihohit (approved):

PR #2221, opened 2026-07-16
redis-test: Add dedicated struct for a Redis server's start commands (2/19)

somechris:
We want to bring `RedisServer::new_with_addr_tls_modules_and_spawner`'s
complexity down and make it easier to read and use.

The main ingredients will be:
* Moving `REDISRS_SERVER_BIN` handling into a dedicated function, so it
  can be re-used and need not be passed to `get_major_version`
* Moving module loading logic into a dedicated function.
* Collect configuration directives and their values in a single
  line. This would turn the current, harder to read:

  ```
  cmnd.arg("--foo")
      .arg("value-foo")
      .arg("--bar")
      .arg("value-bar")
      .arg("--baz")
      .arg("value-baz")
  ```

  into the way simpler

  ```
  cmnd.arg2("--foo", "value-foo")
      .arg2("--bar", "value-bar")
      .arg2("--baz", "value-baz")
  ```

* Harmonize the error handling of a failed spawning. Some callers just
  let things crash, others give a proper error message. Regardless: It's
  needlessly duping logic and should get collected in one place for
  easier maintenance.

That could all be done directly within `RedisServer` and would improve
the overall situation a bit.

But the first three items are are all concerned with building the actual
command line to start the server. So we refactor that part out into a
dedicated `RedisServerCommand` struct. This moves the logic out of
`RedisServer` (rendering `RedisServer` simpler), allows for easy
customization, and gives clear responsibilities.

To keep the diffs managable and reviewable, we tackle the first and last
item (they are the ones that directly affect callers and give a compact
diff) in this commit and leave items 2 and 3 for separate follow-up
commits.

### DIFF
--- a/redis-test/src/cluster.rs
+++ b/redis-test/src/cluster.rs
@@ -171,7 +171,7 @@ impl RedisCluster {
         let max_attempts = 5;
 
         let mut make_server = |port| {
-            RedisServer::new_with_addr_tls_modules_and_spawner(
+            RedisServer::new_with_addr_tls_modules_and_cmd_refiner(
                 ClusterType::build_addr(port),
                 None,
                 tls_paths.clone(),
@@ -210,7 +210,6 @@ impl RedisCluster {
                     }
                     cmd.current_dir(tempdir.path());
                     folders.push(tempdir);
-                    cmd.spawn().unwrap()
                 },
             )
         };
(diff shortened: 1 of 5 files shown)


nihohit:
maybe I'm missing something, but it seems to me like this change just replaces a `process::Command` value with a wrapper, and moves the call to `spawn` from inside the closure to after calling the closure.

We can move the spawn without changing the type, so what's the benefit of the new type?

somechris:
> [...], so what's the benefit of the new type?

Fair point. Maybe the overall goal and strategy was not formulated clearly enough.

I've updated the commit message and PR description.

Does it look better now?

nihohit (approved):

PR #2222, opened 2026-07-16
Makefile: Switch to `--locked` for module tests on RESP3

somechris:
The json module tests ran without `--locked` on RESP3. When adding bloom module tests, this got copy/pasted there.

To limit it creeping further, we add `--locked` to these calls.


somechris:
Just to avoid doubt: This is not in the `RedisServer` refactoring series :grin: , but outside of it against `main`

nihohit (approved):

Issue #2201, opened 2026-07-10
Integrating JSON commands directly into `(Async|)(Typed|)Commands`


somechris:
While most commands get added to `(Async|)(Typed|)Commands` in `commands/mod.rs`, the JSON commands live in `commands/json.rs`.

The setup of JSON commands is very similar to the other commands, but different in subtle ways (e.g.: adding to `Pipeline` and `ClusterPipeline` returns a `RedisResult<&mut Self>` instead of `&mut Self`).

This duplicates much of the logic in `commands/macros.rs` with minor variations.

Hence, JSON commands stick out and seem a bit left behind (E.g.: they only come in untyped variants).

Is it worth to move them into the general `implement_commands!` of `commands/mod.rs`?

I'd be all for it. But it means making the macros of `commands/macros.rs` a bit more complex, as they need to be able to switch between commands that can fail on their own (e.g.: Serialization within `json_arr_append`) and commands that can not (e.g.: those of `commands/mod.rs`)

It should be doable. But functions that can fail on their own need another annotation (e.g.: `Fallible`) to mark them to get different return types.

That would move JSON commands to first class commands and would allow to dedupe much of the macro logic.
Also, it would allow to strip some unneeded `RedisResult` (e.g.: Adding `json_type` to a Pipeline)

WDYT?

nihohit:
Will this be a breaking change, or is it just a matter of moving code around?

In general, I don't see a problem with the suggestion, but the devil is in the details. Feel free to open a PR, and we'll see how this looks in actuality.

somechris:
> Will this be a breaking change, or is it just a matter of moving code around?

As `JsonCommands` and `JsonAsyncCommands` traits would get removed, it's a breaking change.

The commands themselves _can_ be identical.

But I'd prefer to simplify them where possible.

E.g.: When adding to a `Pipeline`.

Normal commands (eg. `set` and `get`) on a `Pipeline` return `&mut Pipeline`.
All JSON commands (eg. `json_set` and `json_get`) currently return `RedisResult<&mut Pipeline>`.

For `json_set`, that's warranted as the internal serialization may fail.

But for `json_get`, it's unwarranted as nothing can fail there. It returns `RedisResult<&mut Pipeline>` simply because the macro cannot differentiate between commands that may fail upon creation and those that cannot.

So I'd suggest to switch the return type for `json_get` on `Pipeline`s from `RedisResult<&mut Pipeline>` to `&mut Pipeline`.

[...]

PR #2228, opened 2026-07-18
Merge `JsonCommands` into `Commands` (1/2)

somechris:
By merging the commands of `JsonCommands` into `Commands` (and accordingly for the async, ... variants), we finally can treat all commands alike and have them in one spot.

This will help with maintenance and keep things alike.

The commands are brought over without modifications (except adding `cfg` guards and a `Generic fallible`) return type to allow easy checking that everything got brought over. Tayloring return types and harmonizing documentation is left for follow-up commits.

The JSON commands are now also available on the typed command implementations (e.g.: `TypedCommands`), but still with generic return types.

This is the first step towards #2201.


redis/src/commands/mod.rs
```diff
@@ -2952,6 +2942,174 @@ assert_eq!(invok_2_res, 5);
         cmd("FLUSHDB").arg(options).take()
     }
 
+    // JSON commands
+
+    /// Append the JSON `value` to the array at `path` after the last element in it.
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    fn json_arr_append<K: ToSingleRedisArg, P: ToSingleRedisArg, V: Serialize>(key: K, path: P, value: &'a V) -> Generic fallible {
```
somechris: Is `fallible` a "good enough" name for the flag that this function may fail?

And where would be a good place to put it?

Putting it at the end (near the affected return type) looked best for me.

But maybe immediately before the `fn` (more in the spirit of reading "fallible function", like in proper English)

```rust
fallible fn json_arr_append<K: ToSingleRedisArg, P: ToSingleRedisArg, V: Serialize>(key: K, path: P, value: &'a V) -> Generic {
```

Maybe immediately after the `fn` (more in the spirit of reading "function that is fallible")

```rust
fn fallible json_arr_append<K: ToSingleRedisArg, P: ToSingleRedisArg, V: Serialize>(key: K, path: P, value: &'a V) -> Generic {
```

redis/src/commands/macros.rs
```diff
@@ -1,27 +1,46 @@
...
     };
 
-    // If return type is specified in the input skeleton, then we will return it in the generated function (note match rule `$rettype:ty`)
+    // If return type is specified in the input skeleton, then we will return it in the generated function (note match rule `$rettype`)
     (
         $lifetime: lifetime
         $(#[$attr:meta])+
         fn $name:ident<$($tyargs:ident : $ty:ident),*>(
-            $($argname:ident: $argty:ty),*) $body:block $rettype:ty
+            $($argname:ident: $argty:ty),*) $body:block $rettype:tt $($fallible:ident)?
```
somechris: `implement_command_async` does not actually need/use `$fallible` at all. It could be removed.

 I added it nonetheless to have the same signature in both `implement_command_sync` and `implement_command_async`.

Is it better to remove the unused `$fallible` or is it better to keep it to keep the macros in sync?

redis/src/commands/macros.rs
```diff
@@ -69,7 +88,7 @@ macro_rules! implement_command_sync {
         ) -> RedisResult<$rettype>
 
         {
-            Cmd::$name($($argname),*).query(self)
+            choose_tt!($($fallible,)? (Cmd::$name($($argname),*)?), (Cmd::$name($($argname),*))).query(self)
```
somechris: Some parts directly use `$body` (e.g.: `implement_command_async`), some parts instead use the wrapper `Cmd::$name` (e.g.: `implement_command_sync`).

To not change too many things in one PR, I now simply extended what was there.

But is there a reason for not using the same for both? If not, should that get harmonized in a follow-up PR? If so, do you prefer `$body` or `Cmd::$name`?

redis/src/commands/mod.rs
StefanPalashev: Can the return type be made `Fallible<Generic>`?

somechris: That is a great idea! Thank you @StefanPalashev. Let's not have a flag, let's use the return type as a flag!

But if so, we could use `RedisResult<Generic>` right away.

Done in 5e26c38

redis/src/commands/macros.rs
somechris: That aspect changed with switching to `RedisResult<...>`.

Hence, resolving.

nihohit:
I do like the end result, but having had quite a frustrating time trying to untangle the commands generating macros in the past, I'm not that eager to make those macros even more complicated - not to mention that this might slow down compilation time, too. Is there maybe another way we can achieve this?

somechris:
> I do like the end result, but having had quite a frustrating time trying to untangle the commands generating macros in the past, I'm not that eager to make those macros even more complicated - not to mention that this might slow down compilation time, too. Is there maybe another way we can achieve this?

Yup. These macros are frustrating :-(

Some time back I played around with the macros in various directions. For me, the key hurdle for simplifications turned out to be adding the `RV: FromRedisValue` only when needed. That would help simplifying a few things, but for that one would have to strip the RV from the `$tyargs:$ty`, if it's there.

And either I use `ident` or `ty` there and lose inspectability (hence, it does not work), or I use `tt` and it gets even more complex and I run into issues with execution orders of nested macros. So it was not worth it.

But you macro-fu is probably better than mine :shrug:

IIRC one could do, is remove a few `choose_if_redisresult` if one used `$body` instead of `Cmd::$name`. But I did not want to switch that in addition to what he PR already changed.

We could generate the code outside of Rust. That would make things simpler as one would not be bound to the macro limitations. But that'd be a big change. Not sure that is worth it.

---

I tried to estimate the effect the change has using `cargo expand` between before/after the macro.rs changes.
That seemed somewhat reasonable as the produced output is the same (with the exception of parantheses). And as they are the same, full compilation would only add more noise.

I could not measure any performance changes. Sometimes that was a few ms faster, sometimes the other.

How do you suggest to measure the difference between before/after the macro changes?

nihohit:
I don't think that the performance of the generated code would be different, because as you said, it's essentially the same generated code. The perf difference was in time to compile - whether the added macro depth affected the time needed to compile the code including generation time, without optimizations (because I assume optimizations will take longer than anything else).

nihohit:
and the `time` linux command is probably good enough for a rough check - if the difference is within noise and isn't consistent to either time, then the compilation time is roughly the same.

somechris:
That sounds like what I did with the `cargo expand` mentioned above.

But I've now redone it with `cargo build`. But of course, the noise is a bit bigger there.

I took two commits:

1. The last commit before the `macro.rs` change (i.e. 4f625a4). I'll call that `before`.
1. The commit that does the `macro.rs` change (i.e. 5e26c38). I'll call that `after_`.

The attached [`macro-comparison.sh`](https://github.com/user-attachments/files/30393898/macro-comparison.sh):
* switches to `before`,
* adds a random function (to force clean re-compilation), and
* runs and records the time of `cargo build`. Then it
* switches to `after_`,
* adds a random function (to force clean re-compilation), and
* runs and records the time of `cargo build`.

It does 2 such rounds silently to warm things up.

Then 5 rounds where it reports the timing.

The difference between `before` and `after_` is lost in the jitter to me.

Here's a run that's overall a bit favourable for `after_`

(`before`: 16,37s avg; `after_`: 16,18s avg):

```
~/macro-comparison.sh
warming up
-- Start --
run 1 / before / 16.34user 1.94system 0:15.94elapsed 114%CPU (0avgtext+0avgdata 947080maxresident)k
run 1 / after_ / 16.23user 2.03system 0:15.02elapsed 121%CPU (0avgtext+0avgdata 966888maxresident)k
run 2 / before / 17.03user 2.08system 0:15.98elapsed 119%CPU (0avgtext+0avgdata 957248maxresident)k
run 2 / after_ / 16.06user 2.08system 0:15.27elapsed 118%CPU (0avgtext+0avgdata 960060maxresident)k
run 3 / before / 16.34user 2.11system 0:15.40elapsed 119%CPU (0avgtext+0avgdata 988192maxresident)k
run 3 / after_ / 16.41user 2.11system 0:15.66elapsed 118%CPU (0avgtext+0avgdata 964920maxresident)k
run 4 / before / 16.28user 2.24system 0:15.73elapsed 117%CPU (0avgtext+0avgdata 956336maxresident)k
run 4 / after_ / 15.80user 2.11system 0:15.24elapsed 117%CPU (0avgtext+0avgdata 970024maxresident)k
run 5 / before / 15.86user 2.08system 0:14.36elapsed 124%CPU (0avgtext+0avgdata 959492maxresident)k
run 5 / after_ / 16.41user 2.15system 0:15.48elapsed 119%CPU (0avgtext+0avgdata 939084maxresident)k
```

But I ran it a few times. Sometimes `before` is on avg a bit quicker. Sometimes `after_`.

somechris:
@nihohit: What's the situation with these typed JSON commands?

For me, the decision on whether the approach of this series is roughly ok, is still missing.

Meanwhile upstream got new json commands, and the command macros themselves changed a lot. So I'd probably have to redo most of it.

So I wondered whether it's better to just abandon it and leave things as they are, or if re-doing it is worth it and it has a chance to land?

nihohit:
after wrestling with the commands macros recently, I think that this solution is ok, and we can move forward with this. I do believe that rebasing will be a lot of work, as you mentioned, so please don't feel obligated to do this. This is still under the "nice to have" umbrella :)

redis/src/commands/macros.rs
somechris: As the changes that happened in the meantime to the macros made it hard to update, and this `$body`/`Cmd::$name` change would help, it gets addressed in #2383.

somechris:
Converting to draft to avoid accidental merges.

(This will get closed once the corresponding commit of the rebased branch gets uploaded)

somechris:
New approach is in #2397.

PR #2229, opened 2026-07-18
Refine return types of JSON commands (2/2)

### LINKED ISSUE #2201: Integrating JSON commands directly into `(Async|)(Typed|)Commands`

somechris:
When moving JSON commands into `Commands`, we moved them using generic
types to make diffs simple. We now refine their return types to simplify
their use.

To properly test them, we switch the tests to the typed interfaces and
test both $-paths and .-paths, as their return types differ.

Fixes #2201


redis/src/commands/mod.rs
```diff
@@ -2952,6 +2949,298 @@ assert_eq!(invok_2_res, 5);
...
+    /// with e.g. `serde_json` you have to use `Vec<T>` for your output type instead of `T`.
+    ///
+    /// ```text
+    /// JSON.GET <key> <path>
+    /// ```
+    ///
+    /// [Redis Docs](https://redis.io/commands/JSON.GET)
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    fn json_get<K: ToSingleRedisArg, P: ToRedisArgs>(key: K, path: P) -> (String) {
```
somechris, line 3156: Not sure about the return type here.

`JSON.GET` basically gives `BulkString`s that contain serialized JSON.

For users, that's probably not what they would expect?

Maybe a `serde_json::Value` would be better?

But that would mean having to parse each value, which maybe not ideal in all cases either?

(There's of course always the untyped `Commands` where one can specify the type)

What would be a good type?

redis/src/commands/mod.rs
```diff
@@ -2952,6 +2949,298 @@ assert_eq!(invok_2_res, 5);
...
+    /// Appends the `json-string` values to the string at `path`.
+    ///
+    /// ```text
+    /// JSON.STRAPPEND <key> <path> <value>
+    /// ```
+    ///
+    /// [Redis Docs](https://redis.io/commands/JSON.STRAPPEND)
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    fn json_str_append<K: ToSingleRedisArg, P: ToSingleRedisArg, V: ToSingleRedisArg>(key: K, path: P, value: V) -> (SingletonFallbackVec<Option<usize>>) {
```
somechris: `JSON.STRAPPEND` expects `value` to be a serialized JSON string. So something like `"foo"` ... with quotes.

`redis-rs` has always used `ToSingleRedisArg` here, while for example in `json_set` or `json_arr_append` we have `Serialize`.

Is it worth changing the type here?

Would you prefer for example to see `Serialize` (or somethignelse) here?

redis/src/types.rs
```diff
@@ -1837,18 +1837,18 @@ impl FromRedisValue for bool {
...
-                } else if &s[..] == "0" {
+                } else if &s[..] == "0" || &s[..] == "false" {
                     Ok(false)
                 } else {
                     crate::errors::invalid_type_error!(v, "Response status not valid boolean");
                 }
             }
             Value::BulkString(ref bytes) => {
-                if bytes == b"1" {
+                if bytes == b"1" || bytes == b"true" {
```
somechris, line 1849: `JSON.TOGGLE` reports the toggled boolean as `BulkString("true")`/`BulkString("false")` for .-paths.

A custom type could parse that. But then I saw that we already parse `BulkString("1")`/`BulkString("0")` (strings, not numbers) to proper `bool`s. So I added the strings here and also for `SimpleString` to make things match up.

But I'm somewhat hesitant to change general value conversion, as it might back-fire for users in unexpected (and hard to check) ways.

But then ... if we already parse number strings, adding a custom type to additionally parse the `true`/`false` strings looks odd?

WDYT?

redis/src/commands/json.rs
```diff
@@ -1,304 +1,351 @@
...
-            )*
-        }
+//! Types for the JSON module
+//!
+//! Some commands in the JSON module return either `T` or `Vec<T>` depending on the path argument.
+//! [`SingletonFallbackVec`] and [`SingletonFirstVec`] abstract this difference away, treating `T`
+//! as singleton `Vec<T>`:
+//!
+//! * [`SingletonFallbackVec`] first tries to parse as `Vec<T>` and falls back to parsing as `T`.
+//! * [`SingletonFirstVec`] first tries to parse as `T` and falls back to parsing as `Vec<T>`.
```
somechris: I rather dislike the names `SingletonFallbackVec` and `SingletonFirstVec`.

They roughly describe what's going on ... but *wurgh* 🤮


Any good naming suggestions?

redis/src/commands/mod.rs
```diff
@@ -2952,6 +2949,298 @@ assert_eq!(invok_2_res, 5);
...
+    /// Reports the type of JSON value at `path`.
+    ///
+    /// ```text
+    /// JSON.TYPE <key> <path>
+    /// ```
+    ///
+    /// [Redis Docs](https://redis.io/commands/JSON.TYPE)
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    fn json_type<K: ToSingleRedisArg, P: ToSingleRedisArg>(key: K, path: P) -> (SingletonFirstVec<Vec<String>>) {
```
somechris: The main return type is `String` here.

I wanted to use an enum here.

But I could not find anything matching in `serde_json`.

So I thought about adding a custom enum here. But then ... if `serde_json` does not provide something ... should we?

And even if ... Redis does not return JSON's standard value productions. Instead of `true` `false` literals it has `boolean`, which is not found on https://json.org. And in addition to `number`, it also has `integer`, which https://json.org only uses to produce `number`, but not as direct type. So I'm not sure.

Should we add a type for it?

If so, `JsonType` is probably a misnomer then ... so ... `RedisJsonType`?

WDYT?

The other thing is that this type effectively comes down to `Vec<Vec<String>>`. But the outer `Vec` is observed only in RESP3 with $-paths:

```
redis-cli -3
127.0.0.1:6379> json.type foo $..a
1) 1) "boolean"
   2) "string"
   3) "integer"
```

In my testing, the outer `Vec` was always empty or a singleton. But then, maybe I just suck at producing good examples.

Judging by the [module's code](https://github.com/RedisJSON/RedisJSON/blob/0459030310f67fdeb5fd9f1113dd69170609a05d/redis_json/src/commands.rs#L1261), the double `Array` looks artificial and I'm tempted to parse it away.

But then, the [`JSON.TYPE` documentation for RESP3](https://redis.io/docs/latest/commands/json.type/#return-information) (the link goes to `RESP2`, switch to `RESP3`) says:

> Array reply of array replies of bulk string replies,

So parsing it to a `Vec<String>` might be more usable but unexpected.

WDYT?

redis/src/commands/json.rs
StefanPalashev: How about `SingletonOrVec<T>` and `VecOrSingleton<T>`?

somechris: :+1:

Done in be0d8ce

somechris:
Converting to draft to avoid accidental merges.

(This will get closed once the corresponding commit of the rebased branch gets uploaded)

somechris:
This got updated and rebased as #2413.

PR #2233, opened 2026-07-19
redis-test/Sentinel: Streamline TLS aspects (8/19)

somechris:
Setting up sentinels unconditionally created certificates, regardless of whether they later got used. This costed time and entropy. And the certificates got passed along unconditionally, which led to suspicious `RedisServer` instantiations.

Hence, we switch to only generating certificates if they are needed.

Furthermore, each of the sentinel helper functions determined whether or not TLS is used by building a default address, which relies on hidden conventions. We reduced this reliance from three places to a single, central place in the main function and pass TLS status to helper functions by whether or not certificates are available.


nihohit (approved):
thanks!

PR #2235, opened 2026-07-19
redis-test/RedisServer: Guard against unused settings (10/19)

somechris:
Instead of silently ignoring unused settings, we panic to alert tests about misconfigurations.

### DIFF
--- a/redis-test/src/server.rs
+++ b/redis-test/src/server.rs
@@ -165,6 +165,24 @@ impl RedisServer {
         modules: &[Module],
         cmd_refiner: impl FnOnce(&mut RedisServerCommand),
     ) -> RedisServer {
+        // Guard against unsupported settings
+        if tls_paths.is_some() && !matches!(addr, ConnectionAddr::TcpTls { .. }) {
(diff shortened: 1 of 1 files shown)


redis-test/src/server.rs
```diff
@@ -165,6 +165,24 @@ impl RedisServer {
         modules: &[Module],
         cmd_refiner: impl FnOnce(&mut RedisServerCommand),
     ) -> RedisServer {
+        // Guard against unsupported settings
```
somechris, line 168: This part duplicates the `!matches!(addr, ConnectionAddr::TcpTls { .. })` condition a bit.

We could de-dupe to something like:


```
if (!matches!(addr, ConnectionAddr::TcpTls { .. })) {
    if tls_paths.is_some() {
      panic!("...");
    }

    if mtls_enabled.is_some() {
        panic!("...");
    }

    if cert_auth_field.is_some() {
        panic!("...");
    }
}
```

But I find that harder to read for such early exits.
So I went with the current formulation.

Which do you prefer?

nihohit (approved):

PR #2236, opened 2026-07-19
redis-test: Add `RedisServerBuilder` (11/19)

somechris:
`RedisServerBuilder` replaces the `RedisServer::new...` methods, whose number of arguments grew over time and made it unwieldingly.

### DIFF
--- a/redis-test/src/server.rs
+++ b/redis-test/src/server.rs
@@ -30,10 +30,117 @@ enum ServerType {
 }
 
 /// Represents a module that can be loaded into the Redis server.
+#[derive(Clone)]
 #[non_exhaustive]
 pub enum Module {
     Bloom,
     Json,
+}
+
+/// A builder for [`RedisServer`]
+///
+/// # Example
+///
+/// ```rust,no_run
+/// use redis_test::server::{RedisServerBuilder, Module};
+///
+/// let server = RedisServerBuilder::new().module(Module::Json).build();
+/// let info = server.connection_info();
+/// // Connect to the server using `info`...
+/// ```
+// Note that this builder is an owned-builder as we want to build in a single chain anyway and do
+// not have to build multiple instances from the same builder. Also, this spares us cloning
+// considerations.
+#[derive(Default)]
+pub struct RedisServerBuilder {
+    address: Option<ConnectionAddr>,
+    config_file: Option<PathBuf>,
+    cert_auth_field: Option<String>,
+    modules: Vec<Module>,
+    mtls: bool,
+    tls_paths: Option<TlsFilePaths>,
+}
+
+impl RedisServerBuilder {
+    /// Starts a fresh builder
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn address(mut self, address: ConnectionAddr) -> Self {
+        self.address = Some(address);
+        self
+    }
+
+    pub fn config(mut self, config_file: PathBuf) -> Self {
+        self.config_file = Some(config_file);
+        self
+    }
+
+    pub fn cert_auth_field(mut self, cert_auth_field: impl Into<String>) -> Self {
+        self.cert_auth_field = Some(cert_auth_field.into());
+        self
+    }
+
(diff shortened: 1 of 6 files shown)


redis-test/src/server.rs
```diff
@@ -30,22 +30,142 @@ enum ServerType {
...
+        self.modules = modules.to_vec();
+        self
+    }
+
+    pub fn mtls(mut self, enable_mtls: bool) -> Self {
+        self.mtls = enable_mtls;
+        self
+    }
+
+    pub fn tls_paths(mut self, tls_paths: TlsFilePaths) -> Self {
```
somechris, line 104: The struct is called `TlsFilePaths`, but we hardly use `tls_file_paths` as parameter. Rather `tls_paths` (the most), `tlspaths`, and `tls_file_params`.

So I took `tls_paths` as this is what we use the most.

redis-test/src/server.rs
```diff
@@ -30,22 +30,142 @@ enum ServerType {
...
+///
+/// ```rust,no_run
+/// use redis_test::server::{RedisServerBuilder, Module};
+///
+/// let server = RedisServerBuilder::new().module(Module::Json).build();
+/// let info = server.connection_info();
+/// // Connect to the server using `info`...
+/// ```
+// Note that this builder is an owned-builder as we want to build in a single chain anyway and do
+// not have to build multiple instances from the same builder. Also, this spares us cloning
```
nihohit, line 52: is there a concrete issue with deriving clone here, or are you just reducing the interface for maintainability's sake?

redis-test/src/server.rs
```diff
@@ -30,22 +30,142 @@ enum ServerType {
...
+        self.mtls = enable_mtls;
+        self
+    }
+
+    pub fn tls_paths(mut self, tls_paths: TlsFilePaths) -> Self {
+        self.tls_paths = Some(tls_paths);
+        self
+    }
+
+    pub fn tls_paths_opt(mut self, opt_tls_paths: Option<TlsFilePaths>) -> Self {
```
nihohit, line 109: is this something we'll want to keep, or is it added just now for backwards compatibility?

redis-test/src/server.rs
somechris: No issue with deriving `Clone`.

To me, Rust seems to more like setters to return `&mut Self` for builders.

That's a bit nicer to use when using them with `if`s etc. But that means that we either have to clone during building or do some stunts to get to the fields' values (e.g.: `Cmd::take` :grin:)

With `RedisServer` we hardly use `if`s though. We want to build in a one-liner anyway. Then the benefit of setters returning `&mut Self` vanishes. And we're left with unneeded cloning. Hence, the setters here return `Self` and spare the clone during building.


Do you suggest to:
* Remove or rephrase the whole comment,
* Remove the `cloning` sentence, or
* Switch to a builder whose setters yield `&mut Self`?

redis-test/src/server.rs
somechris: You mean the name or the funcion as a whole?

I would keep the function.

In our environment, the same test code is run with `tcp` and `tcp+tls`. So we have code that sometimes has to set certificates. Allowing to pass them as `Option` makes the calling code simpler.

Also it's a common pattern for builders.

And for naming ... I was torn between `maybe_...` and `..._option`.

A not too thorough grepping gave:
* 2x `maybe_`
* 2x `_opt`
* 1x `_option`

So the suffix won 3 to 2, with `_opt` being most used ... so I went with `_opt`.

But I'm fine either way.

(My first choice would be `maybe_`, as that reads best for me and I find it very intuitive. Also, it somewhat sticks out (so it's hard to overlook), but not enough to get in the way. And IDEs these days are smart enough to show it for auto-completion when you start typing `tls`.)

redis-test/src/server.rs
nihohit: no need to change anything, I just wanted to understand the background of the comment.

redis-test/src/server.rs
nihohit: I meant the function itself. the name is fine.

PR #2245, opened 2026-07-23
RedisTest: Mark all types as non exhaustive.

nihohit:
Use a builder for the cluster's configuration, and all other types are marked with non_exhaustive.


nihohit:
@somechris FYI

somechris:
> @somechris FYI

Nice! Thanks!

PR #2246, opened 2026-07-24
pipeline: pre-allocate vector capacity in encode_pipeline function

Aditya-9-6:
### Summary
This PR optimizes `encode_pipeline` in `redis/src/pipeline.rs` by pre-allocating the byte buffer capacity using the total sum of command lengths before serializing packed Redis pipeline requests.

### Rationale & Performance Benefit
Previously, `encode_pipeline` initialized an empty `vec![]` and deferred buffer reservation until inside `write_pipeline`.

Pre-allocating `Vec::with_capacity(cmds_len + extra)` upfront:
- Prevents initial buffer reallocation overhead during RESP command encoding.
- Reduces memory fragmentation and improves pipeline encoding throughput.

### DIFF
--- a/redis/src/pipeline.rs
+++ b/redis/src/pipeline.rs
@@ -259,7 +259,9 @@ impl Pipeline {
 }
 
 fn encode_pipeline(cmds: &[Cmd], atomic: bool) -> Vec<u8> {
-    let mut rv = vec![];
+    let cmds_len: usize = cmds.iter().map(cmd_len).sum();
+    let extra = if atomic { 32 } else { 0 };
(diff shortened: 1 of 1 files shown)


nihohit:
how does pre-allocating differ from reserving on an empty vec? as you mention, `write_pipeline` already reserves before writing, so I don't see the benefit here.

Aditya-9-6:
Thanks for pointing that out @nihohit! You are completely right — since `write_pipeline` already calls `rv.reserve(...)` at the start of its execution, pre-calculating the sum in `encode_pipeline` creates redundant iteration over `cmds`.

I will close this PR to keep the existing clean flow. Appreciate the sharp code review!

PR #2247, opened 2026-07-24
docs: Add the need for `non_exhaustive` for `enum`s and `struct`s

somechris:


CODING_CONVENTIONS.md
```diff
@@ -1,5 +1,11 @@
 # Coding Conventions
 
+## General
+
+* All `struct`s with public members and `enum`s have to be marked as `#[non_exhaustive]`.
```
somechris: Should `struct`s without public members (only non-public members) also be `non_exhaustive`? ... Maybe?

`enum` _variants_ are currently not marked `non_exhaustive` ... I guess that should stay that way as they get changed seldom enough, right?

nihohit (approved):
Thanks!

CODING_CONVENTIONS.md
nihohit: > Should structs without public members (only non-public members) also be non_exhaustive?

AFAIK no - `non_exhaustive` essentially behaves as if a struct has private fields, and can't be constructed / deconstructed (without elipsis) because of that. I recall seeing documentation stating that explicitly, but couldn't find it now.

> enum variants are currently not marked non_exhaustive

That is an issue, actually. A struct-like variant is essentially an inner struct, and so it should apply the rule, too.

```rs
#[non_exhaustive]
pub enum Bar {
   Foo { a: A }
}
```
is essentially the same as
```
pub struct Foo {
  pub a: A;
}

#[non_exhaustive]
pub enum Bar {
  Foo(pub Foo)
}
```
only in the second case it's clearer that the struct is missing the `non_exhaustive`.

somechris: > > Should structs without public members (only non-public members) also be non_exhaustive?
> 
> AFAIK no - [...]

I vaguely remembered that adding `#[non_exhaustive]` to a struct with only private members triggered `semver-checks`. But I now tried it out and it does not. :+1:



> > enum variants are currently not marked non_exhaustive
> 
> That is an issue, actually.

Adapted wording to include that.

We currently follow that anywhere ... but it at least sets the direction :+1:

PR #2250, opened 2026-07-25
tests: Migrate `TestContext::with_modules` onto `TestContextBuilder` (15/19)

somechris:
This makes `TestContext::with_modules_and_tls` unused, and we hence drop it.


nihohit (approved):
thanks

Issue #2254, opened 2026-07-26
Speed up TLS tests


somechris:
Running TLS tests locally takes long. It got frustating to the point where I did not run TLS tests for each commit.

This is not ideal, so we should speed up TLS tests and make them less annoying.

---

Looking where the time is spent, it's clearly TLS key generation (certs etc are of course cheap).
We generate new keys for each test. This sucks entropy dry and takes time.

As the keys are raw numbers without cert information, we need not use
fresh numbers for each test. If we'd allow to optionally re-use them,
this should bring numbers down.

---

I've done a proof-of-concept implementation that checks if the environment variables
* `REDISRS_TLS_KEY_CA`
* `REDISRS_TLS_KEY_SERVER`
* `REDISRS_TLS_KEY_CLIENT`

are set. If they are set, the code loads and uses the key material
from that file instead of generating new keys.

This brought rustls tests with all features in CI down from ~550s to ~140s (~75% faster).

That's more in the range of non-TLS tests, which are ~120s and managable.

`build-and-test` time reduced from ~25min to ~12min.

nihohit:
Do you think that the tests should try to create the certificates, or just assume that the certificates were created before the tests ran?
Maybe save the certs in a constant location (/tmp/<date>/certs, or something like it), and then try to create them on test beginning if the files are missing?

somechris:
I'd say to keep the current behaviour as default. That will work for most people without extra steps. And it does not break any existing setup.

For CI, I'd suggest to [create the certs before-hand and re-use](https://github.com/somechris/redis-rs/commit/4370bef9dcec9135366fb591bb5f7846132e907b)

For developers that run tests more often, I'd suggest to create the keys before-hand and re-use them. Just as in the linked commit above. But I would not want to mess with developers' setups. So I would leave it to them, just as we do with JSON or bloom modules.

somechris:
Last commit in the branch got merged. Hence closing. :+1:

nihohit:
Thanks for the dedicated work :)

PR #2255, opened 2026-07-26
redis-test: Add `OpensslCommand` (1/5)

somechris:
We started OpenSSL in 7 places. They differed (seemingly accidentally) in if they checked exit status and how useful error messages were.

To get rid of those accidental differences and improve the signal-to-noise ratio in the code that tries to use OpenSSL, we abstract all the boiler-plate into `OpensslCommand`.

This simplifies and harmonizes OpenSSL use.

This commit is part of #2254.

### DIFF
--- a/redis-test/src/utils.rs
+++ b/redis-test/src/utils.rs
@@ -1,8 +1,10 @@
+use socket2::{Domain, Socket, Type};
+use std::ffi::OsStr;
+use std::io::Write;
 use std::net::{SocketAddr, TcpListener};
 use std::path::PathBuf;
+use std::process::{Command, Output};
 use std::{fs, process};
-
-use socket2::{Domain, Socket, Type};
 use tempfile::TempDir;
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


redis-test/src/utils.rs
```diff
@@ -21,6 +23,71 @@ pub struct ClientCertPaths {
...
+    pub fn arg<S: AsRef<OsStr>>(&mut self, arg: S) -> &mut Self {
+        self.cmd.arg(arg.as_ref());
+        self
+    }
+    pub fn stdin(&mut self, stdin: Vec<u8>) -> &mut Self {
+        self.stdin = Some(stdin);
+        self
+    }
+
+    pub fn spawn(&mut self) -> Output {
```
nihohit, line 50: nit: why mutable reference? will this be reused?

somechris: No, it's not re-used.

But it's meant to be similar to `process::Command`, which is a mutable builder. So I figured a mutable builder is the natural choice for `OpensslCommand` to make it a drop-in replacement.

Also the `if` in [line 208](https://github.com/redis-rs/redis-rs/pull/2255/changes#diff-b95b618e7cd17a6b5abaa52474d2699e73d561d77ba6b85a6f67dfd63dae9b5eR208) is simpler with a mutable Builder. (With an owned one, we'd need an assignment)

And as all the other methods are `&mut self`, so is `spawn`, to allow simple chaining.

(And it will allow us to re-use the `arg2` `arg3` from `RedisServerCommand` in a follow-up commit)

Would you prefer an owned builder?

nihohit: no, I'm fine with this choice.

Issue #2219, opened 2026-07-16
Stabilize test suite


somechris:
Running our test suite 20 times gave 30 failed tests. 😯

With our current settings of 4 retries, non of the 20 runs actually failed.

But having such a flaky test suite is not ideal nonetheless. This is the way of efficient development. 😞

The failing tests were either `sentinel`, `streams idempotency`, or `cluster_async`:

```
________________________________________________________________________________
christian@kokoschka // jobs: 0 // timestamp: 2026-07-16T12:15:22Z // exit code: 0
cwd: ~/rust/redis-rs-tester
grep -r FLAKY ~/redis-rs-flaky-test-logs | cut -f 3- -d : | sort | uniq -c | sort -n
      1 :test_cluster_async cluster_async::test_fail_on_empty_command::case_1_tokio
      1 :test_sentinel async_tests::test_get_replica_clients_report_correct_master_async::case_1_tokio
      1 :test_sentinel async_tests::test_get_replica_clients_with_one_replica_down_async::case_2_smol
      1 :test_sentinel async_tests::test_sentinel_client_async::case_1_tokio
      1 :test_sentinel async_tests::test_sentinel_client_async::case_2_smol
      1 :test_sentinel async_tests::test_sentinel_client_async_with_connection_timeout::case_1_tokio
      1 :test_sentinel async_tests::test_sentinel_client_async_with_response_timeout::case_1_tokio
      1 :test_sentinel async_tests::test_sentinel_client_async_with_response_timeout::case_2_smol
      1 :test_sentinel async_tests::test_sentinel_client_async_with_timeouts::case_1_tokio
      1 :test_sentinel async_tests::test_sentinel_client_async_with_timeouts::case_2_smol
      1 :test_sentinel async_tests::test_sentinel_connect_to_multiple_replicas_async::case_2_smol
      1 :test_sentinel async_tests::test_sentinel_connect_to_random_replica_async::case_1_tokio
      1 :test_sentinel async_tests::test_sentinel_server_down_async::case_1_tokio
      1 :test_sentinel test_get_replica_clients_success
      1 :test_sentinel test_sentinel_client
      1 :test_sentinel test_sentinel_server_down
      2 :test_sentinel async_tests::test_sentinel_client_async_with_connection_timeout::case_2_smol
      2 :test_sentinel test_sentinel_role_no_permission
      2 :test_streams idempotency_tests::test_xcfgset_idempotency_behavior
[...]

PR #2261, opened 2026-07-28
Move `TestContext` to `redis-test` (19/19)

### LINKED ISSUE #2215: RedisServer cleanup & builder, TestContext builder

somechris:
`TestContext` is not only useful in `redis-rs`' own unit tests, but probably also to users for their own integration tests. So we make it publicly available through `redis-test`.

Some parts that rely too much on feature flags, and would require too many changes in `redis-test` stay in `TestContextExt` in `redis/tests` for now. If there is demand, we can evaluate on a function-by-function basis whether bringing them over is worth it.

Fixes #2215


redis/tests/support/cluster.rs
```diff
@@ -18,9 +17,12 @@ use redis::cluster_async::Connect;
 use redis_test::cluster::ClusterType;
 use redis_test::cluster::{RedisCluster, RedisClusterConfiguration};
 use redis_test::server::{RedisServer, use_protocol};
+use redis_test::utils::{build_single_client, start_tls_crypto_provider};
+use redis_test::version::AvailableComponents;
+use redis_test::version::TestContextVersioning;
```
nihohit: nit: combine these 2 lines

.github/workflows/rust.yml
```diff
@@ -251,33 +251,49 @@ jobs:
           fi
           redis-server "${EXTRA_OPTS[@]}" &
 
+      # The examples "should" work without the "-p redis", it's added nonetheless as it helps
```
nihohit: is the change in this file related to the PR, or can it be split?

redis/tests/test_basic.rs
```diff
@@ -29,9 +29,23 @@ mod basic {
         EmbeddingInput, VAddOptions, VEmbOptions, VSimOptions, VectorAddInput, VectorQuantization,
         VectorSimilaritySearchInput,
     };
-    use redis_test::redis_value;
     use redis_test::server::redis_settings;
+    use redis_test::test_context::TestContext;
     use redis_test::utils::get_listener_on_free_port;
+    use redis_test::version::REDIS_CE_7_0;
```
nihohit: maybe use `use redis_test::version::*;` the import cost seems pretty slim here?

nihohit (commented):
I wonder if we should re-export some of the types that are in the new `use` statements in the `support` module - all of the new imports might mean that missing imports in new tests might be harder to resolve for inexperienced devs in the repo.

redis/tests/support/cluster.rs
somechris: Argh. I'm too used to automatic import cleanup through `make fix` :-( I miss this in `redis-rs` :-)

Done in fb8711a :+1:

redis/tests/test_basic.rs
somechris: Done in fb8711a.

.github/workflows/rust.yml
somechris: It's definitely triggered by this commit.

But I'm not sure how to work around it without adding `-p redis` :disappointed: 

In fact, I'm not even sure, what exactly triggers it. I mean, it's certainly a side effect of the feature synchronization between `redis` and `redis-test`. But what's going wrong where?

I had a brief look and could not get to the bottom of this. So I figured the added `-p redis` is harmless enough.

Should I dig further?

somechris:
> I wonder if we should re-export some of the types that are in the new use statements in the support module - all of the new imports might mean that missing imports in new tests might be harder to resolve for inexperienced devs in the repo.

I try to mostly let my IDE handle imports. It figures out where which entity is to be imported from very reliably. :grin: 

But sure, let's re-export things.

But which ones?

* `redis_test::test_context::TextContextBuilder`
* `redis_test::version::REDIS_*`
* `redis_test::version::VALKEY_*`
* `redis_test::version::TextContextVersion`

What else?

And should I add this to this PR, or (to keep this one from growing further) add a new one?

.github/workflows/rust.yml
somechris: The more I think of it, the more I think that something must be off here. I'll set it back to Draft and dig some more.

nihohit:
> What else?

nothing pops to mind ATM

> And should I add this to this PR, or (to keep this one from growing further) add a new one?

I think this PR, otherwise the change will just revert a lot of what you've done here.

Another question - This PR is marked as the final PR in this saga. Are you planning on moving the sentinel & cluster test contexts, too?

.github/workflows/rust.yml
somechris: It seems something goes wrong with the `dev-dependencies` from
`redis-test` to `redis`, when `aio` enabled. Fortunately, these
`dev-dependencies` are not needed in CI.

So I removed them for now in #2271.

This lets CI pass without changes and allows this PR (#2261) to pass without adding `-p redis`.

So, first #2271 has to land. Then this can land.

somechris:
> Another question - This PR is marked as the final PR in this saga. Are you planning on moving the sentinel & cluster test contexts, too?

When we discussed making `TestContext` available, I only meant `TestContext` itself (not the cluster/sentinel contexts). Are the others in a good enough shape to be re-usable for the public?

From my test runs, they are a constant source of having to rerun tests (#2219). We have a crazy high retry of 4. :grin: 

As a user, I'd be frustrated if my tests no longer run reliably once I switch to the official tooling offered by `redis-test`. :shrug:

I do think they should get migrated to `redis-test`. But only once they are stable and work reliably. :+1:

redis/Cargo.toml
```diff
@@ -171,6 +171,7 @@ aio = [
   "tokio-util/codec",
   "combine/tokio",
   "dep:cfg-if",
+  "redis-test/aio",
```
nihohit, line 175: from my experiment, this is the cause of the issue.
It also is incorrect - users who turn on the `aio` feature now will take a dependency on redis-test, even if they don't use it at all.

The `redis-test/aio` dependency probably needs to be added for every relevant test file separately
```toml
[[test]]
name = "test_async"
required-features = ["tokio-comp", "redis-test/aio"]

File: redis/Cargo.toml [hunk 5]
```diff
@@ -129,7 +129,7 @@ vector-sets = ["dep:serde", "serde/derive", "dep:serde_json"]
 cluster = ["dep:crc16", "dep:rand"]
 script = ["dep:sha1_smol"]
 tls-native-tls = ["dep:native-tls"]
-tls-rustls = ["dep:rustls", "rustls/std", "dep:rustls-native-certs"]
+tls-rustls = ["dep:rustls", "rustls/std", "dep:rustls-native-certs", "redis-test/tls-rustls"]
```
nihohit, line 133: same for this

redis/Cargo.toml
somechris: Removing that certainly helps to kill one problem, but it creates others.

> It also is incorrect - users who turn on the aio feature now will take a dependency on redis-test, even if they don't use it at all.

Are you sure?
I've learnt that adding a flag to crates that are `dev-dependencies` does not lift them to normal `dependencies`, but it adds the flag in the `dev-dependencies`.

And `cargo tree` seems to agree as well (compare the `-e dev` and `-e normal` below):

```
________________________________________________________________________________
christian@kokoschka // jobs: 1 // timestamp: 2026-07-31T13:35:30Z // exit code: 0
cwd: ~/rust/redis-rs
cargo tree -p redis -e dev --features=aio,tokio-comp | grep redis-test
├── redis-test v2.0.0-rc.1 (/home/christian/rust/redis-rs/redis-test)

________________________________________________________________________________
christian@kokoschka // jobs: 1 // timestamp: 2026-07-31T13:35:32Z // exit code: 0
cwd: ~/rust/redis-rs
cargo tree -p redis -e normal --features=aio,tokio-comp | grep redis-test

________________________________________________________________________________
christian@kokoschka // jobs: 1 // timestamp: 2026-07-31T13:35:38Z // exit code: 1
cwd: ~/rust/redis-rs
```

I think the `"redis-test/aio"` line does just what it's supposed to do: Enable the `aio` flag in the dev-dependency `redis-test`.

(So no adding of `redis-test/aio` to all `test`s is necessary)

redis/Cargo.toml
somechris: cf https://github.com/redis-rs/redis-rs/pull/2261#discussion_r3690793891

redis/Cargo.toml
nihohit: > but it creates others.

What other problems does it create?

> I've learnt that adding a flag to crates that are dev-dependencies does not lift them to normal dependencies, but it adds the flag in the dev-dependencies.

You're right, I wasn't aware of that. Looks like your solution is the right.

somechris: > > but it creates others.
>
> What other problems does it create?

When removing that `"redis-test/aio"` line for `aio` in `redis/Cargo.toml`, the tests of `redis` that use `aio` no longer enable `aio` in `redis-test`.

With `aio` missing in `redis-test`, `TestContext` no longer offers `async_connection` (that is guarded by `aio`. Before the change it was the `aio` of `redis`, now it's the `aio` of `redis-test`, which is now missing).

Let's see that in real code:

```
________________________________________________________________________________
christian@kokoschka // jobs: 0 // timestamp: 2026-07-31T15:36:34Z // exit code: 0
cwd: ~/rust/redis-rs
git status
HEAD detached at 2fa80420
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   redis/Cargo.toml

no changes added to commit (use "git add" and/or "git commit -a")

________________________________________________________________________________
christian@kokoschka // jobs: 0 // timestamp: 2026-07-31T15:39:41Z // exit code: 0
cwd: ~/rust/redis-rs
git diff
diff --git a/redis/Cargo.toml b/redis/Cargo.toml
index dd6d05da..3d348efe 100644
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -171,7 +171,6 @@ aio = [
   "tokio-util/codec",
   "combine/tokio",
   "dep:cfg-if",
-  "redis-test/aio",
 ]
 
 [dev-dependencies]

________________________________________________________________________________
christian@kokoschka // jobs: 0 // timestamp: 2026-07-31T15:39:47Z // exit code: 0
cwd: ~/rust/redis-rs
cargo nextest run -p redis --all-features --test test_async 2>&1 | head
   Compiling redis v2.0.0-rc.1 (/home/christian/rust/redis-rs/redis)
error[E0599]: no method named `async_connection` found for struct `redis_test::TestContext` in the current scope
   --> redis/tests/test_async.rs:178:27
    |
178 |         let mut con = ctx.async_connection().await.unwrap();
    |                           ^^^^^^^^^^^^^^^^
    |
help: there is a method `connection` with a similar name
    |
178 -         let mut con = ctx.async_connection().await.unwrap();

________________________________________________________________________________
christian@kokoschka // jobs: 0 // timestamp: 2026-07-31T15:39:55Z // exit code: 0
cwd: ~/rust/redis-rs
```

nihohit:
@somechris please rebase this

somechris:
Will do. But i'll take me 2-3 days :+1:

somechris:
> Will do. But i'll take me 2-3 days 👍

It was 2-3 loooooong days :-) But it's rebased now. Thanks for your patience.

nihohit:
Thanks!

Issue #2249, opened 2026-07-25
Make structs with public members and enums `non_exhaustive` for v2


somechris:

somechris:
I have most parts working, and will continue to hack on that the next days 👍

nihohit:
kk. if you want, https://github.com/nihohit/redis-rs/tree/shachar/non-exhaustive already has the majority of the work, and it compiles and tests are passing IIRC.

somechris:
Ouch! Duplicated work. Bummer.

I just wanted to upload my changes ... I should have opened the ticket earlier.

Well. As you reviewed your changes already ... I guess I'll kill my branch :+1: 😄 

Briefly glancing over things, they look similar.

The only thing I did differently is `ReplicaInfo`.

I figured that a `new` is not needed there, because the only test using `ReplicaInfo` is `test_types`.

As that is an integration test, it's affected by `#[non_exhaustive]`.

But `test_types` actually never spawns a `RedisServer` or some such. It's effectively a unit test that's in the wrong place.

So I turned `test_types` into proper unit tests, which they actually are. This requires some re-shuffling around `encode_value`, but now `ReplicaInfo` does not litter a `new` in the main code base and the unit test are now real unit tests.

The linters flag 10 missing structs in your branch (see below). But meh.

Sorry for wasting our time :shrug:

```
________________________________________________________________________________
christian@kokoschka // jobs: 2 // timestamp: 2026-07-25T20:29:48Z // exit code: 0
cwd: ~/rust/redis-rs
git describe
0.21.6-1160-ga8b4f4fc

________________________________________________________________________________
christian@kokoschka // jobs: 2 // timestamp: 2026-07-25T20:29:52Z // exit code: 0
cwd: ~/rust/redis-rs
cargo clippy --workspace --all-targets --all-features --allow-dirty --allow-staged --fix -- -D clippy::exhaustive_enums -D clippy::exhaustive_structs 2>&1 | grep -F -e '-->' -A 3 | grep 'pub \(struct\|enum\)' | tee - >(wc -l)
179 | pub struct SendError;
232 | / pub struct BloomFilterDumpChunk {
6 | pub struct RandomReplicaStrategy;
12 | / pub struct RedisClusterConfiguration {
122 | / pub struct RedisCluster {
28 | / pub struct RedisSentinelCluster {
205 | pub struct SentinelError;
172 | / pub struct RedisServer {
 9 | / pub struct TlsFilePaths {
[...]

PR #2262, opened 2026-07-28
Make all structs `non_exhaustive` that do not require further adaptions (1/15)

somechris:
These are the simple cases. Structs and enums that need code adaptions receive their `non_exhaustive` in dedicated follow-up commits.

This is part of #2249


nihohit:
Thanks!

PR #2263, opened 2026-07-29
redis-test: Allow to re-use existing TLS keys (4/5)

somechris:
As generating TLS keys for each test is time consuming, we allow to optionally skip generating new keys by re-use existing keys.

Set any of the environment variables

* `REDISRS_TLS_KEY_CA`
* `REDISRS_TLS_KEY_SERVER`
* `REDISRS_TLS_KEY_CLIENT`

to the file name of a TLS key, and the test will pick their data for the respective kind of key instead of generating a new one.

This allows to shave ~75% runtime off when running TLS tests and helps to improve the development cycle.

This re-use is not a security changes, as we only re-use the keys (certs etc are still computed afresh for each test).

Also, users have to opt-in. The default is still to generate fresh keys for each test.

This commit is part of #2254.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,29 @@ If the fuzzer finds a crash, in order to reproduce it, run:
     $ cd afl/<target>/
     $ cargo run --bin reproduce -- out/crashes/<crashfile>
 
+### Speeding up TLS tests
+
+TLS tests per-default create fresh keys for each tests. On entropy/resource-limited hosts, this might be time consuming.
+
+To instead re-use the same keys, first generate the needed keys beforehand (this needs to be done only once):
+
+```sh
+mkdir -p "$HOME/redis-rs-keys"
+openssl genrsa -out "$HOME/redis-rs-keys/ca.key" 4096
+openssl genrsa -out "$HOME/redis-rs-keys/client.key" 2048
+openssl genrsa -out "$HOME/redis-rs-keys/server.key" 2048
+```
+
+Then tell your environment to use them for testing (this needs to be executed in each terminal that runs the tests):
+
+```sh
+export REDISRS_TLS_KEY_CA="$HOME/redis-rs-keys/ca.key"
+export REDISRS_TLS_KEY_CLIENT="$HOME/redis-rs-keys/client.key"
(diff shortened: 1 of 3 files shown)


nihohit:
will you added usage of this in CI in a later PR?
Also, this needs to be documented at the crate level, and in the readme under the development clause, so users will be aware of this.

somechris:
> will you added usage of this in CI in a later PR?

Absolutely! :-) That's what commit 5/5 will do.



> Also, this needs to be documented at the crate level, and in the readme under the development clause, so users will be aware of this.

I've added it to the README in be976a9. :+1:

The README gets longer and longer though ... What about moving it's development part into a `DEVELOPMENT.md` at some point :thinking:

I'm not sure about the crate level documentation. That currently does not even mention `REDISRS_REDIS_JSON_PATH`.

And I'm not sure it'd belong there? ... Maybe we could instead link to the `DEVELOPMENT.md` once that is done?

nihohit:
> The README gets longer and longer though ... 

It does, and a lot of the info there is duplication of the info in the docs, which is IMO unnecessary. One of these days I'll find the time for some trimming :)

> What about moving it's development part into a DEVELOPMENT.md at some point 🤔

sounds like a good idea.

> I'm not sure about the crate level documentation. That currently does not even mention REDISRS_REDIS_JSON_PATH.

Just to clarify, I meant the redis-test crate, not the redis crate. redis-test's docs are [pretty minimal](https://docs.rs/redis-test/latest/redis_test/), and I'm that there are plenty of relevant details missing from there, but that's not a reason not to add relevant docs.

somechris:
> > What about moving it's development part into a DEVELOPMENT.md at some point 🤔
> 
> sounds like a good idea.

Opened #2270 so it does not get forgotten so easily.


> > I'm not sure about the crate level documentation. That currently does not even mention REDISRS_REDIS_JSON_PATH.
> 
> Just to clarify, I meant the redis-test crate, [...]. [...] but that's not a reason not to add relevant docs.


Ok. Done in 434e23f :+1:

nihohit:
Thanks!

PR #2268, opened 2026-07-30
Make `SendError` `non_exhaustive` (2/15)

somechris:
This is part of #2249

### DIFF
--- a/redis/src/aio/connection_manager.rs
+++ b/redis/src/aio/connection_manager.rs
@@ -224,7 +224,7 @@ impl ConnectionManagerConfig {
     /// let messages = Arc::new(Mutex::new(Vec::new()));
     /// let config = ConnectionManagerConfig::new().set_push_sender(move |msg|{
     ///     let Ok(mut messages) = messages.lock() else {
-    ///         return Err(redis::aio::SendError);
+    ///         return Err(());
     ///     };
     ///     messages.push(msg);
     ///     Ok(())
--- a/redis/src/aio/mod.rs
+++ b/redis/src/aio/mod.rs
@@ -176,7 +176,16 @@ pub use runtime::prefer_tokio;
 pub(super) use runtime::*;
 
 /// An error showing that the receiver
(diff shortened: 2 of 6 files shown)


nihohit:
Is the `new`/`default` implementation needed just for the doc tests? Because I don't think that users need the ability to create this error.

somechris:
> Is the `new`/`default` implementation needed just for the doc tests?

Yes.


> Because I don't think that users need the ability to create this error.

Mhmm. But that means users can never implement `AsyncPushSender` truly by themselvers, if the need to flag errors.

They have to resort to the provided implementations ... although there is a `Func` variant taking a function :shrug:

Done in `0317c94`

nihohit:
> But that means users can never implement AsyncPushSender truly by themselvers, if the need to flag errors.

🤦 

You're right, I was wrong, users need this ability.

somechris:
> [...], users need this ability.

Ok. Reverted to the previous formulation with `Default` and `new`. :+1:

I think it's good that users _can_ instatiate it.

In the doc examples though ... they actually don't _need_ to. They could yield `Err("failed to lock 'messages'")` or even `Err(())` and it would just work the same, right?

So I figured if the doc examples should better use one of the latter variants? This would make it more obvious to users that they do not _have_ to instantiate `SendError`?

(And this PR has to update them anyway)

WDYT?

nihohit:
> or even Err(())

This is probably the better option. The error handling in the `AsyncPushSender` implementations should probably be reworked in v2 in any case - ATM the errors are silently dropped, and IMO that's not a good behavior. Added that the to v2 issue checklist.

somechris:
> > or even Err(())
> 
> This is probably the better option.

Done in cdd23c9 :+1:

> ATM the errors are silently dropped, and IMO that's not a good behavior.

Agreed. :+1:

nihohit (approved):

nihohit:
Thanks!

PR #2271, opened 2026-07-30
redis-test: Convert dev-dependency on redis to feature synchronization (18.5/19)

somechris:
`redis-test` has a plain dependency on `redis`.

The additional dev-dependency on it with `aio` and `tokio-comp` is
currently not needed for CI, and get's confused Rust about features and
available libraries when trying to move `TestContext` into `redis-test`
(#2261).

There, (with the above dependency still in place) the examples compiled
`redis-test` with `aio` flag. The corresponding `rustc` command had the
required guards for `future` (`--cfg 'feature="futures"'`), but missed
the `--extern futures=...rlib`.

Hence, compiling the examples failed.

(Adding `-p redis` made the `--extern futures=...rlib` appear again.)

It seems something about the additional `redis` dev-dependency in
`redis-test` confused Rust..

So we convert the unconditional `redis` with `aio` and `tokio-comp`
dev-dependency to feature synchronization.

By that, `redis` is only once a dependency for `redis-test` and Rust can
figure things out better.

CI works before and after this change, running the same tests.

When running manually, now both of the following work:

* `cargo nextest run --locked -p redis-test`
* `cargo nextest run --locked -p redis-test --all-features`

So we can manually run also the async test.

(Switching CI to run the async test is unrelated and hence left to
follow-up commits. cf #2272)

### DIFF
--- a/redis-test/Cargo.toml
+++ b/redis-test/Cargo.toml
@@ -21,11 +21,7 @@ socket2 = "0.6"
 rand = "0.10"
 
 [features]
-aio = ["futures", "redis/aio"]
+aio = ["futures", "redis/aio", "redis/tokio-comp"]
 
 [dev-dependencies]
-redis = { version = "2.0.0-rc.1", path = "../redis", features = [
-    "aio",
-    "tokio-comp",
-] }
 tokio = { workspace = true }


nihohit:
This works because `RUST_BACKTRACE=1 cargo nextest run --locked -p redis-test` doesn't enable the `aio` features. If we'll fix the makefile to run `RUST_BACKTRACE=1 cargo nextest run --locked -p redis-test --all-features` this will fail.
Once again, our test coverage is missing critical tests🤦

somechris:
> If we'll fix the makefile to run `RUST_BACKTRACE=1 cargo nextest run --locked -p redis-test --all-features` this will fail.

Sure. If you change the target, the workaround needs to be adapted :grin:

But that's an easy fix.

It's complaining that there is no runtime ... because the
Adding `redis-tokio-comp = ["redis/tokio-comp"]` to `features` in `redis-test/Cargo.toml` brings in the required features and your new `nextest` command works again.

We can even make proper `tokio-comp` and `smol-comp` features and sync them up with `redis`. And we can probably also add the needed deps and switch the `#[tokio::test]` to `#[async_test]` and have it run on tokio and smol.

All that (changing the command to run `redis-test` tests, or enabling a test that's currently not enabled) are things we maybe (?) can/should do at some point.

But they are orthogonal to fixing things for the move of `TestContext` into `redis-test`.

Let's not mix things.

(And _if_ we chose to add `--all-features` to `nextest` in another commet, we can make things work by using feature synchronization instead of the current hard-wired and currently not needed `redis` dev-dependency)

> Once again, our test coverage is missing critical tests🤦

`async_basic_test` was not enabled before this change and is not enabled after it. We can enable it and make it work. But that's orthogonal to this change. ... actually ... I've created #2272 so we don't forget so easily.

nihohit:
I don't think that fixing the CI to run the async tests is orthogonal to this change, because this change actively breaks the `redis-test` tests - even if only for manual running. Going by the principle of [changes like this one](https://github.com/redis-rs/redis-rs/pull/2265), the fact that we don't run some configuration of features in CI, doesn't mean we should break it.

What I'd like to understand is why this blocks moving the TestContext. How can I reproduce the issues you've experienced?

somechris:
> I don't think that fixing the CI to run the async tests is orthogonal to this change, [...]

:grin: We've both spent enough time on this. And I guess I would have added the `--all-features` to the `Makefile` soon anyway.

So let's not get us held up by this. ... I've added a PR 18.75 (#2274, this adds the `--all-features`) between PR 18.5 and PR 19 ... good that floats are dense :grin:


> How can I reproduce the issues you've experienced?

See https://github.com/somechris/redis-rs/tree/nihohit-compile-issue-demo . Its last commit message gives the commands to expose the issue. I guess it can get solved a tons of ways. But the double dependency on `redis` was for sure not ideal. So getting rid of it should be worth it anyways.

somechris:
This got merged through #2274

PR #2273, opened 2026-07-31
ci: Re-use TLS keys for tests (5/5)

### LINKED ISSUE #2254: Speed up TLS tests

somechris:
This reduces e.g. `rustls` tests with all features from 550s to 140s (~75% improvement).

Runtime for a whole `build-and-test` job is reduced from ~25min to ~12min (~50%).

Fixes #2254

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,6 +10,9 @@ env:
   CARGO_TERM_COLOR: always
   REDISRS_REDIS_JSON_PATH: "/tmp/librejson.so"
   REDISRS_REDIS_BLOOM_PATH: "/tmp/redisbloom.so"
+  REDISRS_TLS_KEY_CA: "/tmp/tls-ca.key"
+  REDISRS_TLS_KEY_SERVER: "/tmp/tls-server.key"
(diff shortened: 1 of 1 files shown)


nihohit:
LGTM

PR #2274, opened 2026-07-31
ci: Run `redis-test` tests with both no and all features (18.75/19)

### LINKED ISSUE #2272: Make sure `async_basic_test` is run in CI
Due to running the `redis-test` tests without `--all-features`, `aio` is not enabled and the `async_basic_test` is not run. We should adapt CI to make it run.

somechris:
CI uses `make test`, which ran `redis-test` tests without features, as
there were no default features.

We now make this clearer by adding the `--no-default-features` flag.

To cover also `redis-test` tests that rely on a feature, we add a second
test run with `--all-features`.

This increases test coverage and should help with detecting issues earlier.

Fixes #2272

### DIFF
--- a/redis-test/Cargo.toml
+++ b/redis-test/Cargo.toml
@@ -21,11 +21,7 @@ socket2 = "0.6"
 rand = "0.10"
 
 [features]
-aio = ["futures", "redis/aio"]
+aio = ["futures", "redis/aio", "redis/tokio-comp"]
 
 [dev-dependencies]
-redis = { version = "2.0.0-rc.1", path = "../redis", features = [
-    "aio",
-    "tokio-comp",
-] }
 tokio = { workspace = true }
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,14 @@ test:
 	@RUSTFLAGS="-D warnings" REDISRS_SERVER_TYPE=unix RUST_BACKTRACE=1 cargo nextest run --locked -p redis --all-features --profile unix
 
 	@echo "===================================================================="
-	@echo "Testing redis-test"
+	@echo "Testing redis-test without features"
 	@echo "===================================================================="
-	@RUSTFLAGS="-D warnings" RUST_BACKTRACE=1 cargo nextest run --locked -p redis-test
+	@RUSTFLAGS="-D warnings" RUST_BACKTRACE=1 cargo nextest run --locked -p redis-test --no-default-features
+
+	@echo "===================================================================="
+	@echo "Testing redis-test with all features"
+	@echo "===================================================================="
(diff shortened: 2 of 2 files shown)


redis-test/Cargo.toml
```diff
@@ -21,11 +21,7 @@ socket2 = "0.6"
 rand = "0.10"
 
 [features]
-aio = ["futures", "redis/aio"]
+aio = ["futures", "redis/aio", "redis/tokio-comp"]
 
 [dev-dependencies]
-redis = { version = "2.0.0-rc.1", path = "../redis", features = [
```
nihohit, line 27: see my comments in the other PR - I think the problem was taking non-dev dependencies on `redis-test` in `redis`, and not this edge in the cycle.

https://github.com/redis-rs/redis-rs/pull/2261#discussion_r3690718687
https://github.com/redis-rs/redis-rs/pull/2261#discussion_r3690727814

redis-test/Cargo.toml
```diff
@@ -21,11 +21,7 @@ socket2 = "0.6"
 rand = "0.10"
 
 [features]
-aio = ["futures", "redis/aio"]
+aio = ["futures", "redis/aio", "redis/tokio-comp"]
```
nihohit, line 24: I really don't like this additional dependency, but I'm not sure how we can avoid it without declaring the additional dev dependency, so it looks like this is the least-bad option :(

nihohit (approved):

redis-test/Cargo.toml
somechris: > I really don't like this additional dependency, [...]

Neither do I.

It is, however, just the dependency that was there before. Only phrased differently.

We could also have

```
[features]
aio = ["futures", "redis/aio"]
tokio-comp = ["aio", "redis/tokio-comp"]
```

Then we're basically in the same situation as in `redis` that we have `aio` but we should not use it directly. Only through a flag that provides a runtime.

But that's more change than necessary and I had the impression that you don't want to change too much here :shrug:

Taking a short-cut ... if we want to run `redis-test`'s only `async` test with both `tokio` and `smol` at some point, we probably also need `smol-comp` and all the other features that `redis` has.

Should I bring them all over?

(Then `aio` might be cleaner ... without a runtime ... but it's not needed right now. /me jokingly thinks about orthogonality and what ratholes come with it :grin:)

nihohit: > It is, however, just the dependency that was there before. Only phrased differently.

Not entirely - the `tokio-comp` dependency was there only on `redis-test`'s dev dependency, not every usage of `aio`. 

I'd like to go over it and try to restore this after the `TestContext` change is merged, but IMO it's not the most important issue for either of us to work on ATM.

PR #2278, opened 2026-08-01
Fix cluster mass connection failure by escalating to blocking recovery

### LINKED ISSUE #2277: Cluster commands fail when enough connections go stale (regression from #2120)
Disclaimer: investigation, reproducer, bug report, and associated suggested fix where produced with assistance from an AI agent based off of behaviour observed in production.

## Description

PR #2120 changed cluster reconnection from blocking (synchronous) to background (asynchronous) to fix the issue where a single bad replica blocks all shards. However, this
introduces a regression: when ALL cluster connections go stale simultaneously, commands fail with `BrokenPipe` because the retry loop exhausts before any background `reconnect_loop`
task is polled.

## Root Cause

Prior to #2120, `PollFlushAction::Reconnect(addrs)` entered a blocking `Recover` state via `refresh_connections`, healing the node before the next retry fired. After #2120, it spawns a non-blocking `reconnect_loop` that sleeps on failure (exponential backoff starting at 50ms). Since mock/real connections in the poll loop complete synchronously relative to the executor, all retries fire instantly against remaining stale nodes without ever yielding to the background repair tasks.

In clusters with more nodes than the retry budget (default 16), the retry loop hits 16 different stale-but-"Connected" nodes, burns all retries, and the command fails. `ClusterConnectionNotFound` (which would trigger blocking recovery) is never reached because there are always remaining nodes in `Connected` state that haven't been visited yet.

## Reproduction

The scenario:
1. A cluster with N nodes where N > `DEFAULT_RETRIES` (16)
2. Enough TCP connections become stale so all retries are attempted on stale servers (e.g., server-side idle timeout, host reprovisionning, server pool churn)
3. Client sends a command
4. Each retry picks a random `Connected` node (all stale), gets `BrokenPipe`, spawns background repair
5. After 16 retries, the command fails -- even though background repair would eventually succeed

This is observable in production with clusters of hundreds of nodes after any event that invalidates all (or enough) connections.

A reproducer is available here: https://gist.github.com/chantra/5d7cd4ef8e314bd10ddb2113e3df59b7

* pre_pr2120 with reproducer branch: https://github.com/chantra/redis-rs-redis-rs/tree/pre_pr2120
```
 cargo test --features cluster-async test_repro_all_connections_stale
   Compiling redis v1.3.0 (/Users/mbretelle/devel/redis-rs/redis)
   Compiling redis-test v1.0.4 (/Users/mbretelle/devel/redis-rs/redis-test)
   Compiling fuzz-target-parser v0.1.0 (/Users/mbretelle/devel/redis-rs/afl/parser)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 9.57s
     Running unittests src/main.rs (target/debug/deps/fuzz_target-8d0f26be40b3b017)
...
...

     Running tests/test_cluster_async.rs (target/debug/deps/test_cluster_async-590c78490e770e51)

running 2 tests
test cluster_async::test_repro_all_connections_stale_retries_exhaust_before_recovery ... ok
test cluster_async::test_repro_all_connections_stale_many_nodes_retries_exhausted ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 72 filtered out; finished in 0.01s
...
...
```

Full output: https://gist.github.com/chantra/7832df180015c8f4d27ca0607d5f7c55

* post_pr2120 with reproducer branch: https://github.com/chantra/redis-rs-redis-rs/tree/post_pr2120
```
cargo test --features cluster-async test_repro_all_connections_stale
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.36s
     Running unittests src/main.rs (target/debug/deps/fuzz_target-8d0f26be40b3b017)
...
...
thread 'cluster_async::test_repro_all_connections_stale_many_nodes_retries_exhausted' (12251028) panicked at redis/tests/test_cluster_async.rs:3697:9:
Expected command to succeed after reconnection, but got: mock-io-error
With 20 nodes and 16 retries (default), retries exhaust across stale nodes without ever triggering blocking recovery. This models the production scenario (hundreds of nodes, 16 retries).


failures:
    cluster_async::test_repro_all_connections_stale_many_nodes_retries_exhausted
    cluster_async::test_repro_all_connections_stale_retries_exhaust_before_recovery

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 76 filtered out; finished in 0.02s

error: test failed, to rerun pass `-p redis --test test_cluster_async`
```
Full output: https://gist.github.com/chantra/5f777028c160b3f2ed4b7f5b7363b992

Here is a redacted log showing the order of events: https://gist.github.com/chantra/d70cb6728acbc27ded3415618e631398
Here is a readout of the logs:

- job using redis cluster start after having been idle for an extended period
- `CLUSTER SHARDS` command is sent
- there is an attempt to connect to host 01 -> host 17 (initial host + 16 retries), each of them fail and are the connection is sent to repair in the background (over 2ms)
- job fail with "broken pipe"
- async repairs log lines are still showing up 
- first repaired connection surfaces after 50ms from the initial command
- host13:port13 fails recovery for ever (there is no redis service running there). My understanding is that it should eventually be removed from the topology which would [break the infinite loop trying to connect to a dead host](https://github.com/redis-rs/redis-rs/blob/65a344a01dd6c0df75b17887847bf6cc44eb0966/redis/src/cluster_handling/async_connection/mod.rs#L1263-L1267), but in my observation this never happened, probably a separate issue.

## Expected Behavior

Commands should eventually succeed after reconnection, as they did before #2120.

## Affected Versions

- redis-rs >= 1.4.0 (any version containing commit 7a41fdd)

chantra:
Disclaimer: Analysis + PR were assisted by an AI agent.

## Summary

- Adds a circuit breaker to the `poll_flush` Reconnect handler that detects widespread connection loss
- When less than half the cluster nodes are healthy, escalates from background `reconnect_loop` to blocking `RecoverFuture::Reconnect`
- The blocking path gathers all non-Connected nodes, reconnects them in parallel via `refresh_connections_locked`, and retries with backoff until at least one succeeds
- Preserves the non-blocking background reconnection for single-node failures (the common case #2120 was designed for)
- Includes two regression tests: a minimal 2-node reproducer and a 20-node scale test modeling production clusters

## Problem

PR #2120 introduced background reconnection to avoid blocking all shards when a single replica fails. However, when ALL connections go stale simultaneously (e.g., after network
partition recovery or mass idle timeout), every retry in the retry loop hits a different stale node. The background reconnect_loop tasks sleep on failure (50ms+ backoff) and are
never polled before retries exhaust. In clusters with more nodes than the retry budget (default: 16), commands fail unconditionally.

## Fix

 The circuit breaker checks (via `try_read` on `conn_lock`) whether healthy nodes constitute less than half the cluster after accounting for the current `Reconnect` action. If so, it
 enters a blocking Recover state that:
 1. Gathers ALL non-Connected nodes (including those marked Reconnecting by earlier iterations)
 2. Attempts parallel reconnection via `refresh_connections_locked`
 3. Retries with exponential backoff until at least one node is restored
 4. Only then resumes the poll loop, allowing retried commands to reach healed connections

 Test plan

 - [x] `test_repro_all_connections_stale_retries_exhaust_before_recovery` -- 2-node, retries=2 (FAILS without fix, PASSES with fix)
 - [x] `test_repro_all_connections_stale_many_nodes_retries_exhausted` -- 20-node, default retries=16 (FAILS without fix, PASSES with fix)
 - [x] Both tests PASS on pre-#2120 code (commit f6f64a0), confirming they are valid regression tests

 
Fixes #2277

### DIFF
--- a/redis/src/cluster_handling/async_connection/mod.rs
+++ b/redis/src/cluster_handling/async_connection/mod.rs
@@ -1332,6 +1332,11 @@ struct Message<C> {
 enum RecoverFuture {
     RecoverSlots(BoxFuture<'static, RedisResult<()>>),
     ReconnectInitial(BoxFuture<'static, RedisResult<()>>),
+    /// Targeted blocking reconnection of specific nodes. Used when no
+    /// healthy connections remain and background repair would fail because
+    /// the retry loop would exhaust across stale nodes before any background
+    /// reconnect_loop is polled.
+    Reconnect(BoxFuture<'static, RedisResult<()>>),
 }
 
 enum ConnectionState {
@@ -1505,6 +1510,7 @@ where
             match fut {
                 RecoverFuture::RecoverSlots(fut) => fut.await?,
                 RecoverFuture::ReconnectInitial(fut) => fut.await?,
+                RecoverFuture::Reconnect(fut) => fut.await?,
             }
         }
         Ok(())
@@ -1536,6 +1542,20 @@ where
                     Err(err) => {
                         error!(
                             "async cluster client recovery: reconnect to initial nodes failed: `{err}`"
+                        );
+                    }
+                }
+                self.state = ConnectionState::PollComplete;
+                Ok(())
+            }
+            RecoverFuture::Reconnect(future) => {
+                match ready!(future.as_mut().poll(cx)) {
+                    Ok(()) => {
+                        info!("async cluster client recovery: targeted reconnection complete");
+                    }
+                    Err(err) => {
+                        error!(
+                            "async cluster client recovery: targeted reconnection failed: `{err}`"
                         );
                     }
                 }
@@ -1764,7 +1784,74 @@ where
                     )));
                 }
                 PollFlushAction::Reconnect(addrs) => {
-                    self.register_reconnect_futures(addrs);
+                    // Check whether there are healthy (Connected) nodes remaining
+                    // after accounting for the nodes that are about to be marked as
+                    // Reconnecting. If no healthy nodes will remain, the background
+                    // reconnect_loop approach fails because every retry in the retry
+                    // loop will hit another stale node, exhausting the retry budget
+                    // before any background repair task is polled. In that case we
+                    // escalate to blocking recovery.
+                    let should_block = self.inner.conn_lock.try_read().ok().is_some_and(|guard| {
+                        let total = guard.0.len();
+                        let healthy_count = guard.0.iter().filter(|(addr, state)| {
+                            matches!(state, ConnState::Connected(_)) && !addrs.contains(*addr)
+                        }).count();
+                        // Escalate to blocking recovery if no healthy nodes remain,
+                        // OR if less than half the cluster is healthy (indicating
+                        // widespread connection loss). The latter case prevents retry
+                        // exhaustion in large clusters where nodes > retries.
+                        total > 0 && healthy_count * 2 < total
+                    });
+
+                    if should_block {
+                        warn!(
+                            "async cluster client recovery: no healthy connections remain, \
+                             escalating to blocking reconnection for {} nodes",
+                            addrs.len()
+                        );
+                        let inner = self.inner.clone();
+                        self.state = ConnectionState::Recover(RecoverFuture::Reconnect(
+                            Box::pin(async move {
+                                // Gather ALL nodes that need reconnection (not just the
+                                // ones from this Reconnect action). This includes nodes
+                                // already marked Reconnecting by earlier iterations.
+                                let mut all_addrs = addrs;
+                                {
+                                    let guard = inner.conn_lock.read().await;
+                                    for (addr, state) in guard.0.iter() {
+                                        if !matches!(state, ConnState::Connected(_)) {
+                                            all_addrs.insert(addr.clone());
+                                        }
+                                    }
+                                }
+                                // Retry until at least one connection is established.
+                                let mut backoff = Duration::from_millis(10);
+                                loop {
+                                    {
+                                        let mut connections = inner.conn_lock.write().await;
+                                        inner
+                                            .refresh_connections_locked(
+                                                &mut connections.0,
+                                                all_addrs.clone(),
+                                            )
+                                            .await;
+                                        // Check if at least one node is now Connected
+                                        let any_connected = connections.0.values().any(|state| {
+                                            matches!(state, ConnState::Connected(_))
+                                        });
+                                        if any_connected {
+                                            break;
+                                        }
+                                    }
+                                    boxed_sleep(backoff).await;
+                                    backoff = (backoff * 2).min(Duration::from_secs(1));
+                                }
+                                Ok(())
+                            }),
+                        ));
+                    } else {
+                        self.register_reconnect_futures(addrs);
+                    }
                 }
                 PollFlushAction::ReconnectFromInitialConnections => {
                     warn!("async cluster client recovery: beginning reconnect to initial nodes");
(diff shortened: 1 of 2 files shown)


chantra:
I am partial on this solution, while I think it will minimize the chances of hitting the issue, it does not prevent say a cluster of 1000 nodes with 20 stale nodes to reproduce the issue. There is still a chance to hit the one of the 20 bad node during the 1 + 16 retries connection attempts and never trigger the circuit breaker as 2x980 remaining healthy hosts is greater than 1000 initial pool size.

@nihohit happy to hear what your suggestion would be. Would it make more sense to pair PR#2120 solution for the common case with then a fallback on the older blocking approach? Optimizing for the common case while providing contingency solution?

nihohit:
@chantra the main question I have here - was this fix triggered by a real world issue, or is this a fix to a theoretical problem?
If this was prompted by a real world issue, I'd love to hear exactly what happened.

chantra:
@nihohit real world workload. #2277 goes over the details of the issue, redacted logs showing the series of event with attempt to repair connection is shown in there. The issue is verbose but hopefully contains all the details needed.

To be clear, this PR is more for illustration and to kick discussion, there is a bunch of problem with it.

To answer your question,
At a high level, the program goes like this:
- on start up establish a connection to the redis cluster
- Every $interval time, execute a task performing queries against the redis cluster


Before PR 2120, this has been working fine, supposedly because of the blocking recovery that was happening under the hood.
Since PR 2120, it breaks because by the time the command is sent, the connections have not recovered.
There is likely 2 triggers to this issue that PR2120 did not account for:
- large redis cluster with host:port coming and going.
- Connection being idle for a while increasing the chances of connection to node being stale.

nihohit:
> Connection being idle for a while increasing the chances of connection to node being stale.

This can be fixed with the addition of a configurable, timed heartbeat.

> large redis cluster with host:port coming and going.

By "coming and going" you mean that nodes are added and removed from the cluster?

nihohit:
> When less than half the cluster nodes are healthy, escalates from background reconnect_loop to blocking RecoverFuture::Reconnect

I don't think that this condition is sufficient. If a large cluster has 3 shards and all disconnect, you'd experience the same issue when sending a request to one shard, but because you'd only iterate through a third of the nodes, the condition wouldn't be hit. 
The issue is that ATM the cluster's internal representation has only implicit representation of a shard, so it'll be hard to reflect a state like "disconnected shard".

Did you investigate how other clients handle similar situations? That's a great research task for an AI :)

chantra:
> > Connection being idle for a while increasing the chances of connection to node being stale.
> 
> This can be fixed with the addition of a configurable, timed heartbeat.
> 
> > large redis cluster with host:port coming and going.
> 
> By "coming and going" you mean that nodes are added and removed from the cluster?

yes, hosts may be sent to reprovision, reboot, redis software update....

chantra:
> > When less than half the cluster nodes are healthy, escalates from background reconnect_loop to blocking RecoverFuture::Reconnect
> 
> I don't think that this condition is sufficient. If a large cluster has 3 shards and all disconnect, you'd experience the same issue when sending a request to one shard, but because you'd only iterate through a third of the nodes, the condition wouldn't be hit. The issue is that ATM the cluster's internal representation has only implicit representation of a shard, so it'll be hard to reflect a state like "disconnected shard".

My knowledge of the specifics of redis is pretty weak, so I may be totally wrong here. But isn't having at least 1 valid connection to any hosts in the cluster enough to bootstrap the topology? 
> 
> Did you investigate how other clients handle similar situations? That's a great research task for an AI :)

Good idea, commented with the results in #2277.

nihohit:
> My knowledge of the specifics of redis is pretty weak, so I may be totally wrong here. But isn't having at least 1 valid connection to any hosts in the cluster enough to bootstrap the topology?

It is sufficient technically, but ATM disconnects aren't a sufficient trigger for the client to start a topology refresh. Reconnections only reconnect known disconnected connections.

PR #2296, opened 2026-08-04
Bump `azure_identity` to 1.0 and `azure_core` to 1.1.0

StefanPalashev:
## Summary

The upstream 1.0 release changed how a client certificate is handed to `ClientCertificateCredential`, so `ClientCertificate` is reworked accordingly. It now holds the **decoded** PKCS12 archive instead of a base64-encoded one. 
This is a **breaking change** to the public API of the `entra-id` feature.

The release also changes which TLS implementation is used for the requests to the Entra ID token endpoint.

## Background

`azure_identity` 0.31 took the certificate as a base64-encoded `Secret` and decoded it internally:

```rust
// azure_identity-0.31.0/src/client_certificate_credential.rs
pub fn new(tenant_id: String, client_id: String, certificate: Secret, ...)
// ...
let cert_bytes = base64::decode(certificate.secret())
```

`azure_identity` 1.0.0 takes `SecretBytes` and parses the archive directly, with no decoding step:

```rust
// azure_identity-1.0.0/src/client_certificate_credential.rs
pub fn new(tenant_id: String, client_id: String, certificate: SecretBytes, ...)
// ...
let (key, cert, ca_chain) = parse_certificate(certificate.bytes(), options.password.as_ref())?;
```

`ClientCertificate`'s shape was changed so it follows upstream:

| | Before | After |
| --- | --- | --- |
| Certificate field | `pub base64_pkcs12: String` | `pkcs12: Vec<u8>` (private) |
| Password field | `pub password: Option<String>` | `password: Option<String>` (private) |
| Construction | struct expression | `ClientCertificate::new` |
| Password | `password` field | `ClientCertificate::set_password` |
| `Debug` | derived - could print sensitive data | manual - doesn't print sensitive data|

## API

```rust
// Before:
let certificate_base64 = fs::read_to_string("path/to/base64_pkcs12_certificate")?;
let certificate = ClientCertificate {
    base64_pkcs12: certificate_base64,
    password: None,
};

// After:
let certificate = ClientCertificate::new(fs::read("path/to/pkcs12_certificate")?);
// For a password-protected archive:
let certificate =
    ClientCertificate::new(fs::read("path/to/pkcs12_certificate")?).set_password("your-password");
```

## Behavior

**The bytes have to be the decoded archive.**
Handing a base64-encoded certificate to `ClientCertificate::new` still compiles.
It fails later, when `EntraIdCredentialsProvider::new_client_certificate` parses the archive, with an error originating in `azure_identity`. Because this cannot be caught by the compiler, it is pinned by a dedicated unit test and called out in `version2.md`.

**TLS for the token endpoint changed from `native-tls` to `rustls`.** `azure_core` 0.31 defaulted to `reqwest_native_tls`. 1.1.0 defaults to `reqwest_rustls` and `reqwest_native_tls` no longer exists as a feature. Visible in the lockfile as `hyper-tls` being replaced by `hyper-rustls`, `rustls-platform-verifier` and `webpki-root-certs`.

Consequences:
- Only the HTTPS requests that fetch tokens from Entra ID are affected. The TLS backend for the connection to the server is unchanged and still selected via `tls-native-tls` / `tls-rustls`.
- Combining `entra-id` with `tls-native-tls` links both TLS implementations into the binary.
- `rustls` validates certificates more strictly than OpenSSL, so token acquisition can start failing in environments that intercept TLS traffic. The platform certificate store is used, via `rustls-platform-verifier`.

## Testing

| Test | Type | Covers |
| --- | --- | --- |
| `test_client_certificate_accepts_raw_pkcs12` | unit | A decoded archive is accepted |
| `test_client_certificate_rejects_base64_pkcs12` | unit | A base64-encoded archive is rejected - pins the semantic this PR changes |
| `test_client_certificate_password` | unit | A protected archive fails without a password and succeeds via `set_password` |
| `test_client_certificate_debug_hides_secrets` | unit | `Debug` output contains neither the password nor the certificate bytes |
| `setup_provider_with_client_certificate_and_scopes` | integration | Adapted to the new API |

## Relevant documentation

- [`azure_identity` on docs.rs](https://docs.rs/azure_identity/1.0.0/azure_identity/struct.ClientCertificateCredential.html)
- [`azure_core::credentials::SecretBytes`](https://docs.rs/azure_core/1.1.0/azure_core/credentials/struct.SecretBytes.html)

### DIFF
--- a/redis/src/entra_id.rs
+++ b/redis/src/entra_id.rs
@@ -856,6 +856,12 @@ mod tests {
             ClientCertificate::new(encoded),
         );
         assert!(provider.is_err());
+        assert!(
+            provider
+                .unwrap_err()
+                .to_string()
+                .contains("deserializing PKCS12 from DER failed")
+        );
     }
 
     #[test]
@@ -866,6 +872,12 @@ mod tests {
             ClientCertificate::new(PKCS12_WITH_PASSWORD.clone()),
         );
         assert!(provider.is_err());
+        assert!(
+            provider
+                .unwrap_err()
+                .to_string()
+                .contains("PKCS12 parsing failed")
+        );
 
         let provider = EntraIdCredentialsProvider::new_client_certificate(
             TENANT_ID.to_string(),
--- a/redis/src/entra_id.rs
+++ b/redis/src/entra_id.rs
@@ -40,6 +40,12 @@
 //!
 //! Add the `entra-id` feature to your `Cargo.toml`.
 //!
+//! The feature depends on `azure_identity` and `azure_core`, which use `rustls` together with the
+//! platform's certificate store for the HTTPS requests to the Entra ID token endpoint.
+//! That is independent of the TLS backend used for the connection to the server, so combining `entra-id`
+//! with `tls-native-tls` links both TLS implementations into the binary. Keep in mind that `rustls`
+//! validates certificates more strictly than OpenSSL, which can matter in environments that intercept TLS traffic.
+//!
 //! ## Basic Usage with DeveloperToolsCredential
 //!
 //! ```rust,no_run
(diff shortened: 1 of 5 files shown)


StefanPalashev:
Integration tests run against a real deployment:
{E3DAD535-CAC4-4A77-AD8B-413055425303}

redis/src/entra_id.rs
```diff
@@ -758,6 +834,55 @@ mod tests {
...
+    fn test_client_certificate_rejects_base64_pkcs12() {
+        use base64::Engine;
+        let encoded = base64::engine::general_purpose::STANDARD.encode(PKCS12.as_slice());
+
+        let provider = EntraIdCredentialsProvider::new_client_certificate(
+            TENANT_ID.to_string(),
+            CLIENT_ID.to_string(),
+            ClientCertificate::new(encoded),
+        );
+        assert!(provider.is_err());
```
nihohit, line 858: throughout, assert on the returned error, so that we can stabilize it

StefanPalashev: Added asserts on the underlying messages.

Reminder: All errors returned by the provider are remapped to `RedisError`.
``` rust
 fn convert_error(err: azure_core::Error) -> RedisError {
        RedisError::from((
            ErrorKind::AuthenticationFailed,
            "Entra ID authentication failed",
            format!("{err}"),
        ))
    }
```

nihohit (approved):
Thanks!

PR #2302, opened 2026-08-08
Add `CreateOptions` for the Search module's `FT.CREATE` command [1/8]

StefanPalashev:
# Add `CreateOptions` for the Search module's `FT.CREATE` command

## Summary

Introduce the `search` cargo feature and the [optional_arguments](https://redis.io/docs/latest/commands/ft.create/#optional-arguments) of `FT.CREATE`: `CreateOptions`, `IndexDataType` and `SearchLanguage`.

This is **part 1** of a series that supersedes the original pull request (https://github.com/redis-rs/redis-rs/pull/2015), where the design of these types is described in full. `CreateOptions` leads the series because it doesn't depend on neither the schema nor any field type and thus can be reviewed in isolation.


redis/Cargo.toml
```diff
@@ -126,6 +126,7 @@ bloom = []
 geospatial = []
 json = ["dep:serde", "serde/derive", "dep:serde_json"]
 vector-sets = ["dep:serde", "serde/derive", "dep:serde_json"]
+search = []
```
nihohit: please rename to `search_unfinished` and document that the feature is still in development and shouldn't be used. we'll rename back once enough functionality will be mergedd.

redis/src/commands/search/create/options.rs
```diff
@@ -0,0 +1,482 @@
...
+    score: Option<f64>,
+    score_field: Option<String>,
+    max_text_fields: bool,
+    no_offsets: bool,
+    temporary: Option<u64>,
+    no_highlight: bool,
+    no_fields: bool,
+    no_freqs: bool,
+    stopwords: Vec<String>,
+    skip_initial_scan: bool,
```
nihohit, line 93: are all the value combination valid? if not, does the server return clear errors on invalid combinations?
if not, we might want to have a client-side validation in a fallible build phase.

StefanPalashev: From what I've seen, `max_text_fields` and `no_fields` are mutually exclusive. There's an integration test that covers the scenario at a later stage.

The server does return a clear error in that case.

StefanPalashev: P.S. and this was changed in 8.8, initially, there was no restriction.

redis/src/lib.rs
```diff
@@ -106,6 +106,7 @@
 //! * `uuid`: enables type conversion to UUID (optional)
 //! * `sentinel`: enables high-level interfaces for communication with Redis sentinels (optional)
 //! * `json`: enables high-level interfaces for communication with the JSON module (optional)
+//! * `search_unfinished`: enables high-level interfaces for communication with the Search module (optional)
```
nihohit: either remove this line, or say explicitly that ATM the feature is in an incomplete state.

nihohit: better to say so explicitly

StefanPalashev: Added a note that states that.

PR #2303, opened 2026-08-08
Add the `FT.CREATE` command builder and index schema [2/8]

StefanPalashev:
# Add the `FT.CREATE` command builder and index schema

## Summary

Adds `FtCreateCommand`, `SearchSchema`, the `schema!` macro, the `TEXT` field type and the `ft_create` command. 
Together with the options from the previous part, this makes `FT.CREATE` usable end to end.
`FtCreateCommand` and `SearchSchema` both use the typestate pattern, so a command cannot be built without a schema and a schema cannot be empty both are compile-time errors rather than server-side ones.

This is **part 2 of the series that supersedes #2015**, where the design is described in full.

## Changes relative to #2015

Some review comments from #2015 were addressed here rather than carried over:

| Change | Why |
|---|---|
| `RediSearchSchema` renamed to `SearchSchema` | It was planned for a rename in the future anyway. |
| `FieldDefinition::JustType` and `From<FieldType> for FieldDefinition` removed | They gave every field type a second serialization path. Fields are now always built through their own builder, e.g. `SchemaTextField::new()` |
| `schema!` uses `+` repetition | Replaces the first-element-plus-`*` form; the non-empty guarantee is unchanged |

Dropping `JustType` also removes a case that type-checked but could not work. `FieldType::Vector` has no valid zero-argument form, so `schema! { "v" => FieldType::Vector }` used to serialize to a bare `VECTOR` token that the server rejects.

Everything else is a direct extraction from `types.rs` and `command.rs`, including all comments. The doc examples use `TEXT` only, since `SchemaNumericField` and `SchemaTagField` do not exist yet; they will be restored to the #2015 wording in the part that adds them.

### DIFF
--- a/redis/src/commands/search/create/fields.rs
+++ b/redis/src/commands/search/create/fields.rs
@@ -76,13 +76,13 @@ impl ToRedisArgs for Sortable {
 #[non_exhaustive]
 pub enum Phonetic {
     /// Double metaphone for English
-    DmEn,
+    DmEnglish,
     /// Double metaphone for French
-    DmFr,
+    DmFrench,
     /// Double metaphone for Portuguese
-    DmPt,
+    DmPortuguese,
     /// Double metaphone for Spanish
-    DmEs,
+    DmSpanish,
 }
 
 impl ToRedisArgs for Phonetic {
@@ -91,10 +91,10 @@ impl ToRedisArgs for Phonetic {
         W: ?Sized + RedisWrite,
     {
         out.write_arg(match self {
-            Self::DmEn => b"dm:en",
-            Self::DmFr => b"dm:fr",
-            Self::DmPt => b"dm:pt",
-            Self::DmEs => b"dm:es",
+            Self::DmEnglish => b"dm:en",
+            Self::DmFrench => b"dm:fr",
+            Self::DmPortuguese => b"dm:pt",
+            Self::DmSpanish => b"dm:es",
         });
     }
 }
@@ -341,7 +341,7 @@ mod tests {
         let schema = schema! {
             TEXT_FIELD_NAME => SchemaTextField::new(),
         };
-        let ft_create = FtCreateCommand::new(INDEX_NAME).schema(schema);
+        let ft_create = FtCreateCommand::new(INDEX_NAME, schema);
         assert_eq!(ft_create.into_args(), "FT.CREATE index SCHEMA title TEXT");
     }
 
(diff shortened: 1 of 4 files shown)


redis/src/commands/search/create/fields.rs
```diff
@@ -0,0 +1,495 @@
...
+}
+
+/// Declaring a text attribute as PHONETIC will perform phonetic matching on it in searches by default.
+/// The obligatory argument specifies the phonetic algorithm and language used.
+/// <https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/phonetic_matching/>
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum Phonetic {
+    /// Double metaphone for English
+    DmEn,
```
nihohit: DmEnglish, etc., please. let's make these easy to understand.

redis/src/commands/search/create/fields.rs
```diff
@@ -0,0 +1,495 @@
...
+            sortable: None,
+            no_index: false,
+        }
+    }
+
+    pub fn sortable(mut self, sortable: Sortable) -> Self {
+        if self.base.field_type.is_sortable() {
+            self.sortable = Some(sortable);
+        } else {
+            unreachable!("Field type {:?} is not sortable", self.base.field_type);
```
nihohit, line 169: let's try to avoid production crashes. either we let the server enforce this, or we return a result from the function

redis/src/commands/search/create/schema.rs
```diff
@@ -0,0 +1,202 @@
...
+    fn from(field: SchemaTextField) -> Self {
+        Self::Text(field)
+    }
+}
+
+/// Marker type indicating an empty schema (no fields added yet).
+pub struct Empty;
+
+/// Marker type indicating a non-empty schema (at least one field added).
+pub struct NonEmpty;
```
nihohit: do these need to be public?

redis/src/commands/search/create/schema.rs
```diff
@@ -0,0 +1,202 @@
...
+        Self {
+            fields: Vec::new(),
+            _state: PhantomData,
+        }
+    }
+
+    /// Insert the first field into the schema.
+    ///
+    /// This transitions the schema from `Empty` to `NonEmpty` state.
+    pub fn insert<K: Into<String>, V: Into<FieldDefinition>>(
```
nihohit: I don't understand the need for the typestate pattern here - just require that `new` has to take a key and value, and save the complexity. The `schema` macro already enforces at least one pair.
AFAIS this is true also in the later PRs.

nihohit (approved):

PR #2307, opened 2026-08-10
fix: use a dedicated cached connection per candidate

freedom-winds:
## Summary

Use each Sentinel candidate's matching cached connection when validating candidates with `ROLE`.

Previously, both the synchronous and asynchronous `get_sentinel_client` paths always used the first cache entry while iterating over every configured Sentinel address. If the first reachable endpoint was not a Sentinel, later iterations reused that same connection instead of contacting subsequent candidates, preventing failover to a healthy Sentinel.

This change zips configured Sentinel addresses with their corresponding connection-cache entries, so each candidate is independently connected and validated in both sync and async paths.

## Testing

- `cargo check -p redis --features cluster-async,tokio-comp --lib`

### DIFF
--- a/redis/tests/regression_sentinel_candidate_cache.rs
+++ b/redis/tests/regression_sentinel_candidate_cache.rs
@@ -1,0 +1,56 @@
+#![cfg(feature = "sentinel")]
+
+mod support;
+
+use redis::sentinel::{SentinelClient, SentinelServerType};
+use test_macros::async_test;
+
+use crate::support::TestSentinelContext;
+
+fn client_with_non_sentinel_first(context: &TestSentinelContext) -> (SentinelClient, redis::ConnectionInfo) {
+    let non_sentinel = context.cluster.servers[0].connection_info();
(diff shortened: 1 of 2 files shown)


redis/tests/regression_sentinel_candidate_cache.rs
```diff
@@ -0,0 +1,56 @@
...
+use redis::sentinel::{SentinelClient, SentinelServerType};
+use test_macros::async_test;
+
+use crate::support::TestSentinelContext;
+
+fn client_with_non_sentinel_first(context: &TestSentinelContext) -> (SentinelClient, redis::ConnectionInfo) {
+    let non_sentinel = context.cluster.servers[0].connection_info();
+    let healthy_sentinel = context.sentinels_connection_info()[0].clone();
+    let client = SentinelClient::build(
+        vec![non_sentinel, healthy_sentinel.clone()],
```
nihohit, line 14: the test relies on a bad configuration - passing the address of a non-sentinel server is wrong. Was the intention of the fix to handle this? because IMO this should raise an error, not be silently ignored.

PR #2317, opened 2026-08-21
Add the remaining simple field types for `FT.CREATE` [3/8]

StefanPalashev:
# Add the remaining simple field types for `FT.CREATE`

## Summary

Adds the `TAG`, `NUMERIC`, `GEO` and `GEOSHAPE` schema field types.

This is **part 3 of the series that supersedes #2015**, where the design is described in full.

### DIFF
--- a/redis/src/commands/search/create/fields.rs
+++ b/redis/src/commands/search/create/fields.rs
@@ -323,6 +323,324 @@ impl Default for SchemaTextField {
     }
 }
 
+/// Represents a tag field in the schema.
+#[must_use = "Tag field has no effect unless inserted into a schema"]
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct SchemaTagField {
+    pub(crate) common: SchemaCommonField,
+    separator: Option<char>,
+    case_sensitive: bool,
+    with_suffix_trie: bool,
+    index_empty: bool,
+}
+
+impl SchemaTagField {
+    /// Create a new TAG field.
+    pub fn new() -> Self {
+        Self {
+            common: SchemaCommonField::new(FieldType::Tag),
+            separator: None,
+            case_sensitive: false,
+            with_suffix_trie: false,
+            index_empty: false,
+        }
+    }
+
+    /// Indicates how the text contained in the attribute is to be split into individual tags.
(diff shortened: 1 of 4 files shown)


redis/src/commands/search/create/mod.rs
```diff
@@ -151,6 +152,84 @@ mod tests {
     // Website examples
     // <https://redis.io/docs/latest/commands/ft.create/#examples>
     // ============================================================================
+    #[test]
+    fn test_create_blog_post_index() {
```
nihohit, line 156: this name makes the test sound like an example. what concretely is it testing?

redis/src/commands/search/create/fields.rs
```diff
@@ -323,6 +323,324 @@ impl Default for SchemaTextField {
...
+        self
+    }
+
+    /// Set index missing. This allows searching for missing values - documents that do not contain a specific field.
+    pub fn index_missing(mut self, index_missing: bool) -> Self {
+        self.common = self.common.index_missing(index_missing);
+        self
+    }
+
+    // Sortable is not applicable to geo shape fields.
```
nihohit, line 606: remove

nihohit (commented):
minor comments

redis/src/commands/search/create/mod.rs
StefanPalashev: This test and the ones below it are precisely the examples from the website.
The purpose was to showcase how these examples could be built with the library's API.
Should they be removed?

nihohit: Experience shows that tests aren't a good way to present examples - users don't know where to look for those examples. Documentation works best, and for long examples, the `examples` folder.

nkaradzhov: Moved to docs. For brevity, left only the blog and json examples.

redis/src/commands/search/create/fields.rs
nkaradzhov: Done

nkaradzhov:
Hi @nihohit, both comments addressed :)

nihohit:
thanks!

PR #2318, opened 2026-08-21
Add test harness and CI support for the Search module [4/8]

StefanPalashev:
# Add test harness and CI support for the Search module

## Summary

Lets tests spawn a server with the Search module loaded and adds the integration tests for `FT.CREATE` covering every non-vector field type.

This is **part 4 of the series that supersedes #2015**, where the design is described in full.

## Notes on the layout

The tests are now named `test_module_search_*`, which means the existing `tcp`, `tcp_tls` and `unix` nextest profiles already exclude them through their `not test(test_module)` filters.
The `module_search` profile uses `binary(test_module_search*)`, so splitting the file per command later needs no config change either. 

## Testing

**8 integration tests**, carried over from #2015: an empty index name, a simple create, the full `CreateOptions` matrix, and one test per field type (`TEXT`, `TAG`, `NUMERIC`, `GEO`, `GEOSHAPE`). Each field-type test walks every modifier individually, then cumulatively, then adds each mutually exclusive modifier onto the fully combined base, and finally asserts that `NOINDEX` and `INDEXMISSING` are rejected together.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,6 +10,7 @@ env:
   CARGO_TERM_COLOR: always
   REDISRS_REDIS_JSON_PATH: "/tmp/librejson.so"
   REDISRS_REDIS_BLOOM_PATH: "/tmp/redisbloom.so"
+  REDISRS_REDIS_SEARCH_PATH: "/tmp/redisearch.so"
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.head_ref || github.run_id || github.ref }}"
@@ -34,8 +35,8 @@ jobs:
             {  db-org: redis, db-name: redis, db-version: 6.2.22 },
             {  db-org: redis, db-name: redis, db-version: 7.0.15, json-module-version: v2.6.22 },
             {  db-org: redis, db-name: redis, db-version: 7.4.9, json-module-version: v2.6.22 },
-            {  db-org: redis, db-name: redis, db-version: 8.0.6, json-module-version: bundled, bloom-module-version: bundled },
-            {  db-org: redis, db-name: redis, db-version: 8.8.0, json-module-version: bundled, bloom-module-version: bundled },
+            {  db-org: redis, db-name: redis, db-version: 8.0.6, json-module-version: bundled, bloom-module-version: bundled, search-module-version: bundled },
+            {  db-org: redis, db-name: redis, db-version: 8.8.0, json-module-version: bundled, bloom-module-version: bundled, search-module-version: bundled },
             # valkey-bloom has some quirks with Valkey <8.1, so we only test it on >=8.1
             {  db-org: valkey-io, db-name: valkey, db-version: 7.2.13, json-module-version: v2.6.22 },
             {  db-org: valkey-io, db-name: valkey, db-version: 8.0.9, json-module-version: v2.6.22 },
@@ -74,6 +75,11 @@ jobs:
         if: matrix.config.bloom-module-version == 'bundled'
         run: |
           cp "$HOME/db/redisbloom.so" "$REDISRS_REDIS_BLOOM_PATH"
+
+      - name: Copy bundled RediSearch into place
+        if: matrix.config.search-module-version == 'bundled'
+        run: |
+          cp "$HOME/db/redisearch.so" "$REDISRS_REDIS_SEARCH_PATH"
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # master
@@ -188,6 +194,10 @@ jobs:
       - name: Run RedisBloom tests
         if: matrix.config.bloom-module-version != ''
         run: make test-module-bloom
+
+      - name: Run Search module tests
+        if: matrix.config.search-module-version != ''
+        run: make test-module-search
 
       - name: Cache RedisJSON
         if: matrix.config.json-module-version != '' && matrix.config.json-module-version != 'bundled'
--- a/redis/tests/support/version.rs
+++ b/redis/tests/support/version.rs
@@ -17,6 +17,8 @@ pub const REDIS_CE_8_8: Component = ("redis", (8, 8, 0));
 pub const REDIS_JSON_8_8: Component = ("ReJSON", (8, 8, 0));
 pub const REDIS_BLOOM_ANY: Component = ("redis:bf", (0, 0, 0));
 
+pub const REDIS_SEARCH_8_0: Component = ("search", (8, 0, 0));
+
 // Valkey forked off at Redis 7.2.4 and still reports its Redis version 7.2.4. So tests that run
 // on Redis<=7.2.4 automatically also run on any Valkey server, and we only need version guards for
 // later versions.
@@ -338,7 +340,7 @@ macro_rules! skip_if_context_does_not_support {
 /// ```
 #[macro_export]
 macro_rules! run_test_if_version_supported {
-    ($component:expr) => {{ run_test_if_version_supported!($component, &[]) }};
+    ($component:expr) => {{ $crate::run_test_if_version_supported!($component, &[]) }};
     ($component:expr, $modules:expr) => {{
         let ctx = $crate::support::TestContext::with_modules($modules);
 
--- a/redis/tests/test_module_search.rs
+++ b/redis/tests/test_module_search.rs
@@ -1,0 +1,638 @@
+#![cfg(feature = "search_unfinished")]
+
+mod support;
+use crate::support::*;
+use redis::Commands;
+use redis::schema;
+use redis::search::*;
+use redis_test::server::Module;
+
+static TEXT_FIELD_NAME: &str = "title";
+static NUMERIC_FIELD_NAME: &str = "price";
+static TAG_FIELD_NAME: &str = "condition";
+static GEO_FIELD_NAME: &str = "location";
+static GEOSHAPE_FIELD_NAME: &str = "area";
+
+fn assert_no_index_and_index_missing_exclusivity_for_field(
+    result: redis::RedisResult<String>,
+    field_name: &str,
+) {
+    let server_error = redis::ServerError::try_from(result.unwrap_err()).unwrap();
+    assert!(server_error.details().is_some_and(|details| {
+        details.contains(
+            format!("cannot be defined with both `NOINDEX` and `INDEXMISSING` `{field_name}`")
+                .as_str(),
+        )
+    }));
+}
+
+#[test]
+fn test_module_search_ft_create_with_an_empty_index_name() {
+    let ctx = run_test_if_version_supported!(
+        [&[REDIS_CE_8_0][..], &[REDIS_SEARCH_8_0]],
+        &[Module::Search]
+    );
+    let mut con = ctx.connection();
+    let empty_index_name = "";
+    let options = CreateOptions::new();
+    let schema = schema! {
+        TEXT_FIELD_NAME => SchemaTextField::new()
+    };
+    // Check that the first call succeeds but the second one fails because the index already exists
+    assert_eq!(
+        con.ft_create(empty_index_name, &options, &schema),
+        Ok("OK".to_string())
+    );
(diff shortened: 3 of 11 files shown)


.github/workflows/rust.yml
```diff
@@ -34,8 +35,8 @@ jobs:
             {  db-org: redis, db-name: redis, db-version: 6.2.24 },
             {  db-org: redis, db-name: redis, db-version: 7.0.15, json-module-version: v2.6.22 },
             {  db-org: redis, db-name: redis, db-version: 7.4.11, json-module-version: v2.6.22 },
-            {  db-org: redis, db-name: redis, db-version: 8.0.6, json-module-version: bundled, bloom-module-version: bundled },
-            {  db-org: redis, db-name: redis, db-version: 8.8.2, json-module-version: bundled, bloom-module-version: bundled },
+            {  db-org: redis, db-name: redis, db-version: 8.0.6, json-module-version: bundled, bloom-module-version: bundled, search-module-version: bundled },
+            {  db-org: redis, db-name: redis, db-version: 8.8.2, json-module-version: bundled, bloom-module-version: bundled, search-module-version: bundled },
             # valkey-bloom has some quirks with Valkey <8.1, so we only test it on >=8.1
             {  db-org: valkey-io, db-name: valkey, db-version: 7.2.14, json-module-version: v2.6.22 },
             {  db-org: valkey-io, db-name: valkey, db-version: 8.0.10, json-module-version: v2.6.22 },
```
nihohit, line 42: please bump to 8.1.10 and add search support

.github/workflows/rust.yml
nihohit, line 44: please bump to valkey 9.1.2 and add search support

nihohit, line 43: and remove these docs :)

redis/tests/support/version.rs
```diff
@@ -17,6 +17,8 @@ pub const REDIS_CE_8_8: Component = ("redis", (8, 8, 0));
 pub const REDIS_JSON_8_8: Component = ("ReJSON", (8, 8, 0));
 pub const REDIS_BLOOM_ANY: Component = ("redis:bf", (0, 0, 0));
 
+pub const REDIS_SEARCH_8_0: Component = ("search", (8, 0, 0));
```
nihohit, line 20: make sure that this is also works for valkey search.

redis/tests/test_module_search.rs
```diff
@@ -0,0 +1,638 @@
...
+    let options = CreateOptions::new();
+    let schema = schema! {
+        TEXT_FIELD_NAME => SchemaTextField::new()
+    };
+    // Check that the first call succeeds but the second one fails because the index already exists
+    assert_eq!(
+        con.ft_create(empty_index_name, &options, &schema),
+        Ok("OK".to_string())
+    );
+    con.ft_create::<_, String>(empty_index_name, &options, &schema)
```
nihohit, line 46: check for the correct error

redis/tests/test_module_search.rs
```diff
@@ -0,0 +1,638 @@
...
+#[test]
+fn test_module_search_ft_create_schema_tag_field() {
+    let ctx = run_test_if_version_supported!(
+        [&[REDIS_CE_8_0][..], &[REDIS_SEARCH_8_0]],
+        &[Module::Search]
+    );
+    run_ft_create_schema_tag_field(&mut ctx.connection(), |_| {});
+}
+
+fn run_ft_create_schema_numeric_field<C, F>(con: &mut C, mut on_created: F)
```
nihohit, line 355: why is this a separate function?

redis/tests/test_module_search.rs
```diff
@@ -0,0 +1,638 @@
...
+    );
+}
+
+#[test]
+fn test_module_search_ft_create_schema_numeric_field() {
+    let ctx = run_test_if_version_supported!(
+        [&[REDIS_CE_8_0][..], &[REDIS_SEARCH_8_0]],
+        &[Module::Search]
+    );
+    run_ft_create_schema_numeric_field(&mut ctx.connection(), |_| {});
```
nihohit, line 446: why is this a separate function?

redis/tests/test_module_search.rs
```diff
@@ -0,0 +1,638 @@
...
+    );
+}
+
+#[test]
+fn test_module_search_ft_create_schema_geo_field() {
+    let ctx = run_test_if_version_supported!(
+        [&[REDIS_CE_8_0][..], &[REDIS_SEARCH_8_0]],
+        &[Module::Search]
+    );
+    run_ft_create_schema_geo_field(&mut ctx.connection(), |_| {});
```
nihohit, line 540: same, and below

redis/tests/test_module_search.rs
StefanPalashev: These helper functions extracted the common logic between regular and clustered tests.

Here's the [commit ](https://github.com/redis-rs/redis-rs/pull/2015/changes/6fbaa389ed29e689e555e46eafb2683dc41c3da2) from the original pull request that introduced the change.

Currently, these helper functions are only used by the regular tests, where `on_created` has an empty body. Once the clustered tests land in the final pull request of the series, they will use the same helpers to verify that the indices are propagated correctly across the cluster.

nkaradzhov:
Hi @nihohit, first of all sorry about the delay!

**Addressed:** the correct-error check; the `search` version guard now tells Redis and
Valkey apart (`redis:search` / `valkey:search`, like `bf` already does); the stale
"Skipping ValKey" comments are gone; Valkey is bumped to 8.1.10 and 9.1.2 (`main` went
to 8.0.11/9.0.6 — I kept your 8.1/9.1, tell me if you'd rather match); search runs on
the 9.1.2 row, built from source.

**Tests:** valkey-search does not support all redis search features — no `GEO`/`GEOSHAPE`, no
`SORTABLE UNF`, no `INDEXMISSING`/`NOINDEX`, and it rejects options like `FILTER`. So 6
of 8 tests fail on Valkey. I gate those six to Redis and run the two basic ones on both.
**Q1** — keep the gating, or split each test into a shared core plus a Redis-only rest?

**Build:** Valkey ships no prebuilt binary, so we have to build valkey-search from source, the same way we
already build valkey-bloom.
It is a heavy C++ build. To avoid compiling the world, I follow upstream's own
`ci/build_ubuntu.sh`: it downloads a prebuilt dependency package and builds with
`--use-system-modules`, so we do not build grpc/protobuf/abseil ourselves. I added
two apt packages the runner lacks — `git` and `libbenchmark-dev` (the dependency
package omits benchmark). CI caches the result per version. The 9.1.2 job builds
and loads the module cleanly.
**Q2:** is building from source acceptable here, or would you rather pull
the prebuilt `libsearch.so` out of the [`valkey-bundle` image](https://hub.docker.com/r/valkey/valkey-bundle)?

**Q3** — search on 8.1 too? It needs valkey-search 1.0.x, which requires a vector field
in every index and supports only `TAG`/`NUMERIC` — nothing here runs without a rewrite.

nihohit:
> Q1 — keep the gating, or split each test into a shared core plus a Redis-only rest?

Please either split the test to 2 tests, or add a `if valkey return` statement in the middle of the test.

> Q2: is building from source acceptable here, or would you rather pull
the prebuilt libsearch.so out of the valkey-bundle image?

Either is fine, as long as the result is cached and won't be rebuilt on the next attempt.

> search on 8.1 too? It needs valkey-search 1.0.x, which requires a vector field
in every index and supports only TAG/NUMERIC — nothing here runs without a rewrite.

I'm sorry, I don't understand the issue here. Are there some capabilities of search that are available only in valkey 9 and not in 8.1? Or are you saying that there was a breaking change between `valkey-search` 1.0 and 1.1, and your code only matches 1.1?

nkaradzhov:
@nihohit 
valkey-search 1.1+ requires valkey core 9.0.1+. So, on valkey 8.1 we can only load valkey-search 1.0.x (newest is 1.0.3).

[valkey-search grew a lot between 1.0 and 1.2:](https://github.com/valkey-io/valkey-search/releases/tag/1.1.0)
- 1.0.x requires a VECTOR field in every index — an FT.CREATE with no vector is rejected with "At least one attribute must be indexed as a vector". That requirement was removed in 1.1.0.
- 1.0.x supports only TAG and NUMERIC (plus VECTOR). TEXT arrived in 1.2; GEO/GEOSHAPE exist in no version yet.

Put together: valkey 8.1 → valkey-search 1.0.3 → needs a vector field and has no TEXT. Our FT.CREATE tests build vector-less indexes with TEXT/TAG/NUMERIC/GEO/GEOSHAPE fields, so 1.0.3 rejects them before we test anything.

The client works against both versions, but the test indexes need features 1.0.x doesn't have (schema-only indexes, TEXT). Covering 8.1 would mean a separate test path just for the old module: every index would need a vector field, and could use only TAG or NUMERIC.

Nevertheless it can be done, i just want to confirm with you whether its needed

nihohit:
ATM it isn't needed, but I would appreciate it if down the line the ability to create indexes with vectors will be added and tested against valkey-search 1.0.x

.github/workflows/rust.yml
nkaradzhov: as agreed, bumped, but search support will be added later

.github/workflows/rust.yml
nkaradzhov: done

nkaradzhov: done

redis/tests/support/version.rs
nkaradzhov: done

redis/tests/test_module_search.rs
nkaradzhov: done

redis/tests/support/version.rs
```diff
@@ -17,6 +17,9 @@ pub const REDIS_CE_8_8: Component = ("redis", (8, 8, 0));
 pub const REDIS_JSON_8_8: Component = ("ReJSON", (8, 8, 0));
 pub const REDIS_BLOOM_ANY: Component = ("redis:bf", (0, 0, 0));
 
+pub const REDIS_SEARCH_8_0: Component = ("redis:search", (8, 0, 0));
+pub const VALKEY_SEARCH_ANY: Component = ("valkey:search", (0, 0, 0));
```
nihohit: but ATM we don't support "any" valkey search - we only support 1.1.* and above, right? and we only test 1.2.*, so let's make this correct.

redis/tests/test_module_search.rs
```diff
@@ -0,0 +1,821 @@
...
+            },
+        ),
+        NUMERIC_FIELD_NAME,
+    );
+}
+
+#[test]
+fn test_module_search_ft_create_schema_numeric_field() {
+    // Redis-only: valkey-search rejects some modifiers (e.g. SORTABLE UNF).
+    let ctx = run_test_if_version_supported!(REDIS_SEARCH_8_0, &[Module::Search]);
```
nihohit: isn't numeric supported on valkey search?

redis/tests/test_module_search.rs
```diff
@@ -0,0 +1,821 @@
...
+            },
+        ),
+        TAG_FIELD_NAME,
+    );
+}
+
+#[test]
+fn test_module_search_ft_create_schema_tag_field() {
+    // Redis-only: valkey-search rejects some modifiers (e.g. SORTABLE UNF).
+    let ctx = run_test_if_version_supported!(REDIS_SEARCH_8_0, &[Module::Search]);
```
nihohit: isn't this supported on valkey search?

redis/tests/test_module_search.rs
```diff
@@ -0,0 +1,821 @@
...
+            },
+        ),
+        TEXT_FIELD_NAME,
+    );
+}
+
+#[test]
+fn test_module_search_ft_create_schema_text_field() {
+    // Redis-only: valkey-search rejects some modifiers (e.g. SORTABLE UNF).
+    let ctx = run_test_if_version_supported!(REDIS_SEARCH_8_0, &[Module::Search]);
```
nihohit: same

PR #2328, opened 2026-08-26
Make stream structs `non_exhaustive` (6/15)

somechris:
This is part of #2249

### DIFF
--- a/redis/src/commands/streams.rs
+++ b/redis/src/commands/streams.rs
@@ -174,6 +174,7 @@ impl ToRedisArgs for StreamTrimOptions {
 ///
 /// Supports idempotent message processing to prevent duplicate entries.
 /// See [Redis Streams Idempotency](https://redis.io/docs/latest/develop/data-types/streams/idempotency/)
+#[non_exhaustive]
 #[derive(Debug, Clone)]
 pub enum StreamIdempotencyMode {
     /// Manual mode: Producer provides both producer ID (pid) and idempotent ID (iid)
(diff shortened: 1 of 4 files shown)


redis/src/commands/streams.rs
```diff
@@ -836,6 +843,7 @@ pub struct StreamPendingCountReply {
 /// [`xinfo_stream`]: ../trait.Commands.html#method.xinfo_stream
 /// [`xinfo_stream_with_idempotency`]: ../trait.Commands.html#method.xinfo_stream_with_idempotency
 ///
+#[non_exhaustive]
 #[derive(Default, Debug, Clone)]
 pub struct StreamInfoStreamReply {
```
nihohit, line 848: now we can remove StreamInfoStreamReplyWithIdempotency - please fold that into StreamInfoStreamReply

somechris: I'm all for folding them up. But that's orthogonal no making structs `non_exhaustive` and would break the "commits should only do one thing" approach and thereby hurt retracability.
Can we push the folding into a separate commit?

nihohit: if you prefer that, sure. But that was a big reason for this change, so I didn't want to forget this :)

somechris: Yes, of course.

But it's listed in #2037 as first item. I guess it won't be missed.

And I figured that maybe Stefan may want to grab this, as he knows the history there.

@StefanPalashev  What do you think?

But if that does not work out, I'll fix that next week. :+1:

somechris: No response from @StefanPalashev

So I gave it a shot in #2352

Issue #1159, opened 2024-04-22
Redis TTL command returns overflow values on unsigned int


Onefox:
I am trying to get the
```
 // The redis command returns -2 if the key does not exist.
 // The redis command returns -1 if the key exists but has no associated expire.
```
values back from the libs ttl command.
```
  /// Get the expiration time of a key.
    fn ttl<K: ToRedisArgs>(key: K) {
        cmd("TTL").arg(key)
    }
```
When doing so I am getting back an overflow unsigned int so if using u16 I am getting back 65534 if the key doesn't exist in redis and 65535 if it doesn't have a ttl set.

I saw this comment, but it doesn't look like it was implemented that way or?
https://github.com/redis-rs/redis-rs/issues/158#issuecomment-413346421

Maybe I am also just not using it correctly, I wasn't able to find an example for it.

Onefox:
Sorry just figured out the proper way to get the correct values with a signed int

Onefox:
Sorry just figured out the proper way to get the correct values with a signed int

PR #2332, opened 2026-08-28
fix: reject out-of-range integer and double replies instead of wrapping them

jaideeppyne:
### Does your PR solve an issue?

Not formally, but #1159 ("Redis TTL command returns overflow values on unsigned int") is the same defect. It was closed by the reporter within ten minutes after they worked around it with a signed type, so the underlying problem was never fixed.

### Is this a breaking change?

Yes, in the Hyrum's Law sense: conversions that used to silently produce a wrong number now return an error. No conversion that previously produced a *correct* number changes. Retargeted at `2.0.x` and described in `version2.md`.

---

`FromRedisValue` for the integer types converts `Value::Int` (an `i64`) and `Value::Double` (an `f64`) with `as`. `as` wraps around for integers, and saturates (turning NaN into zero) for floats. Both do so silently.

The most visible consequence is `TTL`, which is documented to reply `-2` when the key does not exist and `-1` when it exists without an expiry:

```rust
let ttl: u64 = con.ttl("key")?;   // 18446744073709551615, no error
```

The same value parsed from a string reply already errors, so the outcome depended purely on how the reply happened to be encoded:

```rust
u64::from_redis_value(Value::Int(-1))                      // Ok(18446744073709551615)
u64::from_redis_value(Value::BulkString(b"-1".to_vec()))   // Err(...)
```

RESP3 makes the same split reachable for scores. `ZSCORE` replies with a double under RESP3 and a bulk string under RESP2, so the identical score converted differently depending on the negotiated protocol version:

```rust
i64::from_redis_value(Value::Double(3e40))                    // Ok(i64::MAX)
i64::from_redis_value(Value::BulkString(b"3e+40".to_vec()))   // Err(...)
i64::from_redis_value(Value::Double(f64::NAN))                // Ok(0)
```

`test_u32` already asserts that `-1` must be an error for a `u32`, but only covered the string arm, so the numeric arms were never checked.

This splits the numeric macro so the integer types range-check both numeric arms, which is what the `SimpleString` and `BulkString` arms already get from `str::parse`. `Value::Int` goes through `try_from`. `Value::Double` still truncates towards zero, but is bounds-checked first; the upper bound is a strict `<` against `MAX as f64 + 1.0` because `<int>::MAX as f64` rounds up to the next power of two for the wider types, and NaN and the infinities fall out of the same comparison. `f32` and `f64` are untouched, and in-range values, including the exact type boundaries, convert as before.

Two regression tests cover the integer and double arms; both fail before this change. Verified against a local redis 8.10.1: `test_types`, `test_basic`, `parser`, `test_bignum`, `test_script`, `test_geospatial` and `test_streams` all pass.

Disclosure: this was AI assisted — an agent found the inconsistency by differential-testing the numeric and string arms of `FromRedisValue` against each other, and drafted the patch; I verified the behaviour and the reasoning before submitting.

### DIFF
--- a/redis/src/types.rs
+++ b/redis/src/types.rs
@@ -1704,7 +1704,7 @@ fn get_owned_inner_value(v: Value) -> Value {
     }
 }
 
-macro_rules! from_redis_value_for_num_internal {
+macro_rules! from_redis_value_for_float_internal {
     ($t:ty, $v:expr) => {{
         let v = if let Value::Attribute {
             data,
@@ -1731,15 +1731,11 @@ macro_rules! from_redis_value_for_num_internal {
     }};
 }
 
-/// Same as `from_redis_value_for_num_internal`, but for integer types.
+/// Same as `from_redis_value_for_float_internal`, but for integer types.
 ///
-/// `Value::Int` holds an `i64` and `Value::Double` holds an `f64`. Neither can
-/// be converted to a narrower or differently signed integer with `as`, which
-/// wraps around for integers, and saturates (turning NaN into zero) for floats.
-/// Both do so silently, so `TTL` replying `:-1` used to be handed back as
-/// `u64::MAX` instead of as an error. These two arms are therefore
-/// range-checked, which is what the `SimpleString` and `BulkString` arms
-/// already get from `str::parse`.
+/// Every arm rejects a value that the requested type cannot represent, so that
+/// a reply converts to the same result whether the server sent it as a number
+/// or as a string.
 macro_rules! from_redis_value_for_int_internal {
     ($t:ty, $v:expr) => {{
         let v = if let Value::Attribute {
@@ -1768,18 +1764,17 @@ macro_rules! from_redis_value_for_int_internal {
                 Err(_) => crate::errors::invalid_type_error!(v, "Could not convert from string."),
             },
             Value::Double(val) => {
-                // Truncate towards zero, as `as` does, but reject anything that
-                // does not fit. `<$t>::MAX as f64` rounds *up* to the next power
-                // of two for the wider integer types, so the upper bound has to
-                // be a strict `<` against `MAX as f64 + 1.0`. NaN and the
-                // infinities fail both comparisons and are rejected too.
-                let truncated = val.trunc();
-                if truncated >= <$t>::MIN as f64 && truncated < <$t>::MAX as f64 + 1.0 {
-                    Ok(truncated as $t)
+                // `<$t>::MAX as f64` rounds *up* to the next power of two for the
+                // wider integer types, so the upper bound is a strict `<` against
+                // `MAX as f64 + 1.0`. NaN and the infinities fail both
+                // comparisons, and a fractional value is not an integer, so all
+                // three are rejected the same way `str::parse` rejects them.
+                if val.fract() == 0.0 && val >= <$t>::MIN as f64 && val < <$t>::MAX as f64 + 1.0 {
+                    Ok(val as $t)
                 } else {
                     crate::errors::invalid_type_error!(
                         v,
-                        "Double is out of range for the requested type."
+                        "Double is not an integer in range for the requested type."
                     )
                 }
             }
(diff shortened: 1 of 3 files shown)


redis/src/types.rs
```diff
@@ -1763,19 +1834,19 @@ impl FromRedisValue for u8 {
...
+from_redis_value_for_int!(i16);
+from_redis_value_for_int!(u16);
+from_redis_value_for_int!(i32);
+from_redis_value_for_int!(u32);
+from_redis_value_for_int!(i64);
+from_redis_value_for_int!(u64);
+from_redis_value_for_int!(i128);
+from_redis_value_for_int!(u128);
 from_redis_value_for_num!(f32);
 from_redis_value_for_num!(f64);
```
nihohit: if this is used only for floats, it should be renamed accordingly

redis/src/types.rs
```diff
@@ -1731,6 +1731,63 @@ macro_rules! from_redis_value_for_num_internal {
     }};
 }
 
+/// Same as `from_redis_value_for_num_internal`, but for integer types.
+///
+/// `Value::Int` holds an `i64` and `Value::Double` holds an `f64`. Neither can
```
nihohit: this looks like the AI habit of documenting previous behavior and why it was wrong. IMO it's not a good coding habit - the reader of the code doesn't see the previous behavior, and the tests should ensure that reverting to the previous implementation won't work, so this is unnecessary.

redis/tests/test_types.rs
```diff
@@ -230,6 +230,91 @@ mod types {
         }
     }
 
+    #[test]
+    fn test_out_of_range_integer_reply_is_an_error() {
+        for parse_mode in [RedisParseMode::Owned, RedisParseMode::Ref] {
+            // `TTL` replies with `:-1` when the key has no expiry and with `:-2`
+            // when it does not exist. Neither fits in an unsigned type, and used
+            // to wrap around to `u64::MAX` / `u64::MAX - 1` instead of erroring.
+            let bad: Result<u64, _> = parse_mode.parse_redis_value(Value::Int(-1));
```
nihohit, line 319: I'd like to test that the behavior of the string representation and the int/double representation is always the same. I think that this even warrants the addition of a dependency like `proptest`, in order to test a wide selection of values.

nihohit (commented):
Thanks, it's a solid beginning!

redis/src/types.rs
jaideeppyne: Renamed to `from_redis_value_for_float` / `from_redis_value_for_float_internal`. `f32` and `f64` are the only users left.

redis/src/types.rs
jaideeppyne: Fair, and you are right about where it came from. Rewritten to say what the macro does rather than what it used to do. The one bit of history I kept is the `MAX as f64 + 1.0` note, since the strict `<` looks like an off-by-one without it.

redis/tests/test_types.rs
jaideeppyne: Added, one property per integer type for the int arm and one for the double arm. I used `quickcheck` rather than `proptest` because it is already a dev-dependency here, so this needs no new dep. Happy to switch if you would rather standardise on proptest.

Writing it turned up two things.

The int arm already agreed everywhere, so those properties pass on the current code and only guard against regressions. The double arm did not: `Value::Double(1.9)` gave `Ok(1)` while `"1.9"` errored, because it truncated. Since you want the two encodings to always match, I made whole numbers the requirement and fractional doubles an error. That is a wider change than the PR started with, so say the word and I will put truncation back and carve the fractional case out of the property instead.

The other one is negative zero. `Double(-0.0)` is zero and converts fine to `u64`, but it writes as `"-0"` and `str::parse` refuses a sign on an unsigned type. That looked like a text artifact rather than a real disagreement about the number, so I normalised it in the test instead of touching the conversion.

Reverting `types.rs` to the base with the new tests in place fails 19 of them.

redis/tests/test_types.rs
```diff
@@ -230,6 +230,86 @@ mod types {
...
+
+        macro_rules! agree_on_double {
+            ($name:ident, $t:ty) => {
+                quickcheck! {
+                    fn $name(val: f64) -> bool {
+                        let from_double = <$t>::from_redis_value(Value::Double(val));
+                        // Neither a non-finite nor a fractional double is an
+                        // integer, and no integer string spells one, so the only
+                        // thing to check is that both are refused.
+                        if !val.is_finite() || val.fract() != 0.0 {
```
nihohit: why don't we compare to the string behavior here?

jaideeppyne: Good point, we can. Every double has a spelling, and the ones that are not whole numbers do not parse as an integer, so both arms are compared directly now and the early return is gone. NaN and the infinities go through the same path `1.9` does.

The one case still handled separately is negative zero: it is zero, but it writes as "-0" and `str::parse` refuses a sign on an unsigned type, so the comparison uses the spelling of the value rather than of the sign.

nihohit (approved):

PR #2337, opened 2026-08-28
fix: accept the RESP3 array-of-pairs shape in map conversions

jaideeppyne:
RESP3 returns `ZRANGE ... WITHSCORES` as an array of `[member, score]` pairs where RESP2 returned one flat array. `Vec<(String, f64)>` handles both, because `from_redis_values` checks `is_collection_of_len`. `HashMap<String, f64>` only handles the flat one, so the same command on the same data parses under RESP2 and errors under RESP3.

```rust
let v = parse_redis_value(b"*2\r\n*2\r\n$1\r\na\r\n,1\r\n*2\r\n$1\r\nb\r\n,2\r\n").unwrap();
Vec::<(String, f64)>::from_redis_value(v.clone()); // Ok
HashMap::<String, f64>::from_redis_value(v);       // Err "Response type not string compatible."
```

Same for ZREVRANGE, ZRANGEBYSCORE, ZREVRANGEBYSCORE, ZPOPMIN/ZPOPMAX with a count, ZRANDMEMBER WITHSCORES, and ZDIFF/ZUNION/ZINTER WITHSCORES. I checked all ten against a live 8.10.1.

There is a structural bug underneath it too: on an even length reply `as_map_iter` pairs adjacent pairs with each other, so `[[a,1],[b,2]]` gives `key=["a",1], value=["b",2]`. It only avoids handing back wrong data because the leaf conversion errors afterwards. `HashMap<Vec<String>, Vec<String>>` does return `{["a","1"]: ["b","2"]}` today.

This teaches `as_map_iter`/`into_map_iter` the nested pair shape using the same `is_collection_of_len(2)` check the tuple path already uses. Both iterator enums are `#[non_exhaustive]`, so the new variant is not breaking. An empty array still parses as an empty map.

I found it by running each command on a RESP2 and a RESP3 connection against the same server state and diffing the parsed results, the same way as #2332.

One behaviour change worth naming: a key or value type that can absorb a two element array, like `HashMap<Vec<String>, Vec<String>>`, previously got `{["a","1"]: ["b","2"]}` from `[[a,1],[b,2]]` and now gets `{["a"]:["1"], ["b"]:["2"]}`. Happy to move this to `2.0.x` with a `version2.md` note if you would rather have it there.

AI assistance disclosure: I used Claude to build the differential harness and draft the patch. I ran and reviewed everything myself.

### DIFF
--- a/redis/tests/test_types.rs
+++ b/redis/tests/test_types.rs
@@ -640,11 +640,9 @@ mod types {
     /// into a map type identically. `Vec<(K, V)>` already handled both.
     #[test]
     fn test_map_from_resp3_array_of_pairs() {
-        use std::collections::HashMap;
+        type MapType = HashMap<String, f64>;
 
-        type Hm = HashMap<String, f64>;
-
-        let expected: Hm = [("a".to_string(), 1.0), ("b".to_string(), 2.5)]
+        let expected: MapType = [("a".to_string(), 1.0), ("b".to_string(), 2.5)]
             .into_iter()
             .collect();
 
@@ -663,11 +661,11 @@ mod types {
 
         for parse_mode in [RedisParseMode::Owned, RedisParseMode::Ref] {
             assert_eq!(
-                parse_mode.parse_redis_value::<Hm>(resp2.clone()),
+                parse_mode.parse_redis_value::<MapType>(resp2.clone()),
                 Ok(expected.clone())
             );
             assert_eq!(
-                parse_mode.parse_redis_value::<Hm>(resp3.clone()),
+                parse_mode.parse_redis_value::<MapType>(resp3.clone()),
                 Ok(expected.clone())
             );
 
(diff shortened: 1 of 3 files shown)


redis/tests/test_types.rs
```diff
@@ -462,6 +462,73 @@ mod types {
         }
     }
 
+    /// RESP3 replaces the flat `[member, score, member, score]` array RESP2 used
+    /// for `ZRANGE ... WITHSCORES` (and ZPOPMIN/ZPOPMAX/ZRANDMEMBER/ZDIFF/ZUNION/
+    /// ZINTER) with an array of `[member, score]` pairs. Both shapes must convert
+    /// into a map type identically. `Vec<(K, V)>` already handled both.
+    #[test]
+    fn test_map_from_resp3_array_of_pairs() {
+        use std::collections::HashMap;
```
nihohit: nit: move outside of this test

redis/tests/test_types.rs
```diff
@@ -462,6 +462,73 @@ mod types {
...
+
+        type Hm = HashMap<String, f64>;
```
nihohit: MapType is better than a two letter acronym

nihohit (commented):
minor comments

redis/tests/test_types.rs
jaideeppyne: Dropped it, `HashMap` is already imported at the top of the module so it was redundant.

redis/tests/test_types.rs
jaideeppyne: Renamed to `MapType`.

I only changed it inside this test. `test_types.rs` has two other tests using a `Hm` alias already, and renaming those looked like unrelated churn for this PR. Happy to do them in a separate one if you want it consistent.

PR #2341, opened 2026-08-28
Mark `ReplicaInfo` as `non_exhaustive` (10/15)

somechris:
This is part of #2249


redis/src/types.rs
```diff
@@ -862,6 +863,17 @@ pub struct ReplicaInfo {
     pub replication_offset: i64,
 }
 
+impl ReplicaInfo {
+    /// Builds a new instance
+    pub fn new<S: Into<String>>(ip: S, port: u16, replication_offset: i64) -> Self {
```
nihohit, line 868: if we move the tests here, we could make this not public :)

(not for this PR)

PR #2351, opened 2026-08-31
Drop unused Rust flags

somechris:
In 7060272 (pre-allocation of commands) unused Rust flags got added, which cargo complains about.

We drop them again.


nihohit (approved):
thanks, that was used for the benchmarks, and was left in accidentally.

PR #2352, opened 2026-08-31
Clean up idempotency for streams

somechris:
`StreamInfoStreamReplyWithIdempotency` got added in `4f11b5a` to allow accessing idempotency fields while staying backwards compatible on the v1 branch.

As v2 is now prepared, backwards-compatibility is no longer a concern, and we clean the idempotency structures up.

### DIFF
--- a/redis/src/commands/streams.rs
+++ b/redis/src/commands/streams.rs
@@ -831,20 +831,42 @@ pub struct StreamPendingCountReply {
     pub ids: Vec<StreamPendingId>,
 }
 
-/// Reply type used with [`xinfo_stream`] command, containing
-/// general information about the stream stored at the specified key.
+/// Reply type used with [`xinfo_stream`] command
+///
+/// It contains general information about the stream stored at the specified key.
 ///
 /// The very first and last IDs in the stream are shown,
 /// in order to give some sense about what is the stream content.
 ///
-/// **Note:** For Redis 8.6+ idempotency tracking fields, use [`StreamInfoStreamReplyWithIdempotency`]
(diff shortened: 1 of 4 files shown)


redis/src/commands/streams.rs
```diff
@@ -831,20 +831,42 @@ pub struct StreamPendingCountReply {
...
+///
+/// // Access idempotency tracking (Only if IDMP was configured for the stream)
+/// println!("Producers tracked: {}", info.pids_tracked);
+/// println!("Idempotent IDs tracked: {}", info.iids_tracked);
+/// println!("Duplicates prevented: {}", info.iids_duplicates);
+/// ```
+// Idempotency fields are optional. Wrapping them by `Option` would mean lots of manual, unnecessary
+// unpacking for callers, as they typically know whether streams are configured to produce relevant
+// IDMP values or not. Also, the `Option` would be inconsistent with other fields. Hence, the IDMP
+// values are not wrapped by `Option`.
```
somechris, line 867: Instinctively, the new `idmp_duration` and other values should be `Option<u32>` instead of `u32`.

And when looking at how the parsing works, this would hold true also for all the old values (`last_generated_value`, ...). But they are plain types as well.

So to stay consistent, the new values are plain `u32` as well.

/me shrugs

nihohit (approved):

redis/src/commands/streams.rs
nihohit: sometimes shrugging is all we have left 
¯\_(ツ)_/¯

PR #2359, opened 2026-08-31
Mark `AclInfo` as `non_exhaustive` (11/15)

somechris:
This is part of #2249

### DIFF
--- a/redis/tests/test_acl.rs
+++ b/redis/tests/test_acl.rs
@@ -1,11 +1,10 @@
 #![cfg(feature = "acl")]
 
 use redis::TypedCommands;
-use redis::acl::{AclInfo, Rule};
+use redis::acl::Rule;
 use redis_test::REDIS_CE_7_2;
 use redis_test::TestContext;
 use redis_test::run_test_if_version_supported;
-
 use std::collections::HashSet;
 
 mod support;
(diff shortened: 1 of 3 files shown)


redis/tests/test_acl.rs
```diff
@@ -59,22 +58,24 @@ fn test_acl_getsetdel_users() {
         Ok(())
     );
     let acl_info = con.acl_getuser("bob").expect("Got user").unwrap();
+    assert_eq!(acl_info.flags, vec![Rule::On]);
     assert_eq!(
```
nihohit, line 62: I'd prefer equality over `acl_info` and not its fields, because then if we add fields, they're added automatically to the equality check.

somechris: Argh! I forgot to add the comment about this PR when I uploaded it yesterday. head -> desk.

So, yes, an equality based check would be better. Agreed, of course. :+1:

But the comment that I forgot that tilts this, is about the surroundings of this test:

```
//TODO: do we need this test?
#[test]
#[ignore]
fn test_acl_getsetdel_users() {
```

So I wanted to ask if we care about this test at all, as it's `ignore`d and has a comment about whether we need it. And that's not fresh, it's from 2021.

Should we remove it?

And if not, is this reasonable to add the builder functions only for this unused test? (they aren't needed elsewhere)

I'd say no, and hence made the test compile without needing adaptions in production code.

If we prefer a proper equality test because some users might need the builder thing anyway, then I'll gladly add it.

(But I'd vote for removing the test, as it's a needless maintenance burden if it's `ignore`d)

nihohit: I think that the test can be removed. It was ignored since before I started maintaining this, and it seems like it tests the logic of the server more than it tests the library.

PR #2360, opened 2026-08-31
redis-test: Harmonize server aliveness check and error messages

somechris:
In both `RedisCluster` and `TestContext`, we checked if a server is alive and reported an error otherwise.

We implement such functionality directly in `RedisServer` to avoid having to manually use the cumbersome process API, and to simplify formatting error messages upon issues.

The error messages now additionally show the command, stdout, stderr, etc. This helps to diagnose server issues.

### DIFF
--- a/redis-test/src/cluster.rs
+++ b/redis-test/src/cluster.rs
@@ -278,37 +278,27 @@ impl RedisCluster {
         };
 
         let verify_server = |server: &mut RedisServer| {
-            let process = &mut server.process;
-            match process.try_wait() {
-                Ok(Some(status)) => {
-                    let log_file_contents = server.log_file_contents();
-                    let err = format!(
-                        "redis server creation failed with status {status:?}.\nlog file: {log_file_contents:?}"
-                    );
-                    Err(err)
-                }
-                Ok(None) => {
-                    // wait for 10 seconds for the server to be available.
-                    let max_attempts = 200;
-                    let mut cur_attempts = 0;
-                    loop {
-                        if cur_attempts == max_attempts {
-                            let log_file_contents = server.log_file_contents();
-                            break Err(format!(
-                                "redis server creation failed: Address {} closed. {log_file_contents:?}",
-                                server.addr
-                            ));
-                        } else if port_in_use(&server.addr.to_string()) {
-                            break Ok(());
-                        }
-                        eprintln!("Waiting for redis process to initialize");
-                        sleep(Duration::from_millis(50));
-                        cur_attempts += 1;
+            if server.is_alive() {
+                // wait for 10 seconds for the server to be available.
+                let max_attempts = 200;
+                let mut cur_attempts = 0;
+                loop {
+                    if cur_attempts == max_attempts {
+                        let process_info = server.stop_with_info();
+                        break Err(format!(
+                            "redis server creation failed: Address {} closed.\n{process_info}",
+                            server.addr
+                        ));
+                    } else if port_in_use(&server.addr.to_string()) {
+                        break Ok(());
                     }
+                    eprintln!("Waiting for redis process to initialize");
+                    sleep(Duration::from_millis(50));
+                    cur_attempts += 1;
                 }
-                Err(e) => {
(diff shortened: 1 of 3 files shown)


redis-test/src/cluster.rs
```diff
@@ -278,37 +278,27 @@ impl RedisCluster {
...
-                                "redis server creation failed: Address {} closed. {log_file_contents:?}",
-                                server.addr
-                            ));
-                        } else if port_in_use(&server.addr.to_string()) {
-                            break Ok(());
-                        }
-                        eprintln!("Waiting for redis process to initialize");
-                        sleep(Duration::from_millis(50));
-                        cur_attempts += 1;
+            if server.is_alive() {
```
nihohit, line 281: why check this once outside of the loop, instead of as the first action in every iteration of the loop?

redis-test/src/server.rs
```diff
@@ -368,17 +380,111 @@ impl RedisServer {
...
-            fs::remove_file(path).ok();
+    /// Stops the server (if running) and optionally yields formatted process information
+    ///
+    /// It's safe to call this function on an already stopped server.
+    ///
+    /// If `format_process_infos` is `true` for an instance's first call to this method, the
+    /// returned string holds the formatted process information (exit code, ...), which is useful
+    /// to build error messages.
+    fn stop_internal(&mut self, format_process_infos: bool) -> String {
+        if let Some(mut process) = self.process.take() {
```
nihohit: please use `let Some(...) else { ... }`, in order to reduce indentation.

redis-test/src/server.rs
```diff
@@ -368,17 +380,111 @@ impl RedisServer {
...
-        if let redis::ConnectionAddr::Unix(ref path) = *self.client_addr() {
-            fs::remove_file(path).ok();
+    /// Stops the server (if running) and optionally yields formatted process information
+    ///
+    /// It's safe to call this function on an already stopped server.
+    ///
+    /// If `format_process_infos` is `true` for an instance's first call to this method, the
+    /// returned string holds the formatted process information (exit code, ...), which is useful
+    /// to build error messages.
+    fn stop_internal(&mut self, format_process_infos: bool) -> String {
```
nihohit: nit: infos -> info. there's no plural form :)

redis-test/src/cluster.rs
somechris: Because I wanted to keep the diff simple and wanted to improve the reporting, not change the logic.

(It was checked only once also in the old code.)

But maybe this and `TestContext::from_server` would benefit from a `RedisServer::wait_until_ready` method to de-dupe this (in a follow-up commit)? ... or even add that directly to building `RedisServer`?

redis-test/src/server.rs
somechris: Thanks. Done in 924f710 :+1:

redis-test/src/server.rs
somechris: In a strict grammatical sense, you're of course right :+1:

In a more lenient software engineering sense thought ... 😉

Regardless: Done in 924f710 :+1:

And I've uploaded #2374 to change previous "infos" use to "info".

redis-test/src/cluster.rs
nihohit: I think that `wait_until_ready` certainly could be helpful.
As for adding that directly when building RedisServer - intuitively that makes sense, but concretely, what's the change? Is it large or small? how much of the control flow does it change?

somechris: I think this could start small (offering to loop until it the server comes up) or clean up more things.

I'm convinced that "improving how the 'server is up' detection works" is worth it. And a few times already, I thought about raising it.

But as it is orthogonal to what this PR is trying to do (improving on reporting and harmonizing process exit detection), I started [a fresh issue (#2377)](https://github.com/redis-rs/redis-rs/issues/2377) for such improvements.

This will keep separate things separate and make the discussion more visible.

nihohit (approved):
thanks!

PR #2375, opened 2026-09-02
redis-test/RedisServer: Reduce server disk access

somechris:
Ideally, we'd like to prohibit all disk access to ease testing. But that's not portable, as for example Redis allows to disable loading dumps by setting `dbfilename` to ``, while Valkey prohibits this.

So we at least get rid of disk access during replication boot-up.


nihohit (approved):
every little bit helps :)

PR #2381, opened 2026-09-04
Mark `ReadCandidates` as `non_exhaustive` (12/15)

somechris:
This is part of #2249

### DIFF
--- a/redis/src/cluster_handling/read_routing/interface.rs
+++ b/redis/src/cluster_handling/read_routing/interface.rs
@@ -177,6 +177,10 @@ impl<'a> ReplicasOnlyCandidates<'a> {
 /// The strategy is only called when there are replicas available for the
 /// target slot. If a slot has no replicas, the caller falls back to the
 /// primary without consulting the strategy.
+#[allow(
+    clippy::exhaustive_enums,
+    reason = "The effects of marking this 'non_exhaustive' are not ideal. See #2390."
+)]
 #[derive(Debug)]
 pub enum ReadCandidates<'a> {
     /// Any node (primary or replica) is acceptable for this read.


redis/examples/latency-aware-routing.rs
```diff
@@ -118,6 +118,7 @@ impl ReadRoutingStrategy for LatencyAwareStrategy {
                 .iter()
                 .min_by_key(|node| latency_of(node))
                 .expect("replicas is non-empty"),
+            &_ => panic!("Unsupported candidate requirement: {candidates:?}"),
```
nihohit, line 121: this makes me think that maybe this type shouldn't be marked as non_exhaustive. I need to read up on best practices, but expecting users to implement a trait handling a value without either an error  response or the ability to exhaustively iterate over the value seems wrong to me.

somechris: Ok, cool.

Then let's skip 12 for now and continue with 13, and 14, as they do not need 12. :+1:

(The final commit 15, which enables linting will have to wait on 12)

nihohit: Yes, I think that this is wise. I'm not sure that this is a bad suggestion, but maybe it requires changing the trait to require a result instead of always returning a value.

somechris: I've created #2390 to track this and will upload a commit to skip the lint for `ReadCandidates`.

This gives some more time to decide on `ReadCandidates`, but will allow us to enable the lints in the last commit of this series (which will guide against `non_exhaustive` regressions)

somechris: > [...] and will upload a commit to skip the lint for `ReadCandidates`.

Done in 447a33f

PR #2382, opened 2026-09-04
ci: Switch to `REDISRS_MODULE_<NAME>_PATH` for modules

### LINKED ISSUE #2373: `REDISRS_REDIS_JSON_PATH` (and other module paths) duplicate `REDIS` and lack a module markers
As we're preparing v2: What about finding a more descriptive name for module env vars that is less focused on Redis and more on what the modules achieve?

---

There were two `REDIS` in the original `REDIS_RS_REDIS_JSON_PATH`.

And that's ok, as the first `REDIS` is because it's for `redis-rs`, the second as it was the path to _Redis_' JSON module.

When adding the bloom filter, we also supported only _Redis_' Bloom module at first. So we also duplicated `REDIS` in `REDIS_RS_REDIS_BLOOM_PATH` to match the JSON module's env var name.

Then support for Valkey's bloom filter got added. And to avoid breaking things, the env var name was kept.

But using `REDIS_RS_REDIS_BLOOM_PATH` for Valkey's Bloom module is of course a misnomer. 💔

But with v2, we could switch to a name that's a more neutral, and a better match 🤔 

(And that would be great before we add search module support)

We could simply drop the latter `REDIS`. We could use a common prefix, so they'd sort better in a table (e.g.: DEVELOPMENT.md). And maybe add a `MODULE` so it's clear that this is a module.

E.g.:

* `REDIS_RS_JSON_PATH` and `REDIS_RS_BLOOM_PATH`: Are neutral, read nicely, but do not sort too well and lack an indicator that they are modules.
* `REDIS_RS_MODULE_PATH_JSON` and `REDIS_RS_MODULE_PATH_BLOOM`: Are neutral, sort nicely, and read ok-ish. It fits `REDIS_RS_SERVER_BIN` in that the third part is `SERVER` there and `MODULE` here. So it clearly says what it is for.
* `REDIS_RS_MODULE_JSON_PATH` and `REDIS_RS_MODULE_BLOOM_PATH`: Same as above, but `PATH` comes last. This might help us if at some point we need to hint CI about the module vendor. Then we could use `REDIS_RS_MODULE_JSON_VENDOR` (or `..._KIND`) and they would still sort nicely.

Maybe any of the above with `MOD` instead of the longer `MODULE`:

* `REDIS_RS_MOD_PATH_JSON` and `REDIS_RS_MOD_PATH_BLOOM`,
* `REDIS_RS_MOD_JSON_PATH` and `REDIS_RS_MOD_BLOOM_PATH`.

WDYT?

(I'd prefer anything without doubled `REDIS` and with some module marker. My favourite would be `REDIS_RS_MODULE_JSON_PATH`)

somechris:
The previous `REDISRS_REDIS_<NAME>_PATH` lacked that the environment variables were for _modules_ and by the duplicated `REDIS` suggested that only _Redis_ modules were supported. But Valkey modules are supported as well.

Fixes #2373

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -59,8 +59,8 @@ The tests need to be able to find Redis' tools and the modules. If automatic det
 | EnvVar Name | Description |
 | --- | --- |
 | `REDISRS_SERVER_BIN` | Binary to start Redis |
-| `REDISRS_REDIS_BLOOM_PATH` | Path to the `bloom` module |
-| `REDISRS_REDIS_JSON_PATH` | Path to the `json` module |
+| `REDISRS_MODULE_BLOOM_PATH` | Path to the `bloom` module |
+| `REDISRS_MODULE_JSON_PATH` | Path to the `json` module |
 
 ### Speeding up TLS tests
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,8 +8,8 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  REDISRS_REDIS_JSON_PATH: "/tmp/librejson.so"
-  REDISRS_REDIS_BLOOM_PATH: "/tmp/redisbloom.so"
+  REDISRS_MODULE_JSON_PATH: "/tmp/librejson.so"
+  REDISRS_MODULE_BLOOM_PATH: "/tmp/redisbloom.so"
   REDISRS_TLS_KEY_CA: "/tmp/tls-ca.key"
   REDISRS_TLS_KEY_SERVER: "/tmp/tls-server.key"
   REDISRS_TLS_KEY_CLIENT: "/tmp/tls-client.key"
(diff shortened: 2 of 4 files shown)


DEVELOPMENT.md
```diff
@@ -58,9 +58,9 @@ The tests need to be able to find Redis' tools and the modules. If automatic det
 
 | EnvVar Name | Description |
 | --- | --- |
+| `REDISRS_MODULE_BLOOM_PATH` | Path to the `bloom` module |
+| `REDISRS_MODULE_JSON_PATH` | Path to the `json` module |
 | `REDISRS_SERVER_BIN` | Binary to start Redis |
-| `REDISRS_REDIS_BLOOM_PATH` | Path to the `bloom` module |
-| `REDISRS_REDIS_JSON_PATH` | Path to the `json` module |
```
somechris, line 63: I'd like to have `REDIS_SERVER_BIN` first, as it's probably the most relevant.
But I'd also like to have the table sorted alphabetically by name.

As having them sorted helps when looking for things, the sorting weighed heavier for me. If you prefer otherwise, let me know and I'll switch.

nihohit: IMO logical importance (`REDIS_SERVER_BIN` first) is better, but I'm fine with other orderings, too.

somechris: Done in 9191a24 :+1:

PR #2386, opened 2026-09-06
commands: Make script commands first class citizens

somechris:
The script commands were generated within `Script` and
`ScriptInvocation`. This made them stick out and required extra
handling, as they failed to hit the fast path, as they were not a simple
`ready_cmd`.
    
We convert them to `ready_cmd`, so they can benefit from the
optimizations of 7060272.

### DIFF
--- a/redis/src/commands/mod.rs
+++ b/redis/src/commands/mod.rs
@@ -2911,8 +2911,8 @@ assert_eq!(invok_res, 3);
 "##)]
     #[cfg(feature = "script")]
     #[cfg_attr(docsrs, doc(cfg(feature = "script")))]
-    fn load_script<>(script: &'a crate::Script) -> Generic {
-        script.load_cmd().take()
+    fn load_script<>(script: &'a crate::Script) -> String {
+        ready_cmd!("SCRIPT", "LOAD", script).take()
     }
 
     /// Invoke a prepared script.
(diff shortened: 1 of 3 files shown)


version2.md
```diff
@@ -10,6 +10,10 @@ redis = "2"
 
 ## Breaking Changes
 
+### Script loading on typed connections returns `String` (Breaking change)
+
+The `script_load` command on typed connections returned a generic `Value`. But as Redis returns a string, the function now also directly returns a `String`.
```
nihohit: the function -> the typed version of the function?

I'd like to be clear that if you use the generic version nothing changed. not sure my suggested phrasing is clear, though.

redis/src/commands/mod.rs
```diff
@@ -2916,8 +2916,8 @@ assert_eq!(invok_res, 3);
 "##)]
     #[cfg(feature = "script")]
     #[cfg_attr(docsrs, doc(cfg(feature = "script")))]
-    fn load_script<>(script: &'a crate::Script) -> Generic {
-        script.load_cmd().take()
+    fn load_script<>(script: &'a crate::Script) -> String {
+        cmd("SCRIPT").arg("LOAD").arg(script).take()
```
nihohit: this doesn't use ready_cmd, although the PR description says it should.

version2.md
somechris: Done in 82b9855 :+1:

redis/src/commands/mod.rs
somechris: Argh?! :fearful:  Done in 82b9855

And additionally added `ToSingleRedisArg` and `args_size`  to `Script`, and switched `ScriptInvocation` as well.

I think with this commit, all commands end in `ready_cmd` (but not all take the fast path yet).

nihohit (approved):

PR #2387, opened 2026-09-06
commands: Eliminate remaining direct `cmd` from `implement_commands`

somechris:
By switching the `cmd`s to single `ready_cmd`, they benefit from the allocation optimizations of 7060272.


nihohit (approved):
thanks!

PR #2392, opened 2026-09-11
redis-test: Parametrize server type and protocol for matrix-based tests

nihohit:
Expose the previously env-var-inferred server setup as explicit options so procedural macros can drive the test matrix:
- make `ServerType` public, add `RedisServerBuilder::server_type` and `TestContextBuilder::server_type`/`protocol`
- add `RedisServer::connection_info_with_protocol` and `RedisServer::get_addr_for_type`
- add `ClusterType` handling to `RedisClusterConfiguration`
- add server-type-aware `RedisSentinelCluster` constructors
- make `use_protocol` report explicitly unset env vars (`Option`)

Update the test support layer to the new API and adapt existing callers of the changed signatures.

This is the first step in the effort to reduce test complexity, see the full change in 
https://github.com/redis-rs/redis-rs/pull/2344

### DIFF
--- a/redis-test/src/server.rs
+++ b/redis-test/src/server.rs
@@ -9,16 +9,17 @@ use crate::utils::{
     CommandMultiArgs, TlsFilePaths, build_keys_and_certs_for_tls, get_random_available_port,
 };
 
-pub fn use_protocol() -> ProtocolVersion {
-    if env::var("PROTOCOL").unwrap_or_default() == "RESP3" {
-        ProtocolVersion::RESP3
-    } else {
-        ProtocolVersion::RESP2
+pub fn use_protocol() -> Option<ProtocolVersion> {
+    match env::var("PROTOCOL").ok().as_deref() {
+        Some("RESP2") => Some(ProtocolVersion::RESP2),
+        Some("RESP3") => Some(ProtocolVersion::RESP3),
+        Some(val) => panic!("Unknown protocol {val:?}"),
+        None => None,
     }
 }
 
 pub fn redis_settings() -> RedisConnectionInfo {
-    RedisConnectionInfo::default().set_protocol(use_protocol())
+    RedisConnectionInfo::default().set_protocol(use_protocol().unwrap_or(ProtocolVersion::RESP2))
 }
 
 /// Get the default host to use for TCP connections.
@@ -26,9 +27,11 @@ pub fn get_default_host() -> String {
     "127.0.0.1".to_string()
 }
 
-#[derive(PartialEq)]
-enum ServerType {
-    Tcp { tls: bool },
+#[derive(PartialEq, Clone, Copy, Debug)]
+#[non_exhaustive]
+pub enum ServerType {
+    Tcp,
+    TcpTls,
     Unix,
 }
 
@@ -56,6 +59,7 @@ pub enum Module {
 // considerations.
 #[derive(Default)]
 pub struct RedisServerBuilder {
+    server_type: Option<ServerType>,
     address: Option<ConnectionAddr>,
     config_file: Option<PathBuf>,
     cert_auth_field: Option<String>,
@@ -68,6 +72,11 @@ impl RedisServerBuilder {
     /// Starts a fresh builder
     pub fn new() -> Self {
         Default::default()
+    }
+
+    pub fn server_type(mut self, server_type: ServerType) -> Self {
+        self.server_type = Some(server_type);
+        self
     }
 
     pub fn address(mut self, address: ConnectionAddr) -> Self {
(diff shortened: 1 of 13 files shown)


redis-test/src/server.rs
```diff
@@ -9,26 +9,29 @@ use crate::utils::{
     CommandMultiArgs, TlsFilePaths, build_keys_and_certs_for_tls, get_random_available_port,
 };
 
-pub fn use_protocol() -> ProtocolVersion {
-    if env::var("PROTOCOL").unwrap_or_default() == "RESP3" {
-        ProtocolVersion::RESP3
-    } else {
-        ProtocolVersion::RESP2
+pub fn use_protocol() -> Option<ProtocolVersion> {
```
somechris, line 12: I probably miss the obvious here, and I'm all for giving users access to raw data, but we still always default to `RESP2`, right?

But now, the defaulting is more explicit as the caller has to do it.

For me, that hurts readability a bit.

So I wondered if this change is needed or is it an artifact from how the whole series got created?

---

I checked the current full branch at 1699fff and still could only find defaulting to RESP2.

Most cases default directly via `use_protocol().unwrap_or(redis::ProtocolVersion::RESP2)`

The other cases are:

* `test_context.rs:162`

  ```
  redis-test/src/test_context.rs-161-        let protocol = protocol
  redis-test/src/test_context.rs:162:            .or_else(use_protocol)
  redis-test/src/test_context.rs-163-            .unwrap_or(ProtocolVersion::RESP2);
  ```

  Which defaults to `use_protocol()` then defaults to `RESP2`.

  So it is the same as `protocol.unwrap_or_else(|| use_protocol().unwrap_or(ProtocolVersion::RESP2))` and again matches the above defaulting.

* `bench_cache.rs:98`

  ```
  redis/benches/bench_cache.rs-97-fn bench_cache(c: &mut Criterion) {
  redis/benches/bench_cache.rs:98:    if use_protocol() != Some(redis::ProtocolVersion::RESP3) {
  redis/benches/bench_cache.rs-99-        return;
  ```

  Here we only check if we're using `RESP3`. That's possible regardless of where we default to `RESP2`


If we left `use_protocol` untouched, is there some use case we don't cover?

nihohit: Just to make sure that I understand your question - you're asking why the function returns an option, and doesn't return a default value if no env variable is set?

The reason is that the macros should generate the RESP2 & RESP3 variants (or the tcp, tls, unix variants) only if the env variables aren't set. If they are set, then the macro should only generate the variant that matches the env variable - that means that we need to be able to differentiate between the default and RESP2 being set by the env var.

BTW, ATM there bugs in the base branch, it's still very much WIP, and as you can see this branch doesn't pass, too.

somechris: Since this commit is only the first in a series, and the full series is (I suppose) in the [`redo-tests` branch in your own fork (1699fff; i.e. #2344)](https://github.com/nihohit/redis-rs/tree/redo-tests), I'll give the links to that branch when relevant. Maybe that's outdated, but that's the best I could find online)

> The reason is that the macros should generate the RESP2 & RESP3
> variants (or the tcp, tls, unix variants) only if the env variables
> aren't set.

Right. But that's functions like [`expand_single_server_test()`](https://github.com/nihohit/redis-rs/blob/1699fff878a7ac398f21b7880e31c6cfa72bc8ad/test-macros/src/single.rs#L129) in the `test-macros` crate, which takes the full matrix of [`single::single_server_matrix()`](https://github.com/nihohit/redis-rs/blob/1699fff878a7ac398f21b7880e31c6cfa72bc8ad/test-macros/src/single.rs#L20) (with both `RESP2` and `RESP3`), and filters it using [`test_env::protocol_enabled()`](https://github.com/nihohit/redis-rs/blob/1699fff878a7ac398f21b7880e31c6cfa72bc8ad/test-macros/src/test_env.rs#L47), which in turn relies on [`test_env::requested_protocol()`](https://github.com/nihohit/redis-rs/blob/1699fff878a7ac398f21b7880e31c6cfa72bc8ad/test-macros/src/test_env.rs#L30).

My point is: None of these four functions ever call `server::use_protocol` from the `redis-test` module. Those four functions handle the checking whether the `PROTOCOL` variable is set/unset on their own. Fully within the `test-macro` crate.

The only callers to `server::use_protocol` that I could find are of the form
```
use_protocol().unwrap_or(...->RESP2)`
```
(or they essentially do the same thing).

Or did I miss callers? :thinking: 

And if the change is not actually needed in the full series (#2344), is it then really necessary?

nihohit: if it won't be used, it isn't necessary, no.

PR #2397, opened 2026-09-14
commands: Merge `JsonCommands` into `Commands` (typed-json 3/6)

somechris:
By merging the commands of `JsonCommands` into `Commands` (and accordingly for the async, ... variants), we finally can treat all commands alike and have them in one spot.

This will help with maintenance and with keeping things alike.

The commands are brought over _without_ modifications (except adding `cfg` guards and a `RedisResult<Generic>` return type) to allow easy checking that everything got brought over correctly.

Tayloring return types, and switching to `ready_cmd` is left to follow-up commits.

With this commit JSON commands are now also available on the typed command implementations (e.g.: `TypedCommands`), but still with generic return types.

This is part of #2201.


somechris:
I'm somewhat scared that Commands or JSON parts get touched at some point and I have to redo this PR once more :-/

Sooo...

@nihohit: ... just to pick up the weekend momentum :grin: ... Do you think you could squeeze in a review of this PR? :thinking:

(If you prefer to split the macro messing with the actual command moving, I'll gladly split it.)

version2.md
```diff
@@ -10,6 +10,22 @@ redis = "2"
 
 ## Breaking Changes
 
+### JSON commands got promoted to standard commands (Breaking Change)
+
+`JsonCommands`, and `JsonAsyncCommands` got merged into `Commands`, and `JsonAsyncCommands`.
+
+**Migration:** Switch from `JsonCommands` and `JsonAsyncCommands` to `Commands` and `AsyncCommands`
```
nihohit, line 17: IMO the documentation should push towards the typed trait variants, or at least mention them.

somechris: With this change, there is little benefit for users in switching to the typed traits. So I'd rather leave it as it.

The next commit though will mess with the return types.

Then there is real benefit to switch to the typed traits. So only then I'd adapt these paragraphs to suggest the typed traits.

Otherwise, if we adapt the paragraph already with this commit, then if I get knocked over by a bus 🚌 today in the evening, and cannot upload the follow-up commit, then people would think: "Why should I switch to typed traits? :shrug:  There is no benefit!"

I'll try to stay clear of busses today in the evening ... but WDYT? :thinking:

nihohit: I prefer that you not get hit, regardless of the state of the documentation :)
But sure, if you prefer to fix this in the next PR, that's fine.

PR #2399, opened 2026-09-15
Stream CLIENT LIST off the wire via Connection::client_list_iter (fixes #2396)

### LINKED ISSUE #2396: CLIENT LIST on a connection with many thousands of clients causes a large, avoidable memory spike
### Description

There's no dedicated `client_list` helper on the connection types, so calling `CLIENT LIST` goes through the generic reply path, which materializes the whole bulk-string reply as an owned buffer (`Value::BulkString(bs.to_vec())` in the parser) before the caller sees anything. Against a server with a large number of connected clients, this reply can be very large, and there's no way to consume it incrementally.

### Observed behavior

Testing with an increasing number of connected clients showed memory usage during a single `CLIENT LIST` call scaling roughly linearly with connection count, reaching multi-GB territory once the connection count climbs into the tens of thousands — realistic for a busy, long-lived server.

### Steps to reproduce

1. Start a Redis server and open several thousand to tens of thousands of client connections against it.
2. From a separate connection, issue `CLIENT LIST` (e.g. via `cmd("CLIENT").arg("LIST").query::<String>(&mut con)`).
3. Observe memory usage before/after the call.

### Suggested fix

A streaming/incremental read path for large bulk-string replies (starting with `CLIENT LIST` as the motivating case) that yields data as it arrives off the socket rather than buffering the full reply, bounding peak memory to roughly the read-buffer size instead of the total client count.

### Related

Filing the same report against `go-redis`, `hiredis`, and `redis-py`, since the underlying cause (bulk-string replies aren't streamable) appears to be common across client implementations rather than specific to this crate.

fcostaoliveira:
## Problem

`CLIENT LIST` goes through the generic reply path, which materializes the entire bulk-string (or, under RESP3, Verbatim String) reply into one owned buffer before any of it is usable. Against a server with a large number of connected clients this reply can be multi-GB, and today there's no way to consume it incrementally. Fixes #2396. The same underlying limitation was independently reported against redis-py: redis/redis-py#4326.

## Approach

Adds `Connection::client_list_iter(&mut self) -> RedisResult<ClientListIter<'_>>`, which reads the `CLIENT LIST` reply directly off the wire in bounded 16KiB chunks and yields one client-info line at a time, instead of buffering the whole reply. Peak memory for the call is roughly the chunk size plus the longest single line, independent of client count.

Supporting pieces, all internal to this PR:
- A private `BulkReplyLines` line-streamer over the RESP bulk-string/verbatim-string wire format, unit-tested against synthetic `Read` sources (including a bounded header-line length, so a misbehaving/malicious peer can't grow it unboundedly).
- `impl Read for ActualConnection` for uniform raw-read access across TCP/TLS/Unix transports.
- `Parser::has_buffered_input()`, a thin wrapper over combine's public `Decoder::buffer()`, used to safely drain any reply debt (`messages_to_skip` from an earlier client-side read timeout) and any already-buffered RESP3 push before issuing `CLIENT LIST`, so the iterator never mislabels stale or out-of-band data as client-list output.
- A clean single-line server-side rejection of `CLIENT LIST` itself (e.g. an ACL/`NOPERM` or `NOAUTH` error) is surfaced as a normal server error with the connection left open and usable -- that case fully and cleanly consumes exactly one line, so the wire stays in sync, unlike a genuine framing error.
- `ClientListIter` drains-on-drop, so abandoning the iterator early (`.take(n)`, an early `break`/`return`) can't desync the connection for the next command.

Handles both RESP2 (plain bulk string) and RESP3 (Verbatim String) reply shapes. Sync `Connection` only; async is left for a follow-up, and there's no equivalent on the sync cluster client either -- the doc comment says so explicitly.

## Measured impact

Added `redis/examples/client_list_memory.rs`, a manual reproduction tool (not part of the automated test suite) that opens a configurable number of idle client connections against a real server and measures this process's own RSS (via `/proc/self/status`, Linux-only) across both a generic `CLIENT LIST` call (`query::<String>()`) and an equivalent `client_list_iter()` call.

Measured on an otherwise-idle 96-core/371GB host, connections confirmed to close cleanly between runs so each measurement is independent, not cumulative:

- 20,000 connections: generic `query()` delta +13056 kB vs `client_list_iter()` delta 0 kB (re-run twice, identical +13056 kB both times)
- 40,000 connections: generic `query()` delta +26112 kB vs `client_list_iter()` delta 0 kB
- 60,000 connections: generic `query()` delta +39936 kB vs `client_list_iter()` delta 0 kB

The generic-path delta scales linearly with connection count (~0.65-0.67 kB/connection at every scale tested), matching the growth pattern described in #2396; `client_list_iter()` shows a 0 kB delta at every scale tested. Reproduce yourself with `cargo run --release --example client_list_memory -- <server-addr> <connection-count>` (see the example for exact usage and a safety note -- it opens real, held-open connections).

## Testing

Unit tests for `BulkReplyLines` against synthetic readers (RESP2/RESP3, nil/empty replies, malformed/truncated input, a bounded header length, a malformed short verbatim-string prefix), integration tests against a live server (RESP2 and RESP3, an early iterator drop, a pending unrelated reply, a prior client-side timeout, and a server-side rejection of `CLIENT LIST` itself), and two deterministic `UnixStream::pair()`-based regression tests covering (a) a RESP3 push arriving while draining stale reply debt, and (b) a client-side read timeout leaving a partial value buffered -- both scenarios that could otherwise silently corrupt or hang the connection.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nihohit:
* is this intentionally a partial implementation on only one connection type? Because you missed cluster, async, and async cluster implementations. Will they be added in future PRs?

* I'm not sure why you chose to fix this only for `CLIENT LIST` - wouldn't it make more sense to allow streaming for many kinds of responses, not just one specific command?

* the documentation is VERY verbose, and includes unnecessary details that don't interest the average user, such as redis-py mentions and issue links. Please reduce it to something more concise. Most users aren't interested in reading multi-paragraph docs.

somechris:
I’m not a maintainer, but it looks like covering this special case grows the 2800 lines file `redis/src/connection.rs` by whopping 1100 lines. :grinning: That looks like a lot for handling a special case.

And briefly glancing over it, it looks like it (currently) covers only synchronous `Connection`s (not asynchronous). :thinking:

In #2396, it’s mentioned that the same issue arises also in Redis drivers for other languages.

Just going out on a limb, but … if it’s just for the client list … wouldn’t it be simpler to offer iterator functionality directly in the Redis server (maybe through a new command). There are already other commands that allow consuming results piece by piece (`SCAN`, `BF.SCANDUMP`) :thinking:

fcostaoliveira:
Thanks both for the detailed look!

**Scope (sync `Connection` only)** — yes, intentional. `MultiplexedConnection`/async and the cluster clients aren't covered by this PR. Happy to open those as follow-ups if there's interest, or a tracking issue to note the gap — your call on whether you'd rather see this land narrow first or want the scope widened before merging.

**Why `CLIENT LIST` specifically, not a general streaming primitive** — fair question, and it applies to both of your comments about the size of this. The actual mechanism (`BulkReplyLines`) isn't `CLIENT LIST`-specific: it streams any RESP bulk-string/verbatim-string reply line-by-line, regardless of command. I only exposed it publicly as `client_list_iter()` because that's the concretely reported case (#2396), and kept the public surface narrow rather than speculatively generalizing to commands nobody's asked for yet. If you'd rather see a general primitive instead (something like `Connection::stream_bulk_reply_lines(cmd) -> impl Iterator<...>`, with `client_list_iter` becoming a thin wrapper over it), I'm glad to reshape it that way — didn't want to guess at the right general API shape without your input first.

**Docs** — trimmed per your note, down from ~9 paragraphs to 4 sentences.

**@somechris on the file-size concern** — of the ~1100 lines added to `connection.rs`, about 650 are tests (several exist specifically because earlier review passes found real bugs in this — a hang, a silent-data-loss case, an unnecessary connection close on a common server error) and ~410 are the actual implementation. Fair to still think that's a lot for one command; see the generalization question above.

**@somechris on a server-side cursor/SCAN-style command instead** — that could well be the better long-term fix, but it'd need to happen in the Redis/Valkey server itself, not this client — redis-rs can't add new server commands. Worth raising on redis/redis or valkey-io/valkey if you want to pursue it; this PR is scoped to what's achievable client-side today.

nihohit:
> If you'd rather see a general primitive instead

Yes, I'd prefer a general primitive, and have wrappers for it for specific commands via the `commands` trait.
Notice that from what I see, most users use the async interfaces, so it's better to have an async implementation first, and only later add the sync implementation.

In order to save time, I recommend proposing a design BEFORE opening a large PR, in order to check that the general direction.

More importantly - please read the [coding guidelines](https://github.com/redis-rs/redis-rs/blob/main/CONTRIBUTING.md#are-ai-contributions-allowed).

> [...] having a conversation with an AI isn't always pleasant - it would be better if you answer PR comments yourself, instead of piping comments and responses.

If this PR isn't important enough to warrant your personal attention, it also isn't important enough to warrant our attention.

PR #2400, opened 2026-09-16
RedisServer: Allow to dump server info during panicking drops (server-info 2/3)

somechris:
When using `redis-test` in integration to develop server modules, the server's stdout/stderr/logs contain valuable information to debug the module. But there was no simple way to report them automatically.

So we dump them during when `drop`ping during a `panic`.

This way, they get reported when an assertion fails, but not during successful test runs.

And as the server info would be overly noisy for failing integration tests of plain applications, we make the dumping optional. So plain applications' integration tests by default do not show it. But it can be turned on for server module projects.


nihohit:
should we dump to `stdout`? maybe let the user pass a file / file descriptor, and choose the output target? WDYT?

somechris:
No we _should_ not dump to `stdout`! :+1:

(But that's a longer and different discussion :grin:)

Right now, though, our tests are riddled with `println`s (102 `println`s and 10 `eprintln`s).

So a `println` looked like the best fit. (If you prefer `eprintln`, then let's use that)

For me, the important part is that it's directly on screen right when `cargo test` fails.

`println` / `eprintln` does that.

`file names` are less ideal as I then need to append `|| cat file` to `cargo test` calls.

`file descriptor`s would work. But to avoid that users have to mess with file descriptors, we'd probably provide a default one. That default would probably be `stdout` / `stderr`, so we're back to square #1.

To be honest, I'd start with `println` / `eprintln`. That's really simple to use, it get's the job done, it fits in, and it's easy to extend in many directions. Everything else is "guessing" what users find most convenient ... I doubt there are many users of the feature anyway. So I'd start minimal and expand once people come with new use cases. WDYT?

nihohit:
Fair enough. I agree that this is exactly what we need for our tests, but as a public API, I think that we need to be open for extension if any user will request that. That being said, I think that `dump_info_on_panicking_drop` should be renamed to mention `stdout`, so that if we ever add `dump_info_on_panicking_drop_to_file`, the distinction will be clear.

somechris:
It seems we really care about making this extendable. So the latest version (a40fe39) switches to an enum that allows to select the output.

That approach:

* is backwards-compatible
* does not change default behaviour
* is extendable (with file name, file descriptor, ...),
* allows to easily get the server info to stdout,
* avoids having dedicated functions for each output flavour.

WDYT?

redis-test/src/server.rs
```diff
@@ -576,6 +607,32 @@ impl Debug for RedisServerCommand {
     }
 }
 
+/// Output specifier that converts to a [`Write`].
+#[non_exhaustive]
+#[derive(Default)]
+pub enum Output {
+    /// Does not output anything
+    ///
+    /// This is the default.
```
nihohit, line 616: I think the line below says the same thing :)

nihohit (approved):
LGTM :)

redis-test/src/server.rs
somechris: Sure. But `#[default]` is for machines. And the `/// This ...` text is for humans (`cargo doc` does not mark the default otherwise, and I think in this case, calling out the default makes sense).

Do you prefer to remove the text?

nihohit: nah, is ok, just funny :)

